PR #22, opened 2021-12-27
feat(datasource,mongo): create mongo datasource and collection and generate the schema from connection

arnaud-moncel:
## Definition of Done

TODO:
generate invers relations

### General

### Security

### DIFF
--- a/packages/_example/src/datasources/live/seed.ts
+++ b/packages/_example/src/datasources/live/seed.ts
@@ -17,7 +17,5 @@ export default async function seed(dataSource: LiveDataSource) {
   } catch {
     // eslint-disable-next-line no-console
     console.log('Please run `docker-compose up -d && yarn db:seed` to run the db with data');
-  } finally {
-    await sequelizeMySql.close();
   }
 }


packages/_example/scripts/db-seed.ts
```diff
@@ -30,6 +32,20 @@ async function createStoreRecords(db: Sequelize, ownerRecords: any[]): Promise<a
...
+
+  for (let i = 0; i < 30; i += 1) {
+    reviewsRecords.push({
+      title: faker.word.adjective(),
+      message: faker.lorem.paragraphs(1),
+      storeId: faker.helpers.randomize(storeRecords.map(({ id }) => id)),
+    });
+  }
+
+  await db.models.review.create(reviewsRecords);
```
romain-gilliotte, line 46: all the other seeds are using the query interface, which is not ready yet for mongo
we should change this later on

packages/datasource-mongoose/src/utils/schema-fields-generator.ts
```diff
@@ -0,0 +1,207 @@
...
+        schemaField.columnType = Array.isArray(schemaField.columnType) ? [type] : type;
+        delete schemaField.schema;
+      }
+    });
+
+    return prebuildFieldsSchema;
+  }
+
+  private static getColumnType(instance: string, field: SchemaType): ColumnType {
+    if (field.options?.enum) return 'Enum';
```
romain-gilliotte, line 58: In mongoose, the enum keyword is a validator, so you can have Date enum, number enums, and so on.
=> afaik we only support string enums (needs to be verified at least), so I don't this works.

To be safe, I would only use the `Enum` type if `instance === 'String'`, and otherwise add a validator to the field

It would also be wise to check that the array does **not** contains nulls

packages/datasource-mongoose/src/utils/schema-fields-generator.ts
```diff
@@ -0,0 +1,207 @@
...
+import { Model, Schema, SchemaType } from 'mongoose';
+
+import FilterOperatorBuilder from './filter-operator-builder';
+import MongooseCollection from '../collection';
+
+type MongooseColumnSchema = ColumnSchema & { schema?: CollectionSchema['fields'] };
+type ModelFields = { [key: string]: SchemaType };
+
+export default class SchemaFieldsGenerator {
+  static buildRelationsInPlace(collections: MongooseCollection[]): void {
```
romain-gilliotte, line 18: this method should be called `addInverseRelationships()`, `addOneToManyRelationships()`, or something similar

packages/datasource-mongoose/src/utils/schema-fields-generator.ts
```diff
@@ -0,0 +1,207 @@
...
+    const relations: [FieldSchema, string][] = [];
+    collections.forEach(collection => {
+      Object.values(collection.schema.fields).forEach(fieldSchema => {
+        if (fieldSchema.type === 'ManyToOne') {
+          relations.push([fieldSchema, collection.name]);
+        }
+      });
+    });
+
```
romain-gilliotte, line 27: why not create the one to many relations on the fly?
the intermediary step of storing in an array and then iterating the array looks useless

packages/datasource-mongoose/src/utils/schema-fields-generator.ts
```diff
@@ -0,0 +1,207 @@
...
+  FieldSchema,
+  PrimitiveTypes,
+  RecordData,
+} from '@forestadmin/datasource-toolkit';
+import { Model, Schema, SchemaType } from 'mongoose';
+
+import FilterOperatorBuilder from './filter-operator-builder';
+import MongooseCollection from '../collection';
+
+type MongooseColumnSchema = ColumnSchema & { schema?: CollectionSchema['fields'] };
```
romain-gilliotte, line 14: This looks a bit unconventional (adding extra fields to our Schema type that no one is reading), but may prove convenient when we implement Story 4, so let's leave it for now.

packages/_example/connected-datasources.puml
```diff
@@ -59,8 +64,9 @@ OWNER "1" -- "*" POST : has
 POST "1" -- "*" COMMENT : has
 STORE "*" -- "1" OWNER : has
 STORE "1" -- "*" DVD : has
-DVD "*" -- "*" RENTALS : has
 STORE "1" -- "1" ADDRESS : has
+STORE "1" -- "*" REVIEW : has
+DVD "*" -- "*" RENTALS : has
```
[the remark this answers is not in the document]
Scra3: really strange, the relation should exist already. Maybe a miss remove on beta .. ?

packages/datasource-mongoose/src/utils/schema-fields-generator.ts
Scra3: yes totally

packages/datasource-mongoose/src/utils/schema-fields-generator.ts
```diff
@@ -0,0 +1,180 @@
...
+    collections.forEach(collection => {
+      Object.values(collection.schema.fields).forEach(fieldSchema => {
+        if (fieldSchema.type === 'ManyToOne') {
+          const foreignCollection = collections.find(c => c.name === fieldSchema.foreignCollection);
+
+          if (!foreignCollection) {
+            throw new Error(`The collection '${fieldSchema.foreignCollection}' does not exist`);
+          }
+
+          foreignCollection.schema.fields[`${collection.name}__oneToMany`] = {
```
romain-gilliotte, line 24: will break if multiple many to one relations are declared in the same collection

romain-gilliotte, line 24: will break if multiple many to one relations are declared in the same collection

Scra3: ty for the catch

PR #242, opened 2022-04-13
fix(datasource sequelize):  build a correct where clause when there is relation on delete and update

Scra3:
### General

### Security

### DIFF
--- a/packages/datasource-sequelize/src/utils/query-converter.ts
+++ b/packages/datasource-sequelize/src/utils/query-converter.ts
@@ -91,24 +91,16 @@ export default class QueryConverter {
     }
 
     const keys = [...model.primaryKeyAttributes];
-    const records = await model.findAll({
-      attributes: keys,
-      where: whereOptions,
-      include,
+    const records = await model.findAll({ attributes: keys, where: whereOptions, include });
+    const conditions = records.map(record => {
+      const equals = keys.map(pk => new ConditionTreeLeaf(pk, 'Equal', record.get(pk)));
+
+      return ConditionTreeFactory.intersect(...equals);
     });
 
-    const conditions = keys.map(pk => {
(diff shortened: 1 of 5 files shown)


packages/datasource-sequelize/src/utils/query-converter.ts
```diff
@@ -71,6 +71,40 @@ export default class QueryConverter {
...
+    const records = await model.findAll({
+      attributes: [model.primaryKeyAttribute],
+      where: whereOptions,
+      include,
+    });
+
+    return QueryConverter.getWhereFromConditionTree(
+      model,
+      new ConditionTreeLeaf(
+        model.primaryKeyAttribute,
```
romain-gilliotte, line 101: I don't think this works with composite pks

romain-gilliotte (changes requested):

packages/datasource-sequelize/src/utils/query-converter.ts
```diff
@@ -71,6 +72,45 @@ export default class QueryConverter {
...
+    }
+
+    const keys = [...model.primaryKeyAttributes];
+    const records = await model.findAll({
+      attributes: keys,
+      where: whereOptions,
+      include,
+    });
+
+    const conditions = keys.map(pk => {
```
romain-gilliotte, line 100: This does not generates the condition we want.

This generates
```
id1 IN (1, 2, 3) AND id2 IN (4, 5, 6)
```

We want
```
(id1 = 1 and id2 = 4) or (id1 = 2 and id2 = 5) or (id1 = 3 and id2 = 6)
```

romain-gilliotte (changes requested):

romain-gilliotte (approved):
Tested on mssql, it works well.
I took the time to rebase it as it was conflicting

PR #243, opened 2022-04-14
fix(datasource sequelize): add include from sort relation for list

Scra3:
### General

### Security

### DIFF
--- a/packages/datasource-sequelize/src/collection.ts
+++ b/packages/datasource-sequelize/src/collection.ts
@@ -50,6 +50,10 @@ export default class SequelizeCollection extends BaseCollection {
       );
     }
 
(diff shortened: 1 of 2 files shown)


packages/datasource-sequelize/src/collection.ts
```diff
@@ -50,6 +50,10 @@ export default class SequelizeCollection extends BaseCollection {
       );
     }
```
romain-gilliotte, line 52: What happens to the generated query when the same include is specified twice? (sort + condition tree)
Did you enable logging on your local project, and check if the generated SQL does the jointure only once?

romain-gilliotte (approved):

PR #246, opened 2022-04-15
fix(datasource sequelize): type converter compute correctly type from sequelize DataTypes

arnaud-moncel:
### General

- [X] Test manually the implemented changes
- [X] Validate the code quality (indentation, syntax, style, simplicity, readability)
- [X] Ensure that Types have been updated according to your changes (if needed)

### Security

### DIFF
--- a/packages/datasource-sequelize/src/utils/type-converter.ts
+++ b/packages/datasource-sequelize/src/utils/type-converter.ts
@@ -1,4 +1,4 @@
-import { AbstractDataType, AbstractDataTypeConstructor, ArrayDataType, DataTypes } from 'sequelize';
+import { AbstractDataTypeConstructor, ArrayDataType, DataType, DataTypes } from 'sequelize';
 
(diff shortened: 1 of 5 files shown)


packages/datasource-sequelize/src/utils/type-converter.ts
```diff
@@ -29,61 +29,56 @@ export default class TypeConverter {
...
+    }
+  }
 
-    if (dataTypeName === 'ARRAY') {
+  public static fromDataType(dataType: DataType): ColumnType {
+    if (dataType instanceof DataTypes.ARRAY) {
       const arrayDataType = dataType as ArrayDataType<AbstractDataTypeConstructor>;
 
-      return [this.fromDataType(arrayDataType.options.type)];
+      return [this.fromDataType(arrayDataType.type)];
```
romain-gilliotte, line 78: Using both `this` and `TypeConverter` in the same function.

Both are OK, but better choose one and stay with it

PR #264, opened 2022-04-25
fix: do not crash when operators are not provided by the datasource

romain-gilliotte:
`ColumnSchema.filterOperators` is marked as optional in `packages/datasource-toolkit/src/interfaces/schema.ts`

When not provided, the field should just be assumed not to support any operator, however currently this cause the agent to crash.

To replicate the issue, comment out the filterOperator definition in any datasource and start the example project

### DIFF
--- a/packages/agent/test/agent/utils/forest-schema/generator-fields_column.test.ts
+++ b/packages/agent/test/agent/utils/forest-schema/generator-fields_column.test.ts
@@ -30,15 +30,24 @@ describe('SchemaGeneratorFields > Column', () => {
         fields: {
           isbn: factories.columnSchema.build({
             columnType: 'String',
+            filterOperators: new Set([
+              'Equal',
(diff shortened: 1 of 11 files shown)


packages/agent/test/agent/utils/forest-schema/generator-fields_column.test.ts
```diff
@@ -30,15 +30,24 @@ describe('SchemaGeneratorFields > Column', () => {
         fields: {
           isbn: factories.columnSchema.build({
             columnType: 'String',
+            filterOperators: new Set([
```
Scra3, line 33: filterOperators: new Set(MAP_ALLOWED_OPERATORS_FOR_COLUMN_TYPE.String) ?

jeffladiray (approved):
LGWF

PR #281, opened 2022-04-29
fix: correctly parse the value of "in" nodes in conditions trees

romain-gilliotte:
### General

### Security


packages/agent/src/agent/utils/condition-tree-parser.ts
```diff
@@ -0,0 +1,75 @@
...
+  static fromPlainObject(collection: Collection, json: unknown): ConditionTree {
+    if (ConditionTreeParser.isLeaf(json)) {
+      const operator = ConditionTreeParser.parseOperator(json.operator);
+      const value = ConditionTreeParser.parseValue(collection, { ...json, operator });
+
+      return new ConditionTreeLeaf(json.field, operator, value);
+    }
+
+    if (ConditionTreeParser.isBranch(json)) {
+      const branch = new ConditionTreeBranch(
```
Scra3, line 21: please add unit test to this case.

packages/agent/src/agent/utils/condition-tree-parser.ts
```diff
@@ -0,0 +1,75 @@
...
+        json.aggregator.toLowerCase() === 'and' ? 'And' : 'Or',
```
Scra3, line 22: if the condition tree from the front passed an unknown aggregator, we convert it to "or". I think we should raise an error.

Scra3: Add tests with "and", "or" and unknown please :)

packages/agent/src/agent/utils/condition-tree-parser.ts
```diff
@@ -0,0 +1,75 @@
...
+      );
+
+      return branch.conditions.length !== 1 ? branch : branch.conditions[0];
+    }
+
+    throw new Error('Failed to instantiate condition tree from json');
+  }
+
+  /** Convert snake_case to PascalCase */
+  private static parseOperator(operator: string): Operator {
```
Scra3, line 33: the method name and the behavior is not really the same. We can remove the method to `parseOperatorToPascalCase` ?

Scra3 (approved):

PR #297, opened 2022-05-09
fix(datasource sql): throw a user firendly error when the provided connection uri is invalid

arnaud-moncel:
### General

- [X] Test manually the implemented changes
- [X] Validate the code quality (indentation, syntax, style, simplicity, readability)
- [X] Ensure that Types have been updated according to your changes (if needed)

### Security

### DIFF
--- a/packages/datasource-sql/src/index.ts
+++ b/packages/datasource-sql/src/index.ts
@@ -4,6 +4,12 @@ import SqlDataSource from './datasource';
 
 // eslint-disable-next-line import/prefer-default-export
 export function createSqlDataSource(connectionUri: string): DataSourceFactory {
+  if (!/.*:\/\//g.test(connectionUri))
+    throw new Error(
+      `Connection Uri "${connectionUri}" provided to Sql data source is not valid.` +
+        ' Should be <dialect>://<connection>.',
+    );
+
   return async (logger: Logger) => {
     const datasource = new SqlDataSource(connectionUri, logger);
     await datasource.build();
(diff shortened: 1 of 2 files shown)


packages/datasource-sql/src/index.ts
```diff
@@ -4,6 +4,12 @@ import SqlDataSource from './datasource';
 
 // eslint-disable-next-line import/prefer-default-export
 export function createSqlDataSource(connectionUri: string): DataSourceFactory {
+  if (!/.*:\/\//g.test(connectionUri))
```
romain-gilliotte, line 7: It looks safer to do something like:

```
try {
  const uri = new URL(connectionUrI);
  if (!['postgres', 'mysql', ....].includes(uri.protocol)) throw new Error('Unsupported protocol machin');
  ...
} catch (e) {
  throw new Error('Malformed uri')
}
```

packages/datasource-sql/test/index.test.ts
```diff
@@ -1,9 +1,20 @@
...
+      expect(() => createSqlDataSource('invalid')).toThrowError(
+        'Connection Uri "invalid" provided to Sql data source is not valid.' +
+          ' Should be <dialect>://<connection>.',
+      );
+    });
+  });
+
+  describe('when the connection Uri is valid', () => {
+    test('should return a data source factory', () => {
+      const factory = createSqlDataSource('postgres://');
```
romain-gilliotte, line 15: this uri is not valid 😅

romain-gilliotte (changes requested):

romain-gilliotte (approved):

PR #362, opened 2022-06-29
feat(importField): make it writable by default

Scra3:
### General

### Security

### DIFF
--- a/docs/agent-customization/fields/computed.md
+++ b/docs/agent-customization/fields/computed.md
@@ -7,7 +7,7 @@ In this first example, we want to display the customer's city in the table-view.
 However, as there is no "city" column on the "customer" database table, we need to retrieve it from the "address" relation.
 
 ```javascript
-collection.importField('city', { path: 'address:city' });
+collection.importField('city', { path: 'address:city', readonly: true });
 ```
 
 {% hint style="info" %}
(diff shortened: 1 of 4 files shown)


packages/agent/src/builder/collection.ts
```diff
@@ -70,6 +70,18 @@ export default class CollectionBuilder<
       enumValues: schema.enumValues,
     });
 
+    this.stack.write.getCollection(this.name).replaceFieldWriting(name, value => {
```
romain-gilliotte, line 73: You don't know that the source field is writable.
this needs to be checked in the schema and throw otherwise

docs/agent-customization/fields/computed.md
```diff
@@ -7,9 +7,14 @@ In this first example, we want to display the customer's city in the table-view.
 However, as there is no "city" column on the "customer" database table, we need to retrieve it from the "address" relation.
 
 ```javascript
-collection.importField('city', { path: 'address:city', readonly: true });
+collection.importField('city', { path: 'address:city' });
```
romain-gilliotte, line 10: i think the readonly option was useful

it prevents customers from writing this:
```importField().replaceFieldWriting().importField().replaceFieldWriting()```

romain-gilliotte (changes requested):
can you re-add the readonly flag?

it avoids needing to

packages/agent/src/builder/collection.ts
Scra3: good catch

docs/agent-customization/fields/computed.md
Scra3: I will add an other PR, because it is already removed.

romain-gilliotte (approved):

PR #367, opened 2022-07-04
fix(builder): data-source charts are lost when mounted on the agent builder

Scra3:
### General

### Security

linked to CU-2ew2e65

### DIFF
--- a/packages/agent/src/builder/agent.ts
+++ b/packages/agent/src/builder/agent.ts
@@ -1,9 +1,8 @@
 import {
-  BaseDataSource,
   ChartDefinition,
   Collection,
+  CompositeDatasource,
   DataSourceFactory,
-  RenameCollectionDataSourceDecorator,
   TCollectionName,
   TSchema,
 } from '@forestadmin/datasource-toolkit';
@@ -31,7 +30,7 @@ import TypingGenerator from './utils/typing-generator';
  *  .start();
  */
 export default class AgentBuilder<S extends TSchema = TSchema> {
-  private readonly compositeDataSource: BaseDataSource<Collection>;
+  private readonly compositeDataSource: CompositeDatasource<Collection>;
   private readonly stack: DecoratorsStack;
   private readonly options: AgentOptions;
   private customizations: (() => Promise<void>)[] = [];
@@ -57,7 +56,7 @@ export default class AgentBuilder<S extends TSchema = TSchema> {
    */
   constructor(options: AgentOptions) {
     this.options = OptionsValidator.withDefaults(options);
-    this.compositeDataSource = new BaseDataSource<Collection>();
+    this.compositeDataSource = new CompositeDatasource<Collection>();
     this.stack = new DecoratorsStack(this.compositeDataSource);
   }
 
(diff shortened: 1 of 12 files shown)


packages/datasource-toolkit/src/composite-datasource.ts
```diff
@@ -0,0 +1,69 @@
...
+    rename?: { [newName: string]: string },
+  ): DataSource {
+    const decorated = new RenameCollectionDataSourceDecorator(dataSource);
+
+    for (const [oldName, newName] of Object.entries(rename ?? {})) {
+      decorated.renameCollection(oldName, newName);
+    }
+
+    return decorated;
+  }
```
romain-gilliotte, line 68: As said on the previous file, can you move this static method to the RenameCollectionDecorator and then call it from the agent builder instead of from here?

packages/agent/src/builder/agent.ts
```diff
@@ -68,16 +67,7 @@ export default class AgentBuilder<S extends TSchema = TSchema> {
...
-      const decorated = new RenameCollectionDataSourceDecorator(dataSource);
-
-      for (const [oldName, newName] of Object.entries(options?.rename ?? {})) {
-        decorated.renameCollection(oldName, newName);
-      }
-
-      for (const collection of decorated.collections) {
-        this.compositeDataSource.addCollection(collection);
-      }
+      this.compositeDataSource.addDataSource(await factory(this.options.logger), options?.rename);
```
romain-gilliotte, line 70: You are creating a useless dependency between the CompositeDataSource and RenameCollectionDataSource.

For now, we've be cautious to make sure that all decorators are independent, let's try to keep it that way

```
const datasource = await factory(this.options.logger);
const renamed = RenameCollectionDataSourceDecorator.rename(dataSource, options?.rename);

this.compositeDataSource.addDataSource(renamed);
```

packages/datasource-toolkit/src/decorators/chart/datasource.ts
```diff
@@ -23,14 +23,20 @@ export default class ChartDataSourceDecorator extends DataSourceDecorator {
...
+    const chartDefinition = this.charts[name];
+
+    if (chartDefinition) {
+      return chartDefinition(new AgentCustomizationContext(this, caller), new ResultBuilder());
     }
 
-    return this.charts[name](new AgentCustomizationContext(this, caller), new ResultBuilder());
+    return super.renderChart(caller, name);
   }
 }
```
romain-gilliotte, line 42: I don't understand what does this change do.
It's the same than before right?

romain-gilliotte (changes requested):

packages/datasource-toolkit/src/decorators/chart/datasource.ts
Scra3: I think it's for remove the negation in the condition.

packages/datasource-toolkit/src/composite-datasource.ts
Scra3: yes

packages/agent/src/builder/agent.ts
Scra3: yes ofc

romain-gilliotte (approved):

PR #371, opened 2022-07-06
fix: skip table without id for DS SQL and throw error if a collection has no id.

Scra3:
### General

### Security


packages/datasource-sql/src/datasource.ts
```diff
@@ -97,18 +99,58 @@ export default class SqlDataSource extends SequelizeDataSource {
...
-        });
+        if (SqlDataSource.hasPrimaryKey(modelDefinition)) {
+          this.sequelize.define(tableName, modelDefinition, {
+            tableName,
+            timestamps: hasTimestamps,
+            paranoid: isParanoid,
+          });
+        } else {
+          tablesWithoutPrimaryKey.push(tableName);
+        }
```
jeffladiray, line 110: This contribution is only focused on Louiss issue.

The overall ticket was more about try catching the model definition instead of manually handling every case - and forward the Sequelize error to the user as a warning

jeffladiray: Let say we have another issue related to a sequelize model definition (Not the one regarding id/PK def), it will still prevent the agent to start.

packages/datasource-toolkit/src/composite-datasource.ts
```diff
@@ -21,6 +21,9 @@ export default class CompositeDatasource<
   ): CompositeDatasource {
     const decoratedDataSource = CompositeDatasource.renameCollections(dataSource, rename);
     this.addCollections(decoratedDataSource);
+    // must be call after the `addCollection` because
+    // some decorators are dependent of other collection when we call the schema getter
+    CompositeDatasource.checkCollectionsHasPrimaryKey(decoratedDataSource);
```
jeffladiray, line 26: Maybe that's just me, but maybe a validation decorator could handle such cases.

Good idea though, as at least one primary key is required for FA to work

jeffladiray (commented):
This PR should target `beta`, and is missing a few test cases to ensure the fix is 👌

PR #372, opened 2022-07-07
refactor(DS-SQL): wrap sequelize datasource to a sequelize datasource builder.

Scra3:
### General

### Security


packages/datasource-sql/src/utils/model-factory.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    if (paranoid) {
+      model = ModelFactory.removeParanoidColumn(model);
+    }
+
+    sequelize.define(tableName, model, {
+      tableName,
+      timestamps,
+      paranoid,
+    });
```
jeffladiray, line 29: That's pure opinion I guess, but I'm not a fan of defining the sequelize model here.
Sequelize model definition for the SQL datasource should be centralized in a single file to prevent having to switch between multiple files + having weird magic happen in some utils.

jeffladiray: For this case, I would feel more confident having this function return `{ tableName, model, timestamps, paranoid }` and define the model in the `definedModels` function of the datasource

packages/datasource-sql/src/utils/model-factory.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  private static removeTimeStampColumns(modelDefinition: ModelAttributes): ModelAttributes {
```
jeffladiray, line 32: Maybe these could be re-ordered (timestamps related grouped :) )

packages/datasource-sql/src/utils/relation-factory.ts
```diff
@@ -0,0 +1,91 @@
+import { Model } from 'sequelize/types/model';
+import { ModelStatic, Sequelize } from 'sequelize';
+
+import { ForeignKeyReference } from './types';
+
+export default class RelationFactory {
+  static build(
```
jeffladiray, line 7: Again, I think that these would be easier to unit test if they do not require a complete instance of sequelize (With already defined models) to work

Scra3: yeah good idea :)

jeffladiray (commented):
My last issue with this is that I tend to think that we shouldn't pass sequelize around too much.

We add issues in the past with Sequelize introducing breaking changes without proper semver respect, and having to find every single piece of code relying on Sequelize and patch them is a pain.

(Eg. Supporting Sequelize > 6.0 < 6.12)

I'm ok with merging this as is if we can't find a better way of isolating sequelize in this refactor right now though (Since I'm aware that we are already passing sequelize around in multiple utils)

romain-gilliotte (approved):
It does work

PR #374, opened 2022-07-11
fix: update code and sequelize version to avoid crash

arnaud-moncel:
### General

### Security

### DIFF
--- a/packages/datasource-live/package.json
+++ b/packages/datasource-live/package.json
@@ -29,9 +29,9 @@
   },
   "devDependencies": {
     "lodash": "^4.17.21",
-    "sequelize": "6.18.0"
+    "sequelize": "6.21.3"
   },
   "peerDependencies": {
-    "sequelize": "6.18.0"
+    "sequelize": "6.21.3"
   }
 }
--- a/packages/datasource-live/package.json
+++ b/packages/datasource-live/package.json
@@ -14,7 +14,8 @@
   "dependencies": {
     "@forestadmin/datasource-sequelize": "1.0.0-beta.33",
     "@forestadmin/datasource-toolkit": "1.0.0-beta.26",
-    "sqlite3": "^5.0.2"
+    "sqlite3": "^5.0.2",
+    "sequelize": "6.21.3"
   },
   "files": [
     "dist/**/*.js",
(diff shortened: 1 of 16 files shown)


packages/datasource-sequelize/src/utils/query-converter.ts
```diff
@@ -44,13 +42,14 @@ export default class QueryConverter {
     operator: Operator,
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     value?: any,
-  ): WhereOperators | OrOperator | Where {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  ): any {
```
jeffladiray, line 46: This is weird though, I'll look if Sequelize provide a better type than `any`

jeffladiray (commented):
I still need to check if everything works as expected for older sequelize version (>6)

packages/datasource-live/package.json
```diff
@@ -29,9 +29,9 @@
   },
   "devDependencies": {
     "lodash": "^4.17.21",
-    "sequelize": "6.18.0"
+    "sequelize": "6.21.3"
   },
   "peerDependencies": {
```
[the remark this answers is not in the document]
romain-gilliotte: For the live datasource and sql datasource it should actually not be a peerdep at all but a simple dependency

Customers don't need to know that we use sequelize (and should be able to use another incompatible version in the same project without interactions)

packages/datasource-sequelize/package.json
```diff
@@ -27,10 +27,10 @@
   },
   "devDependencies": {
     "fishery": "^2.1.0",
-    "sequelize": "6.18.0",
+    "sequelize": "6.21.3",
     "sqlite3": "^5.0.2"
   },
   "peerDependencies": {
-    "sequelize": "6.18.0"
+    "sequelize": "6.21.3"
```
romain-gilliotte, line 34: This should be 6.x.x

Customers will have warning if they don't use this exact version

romain-gilliotte (dismissed):

jeffladiray (commented):
As we discover, using anything regarding sequelize should use the Sequelize instance provided by the user.
I'm pretty sure this is worth mentioning somewhere for futur developers as a comment, since it was a real pain to understand that in the first place.

(The fix looks good to me, but I'll do a few other tests before approving)

jeffladiray (approved):

PR #385, opened 2022-08-19
fix: error message should be handle during authentication

Thenkei:
### General

### Security


packages/agent/src/agent/routes/system/error-handling.ts
```diff
@@ -21,7 +21,7 @@ export default class ErrorHandling extends BaseRoute {
       await next();
     } catch (e) {
       let status = e.status || HttpCode.InternalServerError;
-      let message = 'Unexpected error';
+      let message = e.message || 'Unexpected error';
```
jeffladiray, line 24: Isn't that useless? You are throwing an `UnprocessableError`, and it should forward the message

romain-gilliotte:
I'm closing this PR, as it will no longer be needed if we're getting rid of client_id & agent_url in the config

PR #392, opened 2022-08-24
fix(mongoose): implement route to submit related data edition

romain-gilliotte:
# How to test

The example project does not have any writable embedded relation...
To have one:

- Edit _example/src/forest/agent.ts
- Replace `bills.items` by `bills` on line 45
- You should have an embedded relation on the "account bills" models

### DIFF
--- a/packages/agent/src/agent/routes/modification/update-field.ts
+++ b/packages/agent/src/agent/routes/modification/update-field.ts
@@ -8,6 +8,7 @@ import {
 import { Context } from 'koa';
 import Router from '@koa/router';
 
+import { HttpCode } from '../../types';
 import CollectionRoute from '../collection-route';
 import IdUtils from '../../utils/id';
 import QueryStringParser from '../../utils/query-string';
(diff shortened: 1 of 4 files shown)


packages/agent/src/agent/routes/modification/update-field.ts
```diff
@@ -0,0 +1,57 @@
...
+      await this.services.permissions.getScope(this.collection, context),
+    );
+
+    const caller = QueryStringParser.parseCaller(context);
+    const filter = new Filter({ conditionTree });
+
+    // Load & check record
+    const [record] = await this.collection.list(caller, filter, new Projection(field));
+
+    if (index > record[field]?.length) {
```
Scra3, line 46: Maybe we should also throw an error if the the index is negative ?

packages/agent/src/agent/routes/modification/update-field.ts
```diff
@@ -0,0 +1,57 @@
...
+      throw new ValidationError(`Field '${field}' is too short`);
+    }
+
+    record[field][index] = subRecord;
+
+    // Update record
+    await this.collection.update(caller, filter, { [field]: record[field] });
+
+    context.response.status = 204;
```
Scra3, line 55: All the routes use the HttpCode type :)

Scra3 (approved):

PR #393, opened 2022-08-24
fix(datasource-sql): skip models which do not have primary keys

Scra3:
### General

### Security


romain-gilliotte (approved):
This actually does not skips models with no PK (they crash when the collection is loaded from the frontend though)

It only skips the models that crash within sequelize.

(note that, models which have no pk, BUT which also have a column named "id" do crash on sequelize)

PR #394, opened 2022-08-24
fix: agent url prefix

leonard-henriquez:
### General

### Security

### DIFF
--- a/packages/agent/src/builder/agent.ts
+++ b/packages/agent/src/builder/agent.ts
@@ -146,7 +146,8 @@ export default class AgentBuilder<S extends TSchema = TSchema> {
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   mountOnExpress(express: any): this {
-    express.use('/forest', this.getConnectCallback(false));
+    const { prefix } = this.options;
+    express.use(prefix, this.getConnectCallback(false));
     this.options.logger('Info', `Successfully mounted on Express.js`);
 
     return this;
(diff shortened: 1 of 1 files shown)


packages/agent/src/builder/agent.ts
```diff
@@ -146,7 +146,8 @@ export default class AgentBuilder<S extends TSchema = TSchema> {
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   mountOnExpress(express: any): this {
-    express.use('/forest', this.getConnectCallback(false));
+    const { prefix } = this.options;
+    express.use(prefix, this.getConnectCallback(false));
```
ghusse, line 150: I suggest changing the way the prefix is configured, because the URL must always end with `/forest`, so the last part of the path should not be customizable.

leonard-henriquez: We could either:
- add '/forest' after the prefix
- or run a regex validation that the prefix option ends with '/forest' (and throw an error if it's not the case)

Would any of those options be good for you?

ghusse: I'd rather go with the first solution, but I'll let @jeffladiray or @romain-gilliotte decide what to do with this option.

romain-gilliotte:
Hello @leonard-henriquez 

I'll be committing on top of your contribution so that our internal linting, coverage and documentation requirements are met, and let you review the result

romain-gilliotte:
Closing this, as the issue was resolved in #395

PR #395, opened 2022-08-25
fix(configuration): make 'prefix' option work as intended

romain-gilliotte:
### General

### Security

### DIFF
--- a/docs/getting-started/install/create-your-agent.md
+++ b/docs/getting-started/install/create-your-agent.md
@@ -3,6 +3,7 @@ const { createAgent } = require('@forestadmin/agent');
 
 const agent = createAgent({
   // Mandatory options (those will be provided during onboarding)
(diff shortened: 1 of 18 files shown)


jeffladiray (commented):
The code review is good for me (Thanks for documenting the option with example). I still need to do manual tests on this, just to check that all the cases are covered (And that informations in the forestadmin-schema & serialization process is still working as expected in all cases).

@leonard-henriquez if you have a bit of time to check it out, we would be happy to have your opinion on this.

leonard-henriquez:
The documentation with the examples is really clear and the manual test for my use case worked.
The PR looks good to me 👌 
Thank you very much for your prompt response !!

jeffladiray (approved):
Tried to do my best to manually tests all 4 possible cases, I didn't spot any issue.

Serialization & forestadmin-schema seems to be well formed as well, so ... LGTM 🎉

PR #400, opened 2022-08-30
fix(prefix): fix prefix validation regexp

### LINKED ISSUE #399: Special characters in prefix lead to errors
## Expected behavior

It works

## Actual behavior

Get the following error
```
Error: options.prefix is invalid. It should contain the prefix on which forest admin routes should be mounted (i.e. "/api/v1")
    at OptionsValidator.checkOtherOptions (/workspaces/JabRefOnline/node_modules/@forestadmin/agent/dist/builder/utils/options-validator.js:76:19)
    at OptionsValidator.validate (/workspaces/JabRefOnline/node_modules/@forestadmin/agent/dist/builder/utils/options-validator.js:35:26)
    at AgentBuilder.start (/workspaces/JabRefOnline/node_modules/@forestadmin/agent/dist/builder/agent.js:198:53)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

## Context

Set `prefix: '_admin'`. After removing the regex check at https://github.com/ForestAdmin/agent-nodejs/blob/4b21ca78a4b46d748c8f853e22490c81e6a6bbd3/packages/agent/src/builder/utils/options-validator.ts#L110 it works well, so I think the regex needs to be modified to allow for `_`.

- Package Version:

jeffladiray:
Close #399

Final decision is the one describe [here](https://developers.google.com/maps/url-encoding) without:
- reserved character 
- `.` that shouldn't be supported in prefix.

`\w` handle `[a-zA-Z0-0_]`, so this should be ok

image

### General

### Security

### DIFF
--- a/packages/agent/src/builder/utils/options-validator.ts
+++ b/packages/agent/src/builder/utils/options-validator.ts
@@ -107,7 +107,7 @@ export default class OptionsValidator {
   }
 
   private static checkOtherOptions(options: AgentOptions): void {
-    if (typeof options.prefix !== 'string' || !/^[-/a-z]*$/i.test(options.prefix)) {
+    if (typeof options.prefix !== 'string' || !/^[-~/\w]*$/i.test(options.prefix)) {
       throw new Error(
         'options.prefix is invalid. It should contain the prefix on which ' +
           'forest admin routes should be mounted (i.e. "/api/v1")',
--- a/packages/agent/test/builder/utils/http-driver-options.test.ts
+++ b/packages/agent/test/builder/utils/http-driver-options.test.ts
@@ -141,5 +141,32 @@ describe('OptionsValidator', () => {
         );
       });
     });
+
+    test('should fail when prefix contains not allowed characters', () => {
+      expect(() =>
+        OptionsValidator.validate({
+          ...allOptions,
+          prefix: 'this-should#fail&as-a-prefix!',
+        }),
+      ).toThrow(
+        'options.prefix is invalid. It should contain the prefix on which ' +
+          'forest admin routes should be mounted (i.e. "/api/v1")',
+      );
+    });
+
+    test('should work when prefix is valid', () => {
+      const testPrefix = prefix =>
+        expect(() =>
+          OptionsValidator.validate({
+            ...allOptions,
+            prefix,
+          }),
+        ).not.toThrow();
+
+      testPrefix('api/~v1');
+      testPrefix('api/version-1');
+      testPrefix('foo/bar/baz');
+      testPrefix('foo_bar');
+    });
   });
 });


packages/agent/src/builder/utils/options-validator.ts
```diff
@@ -107,7 +107,7 @@ export default class OptionsValidator {
   }
 
   private static checkOtherOptions(options: AgentOptions): void {
-    if (typeof options.prefix !== 'string' || !/^[-/a-z]*$/i.test(options.prefix)) {
+    if (typeof options.prefix !== 'string' || !/^[-_/a-z]*$/i.test(options.prefix)) {
```
ghusse, line 110: Just passing by, I have 2 questions:

- why validating the prefix? What are the things we want to avoid?
- why don't you allow numbers too? I'm pretty sure we can have some other specificities.

jeffladiray: > why validating the prefix? What are the things we want to avoid?

Issue with customer mixin `agentUrl` & `prefix` - preventing their agent to start without any error. Validating beforehand instead of excepting the agent to throw elsewhere seems to be a good idea.

> why don't you allow numbers too? I'm pretty sure we can have some other specificities.

[Good catch](https://developers.google.com/maps/url-encoding), but again I'm still looking the original URI rfc to make sure we didn't miss anything else - that's why the PR is in draft.

packages/agent/src/builder/utils/options-validator.ts
```diff
@@ -107,7 +107,7 @@ export default class OptionsValidator {
   }
 
   private static checkOtherOptions(options: AgentOptions): void {
-    if (typeof options.prefix !== 'string' || !/^[-/a-z]*$/i.test(options.prefix)) {
+    if (typeof options.prefix !== 'string' || !/^[-~/\w]*$/i.test(options.prefix)) {
```
ghusse, line 110: Is it possible to add unit tests that check that these characters are accepted, and that invalid characters are rejected?

jeffladiray: I've just added tests on the `validate` method, specifically on prefix, to avoid testing this private method.

ghusse: Perfect 👌

ghusse (approved):
- ✅&nbsp;Code review
  - ✅&nbsp; Security concerns
- ✅&nbsp;Automatic tests
  - ✅&nbsp; Unit tests
- ✅&nbsp;Manual tests
- ⚠️&nbsp;PR title

I suggest changing the title to `fix(prefix): allow all possible prefixes as defined in the RFC`

tobiasdiez:
Thanks for the quick fix @jeffladiray!

PR #407, opened 2022-09-06
fix(validator): allow null value for the field value and use strict value for the ids. 

Scra3:
### General

### Security

linked to: #2xwxkwn

### DIFF
--- a/packages/datasource-toolkit/src/validation/field.ts
+++ b/packages/datasource-toolkit/src/validation/field.ts
@@ -23,7 +23,12 @@ export default class FieldValidator {
       }
 
       if (values !== undefined) {
-        values.forEach(value => FieldValidator.validateValue(field, schema, value));
+        values.forEach(value =>
+          FieldValidator.validateValue(field, schema, value, [
+            schema.columnType as PrimitiveTypes,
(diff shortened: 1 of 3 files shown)


packages/datasource-toolkit/src/validation/field.ts
```diff
@@ -63,18 +76,12 @@ export default class FieldValidator {
...
-    if (allowedTypes) {
-      if (!allowedTypes.includes(type)) {
-        throw new ValidationError(`Wrong type for "${field}": ${value}. Expects [${allowedTypes}]`);
-      }
-    } else if (type !== schema.columnType) {
-      throw new ValidationError(
-        `Wrong type for "${field}": ${value}. Expects ${schema.columnType}`,
-      );
+    if (allowedTypes && !allowedTypes.includes(type)) {
+      throw new ValidationError(`Wrong type for "${field}": ${value}. Expects ${allowedTypes}`);
```
Scra3, line 80: I have removed '[]' because it is ambiguous when there is only one allowed type. For example, `[Boolean]` is ambiguous because the user can believe that we are waiting for an boolan array

romain-gilliotte (approved):
It works 👍

PR #408, opened 2022-09-06
fix: close multiple issues with datasource-sql

romain-gilliotte:
### General

### Security


packages/datasource-sql/src/orm-builder/helpers/relation-extractor.ts
```diff
@@ -0,0 +1,80 @@
+import { Relation } from '../types';
+import { Table } from '../../introspection/types';
+
+export default class RelationExtracter {
```
Scra3, line 4: Extracter => Extractor ?

packages/datasource-sql/src/orm-builder/helpers/relation-namer.ts
```diff
@@ -0,0 +1,73 @@
...
+import { Table } from '../../introspection/types';
+
+export default class RelationNamer {
+  static getUniqueRelationNames(table: Table, relations: Relation[]): string[] {
+    const names = relations.map(this.getSimpleName);
+
+    // Search for duplicates
+    const indexesByName: Record<string, number[]> = {};
+
+    for (let i = 0; i < names.length; i += 1) {
```
Scra3, line 13:
```suggestion
    const groupByName = names.reduce((indexesByName, name) => {
      indexesByName[name] ??= [];
      indexesByName[name].push(name);

      return indexesByName;
    }, {} as Record<string, string[]>);```

Comment (Scra3) [reply packages/datasource-sql/src/orm-builder/helpers/relation-namer.ts:13]: We win win one line ...

File: packages/datasource-sql/src/orm-builder/helpers/relation-namer.ts [hunk 3]
```diff
@@ -0,0 +1,73 @@
...
+
+  private static getSimpleName(relation: Relation): string {
+    const { foreignKey } = relation;
+    let name = relation.to;
+
+    if (foreignKey?.length > 3 && foreignKey?.endsWith('_id'))
+      name = foreignKey.substring(0, foreignKey.length - 3);
+
+    if (foreignKey?.length > 2 && foreignKey?.endsWith('Id'))
+      name = foreignKey.substring(0, foreignKey.length - 2);
```
Scra3, line 52: Not very conventional but the case "toto-id" or "toto-Id" is not catch.

packages/datasource-sql/src/orm-builder/helpers/relation-extractor.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    return relations;
+  }
+
+  private static listIndirectRelations(table: Table): Relation[] {
+    const relations: Relation[] = [];
+    const columns = table.columns.filter(c => c.primaryKey && c.constraints.length === 1);
+
+    if (columns.length === 2) {
```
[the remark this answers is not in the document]
romain-gilliotte: 👍

packages/datasource-sql/src/orm-builder/helpers/relation-extractor.ts
romain-gilliotte: 👍

packages/datasource-sql/src/orm-builder/helpers/relation-namer.ts
```diff
@@ -0,0 +1,73 @@
+import { plural, singular } from 'pluralize';
+
+import { Relation } from '../types';
+import { Table } from '../../introspection/types';
+
+export default class RelationNamer {
```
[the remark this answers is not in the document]
romain-gilliotte: 👍

packages/datasource-sql/src/orm-builder/helpers/relation-namer.ts
romain-gilliotte: Not a big fan of using `reduce` in that case for something else than reducing an array.

The rule of thumb I follow, is that if the callback is not pure, I use a for loop

packages/datasource-sql/src/orm-builder/helpers/relation-namer.ts
romain-gilliotte: I did not handle the case, because it's pretty stange to use hyphen in SQL

Those need to be quoted absolutely everywhere so that the database does not understands that `toto-id` means "toto minus id" and throws "columns toto and id are undefined"

Scra3 (approved):

PR #411, opened 2022-09-07
feat: expose typing generator

leonard-henriquez:
When I use the typing generator, it causes my NestJS app to enter in a restart loop because it always rewrites the type file (even if there are no changes), the NestJS file watcher catches this file change and restart the app.
It would be better if the typing generator would write the file only if there are changes but it would create a caching behaviour that has other drawbacks.

Instead, I would prefere to be able to use the TypingGenerator by myself (in a script triggered manually or at startup)

### General

### Security

### DIFF
--- a/packages/agent/src/index.ts
+++ b/packages/agent/src/index.ts
@@ -2,10 +2,13 @@ import { TSchema } from '@forestadmin/datasource-toolkit';
 
 import { AgentOptions } from './types';
 import Agent from './builder/agent';
+import TypingGenerator from './builder/utils/typing-generator';
 
 export { default as Collection } from './builder/collection';
 export { Agent };
 export * from './types';
+
+export { TypingGenerator }
 
 export function createAgent<S extends TSchema = TSchema>(options: AgentOptions): Agent<S> {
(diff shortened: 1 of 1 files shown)


romain-gilliotte:
Hi @leonard-henriquez 

I'm not a big fan of exposing internals from the package as those will show up on the [Api Reference](https://github.com/ForestAdmin/agent-nodejs/wiki/@forestadmin.agent).

It's not really usable right now on Github Wiki, but we have plans to release it in a better format, and link to it where needed from the main documentation.

If NestJS file watcher enters restart loops that's a big issue!
I would want that to work out of the box...

I made this PR: https://github.com/ForestAdmin/agent-nodejs/pull/412

It you want to manually control when the typing file is generated, you can already do that by setting options.typingPath using an environnement variable

```
  const envOptions: AgentOptions = {
        // [...]
        typingsPath: process.env.FOREST_GENERATE_TYPES ? 'src/forest/typings.ts' : null,
  };
```

and then start the agent with `FOREST_GENERATE_TYPES=1 npm start`

PR #413, opened 2022-09-07
fix: allow dev admin editors to see the chart they just created

realSpok:
### General

### Security

### DIFF
--- a/packages/agent/src/agent/services/permissions.ts
+++ b/packages/agent/src/agent/services/permissions.ts
@@ -31,6 +31,13 @@ export default class PermissionService {
   async canChart(context: Context): Promise<void> {
     const chart = { ...context.request.body };
 
+    // If the permissions level already allow the chart, no need to check further
+    const CHARTS_ALLOWED_PERMISSIONS_LEVELS = ['admin', 'editor', 'developer'];
+
+    if (CHARTS_ALLOWED_PERMISSIONS_LEVELS.includes(context.state.user.permissionLevel)) {
+      return;
+    }
+
     // When the server sends the data of the allowed charts, the target column is not specified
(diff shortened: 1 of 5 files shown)


packages/agent/src/agent/services/permissions.ts
```diff
@@ -31,6 +31,13 @@ export default class PermissionService {
   async canChart(context: Context): Promise<void> {
     const chart = { ...context.request.body };
 
+    // If the permissions level already allow the chart, no need to check further
+    const CHARTS_ALLOWED_PERMISSIONS_LEVELS = ['admin', 'editor', 'developer'];
+
```
romain-gilliotte, line 36:
```
  /** Checks that a charting query is in the list of allowed queries */
  async canChart(context: Context): Promise<void> {
    const chart = { ...context.request.body };

    // If the permissions level already allow the chart, no need to check further
    if (!['admin', 'editor', 'developer'].includes(context.state.user.permissionLevel)) {
      // When the server sends the data of the allowed charts, the target column is not specified
      // for relations => allow them all.
      if (chart?.group_by_field?.includes(':'))
        chart.group_by_field = chart.group_by_field.substring(0, chart.group_by_field.indexOf(':'));

      const chartHash = hashObject(chart, {
        respectType: false,
        excludeKeys: key => chart[key] === null,
      });

      await this.can(context, `chart:${chartHash}`);
    }
  }
```

romain-gilliotte (approved):

realSpok:
As discussed, I removed `const CHARTS_ALLOWED_PERMISSIONS_LEVELS`,  and moved the `const chart = { ...context.request.body };` to where it is needed.
merging now

PR #432, opened 2022-09-21
perf(permissions): use a lighter endpoint to retrieve scope configuration

ghusse:
### General

### Security


packages/agent/src/agent/services/authorization/internal/user-permission.ts
```diff
@@ -0,0 +1,33 @@
...
+  private userInfoById: Promise<Map<number, UserPermissionV4>> = null;
+
+  constructor(private readonly options: UserPermissionOptions) {}
+
+  public async getUserInfo(userId: number): Promise<UserPermissionV4 | undefined> {
+    if (
+      !this.lastRetrieval ||
+      this.lastRetrieval.getTime() <
+        Date.now() - this.options.permissionsCacheDurationInSeconds * 1000 ||
+      !(await this.userInfoById).has(userId)
```
Thenkei, line 22: ⚠️ I don't get this code. It seems totally unoptimized to me (maybe it ain't but that's not obvious). 

Why not just resolving `await this.userInfoById` when defining/updating `this.lastRetrieval`. And use the map directly to check `usersMap.has(userId)` and to get a user (usersMap.get(userId)) ?

packages/agent/src/agent/services/authorization/internal/rendering-permission.ts
```diff
@@ -0,0 +1,71 @@
...
+    user,
+    allowRetry,
+  }: {
+    renderingId: string;
+    collectionName: string;
+    user: User;
+    allowRetry: boolean;
+  }): Promise<GenericTree> {
+    const [permissions, userInfo]: [RenderingPermissionV4, UserPermissionV4] = await Promise.all([
+      this.permissionsByRendering.fetch(renderingId),
```
Thenkei, line 53:
```suggestion
      this.renderingPermissions.getRenderingPermission(renderingId),
```

packages/agent/src/agent/services/authorization/internal/rendering-permission.ts
```diff
@@ -0,0 +1,71 @@
...
+>;
+
+export default class RenderingPermissionService {
+  private readonly permissionsByRendering: LruCache<string, RenderingPermissionV4>;
+
+  constructor(
+    private readonly options: RenderingPermissionOptions,
+    private readonly userPermissions: UserPermissionService,
+  ) {
+    this.permissionsByRendering = new LruCache({
```
Thenkei, line 22:
```suggestion
    this.renderingPermissions = new LruCache({
```

packages/agent/src/agent/services/authorization/internal/user-permission.ts
ghusse: Because I don't want to send 2 requests to get the list of users if 2 requests are made on the agent at the same time (for instance list & count).

In order to avoid this, the private property contains the promise to the map, that can be set as soon as the request is sent instead of the moment when the response is received.

packages/agent/src/agent/services/authorization/internal/rendering-permission.ts
ghusse: I prefer the `permissionsByRendering` because it indicates that the cache key is the rendering id.

packages/agent/src/agent/services/authorization/internal/user-permission.ts
ghusse: BTW, I explicitly tested this case & checked that only 1 request is sent.

Thenkei: Well, but the function is called every time getUserInfo is called for every user?

ghusse: Yes, but most of the time the cache will be used and won't be retrieved.

Thenkei: Thanks for adding explanations. It helps a lot when the code is not clear when reading it at first glance.

packages/agent/src/agent/services/authorization/internal/rendering-permission.ts
Thenkei: Well, why the cache key is the rendering id?

ghusse: Because permissions are retrieved by rendering, so the cache key is the parameter that is used to retrieve them.

packages/agent/src/services/authorization/authorization.ts
```diff
@@ -72,4 +77,18 @@ export default class AuthorizationService {
       context.throw(403, 'Forbidden');
     }
   }
+
+  async getScope(collection: Collection, context: Context): Promise<ConditionTree> {
+    const { user } = context.state;
```
Thenkei, line 82: ❔ How does it work?

What is `context.state.user`? How can we trust this value? Is it determined from the Authorization token sent by the front?

packages/agent/test/services/authorization/internal/generate-user-scope.test.ts
```diff
@@ -0,0 +1,243 @@
...
+            operator: 'Equal',
+          },
+          {
+            value: '$currentUser.tags.foo',
+            field: 'bar',
+            operator: 'Equal',
+          },
+        ],
+      };
+
```
Thenkei, line 185: ❔ What about deep tree branches?

```
{
  aggregator: 'And',
  conditions: [
    {
      value: '$currentUser.id',
      field: 'foo',
      operator: 'Equal',
    },
    {
      aggregator: 'Or',
      conditions: [
        {
          value: '$currentUser.id',
          field: 'foo',
          operator: 'Equal',
        },
        {
          value: '$currentUser.tags.tor',
          field: 'tor',
          operator: 'Equal',
        },
      ],
    },
  ],
}
```

Thenkei: Agent v2 `GenericTree` supports this format even if the front does not support it today.

Thenkei (commented):
- ✅&nbsp;Code review
  - ✅&nbsp; Security concerns
- ✅  Automatic tests
  - 🔢   Unit tests - Good job! I just saw the missing tests for the "deep tree branches" case.
- ✅  Manual tests
- ✅  PR title
- ✅  Destination branch

Thenkei (approved):

PR #459, opened 2022-09-29
docs: add forestadmin-schema section

jeffladiray:
### General

### Security


romain-gilliotte:
IMHO, we should consider removing the feature altogether as we're getting threads quite often on the community forums that say "my smartaction/collection/etc.. is not showing up on production", because customers don't version it properly

jeffladiray:
DWO forces schema to be exactly the same between environments. Having small differences between a production and a development environment in the schema would prevent users to push/deploy changes.

Not versionning/save to disk this file could lead to unexpected layout issues (Especially with the SQL datasource I guess).

Maybe we will at some point work on this, but while the `.forestadmin-schema.json` is saved on disk, I think this documentation can be interesting

romain-gilliotte (approved):

PR #462, opened 2022-10-03
feat: allow including/excluding collection when adding datasources

romain-gilliotte:
### General

### Security


packages/datasource-customizer/src/decorators/publication-collection/datasource.ts
```diff
@@ -0,0 +1,46 @@
+import { DataSource } from '@forestadmin/datasource-toolkit';
+import DataSourceDecorator from '../datasource-decorator';
+import PublicationFieldCollectionDecorator from '../publication-field/collection';
+
+export default class PublicationCollectionDataSourceDecorator extends DataSourceDecorator<PublicationFieldCollectionDecorator> {
```
SteveBunlon, line 5: I don't understand the choice of `PublicationCollectionDataSourceDecorator` can you explain please ? (I mean, the naming)

packages/datasource-customizer/src/decorators/publication-collection/datasource.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    // Delete the relations to the collections we're not keeping.
+    for (const collection of Object.values(this._collections)) {
+      for (const [fieldName, field] of Object.entries(collection.schema.fields)) {
+        if (
+          (field.type !== 'Column' && deleted.has(field.foreignCollection)) ||
+          (field.type === 'ManyToMany' && deleted.has(field.throughCollection))
+        )
+          collection.changeFieldVisibility(fieldName, false);
```
SteveBunlon, line 29: Is there a reason why we simply set the visibility to false, and not delete it entirely ?

packages/datasource-customizer/src/datasource-customizer.ts
```diff
@@ -36,11 +37,21 @@ export default class DataSourceCustomizer<S extends TSchema = TSchema> {
...
   addDataSource(factory: DataSourceFactory, options?: DataSourceOptions): this {
     this.customizations.push(async logger => {
-      const dataSource = await factory(logger);
-      const renamedDecorator = new RenameCollectionDataSourceDecorator(dataSource);
-      renamedDecorator.renameCollections(options?.rename);
+      let dataSource = await factory(logger);
 
-      this.compositeDataSource.addDataSource(renamedDecorator);
+      if (options?.include || options?.exclude) {
+        const publicationDecorator = new PublicateCollectionDataSourceDecorator(dataSource);
```
SteveBunlon, line 43: I don't understand Publicate here

docs/datasources/connection/relationships.md
```diff
@@ -1,29 +1,26 @@
+Adding as many datasources as you need is good, inter-connecting them together is better.
```
SteveBunlon, line 1: This is maybe a change you had in mind before this PR. Just to point this out, this does not look to be related to this PR. Is this change intended ?

packages/datasource-customizer/src/decorators/publication-collection/datasource.ts
```diff
@@ -0,0 +1,46 @@
...
+    this.validateCollectionName([...(include ?? []), ...(exclude ?? [])]);
+
+    const deleted = new Set<string>();
+
+    // List collection we're keeping from the white/black list.
+    for (const name of Object.keys(this._collections)) {
+      if ((include && !include.includes(name)) || exclude?.includes(name)) {
+        deleted.add(name);
+      }
+    }
```
SteveBunlon, line 20: This means that exclude and include can both be used ?

packages/datasource-customizer/test/decorators/publication-collection/datasource.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        'Unknown collection name: "unknown"',
+      );
+    });
+
+    it('should be able to remove "librariesBooks" collection', () => {
+      decoratedDataSource.publicate(['libraries', 'books']);
+
+      expect(() => decoratedDataSource.getCollection('librariesBooks')).toThrow();
+      expect(decoratedDataSource.getCollection('libraries').schema.fields).not.toHaveProperty(
+        'myLibrary',
```
SteveBunlon, line 95: Shouldn't it be `myBooks` here ?

SteveBunlon (changes requested):
Could be great to have some tests with the include option, to assert that only the provided collection are being kept in the file `packages/datasource-customizer/test/datasource-customizer.test.ts`.

packages/datasource-customizer/src/datasource-customizer.ts
romain-gilliotte: I renamed it to `keepCollectionsMatching`
Publicate does not exists in english... that's badly translated french

packages/datasource-customizer/src/decorators/publication-collection/datasource.ts
romain-gilliotte: Yep. Not super useful though.

`addDataSource(..., { include: ['steve', 'romain'], exclude: ['romain'] })`
=> Keep only steve

In the initial spec we did months ago we wanted to support regexp matching, which would make a lot more sense but I did not implement it because I got lazy.

`addDataSource(..., { include: [/^payfit_/], exclude: ['payfit_credit_cards'] })`
=> Keep all collection prefixed with `payfit_` but not 'payfit_credit_cards'

packages/datasource-customizer/src/decorators/publication-collection/datasource.ts
romain-gilliotte: It's more complicated than that, we can do a meet.

There is a different schema for each layer in the decorator stack.
This tells that all layers on top of us won't have that field in the schema

packages/datasource-customizer/test/decorators/publication-collection/datasource.test.ts
romain-gilliotte: good catch

docs/datasources/connection/relationships.md
romain-gilliotte: It was intended but you're right it is not really related.

My issue was that this PR was introducing a new option to `addDataSource`, and that
- all documented options were under the `Multiple Datasources` section.
- this new option is not strictly related to customer in that situation

=> I refactored this part of the doc

packages/datasource-customizer/src/decorators/publication-collection/datasource.ts
romain-gilliotte: if you look at the `delete` call on line 35, it's actually the same:
we're removing the field only from our layer (we're deleting in `this`).

So underlying layers don't know

docs/datasources/connection/partial-imports.md
```diff
@@ -0,0 +1,17 @@
...
+
+const agent = createAgent(options);
+const aDataSource = createSqlDataSource('postgres://user:pass@localhost:5432/mySchema');
+const anoterDataSource = createSqlDataSource('postgres://user:pass@localhost:5432/mySchema');
+
+// Rename sqlDataSource collections by providing replacements
+agent
+  .addDataSource(aDataSource, { include: ['books', 'reviews'] })
+  .addDataSource(anoterDataSource, { exclude: ['authors'] });
+```
```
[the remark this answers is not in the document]
romain-gilliotte: 😱

docs/datasources/connection/relationships.md
SteveBunlon: Roger that.

packages/datasource-customizer/src/decorators/publication-collection/datasource.ts
SteveBunlon: Let's keep it that way, the regexp way serves edge cases, that can still be achieved by only providing static include exclude. KISS.

SteveBunlon (approved):
Tested on the test agent, works like a charm ❇️

PR #463, opened 2022-10-03
feat: handle correctly custom action approval request

Thenkei:
### General

### Security


packages/agent/src/routes/modification/action.ts
```diff
@@ -109,6 +113,17 @@ export default class ActionRoute extends CollectionRoute {
     );
   }
 
+  private async middlewareCustomActionApprovalRequestData(context: Context, next: Next) {
+    const approvalRequestDataWithAttributes =
+      this.services.authorization.getApprovalRequestData(context);
+
+    if (approvalRequestDataWithAttributes) {
+      context.request.body.data = approvalRequestDataWithAttributes;
```
ghusse, line 121: Does it work if the body is empty? I don't know if this is something that can happen.

packages/agent/src/services/authorization/authorization.ts
```diff
@@ -62,22 +68,52 @@ export default class AuthorizationService {
...
+        generateCustomActionIdentifier(customActionEvenType, customActionName, collectionName),
+      ))
     ) {
       context.throw(403, 'Forbidden');
     }
   }
 
+  public getApprovalRequestData(context: Context) {
+    const {
+      body: {
```
ghusse, line 101: Maybe you can create a type in order to avoid repeating this structure at 2 different places.

packages/agent/src/services/authorization/authorization.ts
```diff
@@ -62,22 +68,52 @@ export default class AuthorizationService {
...
+        data: {
+          attributes: { requester_id: approvalRequesterId } = { requester_id: null },
+          type = 'custom-action-trigger',
+        } = {},
+      } = {},
+    } = context.request;
+
+    let customActionEvenType = CustomActionEvent.Trigger;
+
+    if (type === 'custom-action-requests') {
```
ghusse, line 82: The type can have another value than this one? It seems weird to have this test here made on the type, that is not explicit on the fact that we have an approval.

packages/agent/src/services/authorization/internal/verify-approval.ts
```diff
@@ -0,0 +1,14 @@
+import jsonwebtoken from 'jsonwebtoken';
+
+import { ActionApprovalJWT } from './types';
+
+export default function verifyAndExtractApproval(approvalRequestToken: string, privateKey: string) {
+  try {
+    return (jsonwebtoken.verify(approvalRequestToken, privateKey) as ActionApprovalJWT).data;
+  } catch (err) {
+    throw new Error(
```
ghusse, line 9: It it possible to declare a specific type for your error, in order to be able to catch it correctly in the `authorization` service, and return the appropriate error?

packages/agent/src/services/authorization/authorization.ts
ghusse: In terms of security, do you think it's possible with this code to forge a request:

- Without a signature
- but with a type set to `custom-action-requests`?

From what I can see, I think it would be possible to bypass the absence of the right to trigger by doing this.

Thenkei: You're right. We have the same issue with users forging a trigger like request.

packages/agent/src/services/authorization/authorization.ts
Thenkei:
```
export type ActionApprovalRequest = {
  body:{
    data: {
      attributes: { signed_approval_request: string };
    };
  };
};
```

Pour faire un test du genre 
```
const isActionApprovalRequest = (context.request as ActionApprovalRequest).signed_approval_request;
```

packages/agent/src/routes/modification/action.ts
Thenkei: 👀  Body will always exist, but data could not. I'll update the code.

```
type ParameterizedContext<StateT = DefaultState, ContextT = DefaultContext, ResponseBodyT = unknown> = ExtendableContext
        & { state: StateT; }
        & ContextT
        & { body: ResponseBodyT; response: { body: ResponseBodyT }; };
```

packages/agent/src/services/authorization/authorization.ts
```diff
@@ -63,19 +74,66 @@ export default class AuthorizationService {
...
+    } = context.request;
+
+    if (signedApprovalRequest) {
+      try {
+        return verifyAndExtractApproval(signedApprovalRequest, this.options.envSecret);
+      } catch (e) {
+        if (e instanceof JTWTokenExpiredError) {
+          context.throw(
+            HttpCode.Forbidden,
+            'Failed to verify approval payload. The signed approval request token as expired..',
```
ghusse, line 122:
```suggestion
            'Failed to verify approval payload. The signed approval request token as expired.',
```

ghusse (approved):
👌 perfect

PR #495, opened 2022-10-14
feat(datasource-sql): add advanced options on createSqlDataSource

romain-gilliotte:
### General

### Security

### DIFF
--- a/packages/datasource-sql/src/index.ts
+++ b/packages/datasource-sql/src/index.ts
@@ -2,25 +2,31 @@ import { DataSourceFactory, Logger } from '@forestadmin/datasource-toolkit';
 import { Sequelize } from 'sequelize';
 import { SequelizeDataSource } from '@forestadmin/datasource-sequelize';
 
+import { DatabaseConnectionOptions as ConnectionOptions } from './types';
 import { Table } from './introspection/types';
 import Introspector from './introspection/introspector';
 import ModelBuilder from './orm-builder/model';
 import RelationBuilder from './orm-builder/relations';
 
-function createEmptySequelize(connectionUri: string, logger: Logger): Sequelize {
-  if (!/.*:\/\//g.test(connectionUri))
+function createEmptySequelize(uriOrOptions: ConnectionOptions, logger: Logger): Sequelize {
+  const logging = (sql: string) => logger?.('Debug', sql.substring(sql.indexOf(':') + 2));
+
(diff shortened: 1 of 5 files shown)


packages/datasource-sql/src/index.ts
```diff
@@ -2,37 +2,43 @@ import { DataSourceFactory, Logger } from '@forestadmin/datasource-toolkit';
 import { Sequelize } from 'sequelize';
 import { SequelizeDataSource } from '@forestadmin/datasource-sequelize';
 
+import { DatabaseConnectionOptions as ConnectionOptions } from './types';
```
SteveBunlon, line 5: I find ConnectionOptions self explanatory, naming it like this directly from the type is ok IMO.

packages/datasource-sql/src/types.ts
```diff
@@ -0,0 +1,24 @@
+import { Options } from 'sequelize/types';
+
+export type DatabaseConnectionOptions =
```
SteveBunlon, line 3: I'm pretty sure you have a reason, however I would like to understand. 

Why are we picking up only certain options ? Couldn't we let the user use everything provided / authorized by sequelize ?

SteveBunlon (changes requested):
I don't how much time that would took, but having an additional datasource on the test agent would be awesome, just to see the new way of creating connections in action

packages/datasource-sql/src/index.ts
romain-gilliotte: done

packages/datasource-sql/src/types.ts
romain-gilliotte: I removed options that:
- Would prevent us from migrating to another ORM or to a newer version of sequelize (`define`, ...)
- Duplicate functionality (`hook`, `typeValidation`, ...)
- Did not make sense in the scope of the sql-connector (`operatorAliases`, `quoteIdentifiers`, ...)

Because I know that some customers like to be able to still do those things (accepting that it risks breaking the build in the future when we upgrade the lib), they can simply use a cast.

But I don't want it to be officially supported.

`createSqlDataSource({typeValidation: true} as ConnectionOptions)` will work, and do what you expect.

SteveBunlon: ok copy that chief

SteveBunlon (approved):

PR #499, opened 2022-10-18
feat(advanced-export): add new plugin which allows better exports

romain-gilliotte:
### General

### Security

### DIFF
--- a/packages/plugin-export-advanced/src/index.ts
+++ b/packages/plugin-export-advanced/src/index.ts
@@ -86,9 +86,9 @@ export async function addExportAdvanced(
 
         const renderer = renderers[format];
         const records = await context.collection.list({}, fields);
-        const excelFile = renderer.handler(records, fields);
+        const output = renderer.handler(records, fields);
 
-        return resultBuilder.file(excelFile, `${filename}${format}`, renderer.mimeType);
+        return resultBuilder.file(output, `${filename}${format}`, renderer.mimeType);
       },
     });
   }
--- a/packages/plugin-export-advanced/src/index.ts
+++ b/packages/plugin-export-advanced/src/index.ts
@@ -1,0 +1,95 @@
+/* eslint-disable import/prefer-default-export */
+import type {
+  ActionGlobal,
+  CollectionCustomizer,
+  DataSourceCustomizer,
+} from '@forestadmin/datasource-customizer';
+
+import renderers from './renderers';
+import { Options } from './types';
+
+function getFields(
+  dataSource: DataSourceCustomizer,
+  collection: CollectionCustomizer,
+  remainingDepth: number,
+): string[] {
+  if (remainingDepth <= 0) return [];
+
+  const result = [];
+
+  for (const [name, field] of Object.entries(collection.schema.fields)) {
+    if (field.type === 'Column') result.push(name);
+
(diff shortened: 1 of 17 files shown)


packages/plugin-export-advanced/test/index.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  beforeEach(() => {
+    customizer = new DataSourceCustomizer();
+    customizer.addDataSource(async () =>
+      factories.dataSource.buildWithCollections([
+        factories.collection.build({
+          name: 'books',
+          schema: {
+            fields: {
+              id: factories.columnSchema.build({ columnType: 'Number', isPrimaryKey: true }),
```
Scra3, line 20: factories.columnSchema.isPrimaryKey().build()

romain-gilliotte: I want number, not uuid

romain-gilliotte: 👍

packages/plugin-export-advanced/src/renderers/csv.ts
```diff
@@ -0,0 +1,15 @@
+import getFieldValue from '../utils/get-field-value';
+
+function escape(value: unknown): string {
+  const str = value !== null && value !== undefined ? String(value) : '';
+  // console.log(value);
+
+  return str.includes(',') ? `"${str.replace('\\', '\\\\').replace('"', '\\"')}"` : str;
+}
+
+export default function render(records: Record<string, unknown>[], projection: string[]): string {
```
Scra3, line 10: we have a csv exporter maybe we can use a part of this code ...

Scra3: Or why don't use this lib : https://github.com/C2FO/fast-csv ?

romain-gilliotte: Adding a new dependency means that
- The behavior of that code won't be obvious anymore: "what is getting escaped?", "which separator does it defaults to?". Answering those questions will need to defocus and go read the documentation of the dependency
- We'll need to update that dependency and handle alerts and pull requests from dependabot

I don't mean that dependencies are bad and that we should reinvent the wheel.

I mean that external dependencies should be used when we gain more than the technical debt that we introduce.

In this case, I thought it would take less time to maintain those ~10 lines of code than to use an external dependency.

For Excel exports, as the file format is much more complicated, of course, I used an external dependency (as rolling out an xlsx writer is non trivial)

packages/plugin-export-advanced/src/index.ts
```diff
@@ -0,0 +1,95 @@
...
+      generateFile: true,
+      form: getForm(dataSource, c, options),
+      execute: async (context, resultBuilder) => {
+        const format = context.formValues.Format || options.format;
+        const fields = context.formValues.Fields || options.fields;
+        const filename = context.formValues.Filename || options.filename;
+
+        const renderer = renderers[format];
+        const records = await context.collection.list({}, fields);
+        const excelFile = renderer.handler(records, fields);
```
DrRaider, line 89: Rename

DrRaider (approved):

PR #501, opened 2022-10-18
fix(line-chart): line chart with null value on group by should return some values  

Scra3:
…played correctly

### General

### Security

linked to CU-35jy8mp


romain-gilliotte (changes requested):
Is is OK for the aggregate method to return null values here?

- If it is the indended behavior of the aggregate method, can this be solved by using the `Present` operator on the filter instead of postprocessing the returned rows?
- If it's not the indended behavior, can we fix the connector instead?

romain-gilliotte (approved):

PR #502, opened 2022-10-20
fix: allow to create a record when the client sends a null relation

Scra3:
### General

### Security

link to CU-33zaehb


packages/agent/src/routes/modification/create.ts
```diff
@@ -119,6 +119,8 @@ export default class CreateRoute extends CollectionRoute {
   }
 
   private getRelationRecord(field: string, id: CompositeId): RecordData {
+    if (!id) return null;
```
romain-gilliotte, line 122: even if it is quite unlikely that someone would risk it, the number zero can be a valid id in most databases system.

I would test that id !== null

romain-gilliotte (changes requested):

romain-gilliotte (approved):

PR #507, opened 2022-10-20
feat(chart): add context variables to chart

ShohanRahman:
### General

### Security


packages/forestadmin-client/test/chart/chart-handler.test.ts
```diff
@@ -0,0 +1,366 @@
...
+        const filter: PlainConditionTreeBranch = {
+          aggregator: 'Or',
+          conditions: [
+            {
+              field: 'name',
+              operator: 'Equal',
+              value: '{{masters.selectedRecord.name}}',
+            },
+          ],
+        };
```
adriguy, line 72: It's very confusing to have this injectedFilter values, with respect to the given filter.
It makes me think that the contextVariableInjector can change the aggregator, the operator, etc!

This comment is true for all the file

adriguy (changes requested):

adriguy (approved):

PR #508, opened 2022-10-20
fix(chart/segment): check that query is valid before executing it

VincentMolinie:
### General

### Security

### DIFF
--- a/packages/agent/test/services/authorization/authorization.test.ts
+++ b/packages/agent/test/services/authorization/authorization.test.ts
@@ -1,9 +1,12 @@
-import { ChartType, CollectionActionEvent } from '@forestadmin/forestadmin-client';
+import {
+  ChainedSQLQueryError,
+  ChartType,
+  CollectionActionEvent,
+  EmptySQLQueryError,
+  NonSelectSQLQueryError,
+} from '@forestadmin/forestadmin-client';
 import { Collection } from '@forestadmin/datasource-toolkit';
 import { Context } from 'koa';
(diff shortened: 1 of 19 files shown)


packages/forestadmin-client/package.json
```diff
@@ -15,7 +15,8 @@
     "jsonwebtoken": "^8.5.1",
     "lru-cache": "^7.3.1",
     "object-hash": "^3.0.0",
-    "superagent": "^7.0.1"
+    "superagent": "^7.0.1",
+    "@forestadmin/datasource-toolkit": "1.0.1-alpha.1"
```
VincentMolinie, line 19: @ghusse is that an issue to pass it as a dependencies ? I had to do that be cause I'm extending `UnprocessableError`

ghusse: Yes, we only wanted to use the types and not require it for implementations.

packages/agent/test/services/authorization/authorization.test.ts
```diff
@@ -1,6 +1,9 @@
 import { ChartType, CollectionActionEvent } from '@forestadmin/forestadmin-client';
 import { Collection } from '@forestadmin/datasource-toolkit';
 import { Context } from 'koa';
+import ChainedSQLQueryError from '@forestadmin/forestadmin-client/src/permissions/errors/chained-sql-query-error';
```
romain-gilliotte, line 4: You should export those errors from the forstadmin/client package.

VincentMolinie: Good catch, it's already exported I used a bad auto import 😅

adriguy (approved):

PR #522, opened 2022-10-31
feat(chart): add support for context variables inside SQL query

VincentMolinie:
### General

### Security

### DIFF
--- a/packages/forestadmin-client/src/charts/chart-handler.ts
+++ b/packages/forestadmin-client/src/charts/chart-handler.ts
@@ -84,8 +84,7 @@ export default class ChartHandlerService implements ChartHandlerInterface {
       const contextVariableRenamed = contextVariableName.replace(/\./g, '_');
       contextVariablesUsed[contextVariableRenamed] = contextVariables.getValue(contextVariableName);
 
-      // $$ converts into a simple $, so to have 2 $ you actually need to write 4
-      return `$$$$${contextVariableRenamed}`;
+      return `$${contextVariableRenamed}`;
     };
 
     const query = ContextVariablesInjector.injectContextInValueCustom(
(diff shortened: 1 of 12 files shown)


packages/forestadmin-client/src/permissions/rendering-permission.ts
```diff
@@ -28,7 +28,7 @@ export default class RenderingPermissionService {
     this.permissionsByRendering = new LruCache({
       max: 256,
       ttl: this.options.permissionsCacheDurationInSeconds * 1000,
-      fetchMethod: renderingId => this.loadPermissions(renderingId),
+      fetchMethod: renderingId => this.loadPermissions(Number(renderingId)),
```
adriguy, line 31: Why?
What happens if `renderingId` is not a number? 🤔

packages/forestadmin-client/src/charts/chart-handler.ts
```diff
@@ -63,4 +63,38 @@ export default class ChartHandlerService implements ChartHandlerInterface {
...
+      requestContextVariables: chartRequest.contextVariables,
+    });
+    const contextVariablesUsed: Record<string, unknown> = {};
+
+    const replaceContextVariable = (contextVariableName: string) => {
+      const contextVariableRenamed = contextVariableName.replace(/\./g, '_');
+      contextVariablesUsed[contextVariableRenamed] = contextVariables.getValue(contextVariableName);
+
+      return `$${contextVariableRenamed}`;
+    };
```
adriguy, line 88: I found a bit weird that this function is responsible of mutating `contextVariablesUsed` value (a bit out of nowhere), don't you think?

packages/forestadmin-client/src/permissions/rendering-permission.ts
VincentMolinie: it will give NaN. THis throw a TS error if I don't put it 🤔

PR #547, opened 2022-11-09
fix(datasource mongoose): use updateOne when there are only one record to allow mongoose hook

arnaud-moncel:
### General

### Security

### DIFF
--- a/packages/datasource-mongoose/src/collection.ts
+++ b/packages/datasource-mongoose/src/collection.ts
@@ -157,7 +157,7 @@ export default class MongooseCollection extends BaseCollection {
         const subdocPatch = buildSubdocumentPatch(path, patch, isLeaf);
 
         if (Object.keys(subdocPatch).length) {
-          if (rootIds.length > 1) {
+          if (ids.length > 1 || rootIds.length > 1) {
             await this.model.updateMany({ _id: rootIds }, subdocPatch, { rawResult: true });
           } else {
             await this.model.updateOne({ _id: rootIds }, subdocPatch, { rawResult: true });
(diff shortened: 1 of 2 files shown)


packages/datasource-mongoose/src/collection.ts
```diff
@@ -126,7 +126,11 @@ export default class MongooseCollection extends BaseCollection {
 
     if (!this.prefix) {
       // We are updating a real document, we can delegate the work to mongoose directly.
-      await this.model.updateMany({ _id: ids }, patch, { rawResult: true });
+      if (ids.length > 1) {
+        await this.model.updateMany({ _id: ids }, patch, { rawResult: true });
+      } else {
+        await this.model.updateOne({ _id: ids }, patch, { rawResult: true });
+      }
     } else {
```
romain-gilliotte, line 134: there are more updates in the else clause

packages/datasource-mongoose/src/collection.ts
```diff
@@ -153,7 +157,11 @@ export default class MongooseCollection extends BaseCollection {
         const subdocPatch = buildSubdocumentPatch(path, patch, isLeaf);
 
         if (Object.keys(subdocPatch).length) {
-          await this.model.updateMany({ _id: rootIds }, subdocPatch, { rawResult: true });
+          if (rootIds.length > 1) {
```
romain-gilliotte, line 160: I would have used `ids.length > 1 ` here as well.
Using rootIds here will cause inconsistent behavior (hooks being called for some records, but not others)

SteveBunlon (approved):
Works great 👍

PR #548, opened 2022-11-10
ci: improve workflow by updating action dependencies and reducing time to spin up docker containers

jeffladiray:
### General

### Security

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,11 +58,6 @@ jobs:
         with:
           path: '**/node_modules'
           key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock', 'packages/**/package.json', 'patches/*.patch') }}
-      - name: Download build
-        uses: actions/download-artifact@v3
-        with:
-          name: build
-          path: 'packages/'
       - name: Lint commit message
         uses: wagoid/commitlint-github-action@v2
       - name: lint Javascript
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -58,14 +58,9 @@ jobs:
         with:
           path: '**/node_modules'
           key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock', 'packages/**/package.json', 'patches/*.patch') }}
-      - name: Download build
```
jeffladiray, line 61: This is not needed during the lint phase

.github/workflows/build.yml
```diff
@@ -89,9 +84,9 @@ jobs:
           name: build
           path: 'packages/'
       - name: Start docker datasource-sql containers
-        run: docker-compose -f ./packages/datasource-sql/docker-compose.yml up -d; sleep 20
```
jeffladiray, line 92: Not sure why we paid for the 2 * 15s. 5s seems good from my tests, but relying on an healthcheck may be a bit cleaner

.github/workflows/build.yml
```diff
@@ -195,8 +190,7 @@ jobs:
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
-    needs: [release]
```
jeffladiray, line 198: I know the release phase also waits from the doc-build and doc-link-validation, so this change is mostly here to ensure a semantical dependency between build the doc and publishing it

package.json
```diff
@@ -42,7 +42,7 @@
     "clean": "lerna run clean && rm -rf coverage api-reference/",
     "docs": "yarn build && typedoc",
     "docs:check-links": "find docs -type f -name '*.md' | xargs -L1 markdown-link-check -q",
-    "lint": "lerna exec eslint src test",
+    "lint": "lerna exec --parallel eslint src test",
```
jeffladiray, line 45: This was initially a test, but it seems to work just fine + allow to parallelize the linting, which is a bit faster on my machine (50s without, 28s with)

arnaud-moncel (approved):

PR #560, opened 2022-11-25
refactor(action-permissions): split concernes and simplify implementation

Thenkei:
### General

### Security

### DIFF
--- a/packages/forestadmin-client/src/permissions/action-permission.ts
+++ b/packages/forestadmin-client/src/permissions/action-permission.ts
@@ -10,40 +10,40 @@ export default class ActionPermissionService {
 
   constructor(private readonly options: ForestAdminClientOptionsWithDefaults) {}
 
-  public canOneOf(userId: string, actionNames: string[]): Promise<boolean> {
+  public canOneOf(roleId: number, actionNames: string[]): Promise<boolean> {
     return this.hasPermissionOrRefetch({
-      userId,
+      roleId,
       actionNames,
       allowRefetch: true,
     });
   }
 
-  public can(userId: string, actionName: string): Promise<boolean> {
+  public can(roleId: number, actionName: string): Promise<boolean> {
     return this.hasPermissionOrRefetch({
-      userId,
+      roleId,
       actionNames: [actionName],
       allowRefetch: true,
     });
   }
 
   private async hasPermissionOrRefetch({
-    userId,
+    roleId,
     actionNames,
     allowRefetch,
   }: {
-    userId: string;
+    roleId: number;
     actionNames: string[];
     allowRefetch: boolean;
   }): Promise<boolean> {
     const permissions = await this.getPermissions();
-    const isAllowed = this.isAllowedOneOf({ permissions, actionNames, userId });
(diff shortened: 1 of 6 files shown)


packages/forestadmin-client/src/permissions/action-permission.ts
```diff
@@ -10,48 +10,48 @@ export default class ActionPermissionService {
 
   constructor(private readonly options: ForestAdminClientOptionsWithDefaults) {}
 
-  public canOneOf(userId: string, actionNames: string[]): Promise<boolean> {
+  public canOneOf(roleId: number, actionNames: string[]): Promise<boolean> {
     return this.hasPermissionOrRefetch({
-      userId,
+      roleId,
```
ghusse, line 15: ⚠️ this is a big change that will need to be correctly handled in `forest-express`. It won't be detected by Typescript as the roleId and userId have the same type, and we are using parameters and not an object with named properties.

packages/forestadmin-client/src/permissions/generate-actions-from-permissions.ts
```diff
@@ -11,13 +11,19 @@ import {
...
 
 export type ActionPermissions = {
   everythingAllowed: boolean;
   actionsGloballyAllowed: Set<string>;
-  actionsAllowedByUser: Map<string, Set<string>>;
+  actionsByRole: Map<string, ActionPermission>;
+};
+
+export type ActionPermission = {
+  allowedRoles: Set<number>;
```
ghusse, line 23: Why do you need an object with only one property here? There is some commented code just below

ghusse (changes requested):
- ⚠️&nbsp;Code review
  - ✅&nbsp; Security concerns
- ✅&nbsp;Automatic tests
  - ✅&nbsp; Unit tests
  - ✖️&nbsp; Integration tests
- ✅&nbsp;Manual tests
- ⚠️&nbsp;PR title
- ✅&nbsp;Destination branch

I have a question regarding a type which seems too complex for what its usage now, and there is some commented code.

We will need to change `forest-express` as well to pass the `roleId` and use this new version.

In this regard, this is a breaking change for `forestadmin-client`, it should be tagged as a new major version.

packages/forestadmin-client/src/permissions/action-permission.ts
Thenkei: No, the end interface didn't change.

packages/forestadmin-client/src/permissions/generate-actions-from-permissions.ts
Thenkei: It's for my own PR that adds all conditionals. I didn't want to make ActionPermission a set of role since this object will have multiple properties in the near future.

Thenkei:
> We will need to change forest-express as well to pass the roleId and use this new version.
> In this regard, this is a breaking change for forestadmin-client, it should be tagged as a new major version.

I didn't change the interface of `forestadmin-client`. It's just internal wiring.

Thenkei:
The exposed functions haven't changed signature: `packages/forestadmin-client/src/permissions/permission-with-cache.ts`

ghusse (approved):
Ok, thanks for the answers

PR #562, opened 2022-11-28
feat: custom action support conditional trigger, requiresApproval and approve

Thenkei:
### General

### Security


packages/agent/src/routes/modification/action/action-authorization.ts
```diff
@@ -0,0 +1,375 @@
...
+import CustomActionRequiresApprovalError from './errors/customActionRequiresApprovalError';
+import CustomActionTriggerForbiddenError from './errors/customActionTriggerForbiddenError';
+import InvalidActionConditionError from './errors/invalidActionConditionError';
+
+type CanPerformCustomActionParams = {
+  caller: Caller;
+  customActionName: string;
+  collection: Collection;
+  filterForCaller: Filter; // Filter that may include scope if any
+  filterForAllCaller: Filter; // Filter without scope
```
DayTF, line 23: Maybe we should use 2 types, one that extends the scope and one that doesn't

packages/agent/src/routes/modification/action/action-authorization.ts
```diff
@@ -0,0 +1,375 @@
...
+   * @returns
+   */
+  private static transformToRolesIdsGroupByConditions<T>(
+    actionConditionsByRoleId?: Map<number, T>,
+  ): {
+    roleIds: number[];
+    condition: T;
+  }[] {
+    if (!actionConditionsByRoleId) {
+      return [];
```
DayTF, line 349: Missing test for this scenario

packages/agent/src/routes/modification/action/action-authorization.ts
```diff
@@ -0,0 +1,375 @@
...
+        conditionalFilter,
+        new Aggregation({
+          operation: 'Count',
+        }),
+      );
+
+      return (rows?.[0]?.value as number) ?? 0;
+    } catch (error) {
+      throw new InvalidActionConditionError();
+    }
```
DayTF, line 334: The error message for this exception is to request help from an administrator but we log nothing nor return any information to the user. Maybe we should log something to help them resolve this.

Thenkei: 💡 SHould we log the collection name and action name in order to ease debuging?

packages/agent/src/routes/modification/action/action-authorization.ts
Thenkei: That was a bit preventive, this does not exist in a real-case scenario.

packages/agent/src/routes/modification/action/action-authorization.ts
Thenkei: There are any differences between those 2 objects, unfortunately. But later (after some refactoring from the dev exp) we should be able to move our logic into a collection decorator that will remove the need to "transport" that information.

packages/agent/src/routes/modification/action/action-authorization.ts
```diff
@@ -0,0 +1,369 @@
...
+      });
+
+    // Optimization - We groupBy conditions to only make the aggregate count once when possible
+    const rolesIdsGroupByConditions =
+      ActionAuthorizationService.transformToRolesIdsGroupByConditions(actionConditionsByRoleId);
+
+    let requestRecordsCount: number;
+    let conditionRecordsCounts: number[];
+
+    if (rolesIdsGroupByConditions.length > 0) {
```
ghusse, line 243: Why not directly

```js
if (!rolesIdsGroupByConditions.length) {
  return roleIdsAllowedToApproveWithoutConditions;
}

const requestRecordsCount, ...conditionRecordsCounts] = await Promise.all([
        ActionAuthorizationService.aggregateCountConditionIntersection(
          caller,
          collection,
          filterForAllCaller,
        ),
        ...rolesIdsGroupByConditions.map(({ condition }) =>
          ActionAuthorizationService.aggregateCountConditionIntersection(
            caller,
            collection,
            filterForAllCaller,
            condition,
          ),
        ),
      ]);

    return rolesIdsGroupByConditions.reduce(
      (roleIdsAllowedToApprove, { roleIds }, currentIndex) => {
        if (requestRecordsCount === conditionRecordsCounts[currentIndex]) {
          roleIdsAllowedToApprove.push(...roleIds);
        }

        return roleIdsAllowedToApprove;
      },
      // Roles  with userApprovalEnabled excluding the one with conditions
      // are allowed to approve by default
      roleIdsAllowedToApproveWithoutConditions,
    );
```

packages/agent/src/routes/modification/action/action-authorization.ts
```diff
@@ -0,0 +1,369 @@
...
+          caller,
+          collection,
+          requestFilter,
+          condition,
+        ),
+      ]);
+
+      // If some records don't match the condition then the user
+      // is not allow to perform the conditional action
+      if (matchingRecordsCount !== requestRecordsCount) {
```
ghusse, line 298: I suggest directly returning `return matchingRecordsCount === matchingRecordsCount` here

packages/agent/src/routes/modification/action/action-authorization.ts
```diff
@@ -0,0 +1,369 @@
...
+   */
+  private static transformToRolesIdsGroupByConditions<T>(
+    actionConditionsByRoleId: Map<number, T>,
+  ): {
+    roleIds: number[];
+    condition: T;
+  }[] {
+    const rolesIdsGroupByConditions = Array.from(
+      actionConditionsByRoleId,
+      ([roleId, condition]) => {
```
ghusse, line 348: I did not even know about this syntax, with the second argument! I learn something every day with you @Thenkei , thanks 🙂

packages/agent/src/routes/modification/action/action.ts
```diff
@@ -33,6 +37,8 @@ export default class ActionRoute extends CollectionRoute {
   ) {
     super(services, options, dataSource, collectionName);
     this.actionName = actionName;
+
+    this.actionAuthorizationService = new ActionAuthorizationService(options.forestAdminClient);
```
ghusse, line 41: why not passing this in the services, already passed in the constructor?

packages/agent/src/routes/modification/action/action.ts
```diff
@@ -50,18 +56,46 @@ export default class ActionRoute extends CollectionRoute {
 
   private async handleExecute(context: Context): Promise<void> {
     const { dataSource } = this.collection;
+
     const caller = QueryStringParser.parseCaller(context);
-    const filter = await this.getRecordSelection(context);
+    const filterForCaller = await this.getRecordSelection(context);
```
ghusse, line 61: Is it worth using `Promise.all` here to make the calls in parallel?

packages/agent/src/routes/modification/action/action.ts
```diff
@@ -167,4 +200,10 @@ export default class ActionRoute extends CollectionRoute {
 
     return filter;
   }
+
+  // Question should we intersectCount here (I know I've already done it then move the code but..)
```
ghusse, line 204: I'm not sure to understand the problem. Can we talk about it?

packages/agent/src/routes/modification/action/action-authorization.ts
Thenkei: Early return!

ghusse: Exactly!

packages/agent/src/routes/modification/action/action-authorization.ts
Thenkei: Lord god. 😨

packages/agent/src/routes/modification/action/action-authorization.ts
Thenkei: It allows direct mapping. 💪

packages/agent/src/routes/modification/action/action.ts
Thenkei: We don't use DI in the agent v2.. except for the `authorizationService` that is used by all other routes.

packages/agent/src/routes/modification/action/action.ts
Thenkei: You're right even if it will not improve performance in this case.

Oh, well I remember I wanted not to compute the `filterForCaller` in the first place and only compute it if needed (when we compute the roleIdsAllowedToApprove).

packages/agent/src/routes/modification/action/action.ts
Thenkei: Sure. Let's talk about it tomorrow.

packages/agent/src/routes/modification/action/action.ts
ghusse: The authorization service is passed in `services`. Why does this authorization service related to actions is not? For me, this service should be either passed as the authorization service, or be a property of the authorization service. In both cases, it should be injected.


In your tests, you are mocking private properties which is not something that I consider correct. Private must remain private.

packages/agent/test/routes/modification/action/action.test.ts
```diff
@@ -113,15 +155,57 @@ describe('ActionRoute', () => {
...
+                ...baseContext.requestBody.data.attributes,
+                values: { firstname: 'John' },
+                requester_id: 42,
+              },
+            },
+          },
+        });
+
+        const assertCanApproveCustomAction = jest.spyOn(
+          // @ts-expect-error: test private attributes
```
ghusse, line 198: Expecting an error here is not correct. We should use this comment for tests that are checking that error cases are correctly handled, for instance when you pass `undefined` instead of a real value and the compiler detects it.

Here, the purpose is totally "normal". We are not testing an error case, and thus I don't think this comment is correct.

In addition to that, I don't think that mocking private properties is a good practice and I would like to find another way of mocking these functions.

Thenkei: It wasn't a mock (only spy to say ok it delegates the work to xxx) but indeed it was not a great practice. I've updated the tests.

packages/agent/src/routes/modification/action/action.ts
Thenkei: Indeed because we use it on all routes except action routes.

It was passed as a service then I rework it to feat the actual agent design. Not sure what's best but we should write something that explains what we should do.

And indeed my tests were bad even is I wasn't mocking anything.

packages/agent/src/routes/modification/action/action-authorization.ts
Thenkei: I might have been wrong. 🙄

packages/agent/src/routes/modification/action/action-authorization.ts
```diff
@@ -0,0 +1,367 @@
...
+      return (rows?.[0]?.value as number) ?? 0;
+    } catch (error) {
+      throw new InvalidActionConditionError();
+    }
+  }
+
+  /**
+   * Given a map it groups keys based on their hash values
+   */
+  private static transformToRolesIdsGroupByConditions<T>(
```
ghusse, line 338: This function is only used once. Do you really need it to be generic?

Thenkei: I did it this way to handle both `RawTreeWithDataSource` and `RawTreeWithInputSource` but I changed types to `RawTreeWithSources` let me change that.

ghusse (approved):
- ✅&nbsp;Code review
  - ✅&nbsp; Security concerns
- ✅&nbsp;Automatic tests
  - ✅&nbsp; Unit tests
  - ✅&nbsp; Integration tests
- ✅&nbsp;Manual tests
- ✅&nbsp;PR title
- ✅&nbsp;Destination branch
- ✅&nbsp;PR linked to the clickup task

PR #565, opened 2022-11-30
feat(plugin-flattener): rename plugin, and add support for nested relations and field flattening

romain-gilliotte:
### General

### Security


docs/agent-customization/plugins/provided/flattener.md
```diff
@@ -0,0 +1,78 @@
...
+const { flattenRelation } = require('@forestadmin/plugin-flattener');
+
+agent.customizeCollection('customer', collection => {
+  return collection.use(flattenRelation, {
+    relationName: 'address',
+
+    // Subfields to flatten (defaults to all subfields of the relation)
+    include: ['street', 'city', 'country:name', 'country:isoCode'],
+
+    // Subfields to exclude from flattening (defaults to none)
```
Scra3, line 61: none is not really js style

packages/plugin-flattener/src/flatten-column/index.ts
```diff
@@ -0,0 +1,78 @@
...
+    collection.addField(path, makeField(options.columnName, path, schema));
+  }
+
+  // Implement write
+  if (!schema.isReadOnly && !options.readonly) {
+    collection.addHook('Before', 'Create', makeCreateHook(paths));
+    collection.addHook('Before', 'Update', makeUpdateHook(collection, options.columnName, paths));
+
+    for (const path of paths) {
+      collection.replaceFieldWriting(path, makeWriteHandler(path));
```
Scra3, line 75: Please explain why the replaceFieldWriting is used.

packages/plugin-flattener/src/flatten-column/customization.ts
```diff
@@ -0,0 +1,82 @@
...
+      allRecordByPatch[hash].records.push(record);
+    }
+
+    // Update the nested column by grouping records by patch
+    // This is done to avoid too many requests to the database (we do one per distinct patch
+    // instead of one per record)
+    const promises = Object.values(allRecordByPatch).map(async ({ records, patch }) => {
+      const conditionTree = ConditionTreeFactory.matchRecords(collection.schema, records);
+
+      await context.collection.update(
```
Scra3, line 74: Is it the same behavior if we return the promise instead to await ?

packages/plugin-flattener/src/flatten-column/helpers.ts
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Given a list of flattened paths, unflatten them in place in the given object.
+ * The values of the flattened paths are deleted from the object in place.
+ *
+ * @example
+ * unflattenPathsInPlace(['address@@@streetName'], { 'address@@@streetName': 'rue de la paix' })
+ * // => { address: { streetName: 'rue de la paix' } }
+ */
+export function unflattenPathsInPlace(paths: string[], patch: RecordData): RecordData {
```
Scra3, line 73: This method does two thinks. To avoid it, maybe we can have a method to extract the externalPatch and an other one to delete the patch used by the external ?

Scra3:
Should we add an example in our _example package ?

packages/plugin-flattener/src/flatten-relation/index.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param options.include The list of fields to import.
+ * @param options.exclude The list of fields to exclude.
+ * @param options.readonly Should the imported fields be read-only?
+ */
+export default async function flattenRelation(
+  dataSource: DataSourceCustomizer,
+  collection: CollectionCustomizer,
+  options?: {
+    relationName: string;
+    include?: string[];
```
Scra3, line 24: If I add some attributes that does not exist, I have no errors.

packages/plugin-flattener/src/flatten-column/index.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+export default async function flattenColumn(
+  dataSource: DataSourceCustomizer,
+  collection: CollectionCustomizer,
+  options?: FlattenColumnOptions,
+): Promise<void> {
+  if (!options || !options.columnName) throw new Error('options.columnName is required.');
+  if (!collection) throw new Error('This plugin can only be called when customizing collections.');
+
+  const paths = optionsToPaths(collection, options);
```
Scra3, line 61: Should we throw an error if the columnName does not exist ?

packages/plugin-flattener/src/flatten-column/index.ts
```diff
@@ -0,0 +1,78 @@
...
+ * @param options.columnName The name of the column to import.
+ * @param options.include The list of fields to import.
+ * @param options.exclude The list of fields to exclude.
+ * @param options.level The maximum level of nested fields to import.
+ * @param options.readonly Should the imported fields be read-only?
+ */
+export default async function flattenColumn(
+  dataSource: DataSourceCustomizer,
+  collection: CollectionCustomizer,
+  options?: FlattenColumnOptions,
```
Scra3, line 56: If I add some attributes in include option that does not exist, I have no errors.
Same for exclude.

Scra3: When I add the readOnly option, the fiels is not disable in the frontend. I can send a value but the field is not updated. Is it a frontend bug ?

packages/plugin-flattener/src/flatten-relation/helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+export function getColumns(schema: CollectionSchema): string[] {
+  return Object.keys(schema.fields).filter(name => schema.fields[name].type === 'Column');
+}
+
+export function getRelation(
+  relationName: string,
+  dataSource: DataSourceCustomizer,
+  collection: CollectionCustomizer,
+): RelationSchema {
+  const parts = relationName.split(':');
```
Scra3, line 16:
```suggestion
  const parts = relationName.split(':');
```
```suggestion
const [field, ...relation] = relationName.split(':')```

Comment (Scra3) [inline packages/plugin-flattener/src/flatten-relation/helpers.ts:16]:
```suggestion
  const [field, ...relation] = relationName.split(':');
```

packages/plugin-flattener/src/flatten-relation/helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+  const relation = collection.schema.fields[parts[0]];
```
Scra3, line 17:
```suggestion
  const relation = collection.schema.fields[relation];
```

packages/plugin-flattener/src/flatten-relation/helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (!relation) message = `Relation ${name} not found`;
+    else if (relation.type === 'Column') message = `${name} is a column, not a relation`;
+    else message = `${name} is not a ManyToOne or OneToOne relation`;
+
+    throw new Error(message);
+  }
+
+  return parts.length > 1
+    ? getRelation(
+        parts.slice(1).join(':'),
```
Scra3, line 32:
```suggestion
       nested.join(':'),
```

packages/plugin-flattener/src/flatten-relation/helpers.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    let message: string;
+    if (!relation) message = `Relation ${name} not found`;
+    else if (relation.type === 'Column') message = `${name} is a column, not a relation`;
+    else message = `${name} is not a ManyToOne or OneToOne relation`;
+
+    throw new Error(message);
+  }
+
+  return parts.length > 1
```
Scra3, line 30: nested > 0

Scra3, line 30:
```suggestion
  return nested.length > 0
```

packages/plugin-flattener/src/flatten-relation/index.ts
```diff
@@ -0,0 +1,47 @@
...
+): Promise<void> {
+  if (!options || !options.relationName) throw new Error('options.relationName is required.');
+  if (!collection) throw new Error('This plugin can only be called when customizing collections.');
+
+  const { relationName, include, exclude, readonly } = options;
+  const relation = getRelation(relationName, dataSource, collection);
+  const foreignCollection = dataSource.getCollection(relation.foreignCollection);
+
+  const columns = new Set(include?.length > 0 ? include : getColumns(foreignCollection.schema));
+  if (exclude?.length > 0) exclude.forEach(column => columns.delete(column));
```
Scra3, line 37:
```suggestion
exclude?.forEach(column => columns.delete(column))
```

packages/plugin-flattener/src/flatten-relation/index.ts
```diff
@@ -0,0 +1,47 @@
...
+ * @param options The options of the plugin.
+ * @param options.relationName The name of the relation to import.
+ * @param options.include The list of fields to import.
+ * @param options.exclude The list of fields to exclude.
+ * @param options.readonly Should the imported fields be read-only?
+ */
+export default async function flattenRelation(
+  dataSource: DataSourceCustomizer,
+  collection: CollectionCustomizer,
+  options?: {
```
Scra3, line 22: I think cx team want to import different lvl of relation. If you have many one to one nested relation you can tell to the plugin to import all the fields. To confirm with us...

packages/plugin-flattener/src/flatten-column/customization.ts
romain-gilliotte: yes, it does the same, but let's go with a return, it saves one line

packages/plugin-flattener/src/flatten-column/index.ts
romain-gilliotte: > When I add the readOnly option, the field is not disabled in the front end. I can send a value but the field is not updated. Is it a front-end bug?

Yes: https://app.clickup.com/t/3dm91gu

packages/plugin-flattener/src/flatten-relation/index.ts
romain-gilliotte: The issue is that I can't know from the plugin which relations are real many to one, and which are generated by the `asModels` option in the mongoose driver.

Importing all relations up to a given level can give very strange results if real `many to one` are used... so I'm stuck there

Scra3 (approved):

PR #571, opened 2022-12-12
perf(customizer): improve CPU usage and response time

romain-gilliotte:
Took one hour to do that with `ab` + `vs code CPU profiler`

The request I'm testing in `ab` is the list of the `account` collection in the `_example` project.

Before:
```
Time taken for tests:   15.564 seconds
Complete requests:      5000
Failed requests:        0
Total transferred:      37740000 bytes
HTML transferred:       36400000 bytes
Requests per second:    321.26 [#/sec] (mean)
Time per request:       31.127 [ms] (mean)
Time per request:       3.113 [ms] (mean, across all concurrent requests)
Transfer rate:          2368.04 [Kbytes/sec] received
```

After:
```
Time taken for tests:   10.487 seconds
Complete requests:      5000
Failed requests:        0
Total transferred:      37515000 bytes
HTML transferred:       36175000 bytes
Requests per second:    476.78 [#/sec] (mean)
Time per request:       20.974 [ms] (mean)
Time per request:       2.097 [ms] (mean, across all concurrent requests)
Transfer rate:          3493.46 [Kbytes/sec] received
```

### DIFF
--- a/packages/agent/src/services/serializer.ts
+++ b/packages/agent/src/services/serializer.ts
@@ -136,17 +136,19 @@ export default class Serializer {
   }
 
   private stripUndefinedsInPlace(record: unknown): void {
-    if (record === null || typeof record !== 'object') {
-      return;
-    }
+    if (record !== null && typeof record === 'object') {
+      const keys = Object.keys(record);
+      let i = keys.length;
 
-    const indexable = record as Record<string, unknown>;
+      // eslint-disable-next-line no-plusplus
+      while (i--) {
+        const key = keys[i];
 
-    for (const key of Object.keys(indexable)) {
-      if (indexable[key] === undefined) {
-        delete indexable[key];
-      } else {
-        this.stripUndefinedsInPlace(indexable[key]);
+        if (record[key] === undefined) {
+          delete record[key];
+        } else {
+          this.stripUndefinedsInPlace(record[key]);
+        }
       }
     }
   }
--- a/packages/agent/src/utils/query-string.ts
+++ b/packages/agent/src/utils/query-string.ts
@@ -20,6 +20,8 @@ const DEFAULT_ITEMS_PER_PAGE = 15;
 const DEFAULT_PAGE_TO_SKIP = 1;
 
 export default class QueryStringParser {
+  private static VALID_TIMEZONES = new Set<string>();
+
   static parseConditionTree(collection: Collection, context: Context): ConditionTree {
     try {
       const filters =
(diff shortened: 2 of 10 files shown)


packages/agent/src/utils/query-string.ts
```diff
@@ -43,19 +45,15 @@ export default class QueryStringParser {
   static parseProjection(collection: Collection, context: Context): Projection {
     try {
       const fields = context.request.query[`fields[${collection.name}]`];
+      if (fields === '' || fields === undefined) return ProjectionFactory.all(collection);
```
romain-gilliotte, line 48: This patch allow to access `collection.schema` once, instead of once per field (and makes the code shorter)

packages/agent/src/utils/query-string.ts
```diff
@@ -117,16 +115,20 @@ export default class QueryStringParser {
       throw new ValidationError('Missing timezone');
     }
 
-    // This is a method to validate a timezone using node only
-    // @see https://stackoverflow.com/questions/44115681
-    if (!Intl || !Intl.DateTimeFormat().resolvedOptions().timeZone) {
-      throw new Error('Time zones are not available in this environment');
-    }
+    if (!QueryStringParser.VALID_TIMEZONES.has(timezone)) {
```
romain-gilliotte, line 118: Add cache to timezone validation.
This ensures that each timezone is validated only once, as the performance profile shows that this is expensive (not sure why)

packages/datasource-customizer/src/decorators/collection-decorator.ts
```diff
@@ -23,9 +23,8 @@ export default class CollectionDecorator implements Collection {
   private lastSubSchema: CollectionSchema;
```
romain-gilliotte, line 23: This assumes that the schema of a connector (the collection at the very bottom of the stack) never changes (which is the case).

This allows to stop recursing all the way down each time we access the schema of a collection, which saves tons of time.

packages/agent/src/services/serializer.ts
```diff
@@ -136,17 +136,19 @@ export default class Serializer {
...
-    }
-
-    const indexable = record as Record<string, unknown>;
-
-    for (const key of Object.keys(indexable)) {
-      if (indexable[key] === undefined) {
-        delete indexable[key];
-      } else {
-        this.stripUndefinedsInPlace(indexable[key]);
+    if (record !== null && typeof record === 'object') {
```
romain-gilliotte, line 139: Maybe we can remove this method all together.
I suspect its only use is to make the tests easier to read.

I just made it a bit faster so that it is no longer on top of the profiler

jeffladiray (approved):

PR #573, opened 2022-12-15
test(customizer): fix segment decorator test

romain-gilliotte:
### General

### Security


packages/datasource-customizer/test/decorators/segment/collection.test.ts
```diff
@@ -1,118 +1,110 @@
...
+            aggregator: 'And',
+            conditions: [
+              { operator: 'Equal', field: 'name', value: 'otherNameValue' },
+              { operator: 'Equal', field: 'name', value: 'aNameValue' },
+            ],
+          },
         });
+      });
 
-        it('should throw an error when a conditionTree is not valid', async () => {
```
romain-gilliotte, line 88: The only real difference (besides factorizing code) is in this test.

The decorator should check that the condition tree generated by the user is valid.
This is testing that the decorators throws when the condition tree generated by the frontend is valid.

jeffladiray (approved):

PR #643, opened 2023-03-13
fix(computed-fks): using computed foreign keys requires implementing the Present operator

romain-gilliotte:
### General

### Security

### DIFF
--- a/docs/agent-customization/relationships/computed-fks.md
+++ b/docs/agent-customization/relationships/computed-fks.md
@@ -23,9 +23,10 @@ agent.customizeCollection('customers', collection => {
     getValues: async (customers, context) => {
       // We're using Forest Admin's query interface (you can use an ORM or a plain SQL query)
       const messages = context.dataSource.getCollection('messages');
+      const conditionTree = { field: 'customer_id', operator: 'In', value: customers.map(r => r.id) }
       const rows = await messages.aggregate(
-        { field: 'customer_id', operator: 'In', value: customers.map(r => r.id) },
-        { operator: 'Max', field: 'id', groups: [{ field: 'customer_id' }] },
+        { conditionTree },
+        { operation: 'Max', field: 'id', groups: [{ field: 'customer_id' }] },
       );
 
       return customers.map(record => {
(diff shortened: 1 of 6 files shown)


packages/agent/test/routes/modification/delete-related.test.ts
```diff
@@ -428,6 +427,7 @@ describe('DissociateDeleteRelatedRoute > delete', () => {
                   ],
                   field: 'myLibrary:id',
                 }),
+                factories.conditionTreeLeaf.build({ operator: 'Present', field: 'libraryId' }),
```
Scra3, line 430: Why lines are moved ?

docs/agent-customization/relationships/computed-fks.md
```diff
@@ -33,7 +34,7 @@ agent.customizeCollection('customers', collection => {
     },
   });
 
-  // Implement the 'In' operator (required).
+  // Implement the 'In' operator.
```
Scra3, line 37: why removed the '(required)' if it is the case ?

packages/datasource-toolkit/src/interfaces/query/filter/factory.ts
```diff
@@ -183,10 +189,21 @@ export default class FilterFactory {
...
-        new ConditionTreeLeaf(relation.foreignKey, 'Present'),
+      const foreignKeySchema = through.schema.fields[relation.foreignKey];
+      let throughTree: ConditionTree = new ConditionTreeLeaf(
+        relation.originKey,
+        'Equal',
+        originValue,
       );
+
+      // Handle null foreign key case only when the datasource supports it.
+      if (foreignKeySchema.type === 'Column' && foreignKeySchema.filterOperators.has('Present')) {
```
Scra3, line 200: Instead to use 'Present', can we replace It by the 'NotEqual' operator ? If it is possible, we can apply the condition for all the cases. And removed the `filter` that removes the 'null' values.
`new ConditionTreeLeaf(relation.foreignKey, 'NotEqual', null),`

romain-gilliotte: If `Present` is not supported, we know that `NotEqual` is not supported either.
They always go together (thanks to the `OperatorReplacement` operator)

docs/agent-customization/relationships/computed-fks.md
romain-gilliotte: I added text on top to specify that all three steps are required
It seemed redundant

packages/agent/test/routes/modification/delete-related.test.ts
romain-gilliotte: because the code changed :)
look at the CollectionUrl in the toolkit

Scra3 (approved):

PR #660, opened 2023-03-28
fix: add options to cast uuid to string when the format is not supported

Scra3:
### General

### Security

### DIFF
--- a/docs/datasources/provided/sequelize.md
+++ b/docs/datasources/provided/sequelize.md
@@ -28,3 +28,15 @@ User.init(
 // Create agent and import collections from sequelize
 const agent = createAgent(options).addDataSource(createSequelizeDataSource(sequelize));
 ```
+
+# Disable UUID type checking
+
+[Postgres UUID type support non-standard UUIDs](https://www.postgresql.org/docs/current/datatype-uuid.html) but forestadmin-agent does not support it.
+To bypass this limitation, you can disable the UUID type checking by giving the `castUuidToString` option to the data source.
+
+```javascript
+/* ... */
+const agent = createAgent(options).addDataSource(
+  createSequelizeDataSource(sequelize, { castUuidToString: true }),
+);
+```
--- a/docs/datasources/provided/sequelize.md
+++ b/docs/datasources/provided/sequelize.md
@@ -31,9 +31,11 @@ const agent = createAgent(options).addDataSource(createSequelizeDataSource(seque
 
 # Disable UUID type checking
 
-[Postgres UUID type support non-standard UUIDs](https://www.postgresql.org/docs/current/datatype-uuid.html) but forestadmin-agent does not support it.
-One way to overcome this restriction is to utilize the `castUuidToString` option for the data source, which disables UUID type verification and transforms UUIDs into strings.
-Despite this conversion, your database will continue to store UUIDs rather than strings, so there is no need for concern.
+Postgres UUID columns accept [non-standard UUIDs](https://www.postgresql.org/docs/current/datatype-uuid.html) which don't pass Forest Admin validation rules.
+
+To overcome this restriction set the `castUuidToString` option to `true`. It will tell your agent to handle all `UUIDs` coming from Postgres as simple `strings`.
+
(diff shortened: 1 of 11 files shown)


packages/datasource-sequelize/src/utils/model-to-collection-schema-converter.ts
```diff
@@ -119,26 +125,50 @@ export default class ModelToCollectionSchemaConverter {
...
+
+  private static convertEnumAttribute(
+    columnType: ColumnType,
+    column: ColumnSchema,
+    attribute: ModelAttributeColumnOptions<Model>,
+  ) {
     if (columnType === 'Enum') {
       column.enumValues = [...attribute.values];
-    } else if (Array.isArray(columnType) && columnType.length === 1 && columnType[0] === 'Enum') {
+    } else if (ModelToCollectionSchemaConverter.isArrayOf(columnType, 'Enum')) {
```
romain-gilliotte, line 141: the others functions in this class use `this.` which is valid

packages/datasource-sequelize/src/utils/model-to-collection-schema-converter.ts
```diff
@@ -119,26 +125,50 @@ export default class ModelToCollectionSchemaConverter {
...
       const arrayType = attribute.type as DataTypes.ArrayDataType<DataTypes.EnumDataType<string>>;
       column.enumValues = [...arrayType.options.type.values];
     }
+  }
 
-    return column;
+  private static convertUuidAttribute(columnType: ColumnType, column: ColumnSchema) {
+    if (columnType === 'Uuid') {
+      column.columnType = 'String';
+      column.filterOperators = TypeConverter.operatorsForColumnType('String');
```
romain-gilliotte, line 150: Did you try to perform a query with a filter which is specific to strings? (like `startwith` or `contains`)

I think it will crash

packages/datasource-sequelize/src/utils/model-to-collection-schema-converter.ts
```diff
@@ -21,6 +22,8 @@ import {
 
 import TypeConverter from './type-converter';
 
+type Options = { castUuidToString: boolean };
```
romain-gilliotte, line 25: you can create and export the options type like in mongoose for customers to use.

This would avoid the duplication

docs/datasources/provided/sequelize.md
```diff
@@ -28,3 +28,16 @@ User.init(
 // Create agent and import collections from sequelize
 const agent = createAgent(options).addDataSource(createSequelizeDataSource(sequelize));
 ```
+
+# Disable UUID type checking
+
+[Postgres UUID type support non-standard UUIDs](https://www.postgresql.org/docs/current/datatype-uuid.html) but forestadmin-agent does not support it.
+One way to overcome this restriction is to utilize the `castUuidToString` option for the data source, which disables UUID type verification and transforms UUIDs into strings.
+Despite this conversion, your database will continue to store UUIDs rather than strings, so there is no need for concern.
```
romain-gilliotte, line 36: The section is a bit dense on gitbook (4 lines in a row).
image

How about something like this?


```suggestion
Postgres UUID columns accept [non-standard UUIDs](https://www.postgresql.org/docs/current/datatype-uuid.html) which don't pass Forest Admin validation rules.

To overcome this restriction set the `castUuidToString` option to `true`. It will tell your agent to handle all `UUIDs` coming from Postgres as simple `strings`.

This won't change how data is saved into your database.
```

docs/datasources/provided/sql.md
```diff
@@ -99,3 +99,15 @@ const agent = createAgent(options).addDataSource(createSqlDataSource({
 ```
 
 Note that under the hood, the data source uses [Sequelize](https://sequelize.org/) to connect to the database. So, you can pass any option that is supported by Sequelize.
+
```
romain-gilliotte, line 102: For the SQL datasource, maybe we should document this with the other options?

romain-gilliotte (changes requested):

PR #662, opened 2023-03-29
fix: handle standalone server errors when agent is starting

Scra3:
### General

### Security

### DIFF
--- a/packages/agent/src/framework-mounter.ts
+++ b/packages/agent/src/framework-mounter.ts
@@ -8,10 +8,10 @@ import path from 'path';
 import { HttpCallback } from './types';
 
 export default class FrameworkMounter {
-  private onStart: ((router: Router) => Promise<void>)[] = [];
-  private onStop: (() => Promise<void>)[] = [];
-  private prefix: string;
-  private logger: Logger;
+  private readonly onStart: ((router: Router) => Promise<void>)[] = [];
+  private readonly onStop: (() => Promise<void>)[] = [];
+  private readonly prefix: string;
+  private readonly logger: Logger;
 
   /** Compute the prefix that the main router should be mounted at in the client's application */
   private get completeMountPrefix(): string {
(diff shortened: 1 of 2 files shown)


packages/agent/test/framework-mounter.test.ts
```diff
@@ -19,34 +19,59 @@ const router = new Router().get('/', async ctx => {
...
+
+        try {
+          mounter.mountOnStandaloneServer(9997, 'localhost');
+          await mounter.start(router);
+          // throw error on start by mount two servers on the same port
+          mounter2.mountOnStandaloneServer(9997, 'localhost');
+          await mounter2.start(router);
+        } catch (e) {
+          // eslint-disable-next-line jest/no-conditional-expect
+          expect(e.message).toStrictEqual('listen EADDRINUSE: address already in use ::1:9997');
```
romain-gilliotte, line 68: Very likely that you are using node 18 locally

IIRC, on older node versions the message will be `listen EADDRINUSE: address already in use 127.0.0.1:9997` and the test will fail.

packages/agent/src/framework-mounter.ts
```diff
@@ -40,12 +40,17 @@ export default class FrameworkMounter {
...
+      server.listen(chosenPort, host, () => {
+        this.logger(
+          'Info',
+          `Successfully mounted on Standalone server (http://${host ?? '0.0.0.0'}:${port})`,
+        );
+        resolve();
+      });
+      server.on('error', reject);
+    });
+    this.onStart.push(async () => isListeningPromise);
```
romain-gilliotte, line 53: Useless `async`.

packages/agent/src/framework-mounter.ts
```diff
@@ -40,12 +40,17 @@ export default class FrameworkMounter {
   mountOnStandaloneServer(port?: number, host?: string): this {
     const chosenPort = port || Number(process.env.PORT) || 3351;
     const server = createServer(this.getConnectCallback(true));
-    server.listen(chosenPort, host);
-
-    this.logger(
-      'Info',
-      `Successfully mounted on Standalone server (http://${host ?? '0.0.0.0'}:${port})`,
-    );
+    const isListeningPromise = new Promise<void>((resolve, reject) => {
```
romain-gilliotte, line 43: Creating the promise in the body of the function will starting listening right away, and not once the function that was pushed on the onStart array is run.

Is that intentional?
I fear that is can cause a race-condition

romain-gilliotte (changes requested):

PR #671, opened 2023-04-18
feat(webhook-custom-actions): use webhook nocode smart actions configured in the frontend

ghusse:
Linked to CU-860qgzbzg

### General

### Security


packages/forestadmin-client/src/model-customizations/types.ts
```diff
@@ -0,0 +1,36 @@
...
+  global = 'global',
+  single = 'single',
+  bulk = 'bulk',
+}
+
+export enum ActionType {
+  webhook = 'webhook',
+}
+
+export type ModelCustomization<TConfiguration> = {
```
Thenkei, line 15: no default value is provided (use unknown)

packages/forestadmin-client/src/model-customizations/types.ts
```diff
@@ -0,0 +1,36 @@
+export enum ModelCustomizationType {
+  action = 'action',
+}
+
+export enum ActionScope {
```
Thenkei, line 5: we don't use enums (save a lots of imports)

packages/agent/src/services/model-customizations/webhook-executor.ts
```diff
@@ -0,0 +1,59 @@
...
+    default:
+      throw new Error(`Unknown scope: ${action.configuration.scope}`);
+  }
+}
+
+async function executeWebhook<S extends TSchema = TSchema>(
+  action: ModelCustomization<WebhookActionConfiguration>,
+  context: ActionContext<S, Extract<keyof S, string>>,
+) {
+  const ids = action.configuration.scope === 'global' ? [] : await context.getRecordIds();
```
Thenkei, line 48: take care this is wrong.

As we were asked to implement it a global action is the same than bulk + select all.
That means that when running the action on a segment, only the segment is considered.

If we change those semantics, the documentation is wrong

packages/agent/src/services/model-customizations/webhook-executor.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  const body = generateBody(action, ids);
+
+  await superagent.post(action.configuration.url).send(body);
+}
+
+export default function createWebhookExecutor<S extends TSchema = TSchema>(
```
Thenkei, line 55: I suppose you do this so that you can add the action directly.
We tend to avoid indirection

Something in the lines of
```
for (const action of actions)
  addAction(ctx => executeWebhook(action, ctx))
```

packages/agent/src/services/model-customizations/webhook-executor.ts
```diff
@@ -0,0 +1,59 @@
+import { ActionContext, TSchema } from '@forestadmin/datasource-customizer';
+import { ModelCustomization, WebhookActionConfiguration } from '@forestadmin/forestadmin-client';
+import superagent from 'superagent';
+
+function generateBody(
+  action: ModelCustomization<WebhookActionConfiguration>,
+  ids: (string | number)[],
+) {
+  switch (action.configuration.scope) {
+    case 'global':
```
Thenkei, line 10: see comment bellow for global (what happens in a segment?)

packages/agent/src/services/model-customizations/webhook-executor.ts
```diff
@@ -0,0 +1,59 @@
...
+    case 'bulk':
+      return {
+        action: {
+          name: action.name,
+          scope: action.configuration.scope,
+        },
+        records: ids.map(id => ({ id })),
+      };
+    case 'single':
+      if (ids.length !== 1) {
```
Thenkei, line 26: useless check

packages/agent/src/services/model-customizations/webhook-executor.ts
```diff
@@ -0,0 +1,59 @@
...
+        throw new Error('Single actions can only be used with one selected record');
+      }
+
+      return {
+        action: {
+          name: action.name,
+          scope: action.configuration.scope,
+        },
+        record: {
+          id: ids[0],
```
Thenkei, line 36: the primary key is not always named id

packages/agent/src/services/model-customizations/webhook-executor.ts
Thenkei, line 48: IIRC getRecordIds will throw an error on collection using a composite primary key
I think you need to handle those cases as well

packages/agent/src/services/model-customizations/action-customization.ts
```diff
@@ -0,0 +1,62 @@
...
+  ForestAdminClient,
+  ModelCustomization,
+  ModelCustomizationType,
+  WebhookActionConfiguration,
+} from '@forestadmin/forestadmin-client';
+
+import createWebhookExecutor from './webhook-executor';
+import { AgentOptionsWithDefaults } from '../../types';
+
+function translateScope(scope: `${ConfigurationScope}`): ActionScope {
```
Thenkei, line 16: why not use the ActionScope type in the first place?

packages/forestadmin-client/src/model-customizations/model-customization-with-cache.ts
```diff
@@ -0,0 +1,21 @@
+import { ModelCustomization, ModelCustomizationService } from './types';
+import { ForestAdminClientOptionsWithDefaults } from '../types';
+import ServerUtils from '../utils/server';
+
+export default class ModelCustomizationWithCacheService implements ModelCustomizationService {
+  private configuration: Promise<ModelCustomization<unknown>[]>;
+
+  constructor(private readonly options: ForestAdminClientOptionsWithDefaults) {}
+
+  async getConfiguration(): Promise<ModelCustomization<unknown>[]> {
```
Thenkei, line 10: useless cache pattern
the method is called only once

packages/agent/src/services/model-customizations/action-customization.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.client = agentOptions.forestAdminClient;
+  }
+
+  public async addWebhookActions(customizer: DataSourceCustomizer<S>) {
+    const actions = await this.getActions<WebhookActionConfiguration>(ActionType.webhook);
+
+    actions.forEach(action => {
+      const collection = customizer.getCollection(action.modelName as Extract<keyof S, string>);
+
+      collection.addAction(action.name, {
```
Thenkei, line 44: This is using the same customizer than the one that is used by the customer to code.
This means that the customer can **code** smart actions that can depend on smart actions that were defined in the frontend.

Not a big deal for smart actions (will just break those actions), but will be for fields (will prevent the server from starting).

```
// we may want to prevent the customer from being able to do that
collection.addAction('myaction-in-code', {
  execute: (ctx) => {
    return ctx.collection.execute('no-code-action-from-backend');
  }
})
````

IMO, we can not care for now.
When we develop the smart fields we'll stack another customizer on top of the current one for no code stuff

**Review by Thenkei — CHANGES_REQUESTED** [review]
Romain made a few comments. Let's talk about it if you want.

File: packages/forestadmin-client/src/model-customizations/types.ts (hunk 2 above)
Comment (ghusse) [reply packages/forestadmin-client/src/model-customizations/types.ts:5]: Enums are not used In other types : we only declare strings that can take the values of enums. From my point of view it combines the best of both worlds.

File: packages/agent/src/services/model-customizations/webhook-executor.ts (hunk 3 above)
Comment (ghusse) [reply packages/agent/src/services/model-customizations/webhook-executor.ts:48]: Are you sure?

File: packages/forestadmin-client/src/model-customizations/model-customization-with-cache.ts (hunk 9 above)
Comment (ghusse) [reply packages/forestadmin-client/src/model-customizations/model-customization-with-cache.ts:10]: For the moment yes, but we will introduce other customizations soon

File: packages/agent/src/services/model-customizations/webhook-executor.ts (hunk 3 above)
Comment (Thenkei) [reply packages/agent/src/services/model-customizations/webhook-executor.ts:48]: @romain-gilliotte seems pretty sure about that.

Comment (Thenkei) [reply packages/agent/src/services/model-customizations/webhook-executor.ts:48]: Let's run a test to re-ensure ourself ?

File: packages/forestadmin-client/src/model-customizations/types.ts (hunk 2 above)
Comment (ghusse) [reply packages/forestadmin-client/src/model-customizations/types.ts]: Done anyway

File: packages/agent/src/services/model-customizations/webhook-executor.ts (hunk 3 above)
Comment (ghusse) [reply packages/agent/src/services/model-customizations/webhook-executor.ts]: I changed it for now

File: packages/agent/src/services/model-customizations/webhook-executor.ts (hunk 4 above)
Comment (ghusse) [reply packages/agent/src/services/model-customizations/webhook-executor.ts]: I changed it, I still need to simplify the interface but the idea is here

File: packages/agent/src/services/model-customizations/webhook-executor.ts (hunk 7 above)
Comment (ghusse) [reply packages/agent/src/services/model-customizations/webhook-executor.ts]: OMG, you're right.

I fixed this in the latest changes

File: packages/agent/src/services/model-customizations/action-customization.ts (hunk 8 above)
Comment (ghusse) [reply packages/agent/src/services/model-customizations/action-customization.ts]: The API uses `single`, `global` and `bulk` everywhere, whereas the agent uses `Single`, `Global` and `Bulk` with a capital first letter :'(

In addition to that, `ActionScope` is declared in the `datasource-toolkit` which is not a dependency of `forestadmin-client`.

File: packages/forestadmin-client/src/model-customizations/model-customization-with-cache.ts (hunk 9 above)
Comment (ghusse) [reply packages/forestadmin-client/src/model-customizations/model-customization-with-cache.ts]: I removed this premature optimization

File: packages/agent/src/services/model-customizations/execute-webhook.ts [hunk 11]
```diff
@@ -0,0 +1,52 @@
...
+export default async function executeWebhook<S extends TSchema = TSchema>(
+  action: WebhookAction,
+  context: ActionContext<S>,
+) {
+  const primaryKeys = SchemaUtils.getPrimaryKeys(context.collection.schema) as TFieldName<
+    S,
+    Extract<keyof S, string>
+  >[];
+
+  const records = await context.getRecords(primaryKeys);
```
Thenkei, line 47: 👌 Super clean !

packages/forestadmin-client/src/model-customizations/model-customization-from-api.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (modelCustomization.type !== 'action') {
+    throw new Error('Only action customizations are supported for now.');
+  }
+
+  const configuration = modelCustomization.configuration as WebhookActionConfigurationApi;
+  const mappedConfiguration: WebhookActionConfiguration = {
+    ...configuration,
+    scope: configuration.scope
+      ? ((configuration.scope.slice(0, 1).toUpperCase() +
+          configuration.scope.slice(1)) as ActionScope)
```
Thenkei, line 24: 🐐

packages/datasource-customizer/src/context/relaxed-wrappers/collection.ts
```diff
@@ -34,6 +35,10 @@ export default class RelaxedCollection<
   private collection: Collection;
   private caller: Caller;
 
+  get schema(): CollectionSchema {
+    return this.collection.schema;
+  }
+
```
Thenkei, line 41: I would have preferred a higher API level getter.

Something like:
```
get getPrimaryKeys() {
  return SchemaUtils.getPrimaryKeys(this.collection.schema);
}
```

Thenkei: But I guess that your version offers more possibilities in the end.

packages/forestadmin-client/src/model-customizations/model-customization-from-api.ts
```diff
@@ -0,0 +1,46 @@
...
+function mapApiValues(
+  modelCustomization: ModelCustomization<WebhookActionConfigurationApi>,
+): ModelCustomization<WebhookActionConfiguration> {
+  if (modelCustomization.type !== 'action') {
+    throw new Error('Only action customizations are supported for now.');
+  }
+
+  const configuration = modelCustomization.configuration as WebhookActionConfigurationApi;
+  const mappedConfiguration: WebhookActionConfiguration = {
+    ...configuration,
```
Thenkei, line 21:
```suggestion
  const mappedConfiguration: WebhookActionConfiguration = {
    ...modelCustomization.configuration,
```

packages/forestadmin-client/src/model-customizations/model-customization-from-api.ts
```diff
@@ -0,0 +1,46 @@
...
+  return {
+    ...modelCustomization,
+    configuration: mappedConfiguration,
+  };
+}
+
+export default class ModelCustomizationFromApiService implements ModelCustomizationService {
+  constructor(private readonly options: ForestAdminClientOptionsWithDefaults) {}
+
+  async getConfiguration(): Promise<WebhookAction[]> {
```
Thenkei, line 37:
```suggestion
  async getConfiguration(): Promise<ModelCustomization[]> {
```

packages/forestadmin-client/src/model-customizations/model-customization-from-api.ts
```diff
@@ -0,0 +1,46 @@
...
+    const result = await ServerUtils.query<ModelCustomization<WebhookActionConfigurationApi>[]>(
```
Thenkei, line 38:
```suggestion
    const result = await ServerUtils.query<ModelCustomization[]>(
```

packages/forestadmin-client/src/model-customizations/model-customization-from-api.ts
```diff
@@ -0,0 +1,46 @@
...
+  WebhookAction,
+  WebhookActionConfiguration,
+  WebhookActionConfigurationApi,
+} from './types';
+import { ForestAdminClientOptionsWithDefaults } from '../types';
+import ServerUtils from '../utils/server';
+
+function mapApiValues(
+  modelCustomization: ModelCustomization<WebhookActionConfigurationApi>,
+): ModelCustomization<WebhookActionConfiguration> {
```
Thenkei, line 14:
```suggestion
function mapApiValues(modelCustomization: ModelCustomization): ModelCustomization {
```

packages/forestadmin-client/src/model-customizations/model-customization-from-api.ts
ghusse: I set this type on purpose, because today we only have one type that can be returned by this function.

packages/forestadmin-client/src/model-customizations/model-customization-from-api.ts
ghusse: This is also intentional here because we can only have one type here.

packages/forestadmin-client/src/model-customizations/model-customization-from-api.ts
ghusse: Same here

packages/forestadmin-client/src/model-customizations/model-customization-from-api.ts
Thenkei: Well that weird since the interface says `ModelCustomization[]`

Thenkei (approved):
- ✅&nbsp;Code review
  - ✖️&nbsp; Security concerns
- ✅&nbsp;Automatic tests
  - ✅&nbsp; Unit tests
  - ✅&nbsp; Integration tests
-✖️&nbsp;Manual tests
- ✅&nbsp;PR title
- ✅&nbsp;Destination branch
- ✅&nbsp;PR linked to the clickup task

PR #692, opened 2023-05-10
feat(events-subscription): allow to instantly refresh permissions when they change

Thenkei:
… 
### General

### Security

### DIFF
--- a/packages/_example/src/forest/agent.ts
+++ b/packages/_example/src/forest/agent.ts
@@ -26,7 +26,7 @@ export default function makeAgent() {
     envSecret: process.env.FOREST_ENV_SECRET,
     forestServerUrl: process.env.FOREST_SERVER_URL,
     isProduction: false,
-    loggerLevel: 'Debug',
+    loggerLevel: 'Info',
     typingsPath: 'src/forest/typings.ts',
   };
 
--- a/packages/_example/src/forest/agent.ts
+++ b/packages/_example/src/forest/agent.ts
@@ -26,7 +26,7 @@ export default function makeAgent() {
     envSecret: process.env.FOREST_ENV_SECRET,
     forestServerUrl: process.env.FOREST_SERVER_URL,
     isProduction: false,
-    loggerLevel: 'Info',
+    loggerLevel: 'Debug',
     typingsPath: 'src/forest/typings.ts',
   };
 
--- a/packages/agent/src/types.ts
+++ b/packages/agent/src/types.ts
@@ -15,6 +15,7 @@ export type AgentOptions = {
   schemaPath?: string;
   typingsPath?: string | null;
   typingsMaxDepth?: number;
+  useServerEvents?: boolean;
   permissionsCacheDurationInSeconds?: number;
   skipSchemaUpdate?: boolean;
   forestAdminClient?: ForestAdminClient;
--- a/packages/agent/src/types.ts
+++ b/packages/agent/src/types.ts
@@ -15,7 +15,7 @@ export type AgentOptions = {
   schemaPath?: string;
   typingsPath?: string | null;
   typingsMaxDepth?: number;
-  useServerEvents?: boolean;
+  instantCacheRefresh?: boolean;
   permissionsCacheDurationInSeconds?: number;
   skipSchemaUpdate?: boolean;
   forestAdminClient?: ForestAdminClient;
--- a/packages/agent/test/utils/http-driver-options.test.ts
+++ b/packages/agent/test/utils/http-driver-options.test.ts
@@ -16,7 +16,7 @@ describe('OptionsValidator', () => {
       expect(options).toHaveProperty('prefix', '');
       expect(options).toHaveProperty('schemaPath', '.forestadmin-schema.json');
       expect(options).toHaveProperty('typingsMaxDepth', 5);
-      expect(options).toHaveProperty('useServerEvents', true);
+      expect(options).toHaveProperty('instantCacheRefresh', true);
       expect(options).toHaveProperty('permissionsCacheDurationInSeconds', 31560000);
       expect(options).toHaveProperty('skipSchemaUpdate', false);
     });
@@ -80,7 +80,7 @@ describe('OptionsValidator', () => {
         const options = OptionsValidator.withDefaults({
           ...mandatoryOptions,
           logger: jest.fn(),
-          useServerEvents: false,
+          instantCacheRefresh: false,
           permissionsCacheDurationInSeconds: 1,
         });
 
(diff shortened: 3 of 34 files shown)


packages/_example/src/forest/agent.ts
```diff
@@ -26,7 +26,7 @@ export default function makeAgent() {
     envSecret: process.env.FOREST_ENV_SECRET,
     forestServerUrl: process.env.FOREST_SERVER_URL,
     isProduction: false,
-    loggerLevel: 'Info',
+    loggerLevel: 'Debug',
```
realSpok, line 29: Should this be kept ?

packages/forestadmin-client/src/forest-admin-client-with-cache.ts
```diff
@@ -66,6 +68,10 @@ export default class ForestAdminClientWithCache implements ForestAdminClient {
   }
 
   public markScopesAsUpdated(renderingId: number | string) {
-    this.renderingPermissionService.invalidateCache(renderingId);
+    if (!this.options.useServerEvents) this.renderingPermissionService.invalidateCache(renderingId);
+  }
+
+  public async subscribeServerEvents() {
```
realSpok, line 74: `subscribeToServerEvents` ?

packages/_example/src/forest/agent.ts
Thenkei: I forgot to remove it. I'm gonna clean it asap.

packages/agent/src/types.ts
```diff
@@ -15,6 +15,7 @@ export type AgentOptions = {
   schemaPath?: string;
   typingsPath?: string | null;
   typingsMaxDepth?: number;
+  useServerEvents?: boolean;
```
ghusse, line 18: If we want to provide an option here, I think we should name it after what it provides for the user, now how it works.

packages/agent/test/utils/http-driver-options.test.ts
```diff
@@ -92,11 +94,24 @@ describe('OptionsValidator', () => {
...
+
+      describe('when using Server Events (useServerEvents=true)', () => {
+        test('should set permissionsCacheDurationInSeconds to 1 year', () => {
+          const options = OptionsValidator.withDefaults({
+            ...mandatoryOptions,
+            useServerEvents: true,
+            permissionsCacheDurationInSeconds: 5 * 60,
+          });
+
+          expect(options).toHaveProperty('permissionsCacheDurationInSeconds', 31560000);
```
ghusse, line 112: I think that these 2 options should be handled separately. If a user sets the `permissionsCacheDurationInSeconds` then we should respect the value that is passed.

If it is not set, then we should use the default value (depending on the fact that `useServerEvents` is true or false)

packages/forestadmin-client/src/events-subscription/index.ts
```diff
@@ -0,0 +1,90 @@
+import EventSource from 'eventsource';
+
+import { RefreshEventsHandlerService, ServerEvent, ServerEventType } from './types';
+import { ForestAdminClientOptionsWithDefaults } from '../types';
+
+export default class EventsSubscriptionService {
+  constructor(
+    private options: ForestAdminClientOptionsWithDefaults,
```
ghusse, line 8:
```suggestion
    private readonly options: ForestAdminClientOptionsWithDefaults,
```

packages/forestadmin-client/src/events-subscription/index.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    const { renderingIds } = JSON.parse(event.data as unknown as string);
+    await this.refreshEventsHandlerService.refreshRenderings(renderingIds);
+  }
+
+  private onEventError(event: { type: string; status?: number; message?: string }) {
+    if (event.status === 502) {
+      this.options.logger(
+        'Debug',
+        'Server Event - Connection lost (ForestAdmin servers are restarting)',
```
ghusse, line 71: I don't think we can be affirmative about the fact that servers are restarting
```suggestion
        'Server Event - Connection lost',
```

packages/forestadmin-client/src/events-subscription/index.ts
```diff
@@ -0,0 +1,90 @@
...
+      headers: { 'forest-secret-key': this.options.envSecret },
+      https: {
+        rejectUnauthorized: process.env.NODE_TLS_REJECT_UNAUTHORIZED !== '0',
+      },
+    };
+    const url = new URL('/liana/v4/subscribe-to-events', this.options.forestServerUrl).toString();
+
+    const source = new EventSource(url, eventSourceConfig);
+
+    source.addEventListener('error', this.onEventError.bind(this));
```
ghusse, line 35: Should we close the eventSource here and try to start another one?

packages/forestadmin-client/src/events-subscription/native-refresh-events-handler-service.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.emit('RefreshCustomizations');
+  }
+
+  public refreshEverything() {
+    this.usersPermissionService.invalidateCache();
+    this.actionPermissionService.invalidateCache();
+    this.renderingPermissionService.invalidateAllCache();
+
+    // Emit RefreshCustomizations event
+    this.emit('RefreshCustomizations');
```
ghusse, line 43: Should it be inside this PR, as this is not handled yet?

packages/forestadmin-client/src/events-subscription/index.ts
Thenkei: It's already handled. The event source tries to reconnect every second. I can look for the exact option to make it clear.

packages/forestadmin-client/src/events-subscription/index.ts
Thenkei: Oh ! Well seen. I did use a trick when doing the switch strategy but I've reverted all the code except this one.

packages/agent/test/utils/http-driver-options.test.ts
Thenkei: I'll change it but it will be weird from a user perceptive. I when my cache to be refresh fast and for sure `useServerEvents` doesn't help that much at the moment.

packages/agent/src/types.ts
Thenkei: @ghusse any idea?

Thenkei: The court has decided it will be named `instantCacheRefresh`. 💪

packages/forestadmin-client/src/events-subscription/index.ts
ghusse: Ok, thanks. Is there a configuration option regarding the delay?

packages/forestadmin-client/src/events-subscription/native-refresh-events-handler-service.ts
Thenkei: I implementes everything than remove some part. Do you think it's an issue since no one is listening for `RefreshCustomizations`? (The basement is here, just the additionnal behavior has not been implemented in this PR since the code was on alpha branch)

packages/forestadmin-client/src/events-subscription/index.ts
Thenkei: I never responded sorry the delay is `1000ms` and not configurable without some magical access to the eventSource object.

ghusse (approved):

PR #709, opened 2023-05-31
chore(forestadmin-client): bump lru-cache to latest version

Thenkei:
### General

### Security

### DIFF
--- a/packages/forestadmin-client/src/permissions/rendering-permission.ts
+++ b/packages/forestadmin-client/src/permissions/rendering-permission.ts
@@ -1,4 +1,4 @@
-import LruCache from 'lru-cache';
+import { LRUCache } from 'lru-cache';
 
 import { hashChartRequest, hashServerCharts } from './hash-chart';
 import isSegmentQueryAllowed from './is-segment-query-authorized';
@@ -23,17 +23,17 @@ export type RenderingPermission = {
 };
 
 export default class RenderingPermissionService {
-  private readonly permissionsByRendering: LruCache<string, RenderingPermission>;
+  private readonly permissionsByRendering: LRUCache<string, RenderingPermission>;
 
   constructor(
     private readonly options: ForestAdminClientOptionsWithDefaults,
     private readonly userPermissions: UserPermissionService,
     private readonly forestAdminServerInterface: ForestAdminServerInterface,
   ) {
-    this.permissionsByRendering = new LruCache({
+    this.permissionsByRendering = new LRUCache({
       max: 256,
       ttl: this.options.permissionsCacheDurationInSeconds * 1000,
-      fetchMethod: renderingId => this.loadPermissions(Number(renderingId)),
+      fetchMethod: async renderingId => this.loadPermissions(Number(renderingId)),
     });
   }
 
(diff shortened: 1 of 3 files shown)


packages/forestadmin-client/src/permissions/rendering-permission.ts
```diff
@@ -125,9 +125,9 @@ export default class RenderingPermissionService {
     segmentQuery: string;
     allowRetry: boolean;
   }): Promise<boolean> {
-    const permissions: RenderingPermission = await this.permissionsByRendering.fetch(
+    const permissions = (await this.permissionsByRendering.fetch(
       `${renderingId}`,
-    );
+    )) as RenderingPermission;
```
ghusse, line 130: Why do you need this type cast here?

Thenkei: Because the fetch method returns `Promise<void | ValueType>`

ghusse: Ok, but it indicates that the function can return nothing if the value is not present in the cache, which is the case also with this line.

Thenkei: Well, it is a bit complicated. 
In our case, the value is fetched (using the fetch method defined for the LRU) and always available (even if it's undefined).

I don't know how to handle correctly those void that won't happen

ghusse: void is undefined.

To be fair, I think that the declaraiton of this function should be `undefined | T` instead of `void | T`.

Thenkei: I agree. I'm gonna submit a PR on LRU cache.

isaacs: The issue is a little bit complicated.

It is allowed for the `fetchMethod` provided to LRUCache to return _actually nothing_ in the event that the fetch is aborted.  An abort signal can be fired if the key is deleted or overwritten during the fetch (which can happen under load if the cache evicts old members which are in the process of being fetched), or if the caller explicitly aborts the fetch by providing an AbortSignal of their own.

It's rather annoying (imo) that TS treats `void` so differently from `undefined` when referring to a value in this way, since it is literally `undefined` in that case.  It is definitely _not_ type safe to assume that the value will always be defined, though.

I'm open to updating LRUCache to make that function `V | undefined`, but it's a SemVer major change, which can be a pain.  (At minimum, you'd have to update LRUCache.)  In the meantime, you can make TS be a bit less snippy about it by doing:

```ts
(permissions === undefined ? undefined : permissions)?.collections?.[collectionName];
```

Thenkei: https://github.com/isaacs/node-lru-cache/issues/302
https://github.com/isaacs/node-lru-cache/pull/303

The owner of LRU Cache is not willing to make any changes that soon. He proposed a patch we could use it with https://www.npmjs.com/package/patch-package.

Thenkei (commented):
🚀  Nice addition in `lru-cache` v10.0.0.

DayTF (approved):

PR #723, opened 2023-06-09
chore: bump all versions to latest minor

romain-gilliotte:
### General

### Security

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "prettier-eslint": "^15.0.1",
     "semantic-release-slack-bot": "^3.5.3",
     "ts-jest": "^29.0.4",
+    "ts-node": "^10.9.1",
     "typedoc": "^0.23.24",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -29,6 +29,7 @@
     "prettier-eslint": "^15.0.1",
     "semantic-release-slack-bot": "^3.5.3",
     "ts-jest": "^29.0.4",
+    "ts-node": "^10.9.1",
```
arnaudbesnier, line 32: What is this addition?

romain-gilliotte: image

Jest requires ts-node to parse typescript config files.
This worked before upgrading, most likely because some other package depended on ts-node (so it was already installed), but no longer does after the upgrade

arnaudbesnier: Ok thanks.
I am sure that a pro-active comment to explain this specific change would have speed up the review process.

arnaudbesnier (approved):

PR #724, opened 2023-06-09
refactor(datasource-sql): improve the uri management

Scra3:
### General

### Security


packages/datasource-sql/src/connection/errors.ts
```diff
@@ -1,18 +1,15 @@
...
-  }
+  if (uriObject.password) uriObject.password = '**sanitizedPassword**';
 
   return uriObject.toString();
 }
 
-export type SourceError = 'Proxy' | 'Database';
-
 class BaseError extends Error {
   public source: SourceError;
```
romain-gilliotte, line 14: we don't use the `public` keyword in this repo

packages/datasource-sql/src/connection/errors.ts
```diff
@@ -27,13 +24,17 @@ class BaseError extends Error {
...
+    if (options && options.uriAsString) {
+      const sanitizedUri = sanitizeUri(options.uriAsString);
       super(`Unable to connect to the given uri: ${sanitizedUri}.`, message);
       this.uri = sanitizedUri;
     } else {
-      super(`Unable to connect to the given uri.`, message);
+      super(`Unable to connect to the given uri/options.`, message);
     }
 
     this.name = this.constructor.name;
```
romain-gilliotte, line 40: you can move this up

packages/datasource-sql/src/connection/errors.ts
```diff
@@ -42,11 +43,12 @@ export class DatabaseConnectError extends BaseError {
...
 export class ProxyConnectError extends BaseError {
-  constructor(message: string, proxyConfig?: ConnectionOptionsObj['proxySocks']) {
+  constructor(message: string, options?: ConnectionOptionsWrapper) {
     const defaultMessage = 'Your proxy has encountered an error.';
 
-    if (proxyConfig) {
-      const sanitizedUri = ProxyConnectError.buildSanitizedUriFromConfig(proxyConfig);
+    if (options && options.uriAsString) {
+      // remove tcp protocol because its not added by the user
+      const sanitizedUri = sanitizeUri(options.proxyUriAsString).replace('tcp://', '');
```
romain-gilliotte, line 51: not really an uri anymore once you remove the protocol :)
(variable name)

packages/datasource-sql/src/connection/handle-errors.ts
```diff
@@ -28,7 +25,7 @@ export default function handleErrors(
   options?: ConnectionOptionsObj,
   proxy?: ReverseProxy,
```
romain-gilliotte, line 26: You are passing the proxy as third argument only to access `proxy.wrapperOption`, which actually is the second argument

packages/datasource-sql/src/connection/handle-errors.ts
```diff
@@ -37,7 +34,10 @@ export default function handleErrors(
       (_, m1, m2) => `${m1} ${m2.toLowerCase()}`,
     );
 
-    throw new DatabaseConnectError(`${nameWithSpaces}: ${error.message}`, options?.uri);
+    throw new DatabaseConnectError(
+      `${nameWithSpaces}: ${error.message}`,
+      options ? new ConnectionOptionsWrapper(options) : null,
```
romain-gilliotte, line 39: If using a wrapper try to use it only once at the top calling function.
When i searching for `new ConnectionOptionsWrapper` I found it in four different files

packages/datasource-sql/src/connection/index.ts
```diff
@@ -16,30 +17,24 @@ export default async function connect(
...
-    const schema = opts.schema ?? getSchema(uri);
     const logging = logger ? getLogger(logger) : false;
+    const wrapper = new ConnectionOptionsWrapper(options);
+    const { uriIfValidOrNull, schemaFromUriOrOptions: schema } = wrapper;
 
     opts.dialectOptions = {
       ...(opts.dialectOptions ?? {}),
-      ...getSslConfiguration(opts.dialect, sslMode, servername, logger),
+      ...wrapper.computeSslConfiguration(opts.dialect, sslMode, uriIfValidOrNull, logger),
     };
```
romain-gilliotte, line 38: Taking data out of an object, and feeding it back in
```
const obj = new machin()
obj.method(obj.truc);
```

packages/datasource-sql/src/connection/preprocess.ts
```diff
@@ -1,11 +1,11 @@
 import type { ConnectionOptions, ConnectionOptionsObj, SslMode } from '../types';
 
 import connect from './index';
-import { getDialect, getUri } from './utils';
+import ConnectionOptionsWrapper from '../connection-options-wrapper';
 
-async function resolveSslMode(uriOrOptions: ConnectionOptionsObj): Promise<SslMode> {
+async function resolveSslMode(wrapper: ConnectionOptionsWrapper): Promise<SslMode> {
```
romain-gilliotte, line 6: Since you switched to using OOP for this module, and already moved the `computeSslConfiguration` code, all this file can move as well.

packages/datasource-sql/src/connection/reverse-proxy.ts
```diff
@@ -38,10 +36,10 @@ export default class ReverseProxy {
 
   get connectionOptions(): ConnectionOptionsObj {
```
romain-gilliotte, line 37: I know this was my code but in hindsight giving so much knowledge about how sequelize options are structured to a class whose role is to spawn a proxy seems wrong.

We just just give host+port and have a `start` and `stop` method here.

I did it because I was lazy, and at the time the `connectionOptions` were a plain object.
But now that `connectionOptions` are a class, we end up with strong coupling between those.

(`OptionWrapper` knows about the `reverseproxy` and the `reverseproxy` knows about the `OptionWrapper`)

packages/datasource-sql/src/connection/reverse-proxy.ts
```diff
@@ -57,63 +55,37 @@ export default class ReverseProxy {
...
+      if (!socks5Proxy?.socket.closed) socks5Proxy?.socket.destroy();
     });
 
     try {
       socks5Proxy = await SocksClient.createConnection({
-        proxy: { ...this.destination.proxySocks, type: 5 },
+        proxy: { ...this.wrapperOptions.options.proxySocks, type: 5 },
         command: 'connect',
-        destination: { host: this.destinationHost, port: this.destinationPort },
+        destination: {
```
romain-gilliotte, line 75: This is good.
You are starting the break the coupling between both classes

(the options should produce the destination host, not the reverse proxy)

packages/datasource-sql/src/connection/utils.ts
```diff
@@ -1,119 +1,5 @@
...
-  return dialect as Dialect;
-}
-
-export function getSchema(uri: string): string {
-  return uri ? new URL(uri).searchParams.get('schema') : null;
-}
-
-export function getLogger(logger: Logger): (sql: string) => void {
+export default function getLogger(logger: Logger): (sql: string) => void {
   return (sql: string) => logger?.('Debug', sql.substring(sql.indexOf(':') + 2));
```
romain-gilliotte, line 4: If you moved all the generation of the sequelize options, you may as well move this last one 😁

PR #741, opened 2023-06-21
fix: add a timeout on superagent call

SkyHuss:
### General

### Security

### DIFF
--- a/packages/forestadmin-client/src/utils/server.ts
+++ b/packages/forestadmin-client/src/utils/server.ts
@@ -12,27 +12,24 @@ export default class ServerUtils {
     path: string,
     headers: Record<string, string> = {},
     body?: string | object,
+    maxTimeAllowed = 10000, // Set a default value if not provided
   ): Promise<T> {
     try {
-      const maxTimeAllowed = 10000;
       const url = new URL(path, options.forestServerUrl).toString();
-      
-      const request = superagent[method](url).timeout(maxTimeAllowed)
+
+      const request = superagent[method](url).timeout(maxTimeAllowed);
 
       request.set('forest-secret-key', options.envSecret);
       if (headers) request.set(headers);
 
       const response = await request.send(body);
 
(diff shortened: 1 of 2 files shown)


packages/forestadmin-client/src/utils/server.ts
```diff
@@ -19,7 +19,13 @@ export default class ServerUtils {
...
       if (headers) request.set(headers);
 
-      const response = await request.send(body);
+      const timeoutPromise = new Promise(resolve => {
+        setTimeout(resolve, 10000);
+      }).then(() => Promise.reject(new Error('The request to ForestAdmin server has timeout')));
+
+      const responsePromise = request.send(body);
+
+      const response = await Promise.race([responsePromise, timeoutPromise]);
```
jeffladiray, line 28: `superagent` as a `timeout()` method we should use, instead of this race condition (See [this documentation](https://ladjs.github.io/superagent/#timeouts))

packages/forestadmin-client/test/utils/server.test.ts
```diff
@@ -91,6 +91,32 @@ describe('ServerUtils', () => {
     await expect(ServerUtils.query(options, 'get', '/endpoint')).rejects.toThrow(unknownError);
   });
 
+  it('should timeout if the server take more than 10s to respond', async () => {
+    const message = 'The request to ForestAdmin server has timeout';
+    nock(options.forestServerUrl, {
+      reqheaders: { 'x-foo': 'bar', 'forest-secret-key': options.envSecret },
+    })
+      .get('/endpoint')
+      .delay(11000)
```
jeffladiray, line 100: We shouldn't write test that need ~10s ish to resolve.

`jest` has a whole lot of timer manipulation that could be used, or the `timeout` method could be mock to avoid wasting test time in CI

jeffladiray (changes requested):

jeffladiray (approved):

PR #743, opened 2023-06-21
chore(ci): improve code coverage output

romain-gilliotte:
This also reduce by quite a lot the number of files that jest needs to scan + the time it takes to generate the coverage

Before: 
image

After:
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -117,7 +117,7 @@ jobs:
           key: ${{ runner.os }}-build-${{ github.sha }}
           fail-on-cache-miss: true
       - name: test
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -142,17 +142,17 @@ jobs:
           CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
         with:
           coverageLocations: |
```
romain-gilliotte, line 144: using clover instead of lcov because the files paths are absolute (they are relative in lcov)

jeffladiray (approved):

PR #747, opened 2023-06-23
feat(agent): no-code update record should handle errors correctly

Thenkei:
### General

### Security

### DIFF
--- a/packages/agent/src/services/model-customizations/actions/update-record/execute-update-record.ts
+++ b/packages/agent/src/services/model-customizations/actions/update-record/execute-update-record.ts
@@ -2,6 +2,10 @@ import { ActionContext, TSchema } from '@forestadmin/datasource-customizer';
 import { ActionResult, BusinessError } from '@forestadmin/datasource-toolkit';
 import { UpdateRecordAction } from '@forestadmin/forestadmin-client';
 
+const genericErrorMessage = actionName =>
+  `The no-code action <strong>${actionName}</strong> cannot be triggered due to a` +
+  ' misconfiguration. Please contact your administrator.';
+
 export default async function executeUpdateRecord<S extends TSchema = TSchema>(
   action: UpdateRecordAction,
   context: ActionContext<S>,
@@ -13,8 +17,8 @@ export default async function executeUpdateRecord<S extends TSchema = TSchema>(
       },
     } = action;
 
-    const fields = fieldsToUpdate.reduce((acc, fieldToUpdate) => {
-      return { ...acc, [fieldToUpdate.fieldName]: fieldToUpdate.value };
+    const fields = fieldsToUpdate.reduce((fieldsAccumulator, fieldToUpdate) => {
+      return { ...fieldsAccumulator, [fieldToUpdate.fieldName]: fieldToUpdate.value };
     }, {});
 
     await context.collection.update(context.filter, fields);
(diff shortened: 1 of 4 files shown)


realSpok:
Nice !
image

packages/agent/src/services/model-customizations/actions/update-record/execute-update-record.ts
```diff
@@ -0,0 +1,38 @@
...
+      return { ...acc, [fieldToUpdate.fieldName]: fieldToUpdate.value };
+    }, {});
+
+    await context.collection.update(context.filter, fields);
+  } catch (error) {
+    if (error instanceof BusinessError) {
+      return {
+        type: 'Error',
+        message:
+          `The no-code action ${action.name} cannot be triggered due to misconfiguration.` +
```
realSpok, line 26:
```suggestion
          `The no-code action ${action.name} cannot be triggered due to a misconfiguration.`
```
It is up for debate but we could remove the 'no-code'. From the user's perspective there are no difference between actions.

packages/agent/src/services/model-customizations/actions/update-record/execute-update-record.ts
```diff
@@ -0,0 +1,38 @@
...
+          ` Please contact your administrator.\n(${error.name}: ${error.message})`,
```
realSpok, line 27: The '\n' line break doesn't seem to be working. Do we need a <br/> ? or something else ?

packages/agent/src/services/model-customizations/actions/update-record/execute-update-record.ts
```diff
@@ -0,0 +1,38 @@
...
+      };
+    }
+
+    return {
+      type: 'Error',
+      message:
+        `The no-code action ${action.name} cannot be triggered due to misconfiguration.` +
```
realSpok, line 34: Same as above

realSpok: We could maybe factorize those 2 messages into a function

packages/agent/src/services/model-customizations/actions/update-record/execute-update-record.ts
```diff
@@ -0,0 +1,38 @@
...
+  context: ActionContext<S>,
+): Promise<ActionResult> {
+  try {
+    const {
+      configuration: {
+        configuration: { fields: fieldsToUpdate },
+      },
+    } = action;
+
+    const fields = fieldsToUpdate.reduce((acc, fieldToUpdate) => {
```
realSpok, line 16: I am not sure what `acc` stands for …

realSpok (changes requested):
- ❗  Code review
  - ✅  Use pure functions when possible
  - ✅  Performance concerns
  - ✅  Security concerns
- ✅ Automatic tests
  - ✅  Unit tests
  - ⤬  Integration tests
- ✅ Manual tests
  -  ✅ Test the functionality
  -  ✅ Test error cases
- ✅  PR title:
- ✅  Destination branch

packages/agent/src/services/model-customizations/actions/update-record/execute-update-record.ts
realSpok: Even if the 'no-code' is a marketing term, I would say it is for the admins, not the users

packages/agent/src/services/model-customizations/actions/update-record/execute-update-record.ts
Thenkei: accumulator 😥

realSpok: ok. I guess I am also not very comfortable with reduce :)
But I get it now. You can leave it like that or write it as `accumulator`, as you prefer

packages/agent/src/services/model-customizations/actions/update-record/execute-update-record.ts
```diff
@@ -0,0 +1,38 @@
...
+): Promise<ActionResult> {
+  try {
+    const {
+      configuration: {
+        configuration: { fields: fieldsToUpdate },
+      },
+    } = action;
+
+    const fields = fieldsToUpdate.reduce((fieldsAccumulator, fieldToUpdate) => {
+      return { ...fieldsAccumulator, [fieldToUpdate.fieldName]: fieldToUpdate.value };
```
realSpok, line 21: `fieldsAccumulator` -> very straigthforward !

PR #752, opened 2023-06-28
feat(config): configuration now allows to differenciate bucket path and database path

SteveBunlon:
This PR has been done to allow customers to differentiate the path in their bucket, and the path stored in the database. 

The new storeAt attribute allows to configure two different logics to compute AWSPath and databasePath. 

Knowing the two paths can now be different, we also need a logic to reconcile the database stored value with the actual path  in the bucket. 

Customers also found really restrictive to be given only the record id in the storeAt conf, I've changed that so it is now asynchronous and provides the context as-well. 

### General

### Security

### DIFF
--- a/packages/plugin-aws-s3/src/types.ts
+++ b/packages/plugin-aws-s3/src/types.ts
@@ -3,11 +3,11 @@ import type {
   TCollectionName,
   TColumnName,
   TFieldName,
+  TPartialSimpleRow,
   TSchema,
 } from '@forestadmin/datasource-customizer';
 import type CollectionCustomizationContext from '@forestadmin/datasource-customizer/dist/context/collection-context';
 import type WriteCustomizationContext from '@forestadmin/datasource-customizer/dist/decorators/write/write-replace/context';
-import type { RecordData } from '@forestadmin/datasource-toolkit';
 
 import { ObjectCannedACL } from '@aws-sdk/client-s3';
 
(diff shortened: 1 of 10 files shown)


packages/plugin-aws-s3/src/types.ts
```diff
@@ -24,7 +27,17 @@ export type Options = {
    * Where should the file be stored on S3?
    * Defaults to '<collection_name>/<field_name>/`.
    */
-  storeAt?: (recordId: string, originalFilename: string) => string;
+  storeAt?: <S extends TSchema = TSchema, N extends TCollectionName<S> = TCollectionName<S>>(
```
SteveBunlon, line 30: Some types are incorrect / missing here, work in progress

SteveBunlon:
Tests are missing, work in progress

packages/plugin-aws-s3/src/types.ts
SteveBunlon: done

romain-gilliotte (changes requested):

packages/plugin-aws-s3/test/index.test.ts
```diff
@@ -184,6 +189,28 @@ describe('plugin-aws-s3', () => {
...
+        },
+      ]);
+
+      expect(mockSend).toHaveBeenCalledWith(
+        expect.objectContaining({
+          ACL: 'private',
+          Body: Buffer.from('Hello World'),
+          Bucket: 'myBucket',
+          ContentType: 'text/plain',
+          Key: 'books/null/myfile.txt',
```
SteveBunlon, line 206: This is an obvious bug that we'll need to fix. On create, upload of the file should be done in an after create hook, otherwise you won't get any record.id because it does not exist.

PR #759, opened 2023-07-06
fix(client): handle more errors on Server Sent Event issues to better reflect what is going on

Thenkei:
### General

### Security

### DIFF
--- a/packages/forestadmin-client/src/events-subscription/index.ts
+++ b/packages/forestadmin-client/src/events-subscription/index.ts
@@ -71,14 +71,22 @@ export default class EventsSubscriptionService implements BaseEventsSubscription
   }
 
   private onEventError(event: { type: string; status?: number; message?: string }) {
-    if (event.status === 502) {
-      this.options.logger('Debug', 'Server Event - Connection lost');
+    const { status, message } = event;
+
+    if ([502, 503, 504].includes(status)) {
+      this.options.logger('Debug', 'Server Event - Connection lost trying to reconnect…');
 
       return;
     }
 
-    if (event.message)
-      this.options.logger('Warn', `Server Event - Error: ${JSON.stringify(event)}`);
+    if (status === 404)
+      throw new Error(
+        'Forest Admin server failed to find the environment ' +
+          'related to the envSecret you configured. ' +
+          'Can you check that you copied it properly during initialization?',
+      );
+
+    if (message) this.options.logger('Warn', `Server Event - Error: ${JSON.stringify(event)}`);
   }
 
   private onEventOpenAgain() {
--- a/packages/forestadmin-client/test/events-subscription/index.test.ts
+++ b/packages/forestadmin-client/test/events-subscription/index.test.ts
@@ -53,7 +53,7 @@ describe('EventsSubscriptionService', () => {
     });
 
     describe('when server events are deactivated', () => {
-      test('should not do anything', async () => {
+      test('should not do anything', () => {
         const eventsSubscriptionService = new EventsSubscriptionService(
           { ...options, instantCacheRefresh: false },
           refreshEventsHandlerService,
@@ -68,7 +68,7 @@ describe('EventsSubscriptionService', () => {
 
   describe('handleSeverEvents', () => {
     describe('on RefreshUsers event', () => {
-      test('should delegate to refreshEventsHandlerService', async () => {
+      test('should delegate to refreshEventsHandlerService', () => {
         const eventsSubscriptionService = new EventsSubscriptionService(
           options,
           refreshEventsHandlerService,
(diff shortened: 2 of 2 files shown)


packages/forestadmin-client/src/events-subscription/index.ts
```diff
@@ -57,6 +59,15 @@ export default class EventsSubscriptionService implements BaseEventsSubscription
     source.addEventListener(ServerEventType.RefreshCustomizations, async () =>
       this.refreshEventsHandlerService.refreshCustomizations(),
     );
+
+    [`exit`, `SIGINT`, `SIGUSR1`, `SIGUSR2`, `uncaughtException`, `SIGTERM`].forEach(eventType => {
+      process.on(eventType, () => {
+        // https://github.com/EventSource/eventsource/blob/79dacfa860751438c116df76f9c5fdb9a2a3d69f/lib/eventsource.js#L392C26-L392C28
+        source.close();
+
+        process.exit(0);
```
ghusse, line 68: The problem with this line, is that the whole application using agent-nodejs won't be able to use process events on its own, as we are forcing `process.exit` here.

I don't know if there is a better solution, but we should use these events with care because they are overriding the default behavior of the app.

ghusse: Maybe we should only expose a `close` function on the agent, which should be called by client's code on exit (if needed).

Thenkei: I was more trying these events than anything else. We could probably forget about `process.exit(0);`.

And, I am not able to reproduce the client issue for now. 😢

packages/forestadmin-client/src/events-subscription/index.ts
```diff
@@ -36,6 +36,8 @@ export default class EventsSubscriptionService implements BaseEventsSubscription
     const url = new URL('/liana/v4/subscribe-to-events', this.options.forestServerUrl).toString();
 
     const source = new EventSource(url, eventSourceConfig);
+    // Override reconnect interval to 5 seconds
+    source.reconnectInterval = 5000;
```
ghusse, line 40: I don't think that this change is related to the intent of handling more errors. Am I wrong?

Thenkei: It's related to the same issue but I'll split it. 
https://app.clickup.com/t/860r920fn

packages/forestadmin-client/src/events-subscription/index.ts
```diff
@@ -71,14 +71,27 @@ export default class EventsSubscriptionService implements BaseEventsSubscription
...
-      this.options.logger('Warn', `Server Event - Error: ${JSON.stringify(event)}`);
+    if (status === 503)
+      throw new Error(
+        'Forest Admin is in maintenance for a few minutes. We are upgrading your experience in ' +
+          'the forest. We just need a few more minutes to get it right.',
+      );
+
+    if (status === 404)
+      throw new Error(
+        'Forest Admin server failed to find the project related to the envSecret you configured.' +
```
ghusse, line 90:
```suggestion
        'Forest Admin server failed to find the environment related to the envSecret you configured.' +
```

packages/forestadmin-client/src/events-subscription/index.ts
```diff
@@ -71,14 +71,27 @@ export default class EventsSubscriptionService implements BaseEventsSubscription
...
+      this.options.logger('Debug', 'Server Event - Connection lost trying to reconnect..');
 
       return;
     }
 
-    if (event.message)
-      this.options.logger('Warn', `Server Event - Error: ${JSON.stringify(event)}`);
+    if (status === 503)
+      throw new Error(
+        'Forest Admin is in maintenance for a few minutes. We are upgrading your experience in ' +
```
ghusse, line 84: Are you sure that upgrades result in 503s? For what I know of the deployment process, I would say that it's transparent for the users.

503's indicates an issue on our side.

Thenkei: I just copied the existing handlers. 👀

packages/forestadmin-client/src/events-subscription/index.ts
```diff
@@ -71,14 +71,22 @@ export default class EventsSubscriptionService implements BaseEventsSubscription
   }
 
   private onEventError(event: { type: string; status?: number; message?: string }) {
-    if (event.status === 502) {
-      this.options.logger('Debug', 'Server Event - Connection lost');
+    const { status, message } = event;
+
+    if (status === 502 || status === 504 || status === 503) {
```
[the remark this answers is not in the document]
DayTF: i'm a freak

PR #762, opened 2023-07-06
refactor(client): allow to close gracefully the current event source

Thenkei:
### General

### Security

### DIFF
--- a/packages/forestadmin-client/src/forest-admin-client-with-cache.ts
+++ b/packages/forestadmin-client/src/forest-admin-client-with-cache.ts
@@ -82,6 +82,10 @@ export default class ForestAdminClientWithCache implements ForestAdminClient {
     await this.eventsSubscription.subscribeEvents();
   }
 
+  public close() {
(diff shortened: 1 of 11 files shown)


packages/forestadmin-client/src/forest-admin-client-with-cache.ts
```diff
@@ -82,6 +82,10 @@ export default class ForestAdminClientWithCache implements ForestAdminClient {
     await this.eventsSubscription.subscribeEvents();
   }
 
+  public closeServerEvents() {
```
ghusse, line 85: It should be named `close` here, as it is the public endpoint for releasing resources.

Thenkei: Hihi.

Thenkei: Indeed, I didn't think about it like that but we can but everything related to close here.

PR #769, opened 2023-07-19
feat: add changed field to action context

arnaud-moncel:
### General

### Security


romain-gilliotte (changes requested):
Not a big fan of using the caller to carry query-specific data.

Its name clearly indicates that it is meant to carry information about who / from where the call is coming (hence call`er`).

Can you add a new parameter to the getForm function?

SkyHuss (approved):
LGTM 👍

PR #780, opened 2023-07-26
feat(datasource-customizer): implement some features of gmail-style search

romain-gilliotte:
### General

### Security


ghusse:
I the following commits, I've added the support of the following syntax:

```
A OR B                                    # A and B being any other supported syntax. Specifies that one of the 2 condition must be true.
"some text with spaces"      # the whole string will be searched in the fields
-"some text"                          # negated search of the whole string
-45                                          # Searches the number "-45"
-"45"                                       # Searches numbers different from 45

relationship.property:42      # Will look for the property "property" in the relationship "relationship". 
property:NULL                      # Will look for null values
-property:NULL                    # Not null values

# Works also with properties
# Will look for a number value that respects the given condition
>42
<42
<=42
>=42

2023                                   # Will look for dates with the given year
>2023                                 # Works for years if the number is > 1800 and < Now + 100 years
>=2023
<2023
<=2023

2023-01-01                        # Will look for dates on the same day
>2023-01-01
>=2023-01-01
<2023-01-01
<=2023-01-01
``` 


The separator `:` has not been used for property paths, because it's already the separator between `property` and `value` in the following syntax: `property:value`.

The same way, the operator `has:property` has been implemented using `property:NULL` or `-property:NULL` instead, to keep a coherent syntax.

**Comment by ghusse** [comment]
I've added the support of the following:

- `A OR B AND C` is now equivalent to `A OR (B AND C)` : operators priority is correctly working
- `(A OR B) AND C` works as expected. It works with nested conditions too.

File: .vscode/settings.json [hunk 1]
```diff
@@ -13,7 +13,7 @@
   "eslint.workingDirectories": ["."],
   "eslint.format.enable": true,
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
realSpok, line 16: Is that related to the PR ?  I'm annoyed because I can't pull without merging my own vscode settings :D

ghusse: No, but vscode keeps changing this parameter. I guess that newer versions changed the boolean into a string to have more flexibility.

I'll make another PR

.github/workflows/build.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         with:
           node-version: 16.14.0
       - name: Start docker containers
-        if: ${{ matrix.package == 'datasource-mongoose' || matrix.package == 'datasource-sql' }}
+        if: ${{ matrix.package == 'datasource-mongoose' || matrix.package == 'datasource-sql' || matrix.package == 'datasource-sequelize' }}
```
realSpok, line 82: There is probably a rebase missing

realSpok: Or maybe something I dont understand

realSpok:
Some food for thought for some frontend improvements.
Gmail has some really neat 'discovery' features for their search
image

realSpok:
I'm seeing an interesting behaviour:
here is the unsearched list
image
if I search for `comment:toto` here is the result
image
I'm not sure why the 3rd line is not there

packages/datasource-customizer/src/decorators/search/Query.g4
```diff
@@ -0,0 +1,39 @@
+grammar Query;
```
realSpok, line 1: Does the file name need to start with a capital letter ?'

packages/datasource-customizer/src/decorators/search/generated-parser/Query.interp
```diff
@@ -0,0 +1,44 @@
+token literal names:
```
realSpok, line 1: How does this work ? I assume that is not code that we have to maintain… :) 
How it is being generated ?

realSpok: Could this code be gitignored and built ?

realSpok: I mean all the code in `generated-parser/`

realSpok: anyway I guess this part should not be reviewed

.github/workflows/build.yml
ghusse: I added integration tests on `datasource-sequelize`, for which I need docker to start containers.

ghusse:
> I'm seeing an interesting behaviour: here is the unsearched list image if I search for `comment:toto` here is the result image
> 
> it seems to be using both the 'normal' search and the gmail-style search at the same time on column `comment`, so we have both lines 1 and 2 matching. However, I'm not sure why the 3rd line is not there

That's normal: the string `comment:toto` actually contains the string `toto`. That's why it's returned.

packages/datasource-customizer/src/decorators/search/generated-parser/Query.interp
ghusse: This is generated by antlr, from the grammar-description file. It seems to be needed by the parser to work properly.

packages/datasource-customizer/src/decorators/search/Query.g4
ghusse: For what I remember, yes. Otherwise the class is not generated with a capital letter.

realSpok, line 1: It would be nice to link to some doc from here. If anybody stumbles upon this file they'll be able to look it up

packages/datasource-customizer/src/decorators/search/custom-parser/custom-error-strategy.ts
```diff
@@ -0,0 +1,8 @@
+import { DefaultErrorStrategy } from 'antlr4';
+
+export default class CustomErrorStrategy extends DefaultErrorStrategy {
+  override reportError(): void {
+    // We don't want console logs when parsing fails
```
realSpok, line 5: What could cause the parser to fail ? What would happen then ? normal search ?

ghusse: If the query does not respect the syntax (unmatched parenthesis). In this case we fallback and consider the whole searched string as a unique token to search.

packages/datasource-customizer/src/decorators/search/filter-builder/build-date-field-filter.ts
```diff
@@ -0,0 +1,238 @@
...
+}
+
+function getPeriodStart(string: string): string {
+  if (isYear(string)) return `${string}-01-01`;
+  if (isYearMonth(string)) return `${string}-01`;
+
+  return string;
+}
+
+function pad(month: number) {
```
realSpok, line 38: you could use native `month.toString().padStart(2, '0')`

realSpok (approved):
- ✅  Code review
  - ✅ Use pure functions when possible
  - ✅  Performance concerns
  - ✅  Security concerns
- ✅ Automatic tests
  - ✅  Unit tests
  - ✅  Integration tests
- ✅ Manual tests
  -  ✅  Test the functionality
  -  ✅  Test error cases
- ✅  PR title
- ✅  Destination branch

Some minor non blocking comments

PR #795, opened 2023-08-28
fix(error-handling): use loger to display context log

Scra3:
### General

### Security


packages/agent/src/routes/system/error-handling.ts
```diff
@@ -29,17 +29,15 @@ export default class ErrorHandling extends BaseRoute {
...
-            detail: this.getErrorMessage(e),
+            detail: this.getErrorMessage(e, context),
             // We needed to maintaining this due to the frontend app/utils/ember-error-util.js
             status,
             ...(data ? { data } : {}),
           },
         ],
       };
 
-      if (!this.options.isProduction) {
```
Scra3, line 40: I have removed this condition because we display the log in "Debug" level. The user can filter it if he wants by adding his logger.

packages/agent/test/routes/system/error-handling.test.ts
```diff
@@ -182,36 +184,19 @@ describe('ErrorHandling', () => {
...
-        use: (handler: Router.Middleware) => {
-          handleError = handler;
-        },
-      });
-    });
 
-    test('it should print stuff to stderr', async () => {
-      const context = createMockContext({ method: 'POST' });
-      const next = jest.fn().mockRejectedValue(new Error('hello'));
+      test('it should print stuff to the logger in debug level', async () => {
```
ghusse, line 188: It should be `logger.error` here instead of `logger.debug`, or is it a customization?

Scra3: We can put in Error level 👍

packages/agent/src/routes/system/error-handling.ts
```diff
@@ -81,21 +79,18 @@ export default class ErrorHandling extends BaseRoute {
...
-    console.error('');
-    console.error('\x1b[31m', error.message, '\x1b[0m');
-    console.error('');
-    console.error(error.stack);
-    console.error(`\x1b[33m===================================\x1b[0m`);
-    console.error('');
+    const errorMessage = `\x1b[31m${error.message}\x1b[0m\n\n${error.stack}\n`;
+    const message = `${path}\n${body ? `${body}\n` : ''}\n${errorMessage}\n`;
+    const endOfException = '\x1b[33m===================================\x1b[0m';
+    this.options.logger('Error', `\n${startException}\n${message}${endOfException}\n`);
```
ghusse, line 94: I think that adding things at the beginning and the end of the error should be done by a default "customizeErrorMessage", but as long as a function is passed we should just use the value that is returned by this function "as is".

This is really important if people are plugging logs extraction.

packages/agent/src/routes/system/error-handling.ts
```diff
@@ -29,17 +29,22 @@ export default class ErrorHandling extends BaseRoute {
...
           },
         ],
       };
 
-      if (!this.options.isProduction) {
-        process.nextTick(() => this.debugLogError(context, e));
-      }
+      process.nextTick(() => {
+        const message =
+          this.options.loggerLevel === 'Debug'
```
ghusse, line 42: Why does it depend on the log level here?

PR #816, opened 2023-09-08
fix(datasource-customizer): add hasFieldChanged function to handle change hook on this field

### LINKED ISSUE #815: usage of changedField inside smart action handlers does not trigger change hook
## Expected behavior

When I use `changedField` property from the context inside smart action handler, I want the field tested to trigger a change hook.

## Actual behavior

It does not trigger a change hook.

```javascript
.addAction('test', {
  scope: 'Global',
  form: [
    {
      label: 'change',
      type: 'String',
    },
    {
      label: 'to change',
      type: 'String',
      isReadOnly: true,
      value: context => {
        if (context.changedField === 'change') {
          return context.formValues.change;
        }
      },
    },
  ],
  execute: () => {},
})
```
image

## Proposal

Add a `hasFieldChanged` function to handle that.
```javascript
.addAction('test', {
  scope: 'Global',
  form: [
    {
      label: 'change',
      type: 'String',
    },
    {
      label: 'to change',
      type: 'String',
      isReadOnly: true,
      value: context => {
        if (context.hasFieldChanged('change')) {
          return context.formValues.change;
        }
      },
    },
  ],
  execute: () => {},
})
```
image

arnaud-moncel:
fix #815

### General

### Security

### DIFF
--- a/packages/datasource-customizer/src/decorators/actions/context/base.ts
+++ b/packages/datasource-customizer/src/decorators/actions/context/base.ts
@@ -20,11 +20,20 @@ export default class ActionContext<
   readonly formValues: any; // eslint-disable-line @typescript-eslint/no-explicit-any
   readonly filter: TFilter<S, N>;
 
+  private _changedField: string;
+
   /**
    * @deprecated use `hasFieldChange` instead. [linked issue](https://github.com/ForestAdmin/agent-nodejs/issues/815).
-   * @todo make it private on the next major version.
+   * @todo remove accessor
    */
-  readonly changedField: string;
+  get changedField() {
+    console.warn(
+      '\x1b[33mwarning:\x1b[0m',
+      'Usage of changedField is deprecated, please use `hasFieldChanged` instead.',
+    );
+
+    return this._changedField;
+  }
 
   private queries: Array<{ projection: Projection; deferred: Deferred<RecordData[]> }>;
   private projection: Projection;
@@ -42,7 +51,7 @@ export default class ActionContext<
     super(collection, caller);
     this.formValues = formValue;
     this.filter = filter;
-    this.changedField = changedField;
+    this._changedField = changedField;
     this.reset();
 
     // Spy on which formValues are accessed to set-up change hooks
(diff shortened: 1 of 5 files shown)


packages/datasource-customizer/src/decorators/actions/context/base.ts
```diff
@@ -19,11 +19,18 @@ export default class ActionContext<
 > extends CollectionCustomizationContext<S, N> {
   readonly formValues: any; // eslint-disable-line @typescript-eslint/no-explicit-any
   readonly filter: TFilter<S, N>;
+
+  /**
+   * @deprecated use `hasFieldChange` instead. [linked issue](https://github.com/ForestAdmin/agent-nodejs/issues/815).
+   * @todo make it private on the next major version.
+   */
   readonly changedField: string;
```
ghusse, line 27: Can't it be replaced by a getter instead, and keep the same API?

packages/datasource-customizer/src/decorators/actions/context/base.ts
```diff
@@ -50,6 +57,12 @@ export default class ActionContext<
           throw new Error('formValues is readonly');
         },
       });
+
+      this.hasFieldChanged = (fieldName: string) => {
+        used.add(fieldName);
```
Thenkei, line 62: This is an underlying mechanism to only trigger recompute on some field change. Basically, the system is here to say whenever this field (those fields) changes I need to recompute form field values.

All this is done by adding a hook change based on the used computed list of fields (name). It only adds fields that will affect output. 

``` 
return (context.hasFieldChanged('A') || context.hasFieldChanged('B') || context.hasFieldChanged('C')) ? 'True' : 'False';
```
Will watch to field change `[A, B, C]`. If `hasFieldChanged('A')` is true then `used` will be only [A]. Since only a change on A field can change the output value (changes on B or C won't change anything).

``` 
return (context.formValues['A'] || context.formValues.['B] || context.formValues.['C']) ? 'True' : 'False';
```
Same behavior with `formValues[fieldName]`.

Thenkei:
I tested it, it's ok (superbe) for me. 👌

Thenkei (approved):
- ✅&nbsp;Code review
  - ✅&nbsp; Security concerns
- ✅&nbsp;Automatic tests
  - ✅&nbsp; Unit tests
  - ✅&nbsp; Integration tests
- ✅&nbsp;Manual tests
- ✅&nbsp;PR title
- ❔&nbsp;PR linked to the clickup task

PR #820, opened 2023-09-12
fix(datasource-mongoose): return values instead of null when querying nested properties on a flattened field

ghusse:
Linked to CU-860rq6m55

### General

### Security

### DIFF
--- a/packages/datasource-mongoose/test/integration/flattener/collection_list.test.ts
+++ b/packages/datasource-mongoose/test/integration/flattener/collection_list.test.ts
@@ -112,4 +112,39 @@ describe('Complex flattening', () => {
 
     expect(moreRecords).toEqual([expect.objectContaining({ _id: `${car._id}.engine` })]);
   });
+
+  describe('asModels', () => {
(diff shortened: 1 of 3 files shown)


packages/datasource-mongoose/test/integration/flattener/collection_list.test.ts
```diff
@@ -112,4 +112,39 @@ describe('Complex flattening', () => {
 
     expect(moreRecords).toEqual([expect.objectContaining({ _id: `${car._id}.engine` })]);
   });
+
+  describe('asModels', () => {
```
Thenkei, line 116: You've done exactly what `asFields` was intended for?

>  // paths that should be moved to the root of the collection
> asFields: ['address'],

ghusse: When using `asModels` and a custom field depending on real nested fields, the fields that are requested internally are constructed this way.

That's why I added this integration test.

DayTF (approved):

PR #862, opened 2023-10-26
fix(authentication): don't crash if an error occurs when initializing the authentication client

ghusse:
Linked to CU-860rxhayg

### General

### Security

### DIFF
--- a/packages/forestadmin-client/src/auth/index.ts
+++ b/packages/forestadmin-client/src/auth/index.ts
@@ -62,12 +62,10 @@ export default class AuthService implements ForestAdminAuthServiceInterface {
   }): Promise<string> {
     await this.createClient();
 
-    const url = this.client.authorizationUrl({
+    return this.client.authorizationUrl({
       scope,
       state,
     });
-
-    return url;
   }
 
   public async generateTokens({
--- a/packages/forestadmin-client/src/auth/index.ts
+++ b/packages/forestadmin-client/src/auth/index.ts
@@ -30,22 +30,7 @@ export default class AuthService implements ForestAdminAuthServiceInterface {
     }
   }
 
-  protected async createClient() {
-    if (this.client) return;
-
-    // We can't use async 'Issuer.discover' because the oidc config is behind an auth-wall.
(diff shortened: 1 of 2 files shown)


packages/forestadmin-client/src/auth/index.ts
```diff
@@ -56,7 +75,7 @@ export default class AuthService implements ForestAdminAuthServiceInterface {
     scope: string;
     state: string;
   }): Promise<string> {
-    if (!this.client) throw new Error('AuthService not initialized');
+    await this.createClient();
 
     const url = this.client.authorizationUrl({
```
Scra3, line 80: If the createClient fails, this.client is undefined

ghusse: If `createClient` fails, it throws an error here. The `try/catch` is only in the init method.

packages/forestadmin-client/src/auth/index.ts
```diff
@@ -12,13 +12,32 @@ export default class AuthService implements ForestAdminAuthServiceInterface {
...
+      // server or network error. We don't want the application to crash.
+      this.options.logger(
+        'Warn',
+        // eslint-disable-next-line max-len
+        `Error while registering the authentication client. Authentication might not work: ${e.message}`,
+      );
+    }
+  }
+
+  protected async createClient() {
```
Scra3, line 33: protected method should be after the public method

packages/forestadmin-client/src/auth/index.ts
Scra3, line 80: optional: you can return here

packages/forestadmin-client/src/auth/index.ts
ghusse: If it's a rule, it should be enforced by eslint.

packages/forestadmin-client/src/auth/index.ts
ghusse: Return what ?

Scra3 (approved):

PR #878, opened 2023-11-14
feat: handle cursor pagination

arnaud-moncel:
### General

### Security


packages/datasource-toolkit/src/interfaces/query/filter/paginated.ts
```diff
@@ -1,25 +1,32 @@
 import Filter, { FilterComponents, PlainFilter } from './unpaginated';
+import Cursor, { PlainCursor } from '../cursor';
 import Page, { PlainPage } from '../page';
 import Sort, { PlainSortClause } from '../sort';
 
+export type PaginationType = 'page' | 'cursor';
```
Thenkei, line 6: 🤝

packages/datasource-toolkit/src/interfaces/query/cursor.ts
```diff
@@ -0,0 +1,13 @@
...
+  before: string;
+  after: string;
+  limit: number;
+
+  constructor(before?: string, after?: string, limit?: number) {
+    this.after = after;
+    this.before = before;
+    this.limit = limit;
+  }
+}
```
Thenkei, line 13: Why did we choose the definition? Could the interface be more lax? (not that easy)

Thenkei: In Elasticsearch we have two ways: 
- https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after `"search_after": [1463538857, "654323"],`
- Point In Time (`pit_id`)
```
"pit": {
    "id":  "46ToAwMDaWR5BXV1aWQyKwZub2RlXzMAAAAAAAAAACoBYwADaWR4BXV1aWQxAgZub2RlXzEAAAAAAAAAAAEBYQADaWR5BXV1aWQyKgZub2RlXzIAAAAAAAAAAAwBYgACBXV1aWQyAAAFdXVpZDEAAQltYXRjaF9hbGw_gAAAAA==", 
    "keep_alive": "1m"
  },
```

We could certainly do something about it in the future if we want to take this option for other DS.

PR #891, opened 2023-11-26
feat(datasource-customizer): add helper to fetch values from selected record in single actions

romain-gilliotte:
This is not much, but it helps when declaring single actions which use data from the selected record.
It's a major eyesore in the codebase I'm currently working in, as the field names are quite long, and this pattern is everywhere!

The templating is so that the output of the function have the proper type

Go from this:
```
{
    label: 'Street',
    isRequired: true,
    type: 'String',
    defaultValue: async (context) =>
      (await context.getRecord(['company:street'])).company?.street,
}
```

to this

```
{
    label: 'Street',
    isRequired: true,
    type: 'String',
    defaultValue: (context) => context.getField('company:street')
}
```

### DIFF
--- a/packages/datasource-customizer/test/decorators/actions/context.test.ts
+++ b/packages/datasource-customizer/test/decorators/actions/context.test.ts
@@ -8,17 +8,37 @@ describe('ActionContext', () => {
   let books: Collection;
 
   beforeEach(() => {
-    books = factories.collection.build({
-      dataSource: factories.dataSource.build(),
-      name: 'books',
-      schema: factories.collectionSchema.build({
-        fields: {
-          id: factories.columnSchema.build({ isPrimaryKey: true }),
-          title: factories.columnSchema.build(),
-        },
+    const dataSource = factories.dataSource.buildWithCollections([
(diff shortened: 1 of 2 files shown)


packages/datasource-customizer/test/decorators/actions/context.test.ts
```diff
@@ -93,4 +93,16 @@ describe('ActionContext', () => {
     await expect(promise1).rejects.toThrow('bad request');
     await expect(promise2).rejects.toThrow('bad request');
   });
+
+  test('should get individual fields', async () => {
```
ghusse, line 97: Can you add a test that is accessing child properties here? I think it's a major part of your contribution, and it should be tested as well.

romain-gilliotte: Sure. Will do!
@scra3 would have told me the same 😁

romain-gilliotte:
It's done!

ghusse (approved):

ghusse:
Thanks!

PR #892, opened 2023-11-29
feat(datasource-customizer): sort enum values in typings file

xiniria:
## Definition of Done

Initially I had implemented sorting for enum values but also plain, nested and flat field names; when rebasing on main I realized someone else had done it in the meantine. Thus there are only additional tests for these, but enum values are still now sorted. This prevents the kind of problem I've had on my project when generating the typings file:

image

### General

### Security

### DIFF
--- a/packages/datasource-customizer/src/typing-generator.ts
+++ b/packages/datasource-customizer/src/typing-generator.ts
@@ -151,10 +151,11 @@ export default class TypingGenerator {
 
     while (queue.length > 0 && result.length < this.options.maxFieldsCount) {
       const { collection, depth, prefix, traversed } = queue.shift();
+      const sortedFields = TypingGenerator.sortedEntries(collection.schema.fields);
 
       if (prefix) {
         result.push(
-          ...TypingGenerator.sortedEntries(collection.schema.fields)
+          ...sortedFields
             .filter(([, schema]) => schema.type === 'Column')
             .map(
               ([name, schema]) => `'${prefix}:${name}': ${this.getType(schema as ColumnSchema)};`,
@@ -164,7 +165,7 @@ export default class TypingGenerator {
 
       if (depth < maxDepth) {
         queue.push(
-          ...TypingGenerator.sortedEntries(collection.schema.fields)
+          ...sortedFields
             .filter(([, schema]) => schema.type === 'ManyToOne' || schema.type === 'OneToOne')
             .map(([name, schema]: [name: string, schema: OneToOneSchema | ManyToOneSchema]) => {
               return {
--- a/packages/datasource-customizer/src/typing-generator.ts
+++ b/packages/datasource-customizer/src/typing-generator.ts
@@ -210,9 +210,11 @@ export default class TypingGenerator {
     }
 
     if (field.columnType === 'Enum') {
+      if (field.enumValues === undefined) return 'string';
+
       return (
-        field.enumValues
-          ?.sort((v1, v2) => v1.localeCompare(v2))
+        [...field.enumValues]
+          .sort((v1, v2) => v1.localeCompare(v2))
           .map(v => `'${v.replace(/'/g, "\\'")}'`)
           .join(' | ') ?? 'string'
       );
--- a/packages/datasource-customizer/src/typing-generator.ts
+++ b/packages/datasource-customizer/src/typing-generator.ts
@@ -158,6 +158,7 @@ export default class TypingGenerator {
       if (depth < maxDepth) {
         queue.push(
           ...Object.entries(collection.schema.fields)
+            .sort(([f1Name], [f2Name]) => f1Name.localeCompare(f2Name))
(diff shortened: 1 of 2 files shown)


packages/datasource-customizer/src/typing-generator.ts
```diff
@@ -18,6 +18,12 @@ export default class TypingGenerator {
     this.options.maxFieldsCount = options.maxFieldsCount ?? this.options.maxFieldsCount;
   }
 
+  private getSortedEntries<T>(
```
ghusse, line 21: Can you rename this function `sortEntries`? A function's name starting with `get` looks like an accessor but in this case, this is not.

We also would need this function to be declared as `static` because it does not use `this` internally.

ghusse (changes requested):
Thanks a lot for your contribution. We'll definitively integrate your changes, as they allow having a more stable output with the typings file.

I have 2 minor concerns regarding the introduced function. If you can do the modifications, I'll merge your PR.

packages/datasource-customizer/src/typing-generator.ts
xiniria: @ghusse done!

packages/datasource-customizer/src/typing-generator.ts
```diff
@@ -203,7 +210,12 @@ export default class TypingGenerator {
     }
 
     if (field.columnType === 'Enum') {
-      return field.enumValues?.map(v => `'${v.replace(/'/g, "\\'")}'`).join(' | ') ?? 'string';
+      return (
+        field.enumValues
+          ?.sort((v1, v2) => v1.localeCompare(v2))
```
Thenkei, line 215: This is the fix.

xiniria:
@ghusse @Thenkei The "Send Coverage" CI step seems to fail every time, is it something you had already seen in the repo?

ghusse:
> @ghusse @Thenkei The "Send Coverage" CI step seems to fail every time, is it something you had already seen in the repo?

Yes, it's working for us, when the CI is run on the original repository.

packages/datasource-customizer/src/typing-generator.ts
```diff
@@ -203,7 +210,12 @@ export default class TypingGenerator {
...
+          .map(v => `'${v.replace(/'/g, "\\'")}'`)
```
Thenkei, line 216:
```suggestion
          ?.map(v => `'${v.replace(/'/g, "\\'")}'`)
          .sort((v1, v2) => v1.localeCompare(v2))
```

packages/datasource-customizer/src/typing-generator.ts
```diff
@@ -157,7 +164,7 @@ export default class TypingGenerator {
 
       if (depth < maxDepth) {
         queue.push(
-          ...Object.entries(collection.schema.fields)
+          ...TypingGenerator.sortedEntries(collection.schema.fields)
```
ghusse, line 167: This is not necessary here to have sorted results

packages/datasource-customizer/src/typing-generator.ts
xiniria: @Thenkei why do you prefer it this way? Normally it should not make any difference on the sorting

packages/datasource-customizer/src/typing-generator.ts
xiniria: Actually removing it breaks the `flat` fields test, so I think it is necessary in order to have ordered `flat` fields

packages/datasource-customizer/src/typing-generator.ts
```diff
@@ -203,7 +210,12 @@ export default class TypingGenerator {
     }
 
     if (field.columnType === 'Enum') {
-      return field.enumValues?.map(v => `'${v.replace(/'/g, "\\'")}'`).join(' | ') ?? 'string';
+      return (
+        field.enumValues
```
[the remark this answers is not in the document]
xiniria: You're right, I had not thought about this use case

packages/datasource-customizer/src/typing-generator.ts
ghusse: Ok, I see.

I don't know if it's better to do it here, or simply at the end of the function with `result.slice().sort()`. This would make it more explicit that we want to return sorted results.

I think that your solution works too, so I'm fine with it. I would just move the call to `sort` at the beginning of the loop to avoid doing it twice on the same elements:

```
    while (queue.length > 0 && result.length < this.options.maxFieldsCount) {
      const { collection, depth, prefix, traversed } = queue.shift();
      const sortedEntries = TypingGenerator.sortedEntries(collection.schema.fields);

       if (prefix) {
        result.push(
          ...sortedEntries
            .filter(/**/)
           .map(/**/)
        )
      }

      if (depth < maxDepth) {
        queue.push(
          ...sortedEntries
          .filter(/**/)
          .map(/**/)
          .filter(/**/)
      }
```

xiniria: @ghusse done!

PR #894, opened 2023-12-07
fix: field formValue is sometimes not correctly provided in execute context

realSpok:
#860t0ghxn

### General

### Security

### DIFF
--- a/packages/agent/src/routes/modification/action/action.ts
+++ b/packages/agent/src/routes/modification/action/action.ts
@@ -98,7 +98,7 @@ export default class ActionRoute extends CollectionRoute {
       this.actionName,
       unsafeData,
       filterForCaller,
-      { shouldDropIfs: false }, // during execute, we need all possible fields
+      { includeHiddenFields: true }, // during execute, we need all possible fields
     );
 
     // Now that we have the field list, we can parse the data again.
@@ -157,7 +157,7 @@ export default class ActionRoute extends CollectionRoute {
       changedField: body.data.attributes.changed_field,
       searchField: body.data.attributes.search_field,
       searchValues,
-      shouldDropIfs: true,
+      includeHiddenFields: false,
     });
 
     context.response.body = {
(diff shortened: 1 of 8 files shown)


packages/agent/src/routes/modification/action/action.ts
```diff
@@ -98,6 +98,7 @@ export default class ActionRoute extends CollectionRoute {
       this.actionName,
       unsafeData,
       filterForCaller,
+      { shouldDropIfs: false }, // during execute, we need all possible fields
```
Thenkei, line 101: `shouldDropIfs` is not straightforward from this perspective. 
Could we change this to `forceReturnAllFields` or something that include the idea of `we need all possible fields` without the technical

Thenkei: What do you think?

Scra3: I think this behavior can cause bugs because we will execute the handler even if the "if" is falsy?

packages/agent/src/routes/modification/action/action.ts
```diff
@@ -98,6 +98,7 @@ export default class ActionRoute extends CollectionRoute {
       this.actionName,
       unsafeData,
       filterForCaller,
+      { includeHiddenFields: true }, // during execute, we need all possible fields
```
Thenkei, line 101: Better

packages/datasource-customizer/src/decorators/actions/collection.ts
```diff
@@ -78,7 +75,7 @@ export default class ActionCollectionDecorator extends CollectionDecorator {
     }
 
     dynamicFields = await this.dropDefaults(context, dynamicFields, formValues);
-    dynamicFields = await this.dropIfs(context, dynamicFields);
+    if (!metas?.includeHiddenFields) dynamicFields = await this.dropIfs(context, dynamicFields);
```
Thenkei, line 78: 👌

Thenkei (approved):

PR #895, opened 2023-12-07
feat: use name to access formValues on top of label

realSpok:
### General

### Security


packages/agent/src/utils/forest-schema/action-values.ts
```diff
@@ -35,6 +35,8 @@ export default class ForestValueConverter {
           data[key] = value;
         }
       }
+
+      data[field.name] = data[key];
```
Thenkei, line 39: `makeFormDataFromFields` and `makeFormDataUnsafe` don't need the same ?

PR #901, opened 2023-12-20
fix: cloud sync error with customization applied to missing collections

realSpok:
## Definition of Done
#860t46nnc

### General

### Security

### DIFF
--- a/packages/agent/src/services/model-customizations/actions/update-record/update-record-plugin.ts
+++ b/packages/agent/src/services/model-customizations/actions/update-record/update-record-plugin.ts
@@ -22,7 +22,7 @@ export default class UpdateRecordActionsPlugin {
     );
 
     actions.forEach(action => {
-      const collection = datasourceCustomizer.collections.find(c => c.name === action.modelName);
+      const collection = datasourceCustomizer.findCollection(action.modelName);
       if (!collection) return;
 
       collection.addAction(action.name, {
--- a/packages/agent/src/services/model-customizations/actions/update-record/update-record-plugin.ts
+++ b/packages/agent/src/services/model-customizations/actions/update-record/update-record-plugin.ts
@@ -22,7 +22,8 @@ export default class UpdateRecordActionsPlugin {
     );
 
     actions.forEach(action => {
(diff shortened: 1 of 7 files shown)


packages/agent/src/services/model-customizations/actions/update-record/update-record-plugin.ts
```diff
@@ -22,7 +22,8 @@ export default class UpdateRecordActionsPlugin {
     );
 
     actions.forEach(action => {
-      const collection = datasourceCustomizer.getCollection(action.modelName);
+      const collection = datasourceCustomizer.collections.find(c => c.name === action.modelName);
```
ghusse, line 25: I think we should rely on methods on the datasourceCustomizer to be sure that the result of `getCollection` will be the same as the result of this find here.

There can be some special cases with removed collections that can return something here and not in `getCollections` if I look at the code.

packages/agent/src/services/model-customizations/actions/webhook/webhook-plugin.ts
```diff
@@ -15,7 +15,8 @@ export default class WebhookActionsPlugin {
   ) => {
     const actions = getActions<WebhookActionConfiguration>('webhook', modelCustomizations);
     actions.forEach(action => {
-      const collection = datasourceCustomizer.getCollection(action.modelName);
+      const collection = datasourceCustomizer.collections.find(c => c.name === action.modelName);
```
ghusse, line 18: We should definitively rely on a factorized function to avoid duplicating the same logic at multiple places.

packages/agent/src/services/model-customizations/actions/update-record/update-record-plugin.ts
Thenkei: Indeed, but we don't work on the same customization layer. 

The issue with getCollection comes from the fact that it doesn't check anything when called. It's only checked afterward. 

_(There are other cases supported for plugins that add or remove collection for example)_

ghusse (approved):

PR #915, opened 2024-01-21
feat(action): allow dynamic form

VincentMolinie:
### General

### Security


Thenkei:
What use case does it leverage? Could you share an explicit example? 😃 

Because the current interface already allows to make dynamic form **behavior**.

VincentMolinie:
One thing to note is that the form will be build on every change if there is a layout change hook on one of the field

PR #931, opened 2024-01-29
feat: expose cloud specific agent interface in cloud toolkit

realSpok:
## Definition of Done
#86bx306x8
### General

### Security

### DIFF
--- a/packages/cloud-toolkit/src/types.ts
+++ b/packages/cloud-toolkit/src/types.ts
@@ -1,0 +1,63 @@
+import type {
+  CollectionCustomizer,
+  DataSourceChartDefinition,
+  Plugin,
+  TCollectionName,
+  TSchema,
+} from '@forestadmin/datasource-customizer';
+
+/**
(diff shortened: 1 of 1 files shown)


packages/cloud-toolkit/src/types.ts
```diff
@@ -0,0 +1,63 @@
...
+  TCollectionName,
+  TSchema,
+} from '@forestadmin/datasource-customizer';
+
+/**
+ * This agent the central object used to customize your cloud project
+ * some methods available in self-hosted, such as:
+ * @method addDataSource are currently not supported in cloud
+ */
+export type Agent<S extends TSchema = TSchema> = {
```
EnkiP, line 14: I think we need a class. If we're using a normal Agent and say it's a cloud Agent as type, typescript will not convert it the cloud agent, it will just say ok, cloud agent is more restrictive than normal Agent, so a normal agent can be a cloud Agent if you want.
But all the methods we want to forbid will still be there

EnkiP (approved):

PR #933, opened 2024-01-29
feat(cloud-toolkit): update typings by fetching introspection on forestadmin server

Scra3:
### General

### Security


packages/cloud-toolkit/src/command.ts
```diff
@@ -0,0 +1,51 @@
...
+import HttpForestServer from './services/http-forest-server';
+import login from './services/login';
+import updateTypings from './services/update-typings';
+
+configDotenv();
+
+const buildHttpForestServer = async () => {
+  if (!process.env.FOREST_SERVER_URL || !process.env.FOREST_ENV_SECRET || !process.env.TOKEN_PATH) {
+    throw new Error(
+      'Missing FOREST_SERVER_URL, FOREST_ENV_SECRET or TOKEN_PATH. Please check your .env file.',
```
realSpok, line 18: we shouldn't throw for missing FOREST_SERVER_URL, it is only useful to us devs

realSpok: TOKEN_PATH as well actually… we can use the same default as the toolbelt

packages/cloud-toolkit/src/command.ts
```diff
@@ -0,0 +1,51 @@
...
+    process.env.FOREST_SERVER_URL,
+    process.env.FOREST_ENV_SECRET,
+    await readFile(path.join(process.env.TOKEN_PATH as string, '.forest.d', '.forestrc'), 'utf8'),
+  );
+};
+
+program
+  .command('update-typings')
+  .description('Update your typings file to synchronize with your datasource')
+  .action(async () => {
```
realSpok, line 32: We could probably call a new function `loginIfNeeded` here, which will check if there is a valid app token before login-in

packages/cloud-toolkit/src/command.ts
```diff
@@ -0,0 +1,51 @@
...
+program
+  .command('bootstrap')
+  .description('Bootstrap your project')
+  .argument(
+    '<Environment secret>',
+    'Environment secret, you can find it in your environment settings',
+  )
+  .action(async (envSecret: string) => {
+    await bootstrap(envSecret);
+    await login();
```
realSpok, line 45: same here `loginIfNeeded` (although less useful since at cloud boostrap they probably wont have signed in with the toolbelt already )

packages/cloud-toolkit/src/command.ts
```diff
@@ -0,0 +1,51 @@
...
+  return new HttpForestServer(
+    process.env.FOREST_SERVER_URL,
+    process.env.FOREST_ENV_SECRET,
+    await readFile(path.join(process.env.TOKEN_PATH as string, '.forest.d', '.forestrc'), 'utf8'),
+  );
+};
+
+program
+  .command('update-typings')
+  .description('Update your typings file to synchronize with your datasource')
```
realSpok, line 31:
```suggestion
  .description('Update your typings file to synchronize code autocompletion with your datasource (whenever its schema changes)')
```

packages/cloud-toolkit/src/services/http-forest-server.ts
```diff
@@ -0,0 +1,35 @@
...
+export default class HttpForestServer {
+  private readonly serverUrl: string | undefined;
+  private readonly secretKey: string | undefined;
+  private readonly bearerToken: string | undefined;
+  constructor(
+    serverUrl: string | undefined,
+    secretKey: string | undefined,
+    bearerToken: string | undefined,
+  ) {
+    this.serverUrl = serverUrl || 'https://api.forestadmin.com';
```
realSpok, line 13: You already have a default :) 
No need to throw !

realSpok (approved):

PR #956, opened 2024-02-01
feat: check code already deployed

EnkiP:
### General

### Security


Thenkei (commented):
This may sound but I do think we should let people re-bootstrap their project if they really want to do so (with a `--force` flag). 

I do think some may loose the project if they don't version it...

Thenkei (approved):
So far so good. Do you want to provide the same experience for the `publish` command in the PR?

PR #987, opened 2024-02-08
feat: display a warning when version is outdated

Scra3:
### General

### Security


Scra3:
Task linked: [CU-86bxd8wky Story #4.6 - Users want to be notified when the cloud-toolkit is not up to date](https://app.clickup.com/t/86bxd8wky)

packages/cloud-toolkit/src/dialogs/check-latest-version.ts
```diff
@@ -0,0 +1,32 @@
...
+import { CloudToolkitVersionError } from '../errors';
+
+export default async function checkLatestVersion(
+  spinner: Ora,
+  customerVersion: string,
+): Promise<void> {
+  let version: string;
+
+  try {
+    version = await latestVersion('@forestadmin/cloud-toolkit');
```
Thenkei, line 13: We must cache this value somehow. In the current state we will have 3 external call to run the `forestadmin:build:package:publish` command.

Scra3: After an internal discussion, we have decided to not use a cache for the moment. It's probably useless because the call is very fast.

packages/cloud-toolkit/src/dialogs/ask-to-overwrite-customizations.ts
```diff
@@ -5,9 +5,9 @@ import { Spinner } from '../types';
 
 export default async function askToOverwriteCustomizations(
   spinner: Spinner,
-  httpServer: HttpServer,
```
Scra3, line 8: we don't want a service inside dialog

packages/cloud-toolkit/src/types.ts
```diff
@@ -82,6 +82,7 @@ export type MakeCommands = {
   distPathManager: DistPathManager;
   logger: Logger;
   login: Login;
+  version: string;
```
Scra3, line 85: now version is injected

Scra3:
Benchmark:
image

packages/cloud-toolkit/src/build-command.ts
```diff
@@ -27,12 +28,15 @@ const logger: Logger = {
 };
 
 export default function buildCommands() {
+  const { version } = JSON.parse(fsSync.readFileSync('package.json', 'utf-8'));
```
realSpok, line 31: this doesn't work
You need something like that:
```suggestion
  const { version } = JSON.parse(
    fsSync.readFileSync(path.join(__dirname, 'package.json'), 'utf-8'),
  );
```

realSpok (approved):

PR #1003, opened 2024-02-14
fix: test package

Scra3:
### General

### Security


packages/cloud-toolkit/src/services/dist-path-manager.ts
```diff
@@ -0,0 +1,17 @@
+import path from 'path';
+
+export default class DistPathManager {
```
EnkiP, line 3: why do we need to have multiple path managers ? could we have 1 with all paths ?

Scra3: I prefer have one to the bootstrap and one for the dist because they are not related

PR #1019, opened 2024-02-17
fix: when chaining commands, the process must exist if fails

Scra3:
### General

### Security

### DIFF
--- a/packages/cloud-toolkit/src/dialogs/action-runner.ts
+++ b/packages/cloud-toolkit/src/dialogs/action-runner.ts
@@ -10,7 +10,13 @@ export default function actionRunner(spinner: Spinner, fn: (...args) => Promise<
 
       if (error instanceof BusinessError) {
         spinner.fail(error.message);
+        // we must exit the process with a non-zero code to indicate an error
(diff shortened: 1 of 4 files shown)


packages/cloud-toolkit/src/dialogs/action-runner.ts
```diff
@@ -10,7 +10,13 @@ export default function actionRunner(spinner: Spinner, fn: (...args) => Promise<
 
       if (error instanceof BusinessError) {
         spinner.fail(error.message);
+        // we must exit the process with a non-zero code to indicate an error
+        // when chaining commands, the process will continue if we don't exit
+        process.exit(1);
```
realSpok, line 15: Does it go through `finally` ?

realSpok: I tested and the spinner does stop but I'm not sure finally is reached

Scra3: ty for the advise, it much better !

realSpok (approved):

PR #1061, opened 2024-03-12
fix(forest-cloud): display an error when the command does not exist

Scra3:
linked to: CU-86bxvf48e
### General

### Security

### DIFF
--- a/packages/forest-cloud/src/commands/logs.ts
+++ b/packages/forest-cloud/src/commands/logs.ts
@@ -81,7 +81,7 @@ export default (program: Command, context: MakeCommands) => {
     )
     .option(
       '-n, --tail <integer>',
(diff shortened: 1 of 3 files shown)


packages/forest-cloud/src/commands/logs.ts
```diff
@@ -81,7 +81,7 @@ export default (program: Command, context: MakeCommands) => {
     )
     .option(
       '-n, --tail <integer>',
-      'Number of lines to show from the end of the logs in the last hour.' +
+      'Number of lines to show from the end of the logs in the last month.' +
```
realSpok, line 84: Is that change expected ?

realSpok: ok as discussed

realSpok (approved):

PR #1099, opened 2024-04-07
fix(datasource-customizer): correct confusing type when using addExternalRelation

romain-gilliotte:
Hello!

I noticed this small issue while working.
Did not want to bother you guys on the community forums for such a small patch

and besides, it's my name on the git blame (yes, I checked)


ghusse:
Thanks for the patch!

PR #1101, opened 2024-04-10
fix: improve error message in the case of outdated introspection

realSpok:
## Definition of Done
#86by7d27e
### General

### Security

### DIFF
--- a/packages/forest-cloud/src/services/update-typings.ts
+++ b/packages/forest-cloud/src/services/update-typings.ts
@@ -6,7 +6,6 @@ import { buildDisconnectedMongooseInstance } from '@forestadmin/datasource-mongo
 import { createMongooseDataSource } from '@forestadmin/datasource-mongoose';
 import { createSequelizeDataSource } from '@forestadmin/datasource-sequelize';
 import { buildDisconnectedSequelizeInstance } from '@forestadmin/datasource-sql';
-import { IntrospectionFormatError } from '@forestadmin/datasource-toolkit';
 import path from 'path';
 
 import { throwIfNoBuiltCode } from './access-file';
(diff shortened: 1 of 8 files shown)


packages/forest-cloud/src/services/update-typings.ts
```diff
@@ -30,12 +32,24 @@ async function buildAgent(
...
+      const mongoose = buildDisconnectedMongooseInstance(introspection);
+      agent.addDataSource(createMongooseDataSource(mongoose, { flattenMode: 'auto' }));
+    } else {
+      const sequelize = await buildDisconnectedSequelizeInstance(introspection, null);
+      agent.addDataSource(createSequelizeDataSource(sequelize));
+    }
+  } catch (e) {
+    const error = e as Error;
+
+    if (error instanceof IntrospectionFormatError) {
```
ghusse, line 46: It will only work if the dependency is shared between `forest-cloud`, `datasource-mongo` and `datasource-sql`. If there is any mismatch in version, this won't work.

Do you think there could be cases where we could have a mismatch?

ghusse: And possibly, with `pnpm` I guess that it could be a problem because all dependencies are installed independently.

realSpok: I could perform another kind of test maybe.
I know `instanceof` compares constructors

Thenkei (approved):

packages/forest-cloud/src/services/update-typings.ts
ghusse: It can be based on `name`

PR #1133, opened 2024-07-03
fix(mongo): mongo datasource throws an error during introspection with mongoose 6 as dependency

Scra3:
linked to: CU-86bzfa0d2
### General

### Security

### DIFF
--- a/packages/datasource-mongo/src/introspection/structure.ts
+++ b/packages/datasource-mongo/src/introspection/structure.ts
@@ -1,6 +1,7 @@
 import type { Binary } from 'bson';
 
 import { ModelStudy, MongoCollection, MongoDb, NodeStudy, Primitive } from './types';
+import VersionManager from '../version-manager';
 
 export default class Structure {
   static async introspect(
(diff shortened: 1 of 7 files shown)


packages/datasource-mongo/package.json
```diff
@@ -15,9 +15,11 @@
     "@forestadmin/datasource-mongoose": "1.7.7",
     "@forestadmin/datasource-toolkit": "1.34.0",
     "json-stringify-pretty-compact": "^3.0.0",
-    "mongoose": "^7.0.1",
     "tunnel-ssh": "^5.1.2"
   },
+  "devDependencies": {
+    "mongoose": "^7.0.1"
+  },
```
Thenkei, line 22: We will need to be careful with cloud ! And also for our client that don't installed `mongoose` when using mongo DS.

Thenkei: Well in DS SQL 👀 

https://github.com/ForestAdmin/agent-nodejs/blob/16b14a50c74e418d9caa83345e883660a576d176/packages/datasource-sql/package.json#L18

Scra3: Fixed. I removed it and tell to our customer to use resolution or override.

Thenkei (approved):
Works for me. :)

PR #1140, opened 2024-07-08
fix(documentDB): prevent field addition from flattener to exceed 30 fields to meet with DocumentDB limitation

SteveBunlon:
### General

### Security

### DIFF
--- a/packages/datasource-mongoose/src/utils/pipeline/reparent.ts
+++ b/packages/datasource-mongoose/src/utils/pipeline/reparent.ts
@@ -74,11 +74,22 @@ export default class ReparentGenerator {
   }
 
   static unflatten(asFields: string[]): PipelineStage[] {
-    return asFields.length
-      ? [
-          { $addFields: Object.fromEntries(asFields.map(f => [f.replace(/\./g, '@@@'), `$${f}`])) },
-          { $project: Object.fromEntries(asFields.map(f => [f, 0])) },
-        ]
-      : [];
+    if (asFields.length === 0) {
+      return [];
+    }
+
+    const unflattenResults = [];
+    const addFields = asFields.map(f => [f.replace(/\./g, '@@@'), `$${f}`]);
+    // DocumentDB limits the addFields stage to 30 fields.
+    const chunkSize = 30;
+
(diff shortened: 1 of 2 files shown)


packages/datasource-mongoose/src/utils/pipeline/reparent.ts
```diff
@@ -74,11 +74,22 @@ export default class ReparentGenerator {
...
-          { $project: Object.fromEntries(asFields.map(f => [f, 0])) },
-        ]
-      : [];
+    if (asFields.length === 0) {
+      return [];
+    }
+
+    const unflattenResults = [];
+    const addFields = asFields.map(f => [f.replace(/\./g, '@@@'), `$${f}`]);
+    // DocumentDB limits the addFields stage to 30 fields.
```
Thenkei, line 83: Is this only a DocumentDB limitation ? Does it have an impact on performance for non DocumentDB DB ?

SteveBunlon: [This](https://docs.aws.amazon.com/documentdb/latest/developerguide/mongo-apis.html#mongo-apis-dababase-aggregation) might also be handy

SteveBunlon: On top of that, I would say that the query system itself is under-performant. I don't why, we do addFields for the flattened fields, where we don't project them. 

If I take a quick look at the query being generated, I have no filters whatsoever, and the system adds my flattened attribute as fields, but doesn't project them in the end 🤔

SteveBunlon: In a more global approach, I tend to think that the query system is under performant. If I don't need any of my flattened fields to be used in a sort or filter, their's no reason to add them using `addFields`. Aliasing them in the $project pipeline would be enough like so : $project: {my@@@nested: $my.nested}. I would rather change this in the regard of performance.

SteveBunlon: I can deep dive into performances if you want me to

Thenkei (approved):
Let's merge it like this and we gonna rework the reparent using only `$project`.

PR #1141, opened 2024-07-10
fix(override-create): clean returned records by removing unknown fields

Scra3:
linked to: CU-86bzh9w5d
### General

### Security

### DIFF
--- a/packages/datasource-customizer/src/decorators/override/collection.ts
+++ b/packages/datasource-customizer/src/decorators/override/collection.ts
@@ -1,4 +1,11 @@
-import { Caller, CollectionDecorator, Filter, RecordData } from '@forestadmin/datasource-toolkit';
+import {
+  Caller,
+  CollectionDecorator,
+  Filter,
+  RecordData,
+  SchemaUtils,
+  ValidationError,
+} from '@forestadmin/datasource-toolkit';
 
 import {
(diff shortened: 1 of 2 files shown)


packages/datasource-customizer/src/decorators/override/collection.ts
```diff
@@ -16,7 +23,10 @@ export default class OverrideCollectionDecorator extends CollectionDecorator {
     if (this.createHandler) {
       const context = new CreateOverrideCustomizationContext(this.childCollection, caller, data);
 
-      return this.createHandler(context);
+      const records = await this.createHandler(context);
+      this.cleanRecordsInPlace(records);
+
+      return records;
```
DayTF, line 29: I would create a new list with only the correct properties instead of mutating the given records

Scra3: I'm not really against but I prefer avoid copy to improve performance. We can have a lot of records here because `create` can be called inside the agent context.

Scra3: Do you have a strong argument to do a copy?

EnkiP: I don't find any, I think your solution is very acceptable

PR #1183, opened 2024-09-25
fix(validation): improve sequelize validation of min max validators

Thenkei:
### General

### Security

### DIFF
--- a/packages/datasource-sequelize/src/utils/model-to-collection-schema-converter.ts
+++ b/packages/datasource-sequelize/src/utils/model-to-collection-schema-converter.ts
@@ -185,11 +185,11 @@ export default class ModelToCollectionSchemaConverter {
       validations.push({
         operator: 'GreaterThan',
         value:
-          (
+          ((
             attribute.validate.min as {
               args: readonly [number];
             }
-          ).args || attribute.validate.min,
+          ).args?.[0] ?? (attribute.validate.min as number)) - 1,
       });
     }
 
@@ -197,11 +197,11 @@ export default class ModelToCollectionSchemaConverter {
       validations.push({
         operator: 'LessThan',
         value:
-          (
+          ((
             attribute.validate.max as {
               args: readonly [number];
             }
-          ).args || attribute.validate.max,
+          ).args?.[0] ?? (attribute.validate.max as number)) + 1,
       });
     }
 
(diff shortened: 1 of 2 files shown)


DayTF (approved):

packages/datasource-sequelize/src/utils/model-to-collection-schema-converter.ts
```diff
@@ -185,23 +185,23 @@ export default class ModelToCollectionSchemaConverter {
       validations.push({
         operator: 'GreaterThan',
```
Thenkei, line 186: I'm gonna have a look at implementing **GreaterThanEqual** and **LessThanEqual**

packages/datasource-sequelize/src/utils/model-to-collection-schema-converter.ts
```diff
@@ -185,23 +185,23 @@ export default class ModelToCollectionSchemaConverter {
       validations.push({
```
Scra3, line 185: Is is working when min or max is equal to 0 ?

Thenkei: Yes, it will say Greater than -1 and Lower than 1. So only 0 will be ok.

Thenkei: We should implement LTE and GTE but it's not easy..

Scra3: I can't comment on it, but if the condition `if(attribute.validate.min)` can't be true when a given value is 0

packages/datasource-sequelize/src/utils/model-to-collection-schema-converter.ts
```diff
@@ -185,23 +185,23 @@ export default class ModelToCollectionSchemaConverter {
...
         operator: 'GreaterThan',
         value:
-          (
+          ((
             attribute.validate.min as {
               args: readonly [number];
             }
-          ).args || attribute.validate.min,
+          ).args?.[0] ?? (attribute.validate.min as number)) - 1,
```
Scra3, line 192: Is args always an Array?

Thenkei:
We will prefer the support of the LTE and GTE operators and validators https://github.com/ForestAdmin/agent-nodejs/pull/1184

Thenkei:
Done with #1184

PR #1199, opened 2024-10-22
feat(forest-cloud): introduce local development environment for cloud projects

Thenkei:
### General

### Security


packages/forest-cloud/src/services/dist-path-manager.ts
```diff
@@ -14,4 +14,8 @@ export default class DistPathManager {
   get distCodeCustomizations(): string {
     return path.join(this.cloudCustomizerPath, 'dist', 'code-customizations');
   }
+
+  get localDatasources(): string {
+    return path.join(this.cloudCustomizerPath, 'datasources.json');
```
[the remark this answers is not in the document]
Thenkei: JS file should looks like.

```
/**
 * @return {Record<string, (import('@forestadmin/forest-cloud').MongoConnectionParams | import('@forestadmin/forest-cloud').SqlConnectionParams)>}
 */
export default () => ({
  main: {
      uri: "postgres://forest:secret@localhost:5435/full-hosted"
  },
  new: {
      uri: "postgres://forest:secret@localhost:5435/full-hosted"
  }
});

```

packages/forest-cloud/src/services/starting-agent-locally.ts
```diff
@@ -0,0 +1,114 @@
...
+>;
+
+// eslint-disable-next-line import/prefer-default-export
+export async function startingAgent(
+  distPathManager: DistPathManager,
+  { forestServerUrl, envSecret },
+  logger: AgentOptions['logger'],
+): Promise<void> {
+  const agentOptions: AgentOptions = {
+    authSecret: 'dev8environment'.repeat(8),
```
DayTF, line 30: nothing really but typo maybe ?
```suggestion
    authSecret: 'dev_environment'.repeat(8),
```

packages/forest-cloud/src/services/http-server.ts
```diff
@@ -154,4 +155,31 @@ export default class HttpServer {
...
   }
+
+  async getOrCreateNewDevelopmentEnvironment() {
+    try {
+      return await handledAxios<{ data: { attributes: { secret_key: string } } }>(
+        {
+          url: `${this.serverUrl}/api/full-hosted-agent/development-environment-for-user`,
+          method: 'POST',
+          headers: this.headers,
+          data: { endpoint: `http://localhost:${process.env.PORT ?? 3351}` },
```
Thenkei, line 166: <3

packages/forest-cloud/src/services/starting-agent-locally.ts
```diff
@@ -0,0 +1,118 @@
...
+
+type Datasources = Record<
+  string,
+  {
+    connectionOptions: ConnectionOptions | ConnectionParams;
+    datasourceSuffix?: string;
+  }
+>;
+
+function getDialectFromConnectionOptions(connectionOptions: ConnectionOptions | ConnectionParams) {
```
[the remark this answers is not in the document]
Thenkei: You right, we have kinda the same one.

packages/forest-cloud/src/commands/start.ts
```diff
@@ -0,0 +1,86 @@
...
+        const vars = await loginIfMissingAuthAndReturnEnvironmentVariables(
+          login,
+          logger,
+          getEnvironmentVariables,
+        );
+        validateEnvironmentVariables(vars);
+
+        // Check that /d.forest/.environments does contains env secret yet
+        if (!fs.existsSync(distPathManager.localCloudEnvironmentConfigPath)) {
+          fs.writeFileSync(distPathManager.localCloudEnvironmentConfigPath, JSON.stringify({}));
```
EnkiP, line 51: maybe put a template like `{ENV_SECRET: "put-your-env-secret-here"}`

packages/forest-cloud/src/services/http-server.ts
```diff
@@ -154,4 +154,31 @@ export default class HttpServer {
...
     return latestVersion(packageName);
   }
+
+  async getOrCreateNewDevelopmentEnvironment() {
+    try {
+      return await handledAxios<{ data: { attributes: { secret_key: string } } }>(
+        {
+          url: `${this.serverUrl}/api/full-hosted-agent/development-environment-for-user`,
+          method: 'POST',
+          headers: this.headers,
```
EnkiP, line 164: why trying to create before trying to get the dev env ?

packages/forest-cloud/src/services/starting-agent-locally.ts
```diff
@@ -0,0 +1,114 @@
...
+
+// eslint-disable-next-line import/prefer-default-export
+export async function startingAgent(
+  distPathManager: DistPathManager,
+  { forestServerUrl, envSecret },
+  logger: AgentOptions['logger'],
+): Promise<void> {
+  const agentOptions: AgentOptions = {
+    authSecret: 'dev8environment'.repeat(8),
+    // New environment variable or fetch it to the server when launching it ? Could be easier from a user perspective (if no dev env we could create it)
```
EnkiP, line 31: Agree, and we could get rid of localCloudEnvironmentsConfig

packages/forest-cloud/src/services/starting-agent-locally.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  return ['mongodb', 'mongodb+srv'].includes(dialect);
+}
+
+async function addDatasourceToAgent(
+  agent: ReturnType<typeof createAgent>,
+  datasources: Datasources,
+) {
+  Object.values(datasources).forEach(async datasource => {
+    const rename = (oldName: string) => `${oldName}${datasource.datasourceSuffix || ''}`;
```
EnkiP, line 94:
```suggestion
    const rename = (oldName: string) => `${oldName || "dev"}${datasource.datasourceSuffix}`;
```

packages/forest-cloud/src/commands/start.ts
Thenkei: No this is an automatically setup file.

packages/forest-cloud/src/services/http-server.ts
Thenkei: Because the dev env does not exist yet most of the time.

packages/forest-cloud/src/services/starting-agent-locally.ts
Thenkei: No 😂
That actually the trick to improve performance. We never remove my initial comment. 🥹

packages/forest-cloud/src/services/http-server.ts
DayTF: We're only calling the createOrGet if the env secret is not present in the config file, the probability of it being a creation rather than a new setup is higher

packages/forest-cloud/src/services/starting-agent-locally.ts
DayTF: I'll remove the comment, indeed having it in the cache is better since we already do a lot of work in Cloud flow commands, I wouldn't want to slow this down.

packages/forest-cloud/src/services/dist-path-manager.ts
```diff
@@ -11,7 +12,31 @@ export default class DistPathManager {
...
   get distCodeCustomizations(): string {
     return path.join(this.cloudCustomizerPath, 'dist', 'code-customizations');
   }
+
+  get distCodeCustomizationsPath(): string {
+    return path.resolve(this.distCodeCustomizations);
+  }
+
+  get localDatasources(): string {
+    return path.join(this.cloudCustomizerPath, 'datasources.js');
```
realSpok, line 28: my file `datasources.js` got generated at the same level as the root, and not inside. I dont think this is expected
image

realSpok: the rest of the code is in a directory a the same level as `datasources.js`

realSpok, line 28: why is this a js file again ? why not ts ? 
The rest of `src/` is ts

packages/forest-cloud/src/commands/start.ts
```diff
@@ -0,0 +1,86 @@
...
+  program
+    .command('start')
+    .description(
+      'Starts the agent locally to allow easier work flow before publish to Cloud Production',
+    )
+    .action(
+      actionRunner(logger.spinner, async () => {
+        await checkLatestVersion(logger.spinner, getCurrentVersion(), HttpServer.getLatestVersion);
+
+        logger.info('Starting agent locally..');
```
realSpok, line 41:
```suggestion
        logger.info('Starting agent locally…');
```

packages/forest-cloud/src/commands/start.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    if (['debug', 'info', 'warn', 'error'].includes(lowerCaseLevel)) {
+      logger[lowerCaseLevel](...args);
+    }
+  };
+
+  program
+    .command('start')
+    .description(
+      'Starts the agent locally to allow easier work flow before publish to Cloud Production',
```
realSpok, line 35:
```suggestion
      'Starts the agent locally to test your code customizations before publishing to your production environment.',
```

packages/forest-cloud/src/services/http-server.ts
```diff
@@ -154,4 +154,31 @@ export default class HttpServer {
...
+          data: { endpoint: `http://localhost:${process.env.PORT ?? 3351}` },
```
realSpok, line 165: there is a dedicated service to inject env variables. could you use it here ?

packages/forest-cloud/src/services/starting-agent-locally.ts
```diff
@@ -0,0 +1,113 @@
...
+    await fs.access(localDatasourcesPath);
+    // eslint-disable-next-line import/no-dynamic-require, global-require, @typescript-eslint/no-var-requires
+    const localDatasources: () => Datasources = require(localDatasourcesPath);
+
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
+    addDatasourceToAgent(agent, localDatasources());
+  } catch (e) {
+    logger('Debug', `Could not load ${localDatasourcesPath}`, e);
+    throw new BusinessError(
+      `No datasources found at ${localDatasourcesPath}.\nPlease provide datasources information..`,
```
realSpok, line 54:
```suggestion
      `No datasources found at ${localDatasourcesPath}.\nPlease provide datasources information…`,
```

packages/forest-cloud/src/services/dist-path-manager.ts
realSpok: Ok, this is due to the require

packages/forest-cloud/src/services/generate-datasource-config-file.ts
```diff
@@ -0,0 +1,24 @@
+import fs from 'fs/promises';
+
+import { BusinessError } from '../errors';
+
+const content = `
```
realSpok, line 5: there is a templates directory where you can put this:  `packages/forest-cloud/src/templates`

packages/forest-cloud/src/commands/start.ts
```diff
@@ -0,0 +1,86 @@
...
+            distPathManager.localCloudEnvironmentConfigPath,
+            JSON.stringify(localCloudEnvironmentsConfig),
+          );
+        }
+
+        // Check that the user has the datasource connection options file
+        if (!fs.existsSync(distPathManager.localDatasourcesPath)) {
+          await generateDatasourceConfigFile(distPathManager.localDatasourcesPath);
+          throw new BusinessError(
+            `Could not find configuration for your local datasource connection options. A new file (${distPathManager.localDatasourcesPath}) has been generated please complete it.`,
```
realSpok, line 75:
```suggestion
            `Could not find configuration for your local datasource connection options. A new file (${distPathManager.localDatasourcesPath}) has been generated. Please complete it with your local database credentials.`,
```

realSpok:
I seem to have a default log level of `debug`, which is very verbose:
image

packages/forest-cloud/src/services/dist-path-manager.ts
realSpok: I think I understand the issue. 
During bootstrap we're 1 directory level higher than during the rest of the commands

realSpok: so you need to use the bootsrap-path-manager

packages/forest-cloud/src/commands/bootstrap.ts
```diff
@@ -58,6 +67,14 @@ export default (program: Command, context: MakeCommands) => {
           }
         }
 
+        // Check that the user has the datasource connection options file
+        if (!fs.existsSync(distPathManager.localDatasourcesPath)) {
+          await generateDatasourceConfigFile(distPathManager.localDatasourcesPath);
+          logger.spinner.info(
+            `Generated a new file (${distPathManager.localDatasourcesPath}) for local development, You can complete it and start an agent locally, with the "start" command.`,
+          );
+        }
```
realSpok, line 76: I think it would be best to move fs logic out of the command, and into a service

packages/forest-cloud/src/services/http-server.ts
```diff
@@ -154,4 +154,31 @@ export default class HttpServer {
   static async getLatestVersion(packageName: string): Promise<string> {
     return latestVersion(packageName);
   }
+
+  async getOrCreateNewDevelopmentEnvironment() {
```
realSpok, line 158: you could pass the `vars` here from the caller (it knows it). Or the PORT directly.

realSpok: this kind of "logic" 
process.env.PORT ?? 3351 
should go in the env vars service

realSpok: what happens if we want to create another enviroment ? 
what happens if the first endpoint port is locally ? can the user change it on forest ?

packages/forest-cloud/src/services/dist-path-manager.ts
DayTF: nicely spotted, updated

DayTF: Yes, the file is outside of the src/, no transpilation is done and we require the content

packages/forest-cloud/src/services/http-server.ts
DayTF: it's not a FOREST env variable, this is done directly by koa if i'm not wrong.

There is no way to create another environment since the rule of one dev env per user per project.
As of now they can't since the product did not greenlight the changes required on the UI Environment settings.

packages/forest-cloud/src/commands/bootstrap.ts
DayTF: Not sure to see the value, as we are not duplicating this behaviour

realSpok: this is more or less how we have done it elsewhere in `forest-cloud`, to move the low level `fs` out of commands and into services
I think it was also easier to test

realSpok: But it is your call

packages/forest-cloud/src/services/generate-datasource-config-file.ts
realSpok: Actually now we have an endpoint that lists datasources. Do we want to call to that to get a more accurate template

Thenkei: I don't think we should mix them.

Thenkei: And you right for you second feedback.

Thenkei: But it's a Nice To Have.

realSpok (approved):
- ✅  Code review
  - ✅  Use pure functions when possible
  - ✅  Performance concerns
  - ✅  Security concerns
- ✅ Automatic tests
  - ✅  Unit tests
  - ⤬  Integration tests
- ✅ Manual tests
  -  ✅  Test the functionality
  -  ✅  Test error cases
- ✅  PR title
- ✅  Destination branch

PR #1211, opened 2024-11-21
chore(capabilities): define native query connection capabilities

DayTF:
### General

- [ ] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)
- [ ] Test manually the implemented changes
- [ ] Validate the code quality (indentation, syntax, style, simplicity, readability)

### Security

- [ ] Consider the security impact of the changes made

PR #1220, opened 2024-11-27
feat(capabilities): define native query connection capabilities

DayTF:
Re-release of #1211  using `feat` tag to release all packages.


jeffladiray (approved):
Previously approved, just for CI purpose.

PR #1223, opened 2024-12-02
fix(sequelize-ds): allow user to create a record when there is a virtual field by removing it from the returned record

Scra3:
linked to: CU-86c173ay7
### General

### Security


packages/datasource-sequelize/test/collection.test.ts
```diff
@@ -115,7 +127,7 @@ describe('SequelizeDataSource > Collection', () => {
     };
 
     it('should delegate work to `sequelize.model.bulkCreate`', async () => {
-      const data = [{ data: 'data' }, { data: 10 }, { data: ['Enum'] }];
```
Scra3, line 118: Data does not exist in the model. I replace it by existing fields.

PR #1228, opened 2024-12-16
fix: upgrade ssh-tunnel dependency to avoid crash when "no response from server"

Scra3:
I upgrade ssh-tunnel dependency because the maintainer merges our contribution to fix this issue: https://github.com/agebrock/tunnel-ssh/pull/121


packages/datasource-mongo/src/connection/create-connection.ts
```diff
@@ -23,6 +23,7 @@ async function createSSHTunnel(uri: string, sshConfig: SshOptions) {
 
   const tunnelOptions: TunnelOptions = {
     autoClose: false,
+    reconnectOnError: false,
```
Scra3, line 26: Type has been changed. The default value is false. The definition type should be optional...

DayTF (approved):

PR #1238, opened 2024-12-31
fix: don't cache rejected promises

BeLi4L:
**Current issue:**
Sometimes our app would be stuck due to the following error in our logs: `"Error: The request to Forest Admin server has timed out while trying to reach https://api.forestadmin.com/liana/v4/permissions/environment"`.
This would make our app unusable (since it can't reach Forest) and we had to restart the pod to make it work again.

**Root cause:**
The code responsible for calling the "https://api.forestadmin.com/liana/v4/permissions/environment" URL, in [ActionPermissionService](https://github.com/ForestAdmin/agent-nodejs/blob/main/packages/forestadmin-client/src/permissions/action-permission.ts#L87), uses a [TTLCache](https://github.com/ForestAdmin/agent-nodejs/blob/main/packages/forestadmin-client/src/utils/ttl-cache.ts), which keeps returning the **same** timed out promise.

**Suggested solution:**
Don't cache rejected promises, so that the following calls will retry until one succeeds.

### DIFF
--- a/packages/forestadmin-client/src/utils/ttl-cache.ts
+++ b/packages/forestadmin-client/src/utils/ttl-cache.ts
@@ -5,11 +5,11 @@
 export default class TTLCache<V> {
   private readonly stateMap = new Map<
     string,
-    { promise: Promise<V | undefined> | V | undefined; expirationTimestamp: number }
+    { promise: Promise<V | undefined>; expirationTimestamp: number }
   >();
 
   constructor(
-    private readonly fetchMethod: (key: string) => Promise<V | undefined> | V | undefined,
+    private readonly fetchMethod: (key: string) => Promise<V | undefined>,
     private readonly ttl = 1000,
   ) {}
 
@@ -23,6 +23,11 @@ export default class TTLCache<V> {
 
     const fetch = this.fetchMethod(key);
     this.stateMap.set(key, { promise: fetch, expirationTimestamp: now + this.ttl });
+
+    fetch.catch(() => {
+      // Don't cache rejected promises
+      this.stateMap.delete(key);
+    });
 
     return fetch;
   }
--- a/packages/forestadmin-client/test/utils/ttl-cache.test.ts
+++ b/packages/forestadmin-client/test/utils/ttl-cache.test.ts
@@ -60,6 +60,22 @@ describe('TTL Cache', () => {
 
       expect(fetchMethod).toHaveBeenCalledTimes(2);
     });
+
+    it('should not cache a rejected promise', async () => {
+      const fetchMethod = jest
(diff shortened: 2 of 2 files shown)


BeLi4L:
Another solution would have been to catch the "Time out" error only in ActionPermissionService, and invalidate the cache there, but I thought it would make sense to put this logic in TTLCache.

Let me know what you think, since I am not that familiar with this project 😉

Scra3:
Hello @BeLi4L,
Thank you so much for your contribution! Your code seems to address a critical bug, and I truly appreciate your effort. I will present your contribution to the tech team in the coming weeks, as many of us are currently on holiday.
Once again, thank you for your help—it’s highly valued and greatly appreciated! 🙏

jeffladiray (approved):
Thanks a lot for this bugfix - We've been on this issue for weeks now, not being able to reproduce it on our internal agent.

Manually tested by forcing `utils/server.ts` to reject via a simulated timeout 

```
      if (
        !path.includes('ip') &&
        !path.includes('hashcheck') &&
        !path.includes('oidc') &&
        !path.includes('liana/v2/renderings') &&
        !path.includes('permissions/users')
      ) {
        const err = new Error();
        err.timeout = true;
        throw err;
      }
``` 
Cache is indeed cleared for the specific key on rejected promise 🙌

(I'm still curious to understand why these timeout appears. Default timeout is ~10s, and all of our monitoring systems seems to indicate pretty good response time (P99>200ms), especially on users/environments set of permissions 🤔)

The CI is failling because of unaccessible credentials from external contribution, so I'll bypass it. Thanks a lot 🙏

BeLi4L:
> (I'm still curious to understand why these timeout appears. Default timeout is ~10s, and all of our monitoring systems seems to indicate pretty good response time (P99>200ms), especially on users/environments set of permissions 🤔)

The TTL is an option given to `TTLCache`, and comes from the [options-validator](https://github.com/ForestAdmin/agent-nodejs/blob/main/packages/agent/src/utils/options-validator.ts#L57
) file:

```ts
// One year cache duration when using events
const DEFAULT_CACHE_DURATION_WITH_EVENTS = 31560000;

// [...]

// When using the event source to refresh cache we set a one year cache duration
copyOptions.permissionsCacheDurationInSeconds = copyOptions.instantCacheRefresh
  ? DEFAULT_CACHE_DURATION_WITH_EVENTS
  : copyOptions.permissionsCacheDurationInSeconds ?? DEFAULT_MINIMUM_CACHE_DURATION * 15;
```

I have no idea why this is done, especially since the `instantCacheRefresh` naming would hint the opposite 😅

PR #1248, opened 2025-01-27
fix(user-permission): don't cache rejected promises

BeLi4L:
(Follows the same issue on action-permission: https://github.com/ForestAdmin/agent-nodejs/pull/1238)

### Current issue:
Sometimes our app would be stuck due to the following error in our logs: "Error: The request to Forest Admin server has timed out while trying to reach https://api.forestadmin.com/liana/v4/permissions/users".
This would make our app unusable (since it can't reach Forest) and we had to restart the pod to make it work again.

### Root cause:
The code responsible for calling the "https://api.forestadmin.com/liana/v4/permissions/users" URL, in [UserPermissionService](https://github.com/ForestAdmin/agent-nodejs/blob/main/packages/forestadmin-client/src/permissions/user-permission.ts), uses a cache which keeps returning the same timed out promise.

### Suggested solution:
Don't cache rejected promises, so that the following calls will retry until one succeeds.

### DIFF
--- a/packages/forestadmin-client/src/permissions/user-permission.ts
+++ b/packages/forestadmin-client/src/permissions/user-permission.ts
@@ -31,7 +31,12 @@ export default class UserPermissionService {
       // the response is not yet available.
       this.userInfoById = this.forestAdminServerInterface
         .getUsers(this.options)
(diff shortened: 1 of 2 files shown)


arnaud-moncel (approved):
🙏

PR #1290, opened 2025-04-24
feat: show a warning in case a mongoose relationship is omitted

realSpok:
## Definition of Done
#86c263jav

### General

### Security

### DIFF
--- a/packages/datasource-customizer/src/decorators/publication/datasource.ts
+++ b/packages/datasource-customizer/src/decorators/publication/datasource.ts
@@ -4,9 +4,9 @@ import PublicationCollectionDecorator from './collection';
 
 export default class PublicationDataSourceDecorator extends DataSourceDecorator<PublicationCollectionDecorator> {
   blacklist: Set<string> = new Set();
-  logger: Logger;
+  logger: Logger | null;
 
-  constructor(childDataSource: DataSource, logger: Logger) {
+  constructor(childDataSource: DataSource, logger?: Logger) {
     super(childDataSource, PublicationCollectionDecorator);
     this.logger = logger;
   }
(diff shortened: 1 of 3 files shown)


packages/datasource-customizer/src/decorators/publication/datasource.ts
```diff
@@ -49,6 +59,17 @@ export default class PublicationDataSourceDecorator extends DataSourceDecorator<
     for (const name of names) this.getCollection(name);
   }
 
+  private warnCollectionWhitelist(excluded: string[], kept: string[]) {
+    if (excluded.length && this.childDataSource.constructor.name === 'MongooseDatasource') {
+      this.logger?.(
+        'Warn',
+        `Using include whitelist for MongooseDatasource may omit virtual collections that define relationships between included collections. Removed collections: '${excluded.join(
```
Scra3, line 66: Is it only for MongooseDatasource ? I thinking about Many to Many

realSpok: I think that sequelize tables are a lot more explicit from the user perspective

realSpok: but I didnt test it

packages/datasource-customizer/src/decorators/publication/datasource.ts
```diff
@@ -49,6 +59,17 @@ export default class PublicationDataSourceDecorator extends DataSourceDecorator<
...
+          "', '",
```
Scra3, line 67: Is it possible to add a back to line because the list could be very large. It can improve the UX by adding this

packages/datasource-customizer/src/decorators/publication/datasource.ts
Scra3, line 66: After merged, could we add a documentation link inside the warning  to document the case ?

packages/datasource-customizer/src/decorators/publication/datasource.ts
realSpok: i think that other messages of this type in the agent do not have line breaks.
having line breaks could mean that the console get overwhelmed with all the collections names.

Scra3 (approved):

PR #1305, opened 2025-05-16
feat(plugin-aws-s3): add force path style parameter to s3 client

JulienRstMttrs:
In plugin-aws-s3 for people using S3 based solution (but not AWS) like Localstack in a local environment we should allow the forcePathStyle option in order to support these solution.

Work perfectly without for getting PreSignedUrl but not with the PutObjectCommand
### General

### Security


JulienRstMttrs:
Thanks a lot @VincentMolinie !

PR #1344, opened 2025-12-03
feat(forest mcp): add the list tools

VincentMolinie:
### General

### Security


packages/mcp-server/src/tools/list.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  return parameters;
+}
+
+export default function declareListTool(mcpServer: McpServer, forestServerUrl: string): void {
+  mcpServer.registerTool(
+    'list',
+    {
+      title: 'List data from the customer agent',
+      description: 'Retrieve a list of data from the specified collection in the customer agent.',
```
EnkiP: I think we should keep the "records" wording, to be consistent with forest docs

PR #1547, opened 2026-04-13
feat(mcp-server): add enabledTools allowlist option

Scra3:
## Summary

- Add `enabledTools` option (allowlist) as an alternative to `disabledTools` (blocklist)
- `enabledTools` is recommended for read-only setups: new tools in future releases won't be automatically enabled
- `enabledTools` takes priority if both options are set (with a warning)
- `describeCollection` is always forced on regardless of configuration
- Startup logs show which tools are enabled and disabled with counts
- Support `FOREST_MCP_ENABLED_TOOLS` env var for standalone mode
- Updated README documenting both approaches with trade-offs

## Test plan

- [x] `enabledTools` only exposes listed tools + `describeCollection` via `tools/list`
- [x] Warning logged when both `enabledTools` and `disabledTools` are set
- [x] `describeCollection` warning preserved when passed to `disabledTools`
- [x] Existing `disabledTools` tests still pass
- [x] Agent passes `enabledTools` through to `ForestMCPServer`
- [x] Build succeeds, lint passes

> [!NOTE]
> ### Replace `disabledTools` denylist with `enabledTools` allowlist in MCP server
> - Replaces the `disabledTools` denylist with an `enabledTools` allowlist across the MCP server, CLI, and agent. Tools must now be explicitly listed to be enabled; unlisted tools are not registered.
> - `describeCollection` is always included regardless of the allowlist. Unknown tool names are ignored with a warning.
> - The CLI now reads `FOREST_MCP_ENABLED_TOOLS` instead of `FOREST_MCP_DISABLED_TOOLS`.
> - `parseDisabledTools` is renamed to `parseToolList` in [parse-tool-list.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1547/files#diff-958d3c5d7b44a4cbbf1aa186175f9a9ea36e5436f796aa608c1c84e5a3c4377d) with no behavior change.
> - Risk: Callers passing `disabledTools` to `mountAiMcpServer` or `FOREST_MCP_DISABLED_TOOLS` in the CLI will have no effect; migration to the new allowlist option is required.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized afac92d.</sup>
> 
>


packages/agent/src/agent.ts
```diff
@@ -208,12 +208,12 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
    * @see {@link https://docs.forestadmin.com/developer-guide-agents-nodejs/agent-customization/ai/mcp-server}
    * @example
    * agent.mountAiMcpServer();
-   * // Or with options:
-   * agent.mountAiMcpServer({ disabledTools: ['create', 'update', 'delete'] });
+   * // Or with options (read-only mode):
```
nbouliol: this is not read only mode, it only mounts the enabled tools

nbouliol (approved):

PR #1561, opened 2026-04-17
fix(mcp server): create activity logs from modelname

nbouliol:
### General

### Security

> [!NOTE]
> ### Fix MCP server activity log creation to use `createMcpActivityLog` instead of `createActivityLog`
> - Adds a new `createMcpActivityLog` method to `ActivityLogsService` and `ForestHttpApi` that POSTs to `/api/activity-logs-requests/mcp` with a flat body including `collectionModelName`.
> - Updates [`activity-logs-creator.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1561/files#diff-8479a6bc2cdaf8ef367e8b278ce15ab7bd390fa48e51363c1f1e068feeea8021) to call `createMcpActivityLog` instead of the generic `createActivityLog`.
> - Changes [`with-activity-log.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1561/files#diff-e4315d245b0931574ae06a34167549cd3e82fb75fd0d8b5a8f593ccb598a9678) to start activity log creation concurrently with the wrapped operation, awaiting the promise only when marking success or failure.
> - Behavioral Change: activity log creation no longer blocks the start of the MCP operation; the two now run concurrently.
>
> 
> #### Changes since #1561 opened
>
> - Added operation-specific labels to activity log context [ef5fc54]
> - Changed `withActivityLog` utility function to synchronously await the creation of a pending activity log before executing the wrapped operation [fa92907]
> - Added `label` field to the context parameter expected by `withActivityLog` in test files [fa92907]
> - Removed `ActivityLogResponse` type from type-only import in `mcp-server` package [dc1e22c]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2f81e96.</sup>
> 
>

### DIFF
--- a/packages/mcp-server/src/utils/with-activity-log.ts
+++ b/packages/mcp-server/src/utils/with-activity-log.ts
@@ -42,10 +42,11 @@ export default async function withActivityLog<T>(options: WithActivityLogOptions
 
   // We want to create the activity log before executing the operation
   // If activity log creation fails, we must prevent the execution of the operation
-  const activityLog = await createPendingActivityLog(forestServerClient, request, action, context);
+  const activityLogPromise = createPendingActivityLog(forestServerClient, request, action, context);
 
   try {
     const result = await operation();
+    const activityLog = await activityLogPromise;
 
     markActivityLogAsSucceeded({
       forestServerClient,
@@ -76,7 +77,7 @@ export default async function withActivityLog<T>(options: WithActivityLogOptions
     markActivityLogAsFailed({
       forestServerClient,
       request,
-      activityLog,
+      activityLog: await activityLogPromise,
       errorMessage,
       logger,
     });
--- a/packages/mcp-server/src/utils/with-activity-log.ts
+++ b/packages/mcp-server/src/utils/with-activity-log.ts
@@ -1,5 +1,5 @@
 import type { ActivityLogAction } from './activity-logs-creator';
-import type { ActivityLogResponse, ForestServerClient } from '../http-client';
+import type { ForestServerClient } from '../http-client';
 import type { Logger } from '../server';
 import type { RequestHandlerExtra } from '@modelcontextprotocol/sdk/shared/protocol.js';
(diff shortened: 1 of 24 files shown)


packages/mcp-server/src/utils/with-activity-log.ts
```diff
@@ -76,7 +77,7 @@ export default async function withActivityLog<T>(options: WithActivityLogOptions
     markActivityLogAsFailed({
       forestServerClient,
       request,
-      activityLog,
+      activityLog: await activityLogPromise,
```
DayTF: The promise could already be resolved here

nbouliol: we use it in both the try and the catch, so we need to keep the promise unresovled

packages/mcp-server/src/utils/with-activity-log.ts
```diff
@@ -42,10 +42,11 @@ export default async function withActivityLog<T>(options: WithActivityLogOptions
 
   // We want to create the activity log before executing the operation
   // If activity log creation fails, we must prevent the execution of the operation
-  const activityLog = await createPendingActivityLog(forestServerClient, request, action, context);
+  const activityLogPromise = createPendingActivityLog(forestServerClient, request, action, context);
```
DayTF: This is not really related to the current PR but, the createPendingActivityLog can create activities without label which is not the case for activities created from the UI for the same type of actions: create, update, deleted, they have, as label, respectively: 'created', 'updated', 'deleted'
See: app/utils/collection-creator/adapter/basic-adapter.js

packages/mcp-server/src/utils/with-activity-log.ts
DayTF: it could also be rejected so it can throw here too I guess

DayTF (approved):

PR #1565, opened 2026-04-22
fix(agent-client): accept composite PKs as arrays

Scra3:
## Summary

- Widen `update`, `delete`, `relation`, `action`, `associate` and `dissociate` on `@forestadmin/agent-client` to accept `string | number | Array<string | number>`.
- Centralize pipe-encoding of composite primary keys inside the library via a new internal helper (`encodeRecordId`) + exported `RecordId` type.
- Backward-compatible: all existing scalar PK callers keep working unchanged.

## Why

The public API was inconsistent on composite primary keys: `list({ filters })` accepts structured multi-field PKs, but `update`, `relation` and `action` all required the caller to pre-pipe-encode the id (`"k1|k2"`). The convention leaked into every consumer — each reinvents its own `encodePk` helper.

This PR makes the convention a library-internal detail. Consumers can now pass `[k1, k2]` natively; the library handles the encoding.

## Follow-up

- Clean up `encodePk` / manual pipe-joining in `packages/workflow-executor/src/adapters/agent-client-agent-port.ts` once the workflow-executor package lands on `main` (currently on its feature branch, so out of scope here).

## Test plan

- [x] `yarn workspace @forestadmin/agent-client build`
- [x] `yarn workspace @forestadmin/agent-client lint` (0 errors)
- [x] `yarn workspace @forestadmin/agent-client test` — 312/312 passing, including new composite cases on update / delete / relation / action / associate / dissociate + 5 unit tests on `encodeRecordId`
- [x] Non-regression on `@forestadmin/mcp-server` — 540/540
- [x] Non-regression on `@forestadmin/agent-testing` — 33/33

> [!NOTE]
> ### Accept composite PKs as arrays in agent-client collection, relation, and action methods
> - Introduces a `RecordId = string | number | Array<string | number>` type and an `encodeRecordId` utility in [record-id.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1565/files#diff-4b446e96098e791d7789156e10404fb0d4148d85b5f02611ab4632934bc684ac) that serializes composite ids to pipe-joined strings (e.g. `[1, 2]` → `"1|2"`).
> - Updates `Collection.update`, `Collection.delete`, `Collection.action`, `Collection.relation`, and all `Relation` methods to accept `RecordId` / `RecordId[]` instead of plain `string | number`.
> - Composite ids are pipe-encoded before being placed in URL paths and request bodies; the encoded form is URL-encoded in paths (e.g. `%7C`).
> - Risk: passing an empty array, a part containing `'|'`, or a `null`/`undefined` part now throws at call time rather than sending a malformed id silently.
>
> 
> #### Changes since #1565 opened
>
> - Renamed `encodeRecordId` to `serializeRecordId` in the `agent-client` package [0dd5d99]
> - renamed local variable in Collection.update method [3db0fd2]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e52b00e.</sup>
> 
>
>

### DIFF
--- a/packages/agent-client/src/record-id.ts
+++ b/packages/agent-client/src/record-id.ts
@@ -1,10 +1,12 @@
 import type { RecordId } from './types';
 
 // Forest Admin backend expects composite PKs pipe-joined (e.g. "k1|k2"). Centralizing the
-// encoding here lets callers pass structured arrays and keeps the convention out of their code.
-// Throws rather than silently corrupting the key when parts are nullish or contain the "|"
-// separator — either case would produce a malformed id that could match the wrong record.
-export default function encodeRecordId(id: RecordId): string {
+// serialization here lets callers pass structured arrays and keeps the convention out of their
+// code. URL-encoding of the resulting string (e.g. "|" → "%7C" in paths) is handled downstream
+// by HttpRequester. Throws rather than silently corrupting the key when parts are nullish or
+// contain the "|" separator — either case would produce a malformed id that could match the
+// wrong record.
+export default function serializeRecordId(id: RecordId): string {
   if (!Array.isArray(id)) return String(id);
   if (id.length === 0) throw new Error('Composite record id cannot be empty');
 
--- a/packages/agent-client/src/record-id.ts
+++ b/packages/agent-client/src/record-id.ts
@@ -1,0 +1,7 @@
+import type { RecordId } from './types';
+
+// Forest Admin backend expects composite PKs pipe-joined (e.g. "k1|k2"). Centralizing the
+// encoding here lets callers pass structured arrays and keeps the convention out of their code.
+export default function encodeRecordId(id: RecordId): string {
+  return Array.isArray(id) ? id.map(v => String(v)).join('|') : String(id);
+}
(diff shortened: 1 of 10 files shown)


packages/agent-client/src/record-id.ts
```diff
@@ -0,0 +1,28 @@
+import type { RecordId } from './types';
+
+// Forest Admin backend expects composite PKs pipe-joined (e.g. "k1|k2"). Centralizing the
+// encoding here lets callers pass structured arrays and keeps the convention out of their code.
+// Throws rather than silently corrupting the key when parts are nullish or contain the "|"
+// separator — either case would produce a malformed id that could match the wrong record.
+export default function encodeRecordId(id: RecordId): string {
```
arnaud-moncel: This function have a misleading name IMO I thought the result was `1%7Cabc` and not `1|abc`.
We should encode like this when we use it on an url.
I assume httpRequester.query do some encode uri so you can just rename the function what do you think ?

Scra3: Good catch — the URL-encoding is indeed already handled downstream by `HttpRequester.escapeUrlSlug` (via `encodeURI`, which turns `|` into `%7C` in paths). Renamed to `serializeRecordId` in 0dd5d99 and updated the header comment to spell out the separation of concerns.

---
_Generated by [Claude Code](https://claude.ai/code)_

arnaud-moncel (approved):

PR #1583, opened 2026-05-18
fix(workflow-executor): align getMcpServerConfigs port with Record shape

hercemer42:
## Summary

- `Runner.fetchRemoteTools()` crashed with `TypeError: configs.map is not a function` on every MCP-typed step. The orchestrator's `GET /liana/mcp-server-configs-with-details` returns `Record<string, ToolConfig>`, but the executor port was typed `Promise<McpConfiguration[]>` and the runner called `.map` on the object.
- Aligns the executor with the wire contract: port now returns `McpServers` (`Record<string, McpServerConfig>`, from `@forestadmin/ai-proxy`), runner wraps the record into `{ configs }` for `loadRemoteTools`, with an empty-Record short-circuit.
- Updates 9 test mock sites (8 `mockResolvedValue([])` plus the `[{ configs: {} }] as never` cast in `runner.test.ts`) to the real Record shape so the contract can't silently regress again. Adds two TDD tests under `MCP lazy loading (via once thunk)` for the non-empty-Record happy path and the empty-Record short-circuit.

## Dependency bumps (heads-up)

`packages/workflow-executor/package.json` bumps:
- `@forestadmin/ai-proxy` 1.8.0 → 1.9.0 — load-bearing for this fix (the `McpServers` type only exists from 1.9.0 onward).
- `@forestadmin/forestadmin-client` 1.39.5 → 1.39.7 — incidental, was already staged in the user's pre-existing main-branch state.

These cannot be split into a separate PR because the workspace-resolved import in `src/ports/workflow-port.ts` would fail to compile against ai-proxy 1.8.0.

## Test plan

- [x] `yarn workspace @forestadmin/workflow-executor jest --runInBand --forceExit --testMatch='**/runner.test.ts'` — 83/83 pass (including 3 new MCP-lazy-loading tests)
- [x] `yarn workspace @forestadmin/workflow-executor jest --runInBand --forceExit` on the 9 changed test files — 361 tests green
- [x] `npx eslint`, `npx prettier --check`, `npx tsc --noEmit` on all 13 changed files — clean
- [x] End-to-end repro against `data-gouv` MCP server on the `executor` dev project — pre-fix run aborted with the generic `"An unexpected error occurred."` (masking the TypeError); post-fix run reports `NoMcpToolsError` cleanly, which is the legitimate downstream behaviour when no tools materialize

Refs: PRD-357

> [!NOTE]
> ### Fix `getMcpServerConfigs` return type to use `McpServers` Record shape in workflow-executor
> - Replaces `McpConfiguration[]` array with `McpServers` (a `Record<string, ToolConfig>` map) as the return type of `getMcpServerConfigs` across [workflow-port.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1583/files#diff-5d7cbcc8082d4fb62a74fb3912a92940e4a2888aa87ec7fb925f3b747948b3f4), [forest-server-workflow-port.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1583/files#diff-70b42aec8454fc9dce637bd3559a6e947379c05eea7f496ee4dac1e9323c8c5d), and the public exports in [index.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1583/files#diff-f3276eeb41201a9f098e7daa674ac8c853f6cda66d051427c7c43c5bec3ebb1c).
> - Updates `Runner.fetchRemoteTools` in [runner.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1583/files#diff-53dd42328518d1424983b180f4f92da938c107f6c821327fd03f8f768ffa7e89) to skip calling `aiModelPort.loadRemoteTools` when the config map is empty, and passes `{ configs }` (the Record wrapped in an object) when non-empty.
> - Bumps `@forestadmin/ai-proxy` to 1.9.0 and `@forestadmin/forestadmin-client` to 1.39.7.
> - Behavioral Change: `loadRemoteTools` is no longer called when the orchestrator returns an empty MCP config map; callers relying on the old array format for `getMcpServerConfigs` will need to update to the Record shape.
>
> 
> #### Changes since #1583 opened
>
> - Fixed `McpStepExecutor.getFilteredTools` to filter tools by `tool.mcpServerId` instead of `tool.sourceId` when a step specifies `mcpServerId`, and updated `NoMcpToolsError` to include diagnostic context showing the requested `mcpServerId` and list of available `mcpServerIds` when filtering yields no matches [53618a7]
> - Added optional `mcpServerId` property to `RemoteTool` class and threaded it through `ServerRemoteTool` and `McpServerRemoteTool` subclass constructors [53618a7]
> - Modified `McpClient` to extract `id` from each server config entry, store it in an internal `mcpServerIdsByName` map, and populate `mcpServerId` on `McpServerRemoteTool` instances during tool loading, and changed `McpConfiguration.configs` type from `MultiServerMCPClient['config']['mcpServers']` to `Record<string, McpServerConfig>` where `McpServerConfig` includes optional `id?: string` [53618a7]
> - Extended `ForestIntegrationConfig` interface with optional `id?: string` and updated `ForestIntegrationClient.loadTools` to thread this `id` as `mcpServerId` parameter to `getZendeskTools`, `getKolarTools`, and `getSnowflakeTools` factory functions, which were each extended to accept optional `mcpServerId` and pass it to `ServerRemoteTool` constructor [53618a7]
> - Added test coverage for `mcpServerId` propagation across `ForestIntegrationClient`, integration tool factories (`getZendeskTools`, `getKolarTools`, `getSnowflakeTools`), `McpClient`, `McpStepExecutor` filtering behavior, and `NoMcpToolsError` message construction [53618a7]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 480148c.</sup>
> 
>
>

### DIFF
--- a/packages/workflow-executor/package.json
+++ b/packages/workflow-executor/package.json
@@ -27,9 +27,9 @@
   },
   "dependencies": {
     "@forestadmin/agent-client": "1.5.6",
-    "@forestadmin/ai-proxy": "1.8.0",
+    "@forestadmin/ai-proxy": "1.9.0",
     "@langchain/openai": "1.2.5",
-    "@forestadmin/forestadmin-client": "1.39.5",
+    "@forestadmin/forestadmin-client": "1.39.7",
     "@koa/bodyparser": "^6.1.0",
     "@koa/router": "^13.1.0",
     "jsonwebtoken": "^9.0.3",
--- a/packages/workflow-executor/src/index.ts
+++ b/packages/workflow-executor/src/index.ts
@@ -68,7 +68,7 @@ export type {
   Limit,
   UpdateRecordQuery,
 } from './ports/agent-port';
-export type { McpConfiguration, McpServers, WorkflowPort } from './ports/workflow-port';
+export type { McpServers, WorkflowPort } from './ports/workflow-port';
 export type { RunStore } from './ports/run-store';
 export type { Logger } from './ports/logger-port';
 export { default as ConsoleLogger } from './adapters/console-logger';
--- a/packages/workflow-executor/src/index.ts
+++ b/packages/workflow-executor/src/index.ts
@@ -68,7 +68,7 @@ export type {
   Limit,
   UpdateRecordQuery,
 } from './ports/agent-port';
-export type { McpConfiguration, WorkflowPort } from './ports/workflow-port';
+export type { McpConfiguration, McpServers, WorkflowPort } from './ports/workflow-port';
 export type { RunStore } from './ports/run-store';
 export type { Logger } from './ports/logger-port';
 export { default as ConsoleLogger } from './adapters/console-logger';
--- a/packages/workflow-executor/src/ports/workflow-port.ts
+++ b/packages/workflow-executor/src/ports/workflow-port.ts
@@ -3,9 +3,9 @@
 import type { AvailableStepExecution, StepUser } from '../types/execution-context';
 import type { CollectionSchema } from '../types/validated/collection';
 import type { StepOutcome } from '../types/validated/step-outcome';
-import type { McpConfiguration, McpServers } from '@forestadmin/ai-proxy';
+import type { McpServers } from '@forestadmin/ai-proxy';
 
-export type { McpConfiguration, McpServers };
+export type { McpServers };
 
 export interface MalformedRunInfo {
   runId: string;
@@ -39,8 +39,6 @@ export interface WorkflowPort {
     stepOutcome: StepOutcome,
   ): Promise<AvailableRunDispatch | null>;
   getCollectionSchema(collectionName: string, runId: string): Promise<CollectionSchema>;
-  // The orchestrator's GET /liana/mcp-server-configs-with-details returns a
-  // Record<string, ToolConfig> map keyed by server name — not an array.
   getMcpServerConfigs(): Promise<McpServers>;
   hasRunAccess(runId: string, user: StepUser): Promise<boolean>;
 }
--- a/packages/workflow-executor/src/ports/workflow-port.ts
+++ b/packages/workflow-executor/src/ports/workflow-port.ts
@@ -3,9 +3,9 @@
 import type { AvailableStepExecution, StepUser } from '../types/execution-context';
 import type { CollectionSchema } from '../types/validated/collection';
 import type { StepOutcome } from '../types/validated/step-outcome';
-import type { McpConfiguration } from '@forestadmin/ai-proxy';
+import type { McpConfiguration, McpServers } from '@forestadmin/ai-proxy';
 
-export type { McpConfiguration };
+export type { McpConfiguration, McpServers };
 
 export interface MalformedRunInfo {
   runId: string;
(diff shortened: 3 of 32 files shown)


packages/workflow-executor/src/ports/workflow-port.ts
```diff
@@ -39,6 +39,8 @@ export interface WorkflowPort {
     stepOutcome: StepOutcome,
   ): Promise<AvailableRunDispatch | null>;
   getCollectionSchema(collectionName: string, runId: string): Promise<CollectionSchema>;
-  getMcpServerConfigs(): Promise<McpConfiguration[]>;
+  // The orchestrator's GET /liana/mcp-server-configs-with-details returns a
+  // Record<string, ToolConfig> map keyed by server name — not an array.
```
hercemer42: Unnecessary comment

packages/workflow-executor/src/index.ts
```diff
@@ -68,7 +68,7 @@ export type {
   Limit,
   UpdateRecordQuery,
 } from './ports/agent-port';
-export type { McpConfiguration, WorkflowPort } from './ports/workflow-port';
+export type { McpConfiguration, McpServers, WorkflowPort } from './ports/workflow-port';
```
hercemer42: Do we still need to import McpConfiguration ? I don't see it in use in this file.

packages/workflow-executor/test/runner.test.ts
```diff
@@ -1162,8 +1162,10 @@ describe('MCP lazy loading (via once thunk)', () => {
       step,
       auth: { forestServerToken: 'test-forest-token' },
     });
-    // Provide a non-empty config so fetchRemoteTools actually calls loadRemoteTools
-    workflowPort.getMcpServerConfigs.mockResolvedValue([{ configs: {} }] as never);
+    const realConfigs = {
+      'data-gouv': { url: 'https://mcp.example.com', type: 'http' as const, headers: {} },
```
hercemer42: make data-gouv more abstract, confusing to use a real world reference

packages/workflow-executor/src/ports/workflow-port.ts
hercemer42: Fair — dropped in 32c1afba. The `McpServers` type plus the route URL in the adapter carry the contract well enough on their own.

packages/workflow-executor/src/index.ts
hercemer42: Good catch — not needed. Before this PR the port returned `McpConfiguration[]` so the re-export was load-bearing. Now the port returns `McpServers` and `McpConfiguration` is only used internally by `AiModelPort` / adapters, imported directly from `@forestadmin/ai-proxy`. Removed from both the index barrel and `workflow-port.ts` in 32c1afba.

packages/workflow-executor/test/runner.test.ts
hercemer42: Agreed — renamed to `mcp-server-1` in both the runner test and the adapter test in 32c1afba.

packages/workflow-executor/package.json
```diff
@@ -27,9 +27,9 @@
   },
   "dependencies": {
     "@forestadmin/agent-client": "1.5.6",
-    "@forestadmin/ai-proxy": "1.8.0",
+    "@forestadmin/ai-proxy": "1.9.0",
     "@langchain/openai": "1.2.5",
-    "@forestadmin/forestadmin-client": "1.39.5",
+    "@forestadmin/forestadmin-client": "1.39.7",
     "@koa/bodyparser": "^6.1.0",
```
hercemer42, line 33: @Scra3 can you confirm it's ok to bump these deps ?

hercemer42: I think it should be ok, the bumps are in main, I don't think it'll break anything in the work branch ?

Scra3: You can bump dependancies without fear :)

matthv (approved):
LGTM. Clean fix 🏅

PR #1584, opened 2026-05-19
fix(workflow-executor): match MCP tools by config id

hercemer42:
## Summary

- ai-proxy: threads a stable DB `id` from each `ToolConfig` entry through `McpClient`, `ForestIntegrationClient` and the integration tool factories (`getZendeskTools`, `getKolarTools`, `getSnowflakeTools`) onto `RemoteTool.id`
- workflow-executor: `McpStepExecutor.getFilteredTools` now matches `tool.id === step.mcpServerId` (was `tool.sourceId`), so workflows targeting a specific MCP server actually resolve their tools instead of always hitting `NoMcpToolsError`
- workflow-executor: `NoMcpToolsError` carries the misconfigured `mcpServerId` and the list of loaded config ids in its technical message; user-facing message stays generic per the dual-message convention

fixes PRD-362

## Stacked on

Targets `fix/prd-357-fetchremotetools-record-shape` (#1583). PR #1583 introduces the `Record<string, ToolConfig>` shape this change consumes. Once #1583 merges into `feat/prd-214-server-step-mapper`, rebase this onto the same base.

## Prerequisite

End-to-end resolution also depends on the orchestrator (forestadmin-server, PRD-360) exposing the DB `id` on each `Record<string, ToolConfig>` entry, including Forest-connector entries. Until that lands, the executor still fires `NoMcpToolsError` with an enriched diagnostic message — which is now actionable from the activity log.

## Test plan

- [ ] CI passes (ai-proxy + workflow-executor)
- [ ] Manual: run a workflow whose MCP step targets a specific MCP server config id — tool resolves and runs
- [ ] Manual: run a workflow whose MCP step targets a Forest-connector-backed config (Zendesk/Kolar/Snowflake) — tool resolves and runs
- [ ] Manual: deliberately set `mcpServerId` to a non-existent id and check the activity log carries the requested id + loaded id list

> [!NOTE]
> ### Fix MCP tool filtering in `McpStepExecutor` to match by config `id` instead of `sourceId`
> - Adds an optional `id` field to `McpServerConfig` and `ForestIntegrationConfig`, which is threaded through tool construction so each `RemoteTool` instance carries a matching `mcpServerId` property.
> - Updates `McpStepExecutor.getFilteredTools` to filter tools by `tool.mcpServerId` instead of `sourceId`, fixing cross-config tool leakage.
> - Improves `NoMcpToolsError` to include the requested `mcpServerId` and list of loaded IDs in its technical message for easier diagnostics.
> - Behavioral Change: steps that previously matched tools by `sourceId` will now only match tools whose `mcpServerId` aligns with the step's `mcpServerId`; steps relying on the old `sourceId` matching will no longer find tools.
>
> 
> #### Changes since #1584 opened
>
> - Renamed MCP server ID variables and map properties to explicitly indicate they store MCP server identifiers [ceacd35]
> - Added explanatory comment to `mcpServerId` property in `RemoteTool` class documenting shared usage across `McpServerRemoteTool` and `ServerRemoteTool` [095113b]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4edc2e4.</sup>
> 
>
>

### DIFF
--- a/packages/ai-proxy/src/forest-integration-client.ts
+++ b/packages/ai-proxy/src/forest-integration-client.ts
@@ -14,6 +14,7 @@ export type CustomConfig = ZendeskConfig | KolarConfig | SnowflakeConfig;
 export type ForestIntegrationName = 'Zendesk' | 'Kolar' | 'Snowflake';
 
 export interface ForestIntegrationConfig {
+  id: string;
   integrationName: ForestIntegrationName;
   config: CustomConfig;
   isForestConnector: true;
@@ -39,16 +40,16 @@ export default class ForestIntegrationClient implements ToolProvider {
   async loadTools(): Promise<RemoteTool[]> {
     const tools: RemoteTool[] = [];
 
-    this.configs.forEach(({ integrationName, config }) => {
+    this.configs.forEach(({ id, integrationName, config }) => {
       switch (integrationName) {
         case 'Zendesk':
-          tools.push(...getZendeskTools(config as ZendeskConfig));
+          tools.push(...getZendeskTools(config as ZendeskConfig, id));
           break;
         case 'Kolar':
-          tools.push(...getKolarTools(config as KolarConfig));
+          tools.push(...getKolarTools(config as KolarConfig, id));
           break;
         case 'Snowflake':
-          tools.push(...getSnowflakeTools(config as SnowflakeConfig));
+          tools.push(...getSnowflakeTools(config as SnowflakeConfig, id));
           break;
         default:
           this.logger?.('Warn', `Unsupported integration: ${integrationName}`);
--- a/packages/ai-proxy/src/forest-integration-client.ts
+++ b/packages/ai-proxy/src/forest-integration-client.ts
@@ -14,7 +14,7 @@ export type CustomConfig = ZendeskConfig | KolarConfig | SnowflakeConfig;
 export type ForestIntegrationName = 'Zendesk' | 'Kolar' | 'Snowflake';
 
 export interface ForestIntegrationConfig {
-  id: string;
+  id?: string;
   integrationName: ForestIntegrationName;
   config: CustomConfig;
   isForestConnector: true;
--- a/packages/ai-proxy/src/forest-integration-client.ts
+++ b/packages/ai-proxy/src/forest-integration-client.ts
@@ -40,16 +40,16 @@ export default class ForestIntegrationClient implements ToolProvider {
   async loadTools(): Promise<RemoteTool[]> {
     const tools: RemoteTool[] = [];
 
-    this.configs.forEach(({ id, integrationName, config }) => {
+    this.configs.forEach(({ id: mcpServerId, integrationName, config }) => {
       switch (integrationName) {
         case 'Zendesk':
-          tools.push(...getZendeskTools(config as ZendeskConfig, id));
+          tools.push(...getZendeskTools(config as ZendeskConfig, mcpServerId));
           break;
         case 'Kolar':
-          tools.push(...getKolarTools(config as KolarConfig, id));
+          tools.push(...getKolarTools(config as KolarConfig, mcpServerId));
           break;
         case 'Snowflake':
-          tools.push(...getSnowflakeTools(config as SnowflakeConfig, id));
+          tools.push(...getSnowflakeTools(config as SnowflakeConfig, mcpServerId));
           break;
         default:
           this.logger?.('Warn', `Unsupported integration: ${integrationName}`);
--- a/packages/ai-proxy/src/mcp-client.ts
+++ b/packages/ai-proxy/src/mcp-client.ts
@@ -8,9 +8,6 @@ import McpServerRemoteTool from './mcp-server-remote-tool';
 
 export type McpServers = MultiServerMCPClient['config']['mcpServers'];
 
-// Base orchestrator entry shape from `@langchain/mcp-adapters`, widened with the optional
-// stable DB `id` exposed by Forest Admin's orchestrator. The id flows down to RemoteTool
-// so the workflow executor can match a step's `mcpServerId` against the config record.
 export type McpServerConfig = MultiServerMCPClient['config']['mcpServers'][string] & {
   id?: string;
 };
--- a/packages/ai-proxy/src/mcp-client.ts
+++ b/packages/ai-proxy/src/mcp-client.ts
@@ -45,7 +45,11 @@ export default class McpClient implements ToolProvider {
           const loadedTools = (await client.getTools()) ?? [];
           const extendedTools = loadedTools.map(
             tool =>
-              new McpServerRemoteTool({ tool, sourceId: name, id: this.idsByServerName[name] }),
+              new McpServerRemoteTool({
+                tool,
+                sourceId: name,
+                mcpServerId: this.idsByServerName[name],
+              }),
           );
           tools.push(...extendedTools);
         } catch (error) {
--- a/packages/ai-proxy/src/mcp-client.ts
+++ b/packages/ai-proxy/src/mcp-client.ts
@@ -18,7 +18,7 @@ export type McpConfiguration = {
 
 export default class McpClient implements ToolProvider {
   private readonly mcpClients: Record<string, MultiServerMCPClient> = {};
-  private readonly idsByServerName: Record<string, string | undefined> = {};
+  private readonly mcpServerIdsByName: Record<string, string | undefined> = {};
   private readonly logger?: Logger;
 
   constructor(config: McpConfiguration, logger?: Logger) {
@@ -26,8 +26,9 @@ export default class McpClient implements ToolProvider {
     // split the config into several clients to be more resilient
     // if a mcp server is down, the others will still work
     Object.entries(config.configs).forEach(([name, serverConfig]) => {
-      const { id, ...rest } = serverConfig as McpServerConfig & Record<string, unknown>;
-      this.idsByServerName[name] = id;
+      const { id: mcpServerId, ...rest } = serverConfig as McpServerConfig &
+        Record<string, unknown>;
+      this.mcpServerIdsByName[name] = mcpServerId;
       this.mcpClients[name] = new MultiServerMCPClient({
         mcpServers: { [name]: rest as McpServerConfig },
         ...config,
@@ -48,7 +49,7 @@ export default class McpClient implements ToolProvider {
               new McpServerRemoteTool({
                 tool,
                 sourceId: name,
-                mcpServerId: this.idsByServerName[name],
+                mcpServerId: this.mcpServerIdsByName[name],
               }),
           );
           tools.push(...extendedTools);
--- a/packages/ai-proxy/src/mcp-client.ts
+++ b/packages/ai-proxy/src/mcp-client.ts
@@ -8,14 +8,20 @@ import McpServerRemoteTool from './mcp-server-remote-tool';
 
 export type McpServers = MultiServerMCPClient['config']['mcpServers'];
 
-export type McpServerConfig = MultiServerMCPClient['config']['mcpServers'][string];
+// Base orchestrator entry shape from `@langchain/mcp-adapters`, widened with the optional
+// stable DB `id` exposed by Forest Admin's orchestrator. The id flows down to RemoteTool
+// so the workflow executor can match a step's `mcpServerId` against the config record.
+export type McpServerConfig = MultiServerMCPClient['config']['mcpServers'][string] & {
+  id?: string;
+};
 
 export type McpConfiguration = {
-  configs: McpServers;
+  configs: Record<string, McpServerConfig>;
 } & Omit<MultiServerMCPClient['config'], 'mcpServers'>;
 
 export default class McpClient implements ToolProvider {
   private readonly mcpClients: Record<string, MultiServerMCPClient> = {};
+  private readonly idsByServerName: Record<string, string | undefined> = {};
   private readonly logger?: Logger;
 
   constructor(config: McpConfiguration, logger?: Logger) {
@@ -23,8 +29,10 @@ export default class McpClient implements ToolProvider {
     // split the config into several clients to be more resilient
     // if a mcp server is down, the others will still work
     Object.entries(config.configs).forEach(([name, serverConfig]) => {
+      const { id, ...rest } = serverConfig as McpServerConfig & Record<string, unknown>;
+      this.idsByServerName[name] = id;
       this.mcpClients[name] = new MultiServerMCPClient({
-        mcpServers: { [name]: serverConfig },
+        mcpServers: { [name]: rest as McpServerConfig },
         ...config,
       });
     });
@@ -39,7 +47,8 @@ export default class McpClient implements ToolProvider {
         try {
           const loadedTools = (await client.getTools()) ?? [];
           const extendedTools = loadedTools.map(
-            tool => new McpServerRemoteTool({ tool, sourceId: name }),
+            tool =>
+              new McpServerRemoteTool({ tool, sourceId: name, id: this.idsByServerName[name] }),
           );
           tools.push(...extendedTools);
         } catch (error) {
--- a/packages/ai-proxy/src/remote-tool.ts
+++ b/packages/ai-proxy/src/remote-tool.ts
@@ -4,18 +4,18 @@ export default abstract class RemoteTool<ToolType = unknown> {
   base: StructuredToolInterface<ToolType>;
   sourceId: string;
   sourceType: string;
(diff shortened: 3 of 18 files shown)


packages/ai-proxy/src/mcp-client.ts
```diff
@@ -8,23 +8,31 @@ import McpServerRemoteTool from './mcp-server-remote-tool';
...
   private readonly mcpClients: Record<string, MultiServerMCPClient> = {};
+  private readonly idsByServerName: Record<string, string | undefined> = {};
   private readonly logger?: Logger;
 
   constructor(config: McpConfiguration, logger?: Logger) {
     this.logger = logger;
     // split the config into several clients to be more resilient
     // if a mcp server is down, the others will still work
     Object.entries(config.configs).forEach(([name, serverConfig]) => {
+      const { id, ...rest } = serverConfig as McpServerConfig & Record<string, unknown>;
```
hercemer42: [Test gap] The destructure-and-spread that strips `id` before passing the config to `MultiServerMCPClient` is the centerpiece of the production change here, but no test asserts the field is actually absent from the payload handed downstream. The two new tests in `mcp-client.test.ts` only check `tools[0].id` on the result. With `MultiServerMCPClient` already mocked, you can exercise this directly: `expect(MultiServerMCPClient).toHaveBeenCalledWith(expect.objectContaining({ mcpServers: { slack: expect.not.objectContaining({ id: expect.anything() }) } }))`. Mitigating factor: `@langchain/mcp-adapters` uses Zod `.strip()` so the field would be silently dropped even without this code — the stripping is belt-and-suspenders. Still, the code change deserves an assertion or someone will revert the destructure under the assumption it's a no-op.

hercemer42: Pushing back — the destructure-and-spread is belt-and-suspenders against `@langchain/mcp-adapters` flipping its Zod schema from `.strip()` to `.passthrough()`, which is an external-dependency concern, not user-visible behavior. The tests do verify the actual outcome (`tools[0].id === 'config-id-42'` after `loadTools`). An `expect.not.objectContaining({ id })` assertion on the `MultiServerMCPClient` constructor pins the *implementation* (destructure vs trust in `.strip()`), and the repo convention is to test behavior, not implementation. If langchain ever flipped that, the right alarm would be an integration test failing on an unexpected upstream request — not a unit test on internal argument shape. Acknowledging the gap but not adding the assertion.

packages/ai-proxy/src/forest-integration-client.ts
```diff
@@ -14,6 +14,7 @@ export type CustomConfig = ZendeskConfig | KolarConfig | SnowflakeConfig;
 export type ForestIntegrationName = 'Zendesk' | 'Kolar' | 'Snowflake';
 
 export interface ForestIntegrationConfig {
+  id: string;
```
hercemer42: Should ID be optional here as it is elsewhere ?

packages/ai-proxy/src/forest-integration-client.ts
```diff
@@ -39,16 +40,16 @@ export default class ForestIntegrationClient implements ToolProvider {
...
+          tools.push(...getZendeskTools(config as ZendeskConfig, id));
           break;
         case 'Kolar':
-          tools.push(...getKolarTools(config as KolarConfig));
+          tools.push(...getKolarTools(config as KolarConfig, id));
           break;
         case 'Snowflake':
-          tools.push(...getSnowflakeTools(config as SnowflakeConfig));
+          tools.push(...getSnowflakeTools(config as SnowflakeConfig, id));
           break;
```
hercemer42, line 53: In this section, if id is undefined or null, no issues ?

packages/ai-proxy/src/mcp-client.ts
```diff
@@ -8,23 +8,31 @@ import McpServerRemoteTool from './mcp-server-remote-tool';
 
 export type McpServers = MultiServerMCPClient['config']['mcpServers'];
 
-export type McpServerConfig = MultiServerMCPClient['config']['mcpServers'][string];
+// Base orchestrator entry shape from `@langchain/mcp-adapters`, widened with the optional
+// stable DB `id` exposed by Forest Admin's orchestrator. The id flows down to RemoteTool
+// so the workflow executor can match a step's `mcpServerId` against the config record.
```
hercemer42: Do we need this elaborate comment ?

packages/ai-proxy/src/remote-tool.ts
```diff
@@ -4,15 +4,21 @@ export default abstract class RemoteTool<ToolType = unknown> {
   base: StructuredToolInterface<ToolType>;
   sourceId: string;
   sourceType: string;
+  // Stable DB id of the originating config entry. Distinct from `sourceId`, which is the
+  // human-readable server/integration name. Use `id` for cross-system references (e.g.
+  // matching a workflow step's mcpServerId to its config); use `sourceId` for display.
```
hercemer42: Unneeded comment

packages/ai-proxy/src/forest-integration-client.ts
hercemer42, line 53: DId we test for what happens when id is undefined or null ? (In the tests, not here)

packages/workflow-executor/src/executors/mcp-step-executor.ts
```diff
@@ -224,13 +224,27 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
     );
   }
 
-  /** Returns tools filtered by mcpServerId (if specified). Throws NoMcpToolsError if empty. */
+  // Match the workflow step's mcpServerId against the stable DB id carried on each tool —
+  // not the display-oriented sourceId — so server-name collisions cannot leak tools across
+  // configs and the same id selects both regular MCP entries and Forest-connector entries.
```
hercemer42: Do we need all this ?

packages/workflow-executor/src/errors.ts
```diff
@@ -203,8 +203,13 @@ export class StepTimeoutError extends WorkflowExecutorError {
...
 export class NoMcpToolsError extends WorkflowExecutorError {
-  constructor() {
-    super('No MCP tools available', 'No tools are available to execute this step.');
+  constructor(requestedId?: string, loadedIds?: readonly string[]) {
+    const technical = requestedId
+      ? `No MCP tools available for mcpServerId="${requestedId}". Loaded MCP config ids: [${(
+          loadedIds ?? []
+        ).join(', ')}]`
+      : 'No MCP tools available';
+    super(technical, 'No tools are available to execute this step.');
```
hercemer42, line 212: Do both messages appear at the same time ? If so, it would seem there is a bit of redunancy in the text ?

packages/workflow-executor/test/executors/mcp-step-executor.test.ts
```diff
@@ -439,14 +440,17 @@ describe('McpStepExecutor', () => {
...
   });
 
-  describe('mcpServerId filter', () => {
-    it('passes only tools from the specified server to the AI', async () => {
-      const toolA = new MockRemoteTool({ name: 'tool_a', sourceId: 'server-A' });
-      const toolB = new MockRemoteTool({ name: 'tool_b', sourceId: 'server-B' });
+  describe('mcpServerId filter (matches by tool.id, not tool.sourceId)', () => {
+    it('passes only tools whose id matches step.mcpServerId to the AI', async () => {
+      // Both tools share sourceId='zendesk' (e.g. same integration server name) but carry
+      // distinct stable DB ids — proves the filter goes through id, not sourceId.
```
hercemer42: Not needed

packages/workflow-executor/test/executors/mcp-step-executor.test.ts
```diff
@@ -455,19 +459,78 @@ describe('McpStepExecutor', () => {
...
       const boundTools = bindTools.mock.calls[0][0] as Array<{ name: string }>;
       const boundNames = boundTools.map(t => t.name);
       expect(boundNames).not.toContain('tool_a');
       expect(boundNames).toContain('tool_b');
       expect(boundNames).toContain('tool_b2');
     });
+
+    it('does not match by sourceId — server-name collisions must not leak tools across configs', async () => {
+      // mcpServerId='server-B' should NOT match a tool that merely has sourceId='server-B' but
+      // a different DB id; only tool.id is authoritative.
```
hercemer42: Not needed

packages/workflow-executor/test/errors.test.ts
```diff
@@ -58,3 +58,45 @@ describe('extractErrorMessage', () => {
...
+
+    expect(err.message).toMatch(/id-A/);
+    expect(err.message).toMatch(/id-B/);
+  });
+
+  it('handles an empty loaded-id list without producing a malformed message', () => {
+    const err = new Ctor('id-missing', []);
+
+    expect(err.message).toMatch(/id-missing/);
+    // Should not produce undefined/null/[object Object] artefacts in the message.
```
hercemer42: No shit

packages/ai-proxy/src/forest-integration-client.ts
hercemer42: **Opus 4.7** — Good catch, applied. Made `ForestIntegrationConfig.id` optional (`id?: string`) to match `McpServerConfig.id` and `RemoteTool.id`. The Forest-connector path now degrades gracefully when PRD-360 isn't deployed (tools carry `id: undefined` → filter misses → diagnostic `NoMcpToolsError`) instead of forcing a type-level requirement. Pushed in 83ab5d3.

packages/ai-proxy/src/forest-integration-client.ts
hercemer42: **Opus 4.7** — No issues. `id` propagates through to `ServerRemoteTool.id` as `undefined`. Downstream, `McpStepExecutor.getFilteredTools` filters with `t.id === step.mcpServerId`: when the step sets `mcpServerId` but tools have `undefined` id, the filter misses → `NoMcpToolsError` fires (which is the expected diagnostic until PRD-360 ships). When the step has no `mcpServerId`, the filter is skipped and all tools are returned regardless of id. Either way the tool itself isn't broken, just unmatchable.

hercemer42: **Opus 4.7** — Yes. Added an explicit test in 83ab5d3 (`forest-integration-client.test.ts` → 'passes undefined to the factory when the config entry has no id') asserting that an entry without `id` results in `getZendeskTools` being called with `undefined` as the second arg. Plus the integration-tool tests in `tools.test.ts` ('leaves RemoteTool.id undefined when no id is provided') verify the same on the receiving end.

packages/ai-proxy/src/mcp-client.ts
hercemer42: **Opus 4.7** — Removed in 83ab5d3.

packages/ai-proxy/src/remote-tool.ts
hercemer42: **Opus 4.7** — Removed in 83ab5d3.

packages/workflow-executor/src/executors/mcp-step-executor.ts
hercemer42: **Opus 4.7** — Removed in 83ab5d3.

packages/workflow-executor/test/executors/mcp-step-executor.test.ts
hercemer42: **Opus 4.7** — Removed in 83ab5d3.

packages/workflow-executor/test/executors/mcp-step-executor.test.ts
hercemer42: **Opus 4.7** — Removed in 83ab5d3.

packages/workflow-executor/test/errors.test.ts
hercemer42: **Opus 4.7** — Removed in 83ab5d3.

packages/workflow-executor/src/errors.ts
hercemer42: **Opus 4.7** — They don't appear together. The dual-message convention separates surfaces:

- **`userMessage`** (`'No tools are available to execute this step.'`) → end-user UI via `stepOutcome.error` → activity log
- **`message`** (`'No MCP tools available for mcpServerId="X". Loaded MCP config ids: [...]'`) → engineer logs via the `logger.error` call I added in `mcp-step-executor.ts`

An end user never sees the technical message; an engineer reading Sentry/structured logs sees the technical message but doesn't care about the user-facing one. The wording overlap ("No MCP tools available" vs "No tools are available") is intentional — both anchor on the same fact in their respective registers (engineering precision vs user-friendly framing). No single observer reads both at the same time, so no redundancy in practice.

packages/ai-proxy/src/mcp-client.ts
```diff
@@ -39,7 +44,8 @@ export default class McpClient implements ToolProvider {
         try {
           const loadedTools = (await client.getTools()) ?? [];
           const extendedTools = loadedTools.map(
-            tool => new McpServerRemoteTool({ tool, sourceId: name }),
+            tool =>
+              new McpServerRemoteTool({ tool, sourceId: name, id: this.idsByServerName[name] }),
```
Scra3: I think id is badly named in the code, It should be name mcpServerId, what do you think?

hercemer42: Well, for me 'id' always means database id, but I can rename if you prefer.

packages/ai-proxy/src/remote-tool.ts
```diff
@@ -4,15 +4,18 @@ export default abstract class RemoteTool<ToolType = unknown> {
   base: StructuredToolInterface<ToolType>;
   sourceId: string;
```
Scra3, line 5: sourceId is currently the mcp server name. And Id the mcp server id.  Could we improve it by adding a commentary or an alias or ... on other idea?
Maybe `id` could be `sourcerMcpServerId` and `sourceId` => `sourceMcpServerName` but maybe it's breaking, we can maybe find a way to add a getter to encapsulate the wrong naming ?

Scra3: get sourceMcpServerName() => return sourceId.

packages/ai-proxy/src/mcp-client.ts
Scra3: Yes i'm agree but we are in the tool class, not in the mcp, by reading `id` we can understand that the `id` of the tool but it's the mcp's id.

packages/ai-proxy/src/remote-tool.ts
hercemer42: Is `RemoteTool.mcpServerId` ok to keep consistent with the other changes ? As for `sourceId` we can but it'll make more changes so if you really want to do that I suggest we open a refacto ticket.

packages/ai-proxy/src/mcp-client.ts
hercemer42: Ok fair point, agreed.  I made the changes - I changed ID to mcpServerId only where it concerns the private maps, and left it alone in the type and wiring - otherwise I'll have to change it in the orchestrator contract, and I think it's correct there as we call the DB on that layer.  Is that ok ?

packages/ai-proxy/src/remote-tool.ts
matthv: If I'm reading this right, `mcpServerId` is accurate on `McpServerRemoteTool` but inherited by `ServerRemoteTool` (Zendesk, Snowflake, ...), whose value is the DB id of an `ai_mcp_configs` row with `isForestConnector: true`. The name reads accurately on half the subclasses, ambiguously on the other half. A dev seeing `tool.mcpServerId === step.mcpServerId` on a Snowflake tool might wonder *"since when is Snowflake an MCP server?"*.

 Would a short JSDoc here clarifying that it covers both user MCP and Forest connectors be enough, or are you leaning toward the broader `sourceId` rename you mentioned? Happy either way.

matthv (approved):

packages/ai-proxy/src/remote-tool.ts
hercemer42: Well, the Forest connectors and user MCP servers live in the same ai_mcp_configs so currently we treat them all as MCP servers.  But yes I think a comment is suitable.  Not pushing for the broader `sourceId` change because its opportunistic and out of scope for this PR, happy to do it in another ticket as mentioned to Alban.

hercemer42: I've added the comment.

PR #1585, opened 2026-05-19
fix(workflow-executor): tolerate LLM field name variations in findField

Scra3:
## Summary

- `findField` previously did exact string matching (`===`) on `displayName` and `fieldName`
- LLMs sometimes return snake_case variants of field names (e.g. `first_name` instead of `firstname`) even when the tool schema constrains the value via `z.literal`
- This adds a normalized fallback that strips separators (`_`, `-`, spaces) and lowercases before matching, so a hallucinated variation no longer kills an otherwise correct step

## Root cause

`invokeWithTool` returns `toolCall.args` directly without re-validating against the Zod schema. The `z.literal` constraint is only used to generate the tool definition sent to the LLM — it is not enforced on the response.

## Test plan
- [ ] Existing `update-record` / `read-record` / `load-related-record` tests pass
- [ ] Manually verify: a workflow with an `account.firstname` field no longer fails when the LLM outputs `first_name`

fixes PRD-214

> [!NOTE]
> ### Fix `findField` in `RecordStepExecutor` to tolerate LLM field name variants
> LLMs may return field names in snake_case, camelCase, lowercase, or hyphenated forms. The `findField` method in [record-step-executor.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1585/files#diff-8a51c158f5f5c6df99cb9483d4d66d54f51671a37f114fafce10ea8b1e51e29c) already uses normalized comparison to handle these cosmetic variants; this PR clarifies the inline comment and adds a parameterized test suite in [update-record-step-executor.test.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1585/files#diff-35ae2d31d3b452c2ac498df9293b2a6209e5a81d3dc793cfdb911260817a6c54) to verify each variant resolves to the correct normalized field name.
>
> 
> #### Changes since #1585 opened
>
> - Modified `RecordStepExecutor.findField` method in `workflow-executor` to implement two-phase field resolution: first attempting exact match on `displayName` or `fieldName`, then computing fuzzy matches using normalized names (lowercased with spaces, underscores, and hyphens removed), returning a field only if exactly one fuzzy match exists, otherwise returning undefined instead of the first fuzzy match [065d4b0]
> - Added test case to `update-record-step-executor.test.ts` in `workflow-executor` under the 'automaticExecution: update direct (Branch B)' suite that constructs a schema with two fields normalizing to the same string and asserts the executor returns an error status [065d4b0]
> - Reformatted field object literals in `update-record-step-executor.test.ts` [e3f8e58]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c3496f0.</sup>
> 
>
>

### DIFF
--- a/packages/workflow-executor/src/executors/record-step-executor.ts
+++ b/packages/workflow-executor/src/executors/record-step-executor.ts
@@ -84,12 +84,18 @@ export default abstract class RecordStepExecutor<
     const normalizeFieldName = (s: string) => s.toLowerCase().replace(/[\s_-]/g, '');
     const normalized = normalizeFieldName(name);
 
-    return (
+    const exact =
       schema.fields.find(f => f.displayName === name) ??
-      schema.fields.find(f => f.fieldName === name) ??
-      schema.fields.find(f => normalizeFieldName(f.displayName) === normalized) ??
-      schema.fields.find(f => normalizeFieldName(f.fieldName) === normalized)
+      schema.fields.find(f => f.fieldName === name);
+    if (exact) return exact;
+
+    const fuzzy = schema.fields.filter(
+      f =>
+        normalizeFieldName(f.displayName) === normalized ||
+        normalizeFieldName(f.fieldName) === normalized,
     );
+
+    return fuzzy.length === 1 ? fuzzy[0] : undefined;
   }
 
   private async toRecordIdentifier(record: RecordRef): Promise<string> {
--- a/packages/workflow-executor/src/executors/record-step-executor.ts
+++ b/packages/workflow-executor/src/executors/record-step-executor.ts
@@ -77,11 +77,10 @@ export default abstract class RecordStepExecutor<
   }
 
   protected findField(schema: CollectionSchema, name: string): FieldSchema | undefined {
-    // The tool definition sent to the LLM is built from z.literal(displayName) — the JSON
-    // Schema does constrain fieldName to exact values. However, invokeWithTool returns
-    // toolCall.args as-is without re-running Zod validation on the response, so the LLM can
-    // silently ignore the constraint and return a formatting variation (e.g. "first_name"
-    // instead of "firstname"). The normalized fallback catches these cosmetic mismatches.
+    // LLMs occasionally return formatting variants of field names (e.g. "first_name" for
+    // "firstname", "full-name" for "Full Name") even though the tool schema declares them
+    // as literals. Fall back to a normalized comparison so a cosmetic variation doesn't
+    // fail an otherwise correct step.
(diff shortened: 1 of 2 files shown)


hercemer42:
[claude-opus-4-7] [Code conventions / defence-in-depth — not blocking]

The normalized fallback runs after both exact passes (`displayName === name`, then `fieldName === name`), so the realistic ambiguity surface is narrower than it first looks: a collision only bites when the LLM hallucinates a form that matches **no** field's exact `displayName` or `fieldName` AND two+ fields happen to normalize to the same string. That's a fairly contrived shape — e.g. fields `{ displayName: "Full Name", fieldName: "fullname" }` and `{ displayName: "FullName", fieldName: "full_name" }` with the LLM returning `"Full-Name"`.

Cheap mitigation if we ever want it: swap the two normalized `.find()` calls for one `.filter()` over both names and return `undefined` when length > 1 — rides on the existing `FieldNotFoundError` path, no new error type. Roughly:

```ts
const exact =
  schema.fields.find(f => f.displayName === name) ??
  schema.fields.find(f => f.fieldName === name);
if (exact) return exact;

const fuzzy = schema.fields.filter(
  f => normalizeFieldName(f.displayName) === normalized
    || normalizeFieldName(f.fieldName) === normalized,
);
return fuzzy.length === 1 ? fuzzy[0] : undefined;
```

Leaving it as-is is also defensible — happy to defer.

packages/workflow-executor/src/executors/record-step-executor.ts
```diff
@@ -77,6 +77,11 @@ export default abstract class RecordStepExecutor<
   }
 
   protected findField(schema: CollectionSchema, name: string): FieldSchema | undefined {
+    // The tool definition sent to the LLM is built from z.literal(displayName) — the JSON
+    // Schema does constrain fieldName to exact values. However, invokeWithTool returns
+    // toolCall.args as-is without re-running Zod validation on the response, so the LLM can
+    // silently ignore the constraint and return a formatting variation (e.g. "first_name"
+    // instead of "firstname"). The normalized fallback catches these cosmetic mismatches.
```
hercemer42: I'd tighten this, for example : 
```
  // LLMs occasionally return formatting variants of field names (e.g. "first_name" for
  // "firstname", "full-name" for "Full Name") even though the tool schema declares them
  // as literals. Fall back to a normalized comparison so a cosmetic variation doesn't
  // fail an otherwise correct step.
```
But it's preferential

hercemer42 (approved):

PR #1586, opened 2026-05-19
fix(workflow-executor): preserve AI suggestion in pending data

Tonours:
## Summary

- Keeps `pendingData` immutable so the AI proposal can be diffed against the user's final choice. User overrides now land in a sibling `userConfirmation: Record<string, unknown>` populated by `patchAndReloadPendingData` from the validated PATCH body.
- `update-record`, `trigger-action` and `load-related-record` executors read user-confirmed values (`value`, `actionResult`, `name`/`displayName`/`selectedRecordId`) from `userConfirmation` with fallback to the AI suggestion in `pendingData`.
- `userConfirmation` is replaced last-write-wins (no spread-merge) to avoid stale keys bleeding across sequential PATCHes.

## Test plan

- [x] `yarn workspace @forestadmin/workflow-executor build`
- [x] `yarn workspace @forestadmin/workflow-executor test` (726 tests pass)
- [x] `yarn workspace @forestadmin/workflow-executor lint`
- [x] Regression tests added: AI suggestion preserved with user-override of `value` / `selectedRecordId` / relation `name`+`displayName`, accept-via-PATCH without override (fallback), rejection via PATCH with `userConfirmation` set.

### DIFF
--- a/packages/workflow-executor/src/executors/base-step-executor.ts
+++ b/packages/workflow-executor/src/executors/base-step-executor.ts
@@ -5,7 +5,10 @@ import type {
   IStepExecutor,
   StepExecutionResult,
 } from '../types/execution-context';
-import type { StepExecutionData, WithUserConfirmation } from '../types/step-execution-data';
+import type {
+  ConfirmableStepExecutionData,
+  StepExecutionData,
+} from '../types/step-execution-data';
 import type { StepDefinition } from '../types/validated/step-definition';
 import type { StepStatus } from '../types/validated/step-outcome';
 import type {
@@ -26,10 +29,6 @@ import {
 } from '../errors';
 import patchBodySchemas from '../http/pending-data-validators';
 import StepSummaryBuilder from './summary/step-summary-builder';
-
-type WithPendingData = StepExecutionData & { pendingData?: object } & WithUserConfirmation;
-
-type PatchBody = Record<string, unknown> & { userConfirmed?: boolean };
 
 export default abstract class BaseStepExecutor<TStep extends StepDefinition = StepDefinition>
   implements IStepExecutor
@@ -193,7 +192,7 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
     error?: string;
   }): StepExecutionResult;
 
-  protected async findPendingExecution<TExec extends WithPendingData>(
+  protected async findPendingExecution<TExec extends ConfirmableStepExecutionData>(
     type: string,
   ): Promise<TExec | undefined> {
     const stepExecutions = await this.context.runStore.getStepExecutions(this.context.runId);
@@ -205,7 +204,7 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
 
   // Preserves the AI suggestion in pendingData: only userConfirmed is mirrored there because
   // handleConfirmationFlow gates on it. The full parsed PATCH body is stored in userConfirmation.
-  protected async patchAndReloadPendingData<TExec extends WithPendingData>(
+  protected async patchAndReloadPendingData<TExec extends ConfirmableStepExecutionData>(
     pendingData?: unknown,
   ): Promise<TExec | undefined> {
     const { type } = this.context.stepDefinition;
@@ -231,18 +230,18 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
       );
     }
 
-    const patchBody = parsed.data as PatchBody;
+    const userConfirmation = parsed.data as TExec['userConfirmation'];
 
     // Last-write-wins: spread-merging would leak stale keys from prior PATCHes.
     const updated: TExec = {
       ...execution,
       pendingData: {
         ...execution.pendingData,
-        ...(patchBody.userConfirmed !== undefined
-          ? { userConfirmed: patchBody.userConfirmed }
+        ...(userConfirmation?.userConfirmed !== undefined
+          ? { userConfirmed: userConfirmation.userConfirmed }
           : {}),
       },
-      userConfirmation: patchBody as TExec['userConfirmation'],
+      userConfirmation,
     };
 
     await this.context.runStore.saveStepExecution(this.context.runId, updated);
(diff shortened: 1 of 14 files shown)


packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -138,21 +138,29 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
+    const { selectedRecordRef, pendingData, userConfirmation } = execution;
 
     if (!pendingData) {
       throw new StepStateError(`Step at index ${this.context.stepIndex} has no pending data`);
     }
 
-    const { name, displayName, selectedRecordId } = pendingData;
+    const isString = (v: unknown): v is string => typeof v === 'string';
+    const isRecordId = (v: unknown): v is Array<string | number> =>
+      Array.isArray(v) && v.every(e => typeof e === 'string' || typeof e === 'number');
```
Scra3: Data are already validated by zod when we receive the http request. I think we don't need to validate two times

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -138,21 +138,29 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
 
-    // Re-derive relatedCollectionName from schema using the (possibly updated) relation name.
-    // `name` is always a fieldName (set from field.fieldName in buildTarget) — search directly.
+    const name = isString(userConfirmation?.name) ? userConfirmation.name : pendingData.name;
+    const displayName = isString(userConfirmation?.displayName)
```
Scra3: displayName seems fully derivable from the schema once we have name, so could we remove it from the HTTP payload entirely? That would avoid having to keep name / displayName in sync and reduce the risk of persisting an inconsistent state.

More generally, it feels like the invariant here should be:

changing only selectedRecordId while keeping the same name is valid,
but if name changes, then selectedRecordId should also be required,
and displayName should be recomputed from the schema rather than trusted from the client.
So maybe we should tighten this in two places:

in the Zod schema, remove displayName and require selectedRecordId whenever name is provided,
in the executor, use name as the source of truth and recompute displayName from the resolved field/schema.

packages/workflow-executor/src/types/step-execution-data.ts
```diff
@@ -14,6 +14,12 @@ interface MutatingStepExecutionData extends BaseStepExecutionData {
   idempotencyPhase?: 'executing' | 'done';
 }
 
+// Parsed PATCH body kept beside `pendingData` so executors can read the user's
+// final input without overwriting the AI suggestion.
+export interface WithUserConfirmation {
+  userConfirmation?: Record<string, unknown>;
```
Scra3: issue (blocking): `Record<string, unknown>` erases the per-step shape that Zod already validated — executors must then add runtime type guards for fields the schema already knows. Export one inferred type per schema in `pending-data-validators.ts` and use them here per step instead of this catch-all.

packages/workflow-executor/src/executors/base-step-executor.ts
```diff
@@ -211,36 +213,44 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
     );
   }
 
+  // Keeps `pendingData` immutable; mirrors `userConfirmed` only because
```
Scra3: nit: "immutable" is misleading — `userConfirmed` is written back into `pendingData` four lines below. The intent is that the original AI suggestion is preserved (not overwritten by a full merge), which is different from immutability.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -138,21 +138,29 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
   ): Promise<StepExecutionResult> {
-    const { selectedRecordRef, pendingData } = execution;
+    const { selectedRecordRef, pendingData, userConfirmation } = execution;
 
     if (!pendingData) {
       throw new StepStateError(`Step at index ${this.context.stepIndex} has no pending data`);
     }
 
-    const { name, displayName, selectedRecordId } = pendingData;
+    const isString = (v: unknown): v is string => typeof v === 'string';
```
Scra3: issue (non-blocking): these guards exist only because `userConfirmation` is typed as `Record<string, unknown>` — Zod already validated that `name` is a `string` and `selectedRecordId` is `Array<string | number>`. Fix the `WithUserConfirmation` typing and both guards can be dropped in favour of direct property access.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
Scra3: suggestion: `displayName` can be re-derived from the schema once `name` is resolved (`field.displayName`), so there is no need to accept it from the PATCH body — trusting client-provided display names risks surfacing stale or inconsistent data.

Scra3 (changes requested):
Two blocking type-safety issues need to be addressed before merge.

- `WithUserConfirmation.userConfirmation` is typed as `Record<string, unknown>`, erasing the per-step shape that Zod already validated — executors end up re-implementing validation with runtime guards.
- `McpStepExecutionData` (line 106 in `step-execution-data.ts`) is the only executor that calls `patchAndReloadPendingData` but does not extend `WithUserConfirmation` — the `userConfirmation` field will be set at runtime but is invisible to TypeScript, breaking the promise of the PR.

Fix: export one inferred type per Zod schema in `pending-data-validators.ts` (e.g. `export type LoadRelatedRecordConfirmation = z.infer<typeof loadRelatedRecordPatchSchema>`) and use those precise types in the interfaces. The runtime guards in `resolveFromSelection` can then be dropped.

Scra3:
https://linear.app/forestadmin/issue/PRD-378/executor-shouldnt-execute-non-automated-tasks

Scra3:
linked to: PRD-378

Scra3 (approved):

PR #1587, opened 2026-05-19
refactor(workflow-executor): suppress spurious timeout log when step fails before timeout

Scra3:
## Summary

- `runWithTimeout()` attached a `.catch()` on `execPromise` before `Promise.race` to prevent `UnhandledPromiseRejection` on late rejections. It always logged `"Step work rejected after timeout — result discarded"` — even when the step failed before the timeout fired, producing a misleading log for any normal step error.
- Added a `timeoutFired` boolean flag set to `true` only inside the `setTimeout` callback. The `.catch()` now skips the log when `timeoutFired` is false (rejection propagates normally through the race; no log needed).
- Added a regression test: verifies the log is NOT emitted when `execPromise` rejects before the timeout.

## Test plan

- [ ] `yarn workspace @forestadmin/workflow-executor test --testPathPattern="base-step-executor"` — all 30 tests green including the new one
- [ ] Existing test `'logs a late rejection of the losing promise as info'` still passes (real timeout case unchanged)
- [ ] New test `'does not log when execPromise rejects before the timeout fires'` passes

> [!NOTE]
> ### Fix spurious 'result discarded' log in `BaseStepExecutor.runWithTimeout` when step fails before timeout
> Adds a `timeoutFired` boolean flag in `runWithTimeout` to distinguish between a step rejection caused by the timeout branch versus a normal pre-timeout failure. The 'Step work rejected after timeout — result discarded' log is now only emitted when the timeout actually fires, not on every rejection. A new test in [base-step-executor.test.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1587/files#diff-d7ab010aad83f0a668d49bd1f5f4eb531ed5ac54ae0068f40e42479a3b7a7cb9) asserts that pre-timeout rejections do not produce this log.
>
> 
> #### Changes since #1587 opened
>
> - Renamed internal timeout tracking flag and updated rejection logging conditions [d0a5054]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 191ddf0.</sup>
> 
>
>

### DIFF
--- a/packages/workflow-executor/src/executors/base-step-executor.ts
+++ b/packages/workflow-executor/src/executors/base-step-executor.ts
@@ -173,9 +173,14 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
     if (!timeoutMs || timeoutMs <= 0) return this.doExecute();
 
     let timer: NodeJS.Timeout | undefined;
+    let timeoutFired = false;
     const execPromise = this.doExecute();
 
+    // .catch() prevents UnhandledPromiseRejection when execPromise rejects after the timeout has
+    // already won the race. Only log when timeoutFired — a rejection before the timeout propagates
+    // normally through the race and needs no log here.
     execPromise.catch(err => {
+      if (!timeoutFired) return;
       this.context.logger.info('Step work rejected after timeout — result discarded', {
         runId: this.context.runId,
         stepId: this.context.stepId,
@@ -188,7 +193,10 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
       return await Promise.race([
         execPromise,
         new Promise<never>((_, reject) => {
-          timer = setTimeout(() => reject(new StepTimeoutError(timeoutMs)), timeoutMs);
+          timer = setTimeout(() => {
+            timeoutFired = true;
+            reject(new StepTimeoutError(timeoutMs));
+          }, timeoutMs);
         }),
       ]);
     } finally {
--- a/packages/workflow-executor/src/executors/base-step-executor.ts
+++ b/packages/workflow-executor/src/executors/base-step-executor.ts
@@ -173,14 +173,11 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
     if (!timeoutMs || timeoutMs <= 0) return this.doExecute();
 
     let timer: NodeJS.Timeout | undefined;
-    let timeoutFired = false;
+    let hasTimeoutFired = false;
     const execPromise = this.doExecute();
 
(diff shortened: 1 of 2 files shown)


packages/workflow-executor/src/executors/base-step-executor.ts
```diff
@@ -173,9 +173,14 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
     if (!timeoutMs || timeoutMs <= 0) return this.doExecute();
 
     let timer: NodeJS.Timeout | undefined;
+    let timeoutFired = false;
     const execPromise = this.doExecute();
 
+    // .catch() prevents UnhandledPromiseRejection when execPromise rejects after the timeout has
+    // already won the race. Only log when timeoutFired — a rejection before the timeout propagates
+    // normally through the race and needs no log here.
```
hercemer42: I think the code is clear enough that the comment isn't necessary, but it's just a preference

packages/workflow-executor/src/executors/base-step-executor.ts
```diff
@@ -173,9 +173,14 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
     if (!timeoutMs || timeoutMs <= 0) return this.doExecute();
 
     let timer: NodeJS.Timeout | undefined;
+    let timeoutFired = false;
```
hercemer42: `hasTimeoutFired` would be clearer

packages/workflow-executor/src/executors/base-step-executor.ts
hercemer42: Oherwise Claude says : New comment partially restates the .catch()/UnhandledPromiseRejection rationale already explained in the comment at lines 168–170; could be trimmed to focus only on the `timeoutFired` guard rationale — file: packages/workflow-executor/src/executors/base-step-executor.ts:179-181

packages/workflow-executor/test/executors/base-step-executor.test.ts
```diff
@@ -453,6 +453,21 @@ describe('BaseStepExecutor', () => {
         process.off('unhandledRejection', unhandled);
       }
     }, 5_000);
+
+    it('does not log when execPromise rejects before the timeout fires', async () => {
```
hercemer42: Test name should express functional intent, for example : `does not log discard message when step rejects before timeout` - preferential

hercemer42:
Considering it's just a log, do we need 'fix' in the commit name ? You could put 'refactor' - which would avoid triggering a build.  No big deal.

hercemer42 (approved):

hercemer42 (approved):

PR #1590, opened 2026-05-21
fix(workflow): replace automatedExecution with executionType

EnkiP:
## Definition of Done

fixes PRD-378

need frontend PR: https://github.com/ForestAdmin/forestadmin/pull/9708
and backend PR: https://github.com/ForestAdmin/forestadmin-server/pull/8254
branch name is always `fix/execution-type`

### General

### Security

> [!NOTE]
> ### Replace `automaticExecution` boolean with `executionType` enum across workflow step executors
> - Introduces `StepExecutionMode` enum (`Manual`, `AutomatedWithConfirmation`, `FullyAutomated`) in [step-definition.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1590/files#diff-66dcc667cb1ebecf3964eecfb28391b1cb81e16c853ad79d4f9017172521aea3) to replace the `automaticExecution` boolean on step definitions.
> - Updates all step executors (`UpdateRecord`, `LoadRelatedRecord`, `TriggerAction`, `MCP`, `Condition`) to gate automation behavior on `executionType === StepExecutionMode.FullyAutomated` instead of `automaticExecution`.
> - Adds `ConditionStepExecutor` support for `executionType=Manual`: returns `awaiting-input` without invoking the AI when no incoming pending data is present.
> - Adds a `warn` method to the `Logger` interface and implements it in `ConsoleLogger` and `PrettyLogger`; the step-definition mapper uses it to log when an unsupported `executionType` is normalized to a fallback.
> - Increases the default polling interval from 5s to 30s in [build-workflow-executor.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1590/files#diff-9a7b00e34df678f7ad856e6420bb042e4f6e01afc81fdc706055f462da01d493).
> - Risk: `executionType` is now a required field on all step definition schemas; any caller passing `automaticExecution` or omitting `executionType` will fail validation.
>
> 
> #### Changes since #1590 opened
>
> - Changed `toStepExecutionMode` function in `workflow-executor` to return `StepExecutionMode | null` instead of `StepExecutionMode | undefined`, explicitly returning `null` when server execution type is unmapped using `EXECUTION_TYPE_TO_MODE[executionType] ?? null` [b7c3582]
> - Modified `normalizeExecutionType` function in `workflow-executor` to accept `StepExecutionMode | null` instead of `StepExecutionMode | undefined` and always return a concrete `StepExecutionMode` by falling back to step type's default execution mode when provided executionType is null or falsy [b7c3582]
> - Updated test expectations in `run-to-available-step-mapper.test.ts` and `step-definition-mapper.test.ts` to expect `FullyAutomated` execution mode instead of `AutomatedWithConfirmation` for read-record steps, and to expect fallback executionType to be present rather than undefined when executionType is omitted [b7c3582]
> - Added explicit `executionType` fields to `StepDefinition` test fixtures across executor test files including `base-step-executor.test.ts`, `guidance-step-executor.test.ts`, `load-related-record-step-executor.test.ts`, `mcp-step-executor.test.ts`, `read-record-step-executor.test.ts`, `step-summary-builder.test.ts`, `trigger-record-action-step-executor.test.ts`, `update-record-step-executor.test.ts`, `workflow-execution.test.ts`, and `runner.test.ts` [b7c3582]
> - Exported `DEFAULT_POLLING_INTERVAL_MS` constant from `build-workflow-executor` module and updated `cli-core` to import and use it for CLI help text default value and removed the hardcoded fallback value in `logStartup` function's `pollingIntervalMs` logging [3631d89]
> - Removed unused `exec` import from `child_process` in `runner.test.ts` [3631d89]
> - Moved `executionType` validation and default value assignment from imperative normalization logic into declarative zod schema definitions per step type [21fc759]
> - Removed `Logger` parameter from step definition mapping and conversion functions throughout the call chain [21fc759]
> - Updated tests to remove logger mocks and reflect silent `executionType` defaulting behavior [21fc759]
> - Added test coverage for `AiModelPortError`, `PendingDataNotFoundError`, and `InvalidPendingDataError` in the `workflow-executor` package [67f723c]
> - Added test coverage for error-wrapping behavior in `InMemoryStore.getStepExecutions` method in the `workflow-executor` package [67f723c]
> - Added test case for `ConsoleLogger.warn()` method [d7a80af]
> - Added test case for `PrettyLogger.warn()` method [d7a80af]
> 
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized f8b0d26. 4 files reviewed, 1 issue evaluated, 0 issues filtered, 1 comment posted</summary>
>
> ### 🗂️ Filtered Issues
>
> </details>
>
>


packages/workflow-executor/src/executors/base-step-executor.ts
```diff
@@ -128,8 +128,10 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
     return Promise.resolve(null);
   }
 
-  // Return null when the frontend performs the action (e.g. TriggerAction with automaticExecution=false)
-  // — the front logs on its side. Override when the executor itself calls the agent.
+  // TODO: What is this method ? why does it always return null ?
+  // Return null when the frontend performs the action (e.g. TriggerAction without
```
Scra3, line 117: It returns null because if you want to create an activity log you should implement in the right step. By default, no activity log

Scra3: Please remove the todo

Scra3 (commented):
Good direction on the enum migration. A few architectural and documentation points before merge.

packages/workflow-executor/src/types/validated/step-definition.ts
```diff
@@ -2,6 +2,8 @@
 
 import { z } from 'zod';
 
+import { ServerStepExecutionTypeEnum } from '../../adapters/server-types';
```
Scra3: issue (non-blocking): `ServerStepExecutionTypeEnum` from the adapter layer leaks into the domain schema — declare a `StepExecutionMode` enum here in `types/validated/` instead and have `step-definition-mapper.ts` be the single translation point. Executors should import from `types/validated/`, never from `adapters/`.

packages/workflow-executor/src/types/validated/step-definition.ts
```diff
@@ -15,11 +17,7 @@ export enum StepType {
 const baseFields = {
   prompt: z.string().optional(),
   aiConfigName: z.string().optional(),
-};
-
-const baseRecordFields = {
-  ...baseFields,
-  automaticExecution: z.boolean().optional(),
+  executionType: z.enum(ServerStepExecutionTypeEnum).optional(),
```
Scra3: nit: `z.enum()` is designed for string-tuple arrays; `z.nativeEnum(StepExecutionMode)` is the idiomatic form for TypeScript `enum` objects — use it once the domain enum is declared above.

packages/workflow-executor/src/executors/update-record-step-executor.ts
```diff
@@ -180,12 +181,12 @@ export default class UpdateRecordStepExecutor extends RecordStepExecutor<UpdateR
...
 
-    // Branch B -- automaticExecution
-    if (step.automaticExecution) {
+    // Branch B -- fully automated execution
+    if (step.executionType === ServerStepExecutionTypeEnum.FullyAutomated) {
       return this.resolveAndUpdate(target);
     }
 
-    // Branch C -- Awaiting confirmation
+    // Branch C -- Awaiting confirmation (also covers Manual fallback)
```
Scra3: issue (non-blocking): `also covers Manual fallback` hides an intentional design choice — document explicitly that `manual` and `automatedWithConfirmation` share Branch C because the AI suggestion is always computed and the distinction is UI-only (how the front presents the confirmation to the user).

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -78,12 +79,12 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
 
-    // Branch B -- automaticExecution
-    if (step.automaticExecution) {
+    // Branch B -- fully automated execution
+    if (step.executionType === ServerStepExecutionTypeEnum.FullyAutomated) {
       return this.resolveAndLoadAutomatic(target);
     }
 
-    // Branch C -- pre-fetch candidates, await user confirmation
+    // Branch C -- pre-fetch candidates, await user confirmation (also covers Manual fallback)
```
Scra3: issue (non-blocking): same as update-record — document why `manual` and `automatedWithConfirmation` intentionally share Branch C.

packages/workflow-executor/src/executors/trigger-record-action-step-executor.ts
```diff
@@ -128,7 +131,8 @@ export default class TriggerRecordActionStepExecutor extends RecordStepExecutor<
       return this.executeOnExecutor(target);
     }
 
-    // Branch C -- Awaiting confirmation (frontend executes the action, including forms)
+    // Branch C -- Awaiting confirmation (frontend executes the action, including forms).
```
Scra3: issue (non-blocking): same — document the intentional equivalence of `manual` and `automatedWithConfirmation` in Branch C.

packages/workflow-executor/src/executors/mcp-step-executor.ts
```diff
@@ -87,12 +88,12 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
...
     const target: McpToolCall = { name: toolName, sourceId: selectedTool.sourceId, input: args };
 
-    if (this.context.stepDefinition.automaticExecution) {
+    if (this.context.stepDefinition.executionType === ServerStepExecutionTypeEnum.FullyAutomated) {
       // Branch B -- direct execution
       return this.executeToolAndPersist(target);
     }
 
-    // Branch C -- Awaiting confirmation
+    // Branch C -- Awaiting confirmation (also covers Manual fallback)
```
Scra3: issue (non-blocking): same — document the intentional equivalence of `manual` and `automatedWithConfirmation` in Branch C.

packages/workflow-executor/src/executors/update-record-step-executor.ts
Scra3: Also, add a warning logs if we receive "manual". Add it on each step

packages/workflow-executor/src/types/validated/step-definition.ts
EnkiP: what about the commentary: Use z.enum(EnumObject), not z.nativeEnum — the latter is deprecated in zod 4.

packages/workflow-executor/src/types/validated/step-definition.ts
```diff
@@ -12,14 +12,57 @@ export enum StepType {
...
+ */
+export enum StepExecutionMode {
+  Manual = 'manual',
+  AutomatedWithConfirmation = 'automated-with-confirmation',
+  FullyAutomated = 'fully-automated',
+}
+
+// Declares which execution modes are valid per step type and the fallback to use otherwise.
+// Enforced at the adapter boundary so executors can assume the configured mode is supported.
+export const SUPPORTED_EXECUTION_MODES: Record<
```
Scra3: `SUPPORTED_EXECUTION_MODES` is a translation rule (which execution modes are valid when mapping server → domain), not a domain type. It is only used by `normalizeExecutionType()` in `step-definition-mapper.ts`. Consider moving it there to keep this file pure types/schemas.

packages/workflow-executor/src/types/validated/step-definition.ts
```diff
@@ -12,14 +12,57 @@ export enum StepType {
...
+  { supported: readonly StepExecutionMode[]; fallback: StepExecutionMode }
+> = {
+  [StepType.Condition]: {
+    supported: [StepExecutionMode.Manual, StepExecutionMode.FullyAutomated],
+    fallback: StepExecutionMode.FullyAutomated,
+  },
+  [StepType.ReadRecord]: {
+    supported: [StepExecutionMode.AutomatedWithConfirmation, StepExecutionMode.FullyAutomated],
+    fallback: StepExecutionMode.AutomatedWithConfirmation,
+  },
```
Scra3: `SUPPORTED_EXECUTION_MODES` for `ReadRecord` is inconsistent in two ways:

1. **`supported` includes `AutomatedWithConfirmation`** — but the server type (`WorkflowTaskGetData`) constrains `executionType` to `FullyAutomated` only. The server will never send `AutomatedWithConfirmation` for a `get-data` task.

2. **`fallback` is `AutomatedWithConfirmation`** — should be `FullyAutomated`. If `normalizeExecutionType` were ever updated to apply the fallback for `undefined` values (to handle legacy runs), legacy `ReadRecord` steps would incorrectly be treated as `AutomatedWithConfirmation`.

No functional impact today since `ReadRecordStepExecutor` has no branching on `executionType` — but `SUPPORTED_EXECUTION_MODES` is wrong and will mislead anyone who reads it.

All other step types look correct (checked `UpdateRecord`, `TriggerAction`, `LoadRelatedRecord`, `Mcp`, `Condition`, `Guidance`).

Suggested fix:
```ts
[StepType.ReadRecord]: {
  supported: [StepExecutionMode.FullyAutomated],
  fallback: StepExecutionMode.FullyAutomated,
},
```

packages/workflow-executor/src/types/validated/step-definition.ts
```diff
@@ -12,14 +12,57 @@ export enum StepType {
...
 const baseFields = {
   prompt: z.string().optional(),
   aiConfigName: z.string().optional(),
-};
-
-const baseRecordFields = {
-  ...baseFields,
-  automaticExecution: z.boolean().optional(),
+  // Use z.enum(EnumObject), not z.nativeEnum — the latter is deprecated in zod 4.
+  executionType: z.enum(StepExecutionMode).optional(),
```
Scra3: `executionType` can be made non-optional.

The server GraphQL subscription now declares `executionType: String!` (non-nullable), and the bpmn-parser sets the field on every step type without exception. Step definitions are parsed from the BPMN at execution time — not read from stored run history — so even runs that started before this deployment will receive a step definition with `executionType` once the backend is deployed.

The legacy-compatibility concern (runs without the field) no longer applies.

```ts
executionType: z.enum(StepExecutionMode), // drop .optional()
```

packages/workflow-executor/src/executors/condition-step-executor.ts
```diff
@@ -61,6 +62,14 @@ export default class ConditionStepExecutor extends BaseStepExecutor<ConditionSte
   protected async doExecute(): Promise<StepExecutionResult> {
     const { stepDefinition: step, incomingPendingData } = this.context;
 
+    // Manual mode: the user picks the option from the frontend. Wait for their input
+    // without ever calling the AI.
+    const isManual = step.executionType === StepExecutionMode.Manual;
+
+    if (isManual && incomingPendingData === undefined) {
```
Scra3, line 69: **Dead code** — `isManual` can never be `true` here.

The executor only receives steps with `sendToExecutor: true`, which the orchestrator sets exclusively for `FullyAutomated` and `AutomatedWithConfirmation` steps. Manual condition steps get `runState = Started` and `sendToExecutor: false` in the history-navigator — they never reach the executor.

Since the backend parser only emits `Manual` or `FullyAutomated` for conditions (never `AutomatedWithConfirmation`), any condition arriving here is guaranteed to be `FullyAutomated`. This block is unreachable.

packages/workflow-executor/src/types/validated/step-definition.ts
```diff
@@ -12,14 +12,57 @@ export enum StepType {
...
+}
+
+// Declares which execution modes are valid per step type and the fallback to use otherwise.
+// Enforced at the adapter boundary so executors can assume the configured mode is supported.
+export const SUPPORTED_EXECUTION_MODES: Record<
+  StepType,
+  { supported: readonly StepExecutionMode[]; fallback: StepExecutionMode }
+> = {
+  [StepType.Condition]: {
+    supported: [StepExecutionMode.Manual, StepExecutionMode.FullyAutomated],
```
Scra3: `Manual` shouldn't be in `supported` here — the executor never receives Manual condition steps (`sendToExecutor: false` in the history-navigator for `Manual`). The only value that can arrive at the executor for a condition is `FullyAutomated`.

Should be:
```ts
[StepType.Condition]: {
  supported: [StepExecutionMode.FullyAutomated],
  fallback: StepExecutionMode.FullyAutomated,
},
```

packages/workflow-executor/src/adapters/step-definition-mapper.ts
```diff
@@ -18,22 +24,60 @@ const TASK_TYPE_TO_STEP_TYPE: Record<ServerTaskType, StepType> = {
...
+  [ServerStepExecutionTypeEnum.Manual]: StepExecutionMode.Manual,
+  [ServerStepExecutionTypeEnum.AutomatedWithConfirmation]:
+    StepExecutionMode.AutomatedWithConfirmation,
+  [ServerStepExecutionTypeEnum.FullyAutomated]: StepExecutionMode.FullyAutomated,
+};
+
+function toStepExecutionMode(
+  executionType: ServerStepExecutionTypeEnum | undefined,
+): StepExecutionMode | undefined {
+  return executionType === undefined ? undefined : EXECUTION_TYPE_TO_MODE[executionType];
```
Scra3: **Silent fallthrough on unknown server value.**

If the server sends an `executionType` string not in `ServerStepExecutionTypeEnum`, `EXECUTION_TYPE_TO_MODE[executionType]` returns `undefined` at runtime. `toStepExecutionMode` then returns `undefined`, and `normalizeExecutionType` short-circuits at the `if (executionType === undefined) return undefined` guard — logging no warning and treating the unknown value exactly like a legacy workflow that never set `executionType`.

The warning path (lines 52-55) is never reached in this case.

Fix: distinguish "server sent `undefined`" from "server sent an unrecognised string":
```ts
function toStepExecutionMode(
  executionType: ServerStepExecutionTypeEnum | undefined,
): StepExecutionMode | null | undefined {
  if (executionType === undefined) return undefined; // legacy
  return EXECUTION_TYPE_TO_MODE[executionType] ?? null; // null = known-unknown
}
```
Then `normalizeExecutionType` can warn and apply the fallback for `null` the same way it does for an unsupported mode.

packages/workflow-executor/src/adapters/server-types.ts
```diff
@@ -13,61 +13,129 @@ export interface ServerWorkflowTransition {
...
+  automaticCompletion: false;
+}
+
+export interface ServerWorkflowEscalation extends ServerWorkflowStepBase {
+  type: ServerStepTypeEnum.Escalation;
   prompt: string;
-  outgoing: ServerWorkflowTransition;
+  outgoing: [ServerWorkflowTransition];
+  inboxId: string | null;
+}
```
Scra3, line 126: `ServerWorkflowEscalation` is the only step variant that doesn't narrow `executionType` — it inherits the unconstrained `executionType: ServerStepExecutionTypeEnum` from `ServerWorkflowStepBase`. Every other variant constrains it (e.g. `End` and `StartSubWorkflow` → `Manual`, `Condition` → `Manual | FullyAutomated`).

Should be:
```ts
export interface ServerWorkflowEscalation extends ServerWorkflowStepBase {
  type: ServerStepTypeEnum.Escalation;
  executionType: ServerStepExecutionTypeEnum.Manual;
  ...
}
```
No runtime impact today (Escalation throws `UnsupportedStepTypeError` before mapping), but the missing constraint is misleading and could hide a bug if the type is ever used elsewhere.

packages/workflow-executor/src/adapters/step-definition-mapper.ts
```diff
@@ -18,22 +24,60 @@ const TASK_TYPE_TO_STEP_TYPE: Record<ServerTaskType, StepType> = {
...
+  executionType: StepExecutionMode | undefined,
+  logger: Logger,
+): StepExecutionMode | undefined {
+  if (executionType === undefined) return undefined;
+  const { supported, fallback } = SUPPORTED_EXECUTION_MODES[stepType];
+  if (supported.includes(executionType)) return executionType;
+
+  logger.warn(
+    `Step type "${stepType}" received unsupported executionType=${executionType}; falling back to ${fallback}`,
+    { stepType, configuredExecutionType: executionType, supportedExecutionTypes: supported },
```
Scra3: **Warning non corrélable en prod** — le message de fallback ne contient pas le `runId` ni le `stepIndex`, qui sont les deux clés qu'on utilise pour tracer un problème dans les logs.

La signature de `normalizeExecutionType` ne les reçoit pas, mais tous les call sites dans `run-to-available-step-mapper.ts` ont ces valeurs disponibles (`run.id`, `pending.stepIndex`). Suggère de les passer en paramètre ou de les inclure dans le contexte du logger :
```ts
logger.warn(
  `Step type "${stepType}" received unsupported executionType=${executionType}; falling back to ${fallback}`,
  { runId, stepIndex, stepType, configuredExecutionType: executionType, supportedExecutionTypes: supported },
);
```

packages/workflow-executor/src/adapters/run-to-available-step-mapper.ts
```diff
@@ -75,9 +76,12 @@ function toStepOutcome(s: ServerStepHistory): StepOutcome {
   return { type: 'record', ...baseFromCtx, status } satisfies RecordStepOutcome;
 }
 
-function tryMapStep(s: ServerStepHistory): Step | null {
+function tryMapStep(s: ServerStepHistory, logger: Logger): Step | null {
   try {
-    return { stepDefinition: toStepDefinition(s.stepDefinition), stepOutcome: toStepOutcome(s) };
+    return {
+      stepDefinition: toStepDefinition(s.stepDefinition, logger),
```
Scra3: **Warning émis pour chaque step de l'historique, pas seulement le step pending.**

`toStepDefinition` est appelé pour chaque step de l'historique dans `toPreviousSteps` (via `tryMapStep` ligne 82 et `toStepOutcome` ligne 42) — soit deux fois par step d'historique — et une fois pour le step pending (ligne 158).

Si un step de l'historique a un `executionType` non supporté, le warning de fallback se déclenche pour chaque step de l'historique qui le contient. Le warning qui importe (celui du step pending à la ligne 158) est noyé dans le bruit.

Suggère de ne logguer le warning que pour le step pending, par exemple en passant un flag `warnOnFallback` à `toStepDefinition`, ou en loggant seulement au call site de la ligne 158.

packages/workflow-executor/test/executors/condition-step-executor.test.ts
```diff
@@ -386,4 +386,70 @@ describe('ConditionStepExecutor', () => {
       expect(runStore.saveStepExecution).not.toHaveBeenCalled();
     });
   });
+
+  describe('executionType=Manual', () => {
```
Scra3, line 392: Ces trois tests couvrent des chemins de code morts.

Le bloc `executionType=Manual` dans `condition-step-executor.ts` ne peut jamais s'exécuter : les conditions Manual reçoivent `sendToExecutor: false` dans le history-navigator et n'arrivent donc jamais à l'executor (cf. commentaire sur la ligne 69 de `condition-step-executor.ts`).

Ces tests donnent une fausse impression de couverture — ils valident un invariant qui n'existe pas en prod. À supprimer en même temps que le dead code.

packages/workflow-executor/src/executors/condition-step-executor.ts
EnkiP: ❌

packages/workflow-executor/src/types/validated/step-definition.ts
EnkiP: ❌  Claude misunderstand the "sendToExecutor" param

packages/workflow-executor/src/adapters/server-types.ts
EnkiP: I see we have ServerWorkflowEscalation but no StepType variant for escalation. Let's tackle this in an escalation specific PR

packages/workflow-executor/src/adapters/step-definition-mapper.ts
EnkiP: galère de récupérer le runId et stepIndex, et pas très important anyway

packages/workflow-executor/src/adapters/run-to-available-step-mapper.ts
EnkiP: I set the logger as optional and removed it from the loops

packages/workflow-executor/test/executors/condition-step-executor.test.ts
EnkiP: Non, condition manual =>
- renvoie awaiting-input si appelé vide
- save le selectedOption is appelé avec userInput

Scra3 (approved):

PR #1592, opened 2026-05-22
refactor(workflow-executor): add stepType to all step execution logs

Scra3:
## Summary

- Destructures `stepType` once at the top of `execute()` and uses it uniformly across all log calls (cache hit, success, timeout, `WorkflowExecutorError`, unexpected error) — eliminates the repetition of `this.context.stepDefinition.type` in every catch branch

## Test plan

- [ ] Run `yarn workspace @forestadmin/workflow-executor test` — all tests pass
- [ ] Verify log output includes `stepType` on all execution outcomes

> [!NOTE]
> ### Add `stepType` to all step execution logs in `BaseStepExecutor`
> - Introduces a `logCtx` private getter on `BaseStepExecutor` that returns `{ runId, stepId, stepIndex, stepType }`, replacing repeated inline context objects across all log calls.
> - Updates error log paths to populate the `cause` field using `extractErrorMessage`, so non-Error causes are stringified and undefined causes remain `undefined`.
> - Updates `extractErrorMessage` in [errors.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1592/files#diff-0109f91a733affdd89e2fae234f7010a44b4b923e1d56be5c84f8014b219cef0) to return `undefined` when called with `undefined`, rather than the string `'undefined'`.
>
> 
> #### Changes since #1592 opened
>
> - Updated test assertions in `BaseStepExecutor.execute` test suite to verify that `stepType` is included in log contexts [7bfb02c]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d030c27.</sup>
> 
>
>

### DIFF
--- a/packages/workflow-executor/src/executors/base-step-executor.ts
+++ b/packages/workflow-executor/src/executors/base-step-executor.ts
@@ -325,6 +325,7 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
 
   private get logCtx() {
     const { runId, stepId, stepIndex, stepDefinition } = this.context;
+
     return { runId, stepId, stepIndex, stepType: stepDefinition.type };
   }
 }
--- a/packages/workflow-executor/src/executors/base-step-executor.ts
+++ b/packages/workflow-executor/src/executors/base-step-executor.ts
@@ -41,11 +41,6 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
     this.agentPort = context.agentPort;
   }
 
-  private get logCtx() {
-    const { runId, stepId, stepIndex, stepDefinition } = this.context;
-    return { runId, stepId, stepIndex, stepType: stepDefinition.type };
-  }
-
   async execute(): Promise<StepExecutionResult> {
     const { baseRecordRef } = this.context;
 
@@ -326,5 +321,10 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
     tool: DynamicStructuredTool,
   ): Promise<T> {
     return (await this.invokeWithTools<T>(messages, [tool])).args;
+  }
+
+  private get logCtx() {
+    const { runId, stepId, stepIndex, stepDefinition } = this.context;
+    return { runId, stepId, stepIndex, stepType: stepDefinition.type };
   }
 }
--- a/packages/workflow-executor/src/executors/base-step-executor.ts
+++ b/packages/workflow-executor/src/executors/base-step-executor.ts
@@ -97,7 +97,7 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
       this.context.logger.error('Unexpected error during step execution', {
         ...this.logCtx,
         error: extractErrorMessage(error),
-        cause: errorCause !== undefined ? extractErrorMessage(errorCause) : undefined,
+        cause: extractErrorMessage(errorCause),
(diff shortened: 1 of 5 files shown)


packages/workflow-executor/test/executors/base-step-executor.test.ts
```diff
@@ -270,6 +284,26 @@ describe('BaseStepExecutor', () => {
           expect.objectContaining({ cause: 'root cause' }),
         );
       });
+
+      it('omits cause from log when error has no cause', async () => {
```
hercemer42: [claude-opus-4-7] **[Should fix]** Test name is misleading. The implementation at `base-step-executor.ts:97-102` always sets `cause: extractErrorMessage(errorCause)` in the log payload — when the cause is undefined, the property is **present with value `undefined`**, not omitted.

Also, `expect.objectContaining({ cause: undefined })` passes whether the field is absent OR present-with-undefined, so the assertion can't actually distinguish the two cases. The test still catches a regression from `undefined` → `'undefined'` (the original bug), so it has value — just rename to something like `'sets cause to undefined when error has no cause'` to match real behavior.

packages/workflow-executor/src/executors/base-step-executor.ts
```diff
@@ -59,9 +56,7 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
 
       if (cached) {
         this.context.logger.info('Step execution completed (replayed from cache)', {
```
hercemer42, line 58: [claude-opus-4-7] **[Preferential]** This cache-replay log path now includes `stepType` via `...this.logCtx`, but no test asserts it. The new test `'includes stepType in info logs for started and completed steps'` (test file:245) only exercises the non-cached branch (`checkIdempotency()` returns null by default in `TestableExecutor`).

A future refactor that strips `logCtx` from this site would go undetected. Worth a small test that overrides `checkIdempotency()` to return a cached result and asserts `stepType` on the replay log.

packages/workflow-executor/src/executors/base-step-executor.ts
```diff
@@ -94,10 +84,8 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
       if (error instanceof WorkflowExecutorError) {
         if (error.cause !== undefined) {
           this.context.logger.error(error.message, {
```
hercemer42, line 86: [claude-opus-4-7] **[Preferential]** This WorkflowExecutorError-with-cause log now includes `stepType` via `...this.logCtx`, but the existing `'logs cause when WorkflowExecutorError has a cause'` test doesn't assert it. One-line addition to that test's `objectContaining` would close the gap.

packages/workflow-executor/src/executors/base-step-executor.ts
```diff
@@ -179,9 +165,7 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
     execPromise.catch(err => {
       if (!hasTimeoutFired) return;
       this.context.logger.info('Step work rejected after timeout — result discarded', {
```
hercemer42, line 167: [claude-opus-4-7] **[Preferential]** This late-rejection log includes `stepType` via `...this.logCtx`, but the existing test for the post-timeout-rejection path doesn't assert it. One-line addition to that test's `objectContaining`.

hercemer42 (approved):

PR #1594, opened 2026-05-26
feat(workflow-executor): serialize recordId as pipe string at front/orchestrator boundaries

Scra3:
## Summary

- Adds `serializeRecordId` / `deserializeRecordId` utility in `src/adapters/record-id-serializer.ts`
- Applies deserialization in `run-to-available-step-mapper.ts`: splits pipe-separated `selectedRecordId` from the orchestrator into a proper array (`'id1|id2'` → `['id1', 'id2']`)
- Applies serialization in `executor-http-server.ts` via `src/http/step-serializer.ts`: converts `recordId` arrays to pipe strings in `GET /runs/:runId` responses (`['id1', 'id2']` → `'id1|id2'`)
- Applies deserialization in `pending-data-validators.ts`: accepts `selectedRecordId` as a pipe string from POST trigger body and transforms to array

## Test plan

- [ ] `record-id-serializer.test.ts` — unit tests for the two utility functions
- [ ] `run-to-available-step-mapper.test.ts` — new test for composite key `'pk1|pk2'` → `['pk1', 'pk2']`
- [ ] `executor-http-server.test.ts` — serialization tests for update-record and load-related-record steps
- [ ] `pending-data-validators.test.ts` — deserialization tests for pipe-string `selectedRecordId`

788 tests passing.

fixes PRD-214

> [!NOTE]
> ### Serialize `RecordId` as a pipe-separated string at workflow executor HTTP and orchestrator boundaries
> - Introduces `serializeRecordId` and `deserializeRecordId` utilities in [`record-id-serializer.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1594/files#diff-20b96e8234208de964d239b8a6a0fefc80ff60210c48259a44ab8334cd493000) to convert between `RecordId` arrays and pipe-joined strings (e.g. `'a|b'`).
> - The `GET /runs/:runId` response now serializes all `recordId` fields (selected record refs, load-related-record candidates, suggested records, and execution results) to pipe-separated strings via [`step-serializer.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1594/files#diff-f86794a68aa45c550e80ed7c315a5c890213e3ebaebb741bcd70d963a7ca15d0).
> - `POST /runs/:runId/trigger` now expects `selectedRecordId` as a pipe-separated string and deserializes it into an array, validated in [`pending-data-validators.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1594/files#diff-3ea7ed8e1b0fbeae14884ba0498905c951e4a071e80bac7cb8408ab76d912198).
> - Step executors (`load-related-record`, `read-record`, `update-record`, `trigger-action`) now pass the full `RecordId` array to activity log args instead of only the first segment; the activity log adapter serializes it to a pipe string before sending.
> - `toAvailableStepExecution` in [`run-to-available-step-mapper.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1594/files#diff-d8e6d94d08cd810252fadde0e0f846eca9f05c1b39b880ed4bb1d2357ca64abf) now throws `InvalidStepDefinitionError` when `selectedRecordId` is missing or empty, and deserializes the pipe string into an array.
> - Behavioral Change: composite record IDs previously truncated to their first segment are now fully preserved across the HTTP boundary.
>
> 
> #### Changes since #1594 opened
>
> - Modified `serializeRecordId` function in `workflow-executor` to validate that no record ID segment contains the pipe separator and throw `RecordIdSerializationError` when validation fails, while stringifying each segment individually before joining [b79ad21]
> - Strengthened validation of record ID segments across multiple schemas to reject empty strings [b79ad21]
> - Added test coverage for record ID serialization validation and step serialization behavior [b79ad21]
> - Removed explanatory comments from recordId serialization utilities and related types [62001d9]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized eabb6ec.</sup>
> 
>
>

### DIFF
--- a/packages/workflow-executor/src/adapters/record-id-serializer.ts
+++ b/packages/workflow-executor/src/adapters/record-id-serializer.ts
@@ -1,0 +1,7 @@
+export function serializeRecordId(recordId: Array<string | number>): string {
+  return recordId.map(String).join('|');
+}
+
+export function deserializeRecordId(value: string): Array<string | number> {
+  return value.split('|');
+}
--- a/packages/workflow-executor/src/adapters/record-id-serializer.ts
+++ b/packages/workflow-executor/src/adapters/record-id-serializer.ts
@@ -1,14 +1,30 @@
 import type { RecordId } from '../types/validated/collection';
 
+import { RecordIdSerializationError } from '../errors';
+
 // Wire format for (composite) record ids exchanged with the front and orchestrator: the segments
 // are joined with '|'. This is a cross-system contract (orchestrator + front + agent-client all use
-// it). Limitation: a primary-key value that itself contains '|' is not round-trip safe — accepted,
-// as it is the convention used throughout the Forest stack.
+// it). Throws — rather than silently corrupting the key — when a part contains the '|' separator,
+// since that would produce a malformed id that round-trips to the wrong record. Mirrors the
+// hardened serializer in `@forestadmin/agent-client` (src/record-id.ts).
 export function serializeRecordId(recordId: RecordId): string {
-  return recordId.join('|');
+  return recordId
+    .map(part => {
+      const serialized = String(part);
+
+      if (serialized.includes('|')) {
+        throw new RecordIdSerializationError(serialized);
+      }
+
+      return serialized;
+    })
+    .join('|');
 }
 
 // Always returns strings: ids travel as opaque strings across systems (no number round-trip).
+// Stays intentionally permissive — empty/malformed segments (e.g. from a partial 'a|' wire value)
+// are rejected at the validation boundaries that consume the result: `RecordRefSchema.recordId`
+// (mapper output) and the `selectedRecordId` schema in `pending-data-validators.ts` (front input).
 export function deserializeRecordId(value: string): string[] {
   return value.split('|');
 }
--- a/packages/workflow-executor/src/adapters/record-id-serializer.ts
+++ b/packages/workflow-executor/src/adapters/record-id-serializer.ts
@@ -1,8 +1,10 @@
+import type { RecordId } from '../types/validated/collection';
+
 // Wire format for (composite) record ids exchanged with the front and orchestrator: the segments
 // are joined with '|'. This is a cross-system contract (orchestrator + front + agent-client all use
 // it). Limitation: a primary-key value that itself contains '|' is not round-trip safe — accepted,
 // as it is the convention used throughout the Forest stack.
-export function serializeRecordId(recordId: Array<string | number>): string {
+export function serializeRecordId(recordId: RecordId): string {
   return recordId.join('|');
 }
 
--- a/packages/workflow-executor/src/adapters/record-id-serializer.ts
+++ b/packages/workflow-executor/src/adapters/record-id-serializer.ts
@@ -2,11 +2,6 @@ import type { RecordId } from '../types/validated/collection';
 
 import { RecordIdSerializationError } from '../errors';
 
-// Wire format for (composite) record ids exchanged with the front and orchestrator: the segments
-// are joined with '|'. This is a cross-system contract (orchestrator + front + agent-client all use
-// it). Throws — rather than silently corrupting the key — when a part contains the '|' separator,
-// since that would produce a malformed id that round-trips to the wrong record. Mirrors the
-// hardened serializer in `@forestadmin/agent-client` (src/record-id.ts).
 export function serializeRecordId(recordId: RecordId): string {
   return recordId
     .map(part => {
@@ -21,10 +16,6 @@ export function serializeRecordId(recordId: RecordId): string {
     .join('|');
 }
 
-// Always returns strings: ids travel as opaque strings across systems (no number round-trip).
-// Stays intentionally permissive — empty/malformed segments (e.g. from a partial 'a|' wire value)
-// are rejected at the validation boundaries that consume the result: `RecordRefSchema.recordId`
-// (mapper output) and the `selectedRecordId` schema in `pending-data-validators.ts` (front input).
 export function deserializeRecordId(value: string): string[] {
   return value.split('|');
 }
--- a/packages/workflow-executor/src/adapters/record-id-serializer.ts
+++ b/packages/workflow-executor/src/adapters/record-id-serializer.ts
@@ -1,7 +1,12 @@
+// Wire format for (composite) record ids exchanged with the front and orchestrator: the segments
+// are joined with '|'. This is a cross-system contract (orchestrator + front + agent-client all use
+// it). Limitation: a primary-key value that itself contains '|' is not round-trip safe — accepted,
+// as it is the convention used throughout the Forest stack.
 export function serializeRecordId(recordId: Array<string | number>): string {
-  return recordId.map(String).join('|');
+  return recordId.join('|');
 }
 
-export function deserializeRecordId(value: string): Array<string | number> {
+// Always returns strings: ids travel as opaque strings across systems (no number round-trip).
+export function deserializeRecordId(value: string): string[] {
   return value.split('|');
 }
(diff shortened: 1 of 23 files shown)


Scra3:
https://linear.app/forestadmin/issue/PRD-399/refactor-align-front-end-composite-key-handling-with-executor-pipe

packages/workflow-executor/src/adapters/record-id-serializer.ts
```diff
@@ -0,0 +1,14 @@
...
+// are joined with '|'. This is a cross-system contract (orchestrator + front + agent-client all use
+// it). Limitation: a primary-key value that itself contains '|' is not round-trip safe — accepted,
+// as it is the convention used throughout the Forest stack.
+export function serializeRecordId(recordId: RecordId): string {
+  return recordId.join('|');
+}
+
+// Always returns strings: ids travel as opaque strings across systems (no number round-trip).
+export function deserializeRecordId(value: string): string[] {
+  return value.split('|');
```
hercemer42, line 20: **[claude-opus-4-8] Should fix**

`deserializeRecordId` is a bare `split('|')` with no per-segment validation. The new guard in `run-to-available-step-mapper.ts` only rejects an empty/undefined `selectedRecordId` — it does not catch partial pipes (`'a|'` -> `['a','']`, `'|'` -> `['','']`, `'a||b'` -> `['a','','b']`). `RecordRefSchema.recordId` (`collection.ts:96`) is `z.array(...).min(1)` with no per-element non-empty constraint, so empty segments pass validation on all three wire paths (this mapper, the `POST /trigger` override in `pending-data-validators.ts:47`, and the activity-log port). An empty PK segment then queries the datasource with `''`, silently matching no/the wrong record. Low probability (requires malformed wire input) but it's exactly the loud failure the new guard intends — consider rejecting empty segments here, or tightening the schema's string branch to `.min(1)`.

packages/workflow-executor/src/adapters/record-id-serializer.ts
```diff
@@ -0,0 +1,14 @@
+import type { RecordId } from '../types/validated/collection';
+
+// Wire format for (composite) record ids exchanged with the front and orchestrator: the segments
+// are joined with '|'. This is a cross-system contract (orchestrator + front + agent-client all use
+// it). Limitation: a primary-key value that itself contains '|' is not round-trip safe — accepted,
+// as it is the convention used throughout the Forest stack.
+export function serializeRecordId(recordId: RecordId): string {
```
hercemer42, line 5: **[claude-opus-4-8] Should fix / discuss**

This re-implements an existing Forest convention more weakly than the two canonical versions. Composite PKs are pipe-joined because the public REST/JSON:API forces ids into a single URL path segment + JSON:API `id` string — real at the orchestrator boundary, but not on the executor's own front-facing API (which already used arrays). The canonical unpacker `agent/src/utils/id.ts` (`unpackId`, :37) is schema-aware: checks segment count vs PK count (:45), casts numeric PKs back to `Number` (:53), validates each value; `agent-client/src/record-id.ts` throws on `|`-in-value/nullish. This `serializeRecordId`/`deserializeRecordId` does none of that — bare join/split, no count check, no number recovery, no validation. Either reuse the hardened serializer, or keep the front boundary as arrays (the contract doc still documents `Array<string|number>`).

packages/workflow-executor/src/adapters/run-to-available-step-mapper.ts
```diff
@@ -139,6 +140,15 @@ export default function toAvailableStepExecution(
     );
   }
 
+  // selectedRecordId is typed `string` but arrives unvalidated from the wire — guard so an empty
+  // value fails loudly instead of deserializing to a bogus [''] baseRecordRef (or throwing a raw
+  // TypeError on undefined). Mirrors the collectionName/collectionId guards above.
+  if (!run.selectedRecordId) {
```
hercemer42, line 143: **[claude-opus-4-8] Preferential**

This guard runs before the `pending`/`pending.done` early-return below (`if (!pending || pending.done) return null`). Previously a terminal/done run lacking `selectedRecordId` was silently skipped (returned `null`); now it throws `InvalidStepDefinitionError` and is classified as `malformed`. `selectedRecordId` is a required `string`, so this is a low-probability edge — but if preserving the prior "skip terminal runs quietly" behavior matters, place the guard after the `pending`/`done` check.

packages/workflow-executor/src/http/step-serializer.ts
```diff
@@ -0,0 +1,54 @@
...
+    case 'trigger-action':
+      return { ...step, selectedRecordRef: serializeRecordRef(step.selectedRecordRef) };
+
+    case 'load-related-record': {
+      const result: Record<string, unknown> = {
+        ...step,
+        selectedRecordRef: serializeRecordRef(step.selectedRecordRef),
+      };
+
+      if (step.pendingData) {
```
hercemer42, line 23: **[claude-opus-4-8] Should fix (tests)**

There is no dedicated test for `step-serializer.ts`; it's only exercised by one all-fields-present GET test in `executor-http-server.test.ts`. The `load-related-record` sub-branches are behaviorally untested even though line coverage looks complete: `suggestedRecord` absent (line 32), `executionResult: { skipped: true }` (the false side of `'record' in executionResult`, line 41), and `pendingData` absent (this `if`). Per CLAUDE.md test conventions (cover edge cases, verify behavior not just lines), add a `step-serializer.test.ts` asserting these branches round-trip without fabricating `undefined` keys.

hercemer42 (changes requested):

hercemer42 (approved):

packages/workflow-executor/src/ports/activity-log-port.ts
```diff
@@ -1,9 +1,11 @@
+import type { RecordId } from '../types/validated/collection';
+
 export interface CreateActivityLogArgs {
   renderingId: number;
   action: string;
   type: 'read' | 'write';
   collectionId?: string;
-  recordId?: string | number;
+  recordId?: RecordId;
```
EnkiP, line 8: shouldn't we still have string | number in activity logs ?

packages/workflow-executor/src/types/validated/collection.ts
```diff
@@ -94,12 +94,15 @@ export type CollectionSchema = z.infer<typeof CollectionSchemaSchema>;
...
-    recordId: z.array(z.union([z.string(), z.number()])).min(1),
+    recordId: z.array(z.union([z.string().min(1), z.number()])).min(1),
     // Index of the workflow step that loaded this record.
     stepIndex: z.number().int().nonnegative(),
   })
   .strict();
 export type RecordRef = z.infer<typeof RecordRefSchema>;
 
+// A record's primary key: one segment for a simple key, several for a composite key.
+export type RecordId = RecordRef['recordId'];
```
EnkiP, line 105: I think it was clearer to leave Array<string | number>

EnkiP: maybe just renaming it RecordIdArray is ok

Scra3: I prefer to have one source of truth.

packages/workflow-executor/src/ports/activity-log-port.ts
Scra3: It is — the array is only the internal representation. The adapter serializes it to the pipe wire format before sending (forestadmin-client-activity-log-port.ts:39: `recordId: args.recordId?.length ? serializeRecordId(args.recordId) : undefined`), so the activity-log payload still goes out as a string. `RecordId` (array of segments) also lets it carry composite primary keys, which a scalar `string | number` couldn't.

Scra3: RecordId will be formatted when we send the request.

PR #1599, opened 2026-05-27
fix(ai-proxy): route Forest connectors via tool-provider split in AiClient

christophebrun-forest:
## Summary

- `AiClient.loadRemoteTools` now delegates to `createToolProviders` (same path as `Router` and `ToolSourceChecker`), so Forest-hosted connectors (Zendesk, Kolar, Snowflake) are routed to `ForestIntegrationClient` instead of being forwarded to `MultiServerMCPClient`.
- Before this fix, any workflow run whose `getMcpServerConfigs()` response included a Forest connector crashed at step setup with `ZodError invalid_union mcpServers.<name>` — Forest connector entries have neither `command`+`args` (stdio) nor `url` (HTTP), so the `@langchain/mcp-adapters` schema rejected them.
- Propagates the cleaner `Record<string, ToolConfig>` shape through `AiModelPort`, the three adapters (`ServerAiAdapter`, `AiClientAdapter`, `AlwaysErrorAiModelPort`), `Runner.fetchRemoteTools` (drops the now-redundant `{ configs }` wrapping), and `WorkflowPort.getMcpServerConfigs` (was mistyped as `McpServers` while the orchestrator already returns a mixed record).

Fixes [PRD-400](https://linear.app/forestadmin/issue/PRD-400/workflow-executor-forest-connectors-zendeskkolarsnowflake-crash).

## Test plan

- [x] `yarn workspace @forestadmin/ai-proxy test` — 400 passing (new regression test in `ai-client.test.ts` covers Forest connector configs lacking MCP transport fields).
- [x] `yarn workspace @forestadmin/workflow-executor test` — 786 passing.
- [x] `yarn workspace @forestadmin/ai-proxy lint && yarn workspace @forestadmin/workflow-executor lint` — clean (preexisting warnings only).
- [x] `yarn workspace @forestadmin/ai-proxy build && yarn workspace @forestadmin/workflow-executor build` — clean.
- [ ] Smoke test against the dev orchestrator with a workflow that has a Zendesk step: the executor must load the Zendesk tools and the step must reach AI selection without the Zod crash.

> [!NOTE]
> ### Route Forest connectors via a generic tool-provider split in `AiClient`
> - Replaces the single `McpClient` field in `AiClient` with a `toolProviders` array, allowing multiple provider kinds (e.g. MCP and Forest) to be loaded together via `createToolProviders`.
> - `loadRemoteTools` now accepts `Record<string, ToolConfig>` instead of `McpConfiguration`, creates providers for each config entry, flattens their tools, and disposes any previously loaded providers before replacing them.
> - All port interfaces (`AiModelPort`, `WorkflowPort`) and adapters are updated to use `Record<string, ToolConfig>` end-to-end, removing the `McpServers` wrapper type.
> - `disposeToolProviders` uses `Promise.allSettled` to isolate per-provider failures and log each rejection individually without interrupting others.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized fcc1305.</sup>
> 
>

### DIFF
--- a/packages/ai-proxy/src/ai-client.ts
+++ b/packages/ai-proxy/src/ai-client.ts
@@ -1,12 +1,14 @@
-import type { McpConfiguration } from './mcp-client';
 import type { AiConfiguration } from './provider';
+import type RemoteTool from './remote-tool';
+import type { ToolProvider } from './tool-provider';
+import type { ToolConfig } from './tool-provider-factory';
 import type { Logger } from '@forestadmin/datasource-toolkit';
 import type { BaseChatModel } from '@langchain/core/language_models/chat_models';
 
 import { createBaseChatModel } from './create-base-chat-model';
 import { AINotConfiguredError } from './errors';
 import getAiConfiguration from './get-ai-configuration';
-import McpClient from './mcp-client';
+import { createToolProviders } from './tool-provider-factory';
 import validateAiConfigurations from './validate-ai-configurations';
 
 // eslint-disable-next-line import/prefer-default-export
@@ -14,7 +16,7 @@ export class AiClient {
   private readonly aiConfigurations: AiConfiguration[];
   private readonly logger?: Logger;
   private readonly modelCache = new Map<string, BaseChatModel>();
-  private mcpClient?: McpClient;
+  private toolProviders: ToolProvider[] = [];
 
   constructor(params?: { aiConfigurations?: AiConfiguration[]; logger?: Logger }) {
     this.aiConfigurations = params?.aiConfigurations ?? [];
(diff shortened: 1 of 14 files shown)


packages/ai-proxy/test/ai-client.test.ts
```diff
@@ -313,95 +247,79 @@ describe('closeConnections', () => {
...
-  it('closes the McpClient', async () => {
-    const disposeMock = jest.fn();
-    jest.mocked(McpClient).mockImplementation(
-      () =>
-        ({
-          loadTools: jest.fn().mockResolvedValue([]),
-          dispose: disposeMock,
-        } as unknown as McpClient),
-    );
+  it('disposes every stored provider', async () => {
```
hercemer42, line 250: **Claude Opus 4.7** · **Should fix**

The dispose-failure tests all use a single provider, and this two-provider test has both `dispose` resolve — so nothing pins the `Promise.allSettled` isolation contract that this PR introduces. If `disposeToolProviders` were refactored back to `Promise.all`, a leading rejection would skip the second provider's disposal and every test here would still pass. Add a case with two providers where the first `dispose` rejects and the second resolves, then assert the second's `dispose` still ran and the logger was called once with the rejected reason.

packages/ai-proxy/src/ai-client.ts
```diff
@@ -36,32 +38,34 @@ export class AiClient {
...
-    await this.disposeMcpClient('Error closing previous MCP connection');
+  async loadRemoteTools(configs: Record<string, ToolConfig>): Promise<RemoteTool[]> {
+    await this.disposeToolProviders('Error closing previous remote tool connection');
 
-    const newClient = new McpClient(mcpConfig, this.logger);
-    const tools = await newClient.loadTools();
-    this.mcpClient = newClient;
+    const providers = createToolProviders(configs, this.logger);
+    const toolsByProvider = await Promise.all(providers.map(p => p.loadTools()));
+    this.toolProviders = providers;
```
hercemer42, line 46: **Claude Opus 4.7** · **Preferential**

`this.toolProviders = providers` runs only after `await Promise.all(...)` resolves. If any provider's `loadTools()` rejects, the created providers are never stored, so a provider that already opened connections can't be disposed via `closeConnections()`. This is currently latent — `ForestIntegrationClient.loadTools` opens nothing and `McpClient.loadTools` swallows per-server errors rather than rejecting — and it matches the previous single-`McpClient` pattern, so not a regression. Defensive option: assign before awaiting, or wrap the load in a try/catch that disposes the freshly-created providers before rethrowing. (The test at line 229 currently codifies the drop-on-failure behaviour.)

packages/workflow-executor/src/adapters/forest-server-workflow-port.ts
```diff
@@ -203,10 +203,11 @@ export default class ForestServerWorkflowPort implements WorkflowPort {
     );
   }
 
-  async getMcpServerConfigs(): Promise<McpServers> {
+  async getMcpServerConfigs(): Promise<Record<string, ToolConfig>> {
```
hercemer42, line 206: `getToolServerConfigs` would be more consistent, but I think it's out of scope for this PR.  If you agree, perhaps file a refactor ticket ?

packages/ai-proxy/test/ai-client.test.ts
hercemer42: (Claude's human here) - Good safeguard to have but it's more of a preferential for me

hercemer42 (approved):

PR #1610, opened 2026-05-29
fix: load related record

EnkiP:
### General

### Security

> [!NOTE]
> ### Rework load-related-record step to support field selection, candidate ranking, and x-to-one fetching
> - Redesigns the `LoadRelatedRecordStepExecutor` to produce richer `pendingData` containing `availableFields`, `suggestedField`, `availableRecordIds` (with optional `referenceFieldValue`), and `suggestedRecord` instead of the previous flat shape.
> - Adds `getSingleRelatedData` to `AgentPort` and `AgentClientAgentPort` for fetching x-to-one (BelongsTo/HasOne) relations via parent-record projection rather than a list query.
> - Supports mid-step field switching: a patch with only `fieldName` (no `userConfirmed`) triggers candidate refresh and returns `awaiting-input` without completing the step.
> - Normalizes `relatedCollectionName` in `getCollectionSchema` to strip trailing target-key suffixes (e.g. `store.id` → `store`) and surfaces an optional `referenceField` from collection schemas.
> - Updates `loadRelatedRecordPatchSchema` HTTP validator to accept preview patches (`fieldName`-only) and rename `name` → `fieldName`; `selectedRecordId` is only required when confirming with a relation override.
> - Risk: `getRelatedData` now requires callers to supply `relatedSchema`; any adapter calling it without the new argument will break at runtime.
>
> 
> #### Changes since #1610 opened
>
> - Updated comment in `stripReferenceKey` utility function to clarify that the related primary key is derived from the schema's `primaryKeyFields` rather than from a `relatedPrimaryKey` field [88bbefa]
> - Removed `relatedPrimaryKey` property from mocked field definitions across test cases in `ForestServerWorkflowPort.getCollectionSchema` test suite and updated assertions to verify removal of arbitrary unknown field keys [88bbefa]
> - Added character and field limits to `LoadRelatedRecordStepExecutor.selectBestRecordIndex` method in `workflow-executor` package [c7ce762]
> - Capped field selection count in `LoadRelatedRecordStepExecutor` AI field-selection routine in `workflow-executor` package [c7ce762]
> - Added tests verifying AI payload bounding behavior in `workflow-executor` package [c7ce762]
> - Added run correlation context to AI prompt truncation warnings in `LoadRelatedRecordStepExecutor.doExecute` method by spreading `logCtx` into logger metadata, and removed trailing explanatory comment from `MAX_RELEVANT_FIELDS` constant [2d9583c]
> - Added test coverage for truncation warning log context, AI-selected index mapping after prompt truncation, and value clamping behavior for non-string field types in `LoadRelatedRecordStepExecutor` test suite [2d9583c]
> - Changed visibility of `BaseStepExecutor.logCtx` getter from private to protected [2d9583c]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7b222de.</sup>
> 
>
>

### DIFF
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -110,24 +110,32 @@ export default class AgentClientAgentPort implements AgentPort {
     });
   }
 
-  // Returns raw rows as deserialized by agent-client (camelCase keys, no PK extraction).
-  // The caller resolves the related collection's schema and maps rows → RecordData; keeping
-  // schema-aware mapping out of the port avoids the silent fallback when the related
-  // collection isn't in the cache.
   async getRelatedData(
-    { collection, id, relation, limit, fields }: GetRelatedDataQuery,
+    { collection, id, relation, relatedSchema, limit, fields }: GetRelatedDataQuery,
     user: StepUser,
-  ): Promise<Record<string, unknown>[]> {
+  ): Promise<RecordData[]> {
     return this.callAgent('getRelatedData', async () => {
       const client = this.createClient(user);
-
-      return client
+      const rows = await client
         .collection(collection)
         .relation(relation, id)
         .list<Record<string, unknown>>({
           ...(limit !== null && { pagination: { size: limit, number: 1 } }),
           ...(fields?.length && { fields }),
         });
+
+      return rows.map(row => {
+        const restored = restoreFieldNames(
+          row,
+          relatedSchema.fields.map(f => f.fieldName),
+        );
+
+        return {
+          collectionName: relatedSchema.collectionName,
+          recordId: relatedSchema.primaryKeyFields.map(f => restored[f] as string | number),
+          values: restored,
+        };
+      });
     });
   }
 
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -25,7 +25,7 @@ import {
 // The agent-client HTTP layer deserializes JSON:API responses with camelCase keys.
 // Field names in the schema and in GetRecordQuery.fields use the original format (e.g. snake_case).
 // This function restores the original field names so callers can look up values by schema fieldName.
-function restoreFieldNames(
+export function restoreFieldNames(
   values: Record<string, unknown>,
   originalFieldNames: string[] | undefined,
 ): Record<string, unknown> {
@@ -110,37 +110,24 @@ export default class AgentClientAgentPort implements AgentPort {
     });
   }
 
+  // Returns raw rows as deserialized by agent-client (camelCase keys, no PK extraction).
+  // The caller resolves the related collection's schema and maps rows → RecordData; keeping
+  // schema-aware mapping out of the port avoids the silent fallback when the related
+  // collection isn't in the cache.
   async getRelatedData(
     { collection, id, relation, limit, fields }: GetRelatedDataQuery,
     user: StepUser,
-  ): Promise<RecordData[]> {
+  ): Promise<Record<string, unknown>[]> {
     return this.callAgent('getRelatedData', async () => {
       const client = this.createClient(user);
-      const parentSchema = this.resolveSchema(collection);
-      const relationField = parentSchema.fields.find(f => f.fieldName === relation);
-      const relatedCollectionName = relationField?.relatedCollectionName ?? relation;
-      const relatedSchema = this.resolveSchema(relatedCollectionName);
 
-      const records = await client
+      return client
         .collection(collection)
         .relation(relation, id)
         .list<Record<string, unknown>>({
           ...(limit !== null && { pagination: { size: limit, number: 1 } }),
           ...(fields?.length && { fields }),
         });
-
-      return records.map(record => {
-        const restored = restoreFieldNames(record, [
-          ...relatedSchema.primaryKeyFields,
-          ...(fields ?? []),
-        ]);
-
-        return {
-          collectionName: relatedSchema.collectionName,
-          recordId: relatedSchema.primaryKeyFields.map(f => restored[f] as string | number),
-          values: restored,
-        };
-      });
     });
   }
 
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -149,14 +149,17 @@ export default class AgentClientAgentPort implements AgentPort {
     user: StepUser,
   ): Promise<RecordData | null> {
     return this.callAgent('getSingleRelatedData', async () => {
-      const projectedFields = Array.from(
-        new Set([...relatedSchema.primaryKeyFields, ...(fields ?? [])]),
-      );
+      // The agent can't parse multiple sub-fields on one relation in a single projection
+      // (`fields[store]=id,name` is read as a single field name → ValidationError). The linkage
+      // `id` carries the (packed) related PK regardless of projection, so project at most ONE
+      // field: the requested reference field for display, else a single PK field just to pull the
+      // relation into the response.
+      const projectedField = fields?.[0] ?? relatedSchema.primaryKeyFields[0];
       const parent = await this.getRecord(
         {
           collection,
           id,
-          fields: projectedFields.map(f => `${relation}@@@${f}`),
+          fields: [`${relation}@@@${projectedField}`],
         },
         user,
       );
@@ -166,7 +169,7 @@ export default class AgentClientAgentPort implements AgentPort {
 
       if (!linkage || !packedId) return null;
 
-      const restored = restoreFieldNames(linkage, projectedFields);
+      const restored = restoreFieldNames(linkage, [projectedField]);
 
       return {
         collectionName: relatedSchema.collectionName,
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -23,6 +23,10 @@ import {
   extractErrorMessage,
 } from '../errors';
 
+function toCamelCase(name: string): string {
+  return name.replace(/_([a-zA-Z0-9])/g, (_, c: string) => c.toUpperCase());
+}
+
 // The agent-client HTTP layer deserializes JSON:API responses with camelCase keys.
 // Field names in the schema and in GetRecordQuery.fields use the original format (e.g. snake_case).
 // This function restores the original field names so callers can look up values by schema fieldName.
@@ -35,8 +39,7 @@ function restoreFieldNames(
   const camelToOriginal: Record<string, string> = {};
 
   for (const name of originalFieldNames) {
-    const camelName = name.replace(/_([a-zA-Z0-9])/g, (_, c: string) => c.toUpperCase());
-    camelToOriginal[camelName] = name;
+    camelToOriginal[toCamelCase(name)] = name;
   }
 
   return Object.fromEntries(Object.entries(values).map(([k, v]) => [camelToOriginal[k] ?? k, v]));
@@ -164,7 +167,11 @@ export default class AgentClientAgentPort implements AgentPort {
         user,
       );
 
-      const linkage = parent.values[relation] as Record<string, unknown> | null | undefined;
+      // agent-client camelCases relation keys; look the linkage up under the camelCased name.
+      const linkage = parent.values[toCamelCase(relation)] as
+        | Record<string, unknown>
+        | null
+        | undefined;
       const packedId = linkage?.id as string | undefined;
 
       if (!linkage || !packedId) return null;
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -4,6 +4,7 @@ import type {
   GetActionFormInfoQuery,
   GetRecordQuery,
   GetRelatedDataQuery,
+  GetSingleRelatedDataQuery,
   UpdateRecordQuery,
 } from '../ports/agent-port';
 import type SchemaCache from '../schema-cache';
@@ -136,6 +137,42 @@ export default class AgentClientAgentPort implements AgentPort {
           values: restored,
         };
       });
+    });
+  }
+
+  // xToOne relations have no /relationships/<relation> route on the agent. We read the
+  // parent record with a `<relation>@@@<field>` projection and unpack the relation linkage
+  // jsonapi-serializer emits as a nested object on the parent (with the related PK packed
+  // under "id" when composite).
+  async getSingleRelatedData(
+    { collection, id, relation, relatedSchema, fields }: GetSingleRelatedDataQuery,
+    user: StepUser,
+  ): Promise<RecordData | null> {
+    return this.callAgent('getSingleRelatedData', async () => {
+      const projectedFields = Array.from(
+        new Set([...relatedSchema.primaryKeyFields, ...(fields ?? [])]),
+      );
+      const parent = await this.getRecord(
+        {
+          collection,
+          id,
+          fields: projectedFields.map(f => `${relation}@@@${f}`),
+        },
+        user,
+      );
+
+      const linkage = parent.values[relation] as Record<string, unknown> | null | undefined;
+      const packedId = linkage?.id as string | undefined;
+
+      if (!linkage || !packedId) return null;
+
+      const restored = restoreFieldNames(linkage, projectedFields);
+
+      return {
+        collectionName: relatedSchema.collectionName,
+        recordId: packedId.split('|'),
+        values: restored,
+      };
     });
   }
 
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -26,7 +26,7 @@ import {
 // The agent-client HTTP layer deserializes JSON:API responses with camelCase keys.
 // Field names in the schema and in GetRecordQuery.fields use the original format (e.g. snake_case).
 // This function restores the original field names so callers can look up values by schema fieldName.
-export function restoreFieldNames(
+function restoreFieldNames(
   values: Record<string, unknown>,
   originalFieldNames: string[] | undefined,
 ): Record<string, unknown> {
--- a/packages/workflow-executor/src/executors/load-related-record-step-executor.ts
+++ b/packages/workflow-executor/src/executors/load-related-record-step-executor.ts
@@ -38,7 +38,7 @@ Choose the record that best matches the user request based on the provided field
 
 // Bound only what is sent to the AI in selectBestRecordIndex — the full candidate list is
 // always returned to the front via availableRecordIds. These cap the prompt size, not the data.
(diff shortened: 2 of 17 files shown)


packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -104,41 +152,172 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
+    const parent = await this.agentPort.getRecord(
+      {
+        collection: target.selectedRecordRef.collectionName,
+        id: target.selectedRecordRef.recordId,
+        fields,
+      },
+      this.context.user,
+    );
+
+    const relation = parent.values[target.name] as Record<string, unknown> | null | undefined;
```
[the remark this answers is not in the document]
EnkiP: fix it for me

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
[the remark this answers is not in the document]
EnkiP: out of scope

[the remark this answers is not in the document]
EnkiP: out of scope

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -104,41 +152,172 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
+    const packedId = relation?.id as string | undefined;
+
+    if (!packedId) {
+      throw new RelatedRecordNotFoundError(target.selectedRecordRef.collectionName, target.name);
+    }
+
+    return {
+      recordId: packedId.split('|'),
+      referenceFieldValue: referenceField
```
hercemer42, line 294: **[Claude Opus 4.8] — Should fix**

`extractReferenceFieldValue(relation, referenceField)` reads the reference field off the nested relation object, but that object is the raw deserialized relationship — its attribute keys are camelCased (`full_name` → `fullName`) and are never passed through `restoreFieldNames`. So a snake_case `referenceField` always yields `undefined` → `null`, and the human-readable label silently never renders on the xToOne path.

The comment at lines 212-215 ("fetchRelatedData has already restored field names … so reading r.values[field] works") holds for the to-many path but not for this one. Camel-case `referenceField` before reading, or run the nested object through `restoreFieldNames`. Same root cause as the relation-key Must-fix above. The test at line ~1594 hand-builds the `getRecord` result already in restored form, so it can't catch this.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -149,10 +328,26 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
-    const name = userConfirmation?.name ?? pendingData.name;
-    const selectedRecordId = userConfirmation?.selectedRecordId ?? pendingData.selectedRecordId;
+    // If the user switched relations, look up the chosen displayName in availableFields to
+    // recover its frozen technical name. Otherwise fall back to the AI's suggestion.
+    const relationRef = userConfirmation?.fieldDisplayName
+      ? pendingData.availableFields.find(f => f.displayName === userConfirmation.fieldDisplayName)
+      : pendingData.suggestedField;
 
-    // Re-derive relatedCollectionName and displayName because the user may have swapped the relation.
+    if (!relationRef) {
```
hercemer42, line 314: **[Claude Opus 4.8] — Should fix (coverage)**

This reachable error path is untested. The validator refine in `pending-data-validators.ts` accepts `{ userConfirmed: true, fieldDisplayName: <stale/renamed>, selectedRecordId }` without checking the name is a real entry in `availableFields`, so a stale relation from the frontend lands here and throws `StepStateError`. The preview twin is tested ("returns error when the previewed relation does not exist") but this confirm-with-override branch is not — it's the one genuinely reachable uncovered line in the file. Per CLAUDE.md (changed files ≥90% covered; cover error edge cases), add a test asserting `status === 'error'` and that `saveStepExecution` is not called.

packages/workflow-executor/src/types/step-execution-data.ts
```diff
@@ -133,11 +133,26 @@ export interface RecordStepExecutionData extends BaseStepExecutionData {
...
-  // AI-selected initially; frontend can override via userConfirmation.selectedRecordId.
-  selectedRecordId: Array<string | number>;
+// One row of `availableRecordIds`. `referenceFieldValue` is the layout-level reference
+// field (collection.displayFieldName on the orchestrator side) — when the related
+// collection has one configured and the agent returns a value, the frontend shows it
+// instead of the raw id. Both `null` (no reference field configured) and `undefined`
+// (configured but value missing) collapse to the same "fall back to id" rendering.
+export interface LoadRelatedRecordCandidate {
+  recordId: Array<string | number>;
+  referenceFieldValue?: string | null;
```
hercemer42: **[Claude Opus 4.8] — Preferential**

`referenceFieldValue?: string | null` is a three-state type, but every producer (`toCandidate`, `fetchXToOneCandidate`) and `extractReferenceFieldValue` only ever emit `string | null` — `undefined` is never produced, and the doc comment says `null`/`undefined` collapse to the same rendering anyway. Tightening to a required `referenceFieldValue: string | null` removes a representable-but-never-produced state with zero behavior change.

hercemer42:
Claude has added a lot of comments, but your code is readable as is, so in my opinion they aren't necessary.  I'd recommend to keep them to a minimum.

hercemer42 (changes requested):

packages/workflow-executor/src/http/pending-data-validators.ts
```diff
@@ -23,25 +23,45 @@ const triggerActionPatchSchema = z
...
   .object({
-    userConfirmed: z.boolean(),
+    userConfirmed: z.boolean().optional(),
     // User may intentionally switch to a different relation than the one the AI selected.
-    // The executor re-derives relatedCollectionName and displayName from FieldSchema when
-    // processing the confirmation.
-    name: z.string().min(1).optional(),
+    // Sent as the displayName; the executor re-derives the technical fieldName and
+    // relatedCollectionName from the live schema when processing the confirmation.
+    fieldDisplayName: z.string().min(1).optional(),
```
Scra3: Why don't send directly the technical name?

packages/workflow-executor/src/ports/agent-port.ts
```diff
@@ -25,7 +25,9 @@ export type GetActionFormInfoQuery = { collection: string; action: string; id: I
 export interface AgentPort {
   getRecord(query: GetRecordQuery, user: StepUser): Promise<RecordData>;
   updateRecord(query: UpdateRecordQuery, user: StepUser): Promise<RecordData>;
-  getRelatedData(query: GetRelatedDataQuery, user: StepUser): Promise<RecordData[]>;
+  // Returns raw rows from the agent (camelCase keys, no PK extraction). The caller is
+  // responsible for resolving the related collection's schema and mapping rows → RecordData.
+  getRelatedData(query: GetRelatedDataQuery, user: StepUser): Promise<Record<string, unknown>[]>;
```
Scra3: Why it's not RecordData[] to return ?

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -245,25 +439,50 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
         limit,
       },
       this.context.user,
     );
 
-    if (relatedData.length === 0) {
-      throw new RelatedRecordNotFoundError(selectedRecordRef.collectionName, name);
-    }
+    return rows.map(row => {
+      const restored = restoreFieldNames(
```
Scra3: Don't use something from adapter inside business domain. If want to format data, you must to do it inside adapter/port. It's not the job of the step business :)

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -104,41 +152,172 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
+        `Cannot load xToOne relation "${target.name}" on collection ` +
+          `"${target.selectedRecordRef.collectionName}": missing relatedPrimaryKey in schema.`,
+      );
+    }
+
+    // Resolve the related schema for the optional referenceField. Cached after the first
+    // lookup, so the extra fetch only pays once per related collection per run.
+    const relatedSchema = await this.getCollectionSchema(target.relatedCollectionName);
+    const referenceField = relatedSchema.referenceField ?? null;
+    const fields = [`${target.name}@@@${target.relatedPrimaryKey}`];
```
Scra3: Same than my previous comment, this syntaxe is related to HTTP transport. It must be move to the port.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
```diff
@@ -110,37 +110,24 @@ export default class AgentClientAgentPort implements AgentPort {
...
 
+  // Returns raw rows as deserialized by agent-client (camelCase keys, no PK extraction).
+  // The caller resolves the related collection's schema and maps rows → RecordData; keeping
+  // schema-aware mapping out of the port avoids the silent fallback when the related
+  // collection isn't in the cache.
   async getRelatedData(
     { collection, id, relation, limit, fields }: GetRelatedDataQuery,
     user: StepUser,
-  ): Promise<RecordData[]> {
+  ): Promise<Record<string, unknown>[]> {
```
Scra3: It returns RecordData[], why it's moved to an other type ?

packages/workflow-executor/src/http/pending-data-validators.ts
EnkiP: The frontend needs to display the displayName, we don't really need the tech name for anything

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -55,6 +64,15 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
     );
 
     if (pending) {
+      // Branch A-preview -- user switched relation without confirming: re-list candidates
+      // for the new relation, refresh pendingData, stay awaiting-input. Detected by a
+      // patch carrying `fieldDisplayName` but no `userConfirmed`.
+      const conf = pending.userConfirmation;
```
Scra3, line 76: This preview-vs-confirm discrimination duplicates the `refine` in `http/pending-data-validators.ts` (loadRelatedRecordPatchSchema) — the two encode the same rule and can
  drift if it changes; consider a single source of truth (tagged union at parse time, or a shared predicate).

Scra3 (changes requested):

packages/workflow-executor/src/http/pending-data-validators.ts
Scra3: Prefer to use fieldName, it's the unique key. DisplayName can be not unique. Also it can change during the process. This field is used to display nicely the field, it's a cosmetic field

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -35,6 +40,10 @@ Choose the record that best matches the user request based on the provided field
 interface RelationTarget extends RelationRef {
   selectedRecordRef: RecordRef;
   relationType?: 'BelongsTo' | 'HasMany' | 'HasOne' | 'BelongsToMany';
+  relatedCollectionName: string;
+  // Primary key field name on the related collection — supplied by the orchestrator's
+  // schema. Required for the xToOne projection syntax ('<relation>@@@<pk>').
+  relatedPrimaryKey?: string;
```
Scra3: Is it possible to be computedKey?

packages/workflow-executor/src/ports/agent-port.ts
EnkiP: Because I wanted to avoid having to pass the schema to the agent-port, but it's ok, I change

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
EnkiP: I don't see how to move this logic to the agentPort. Here we are fetching some fields in the relation from the parent record, to get the id and the referenceField. At this step we don't have the id so we can't fetch the child record directly.
This logic is heavily related to the loadRelatedRecord step so I think it's ok to leave it there

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
EnkiP: It wasn't, the restoreFieldNames was working only for the already fetched collection schema, which wasn't the case for relationships

EnkiP: I fixed it, & restored the RecordData[] type

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
EnkiP: I don't understand, nothing seems repeated, we simply do different things according to the user input

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
Scra3: Yes that why adapter and port exists ! We don't want to be linked to the agent format. Adapter and port encapsulate how we communicate with the outside. This logic must be in the adapter :/

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -35,6 +40,10 @@ Choose the record that best matches the user request based on the provided field
 interface RelationTarget extends RelationRef {
   selectedRecordRef: RecordRef;
   relationType?: 'BelongsTo' | 'HasMany' | 'HasOne' | 'BelongsToMany';
+  relatedCollectionName: string;
+  // Primary key field name on the related collection — supplied by the orchestrator's
+  // schema. Required for the xToOne projection syntax ('<relation>@@@<pk>').
+  computedKey?: string;
```
Scra3: change the name: Type is recordId I think

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
```diff
@@ -25,7 +26,7 @@ import {
 // The agent-client HTTP layer deserializes JSON:API responses with camelCase keys.
 // Field names in the schema and in GetRecordQuery.fields use the original format (e.g. snake_case).
 // This function restores the original field names so callers can look up values by schema fieldName.
-function restoreFieldNames(
+export function restoreFieldNames(
```
Scra3: Export to remove

packages/workflow-executor/src/types/step-execution-data.ts
```diff
@@ -133,11 +133,26 @@ export interface RecordStepExecutionData extends BaseStepExecutionData {
...
+  recordId: Array<string | number>;
+  referenceFieldValue: string | null;
+}
+
+export interface LoadRelatedRecordPendingData {
+  // All relation fields available on the source collection (name + displayName each).
+  availableFields: RelationRef[];
+  // AI-selected relation from `availableFields`.
+  suggestedField: RelationRef;
+  // First 50 records of `suggestedField` paired with their reference-field value.
```
Scra3: If we change to 40 or whatever, the comment will be outdated. Remove it plz :)

packages/workflow-executor/src/types/step-execution-data.ts
```diff
@@ -133,11 +133,26 @@ export interface RecordStepExecutionData extends BaseStepExecutionData {
 
 // -- Load Related Record --
 
-export interface LoadRelatedRecordPendingData extends RelationRef {
-  // undefined when not computed (record has no non-relation fields).
-  suggestedFields?: string[];
-  // AI-selected initially; frontend can override via userConfirmation.selectedRecordId.
-  selectedRecordId: Array<string | number>;
+// One row of `availableRecordIds`. `referenceFieldValue` is the layout-level reference
```
Scra3: We can remove all the comment.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -266,6 +395,23 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
     return relatedData.map(r => this.toRecordRef(r));
   }
 
+  private async fetchRelatedData(
```
Scra3, line 432: I think we can remove this method. It does nothing

EnkiP: it is used at 2 places, it's not big but it simplifies a little the call to the agentPort

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
```diff
@@ -144,6 +140,42 @@ export default class AgentClientAgentPort implements AgentPort {
...
+  // parent record with a `<relation>@@@<field>` projection and unpack the relation linkage
+  // jsonapi-serializer emits as a nested object on the parent (with the related PK packed
+  // under "id" when composite).
+  async getSingleRelatedData(
+    { collection, id, relation, relatedSchema, fields }: GetSingleRelatedDataQuery,
+    user: StepUser,
+  ): Promise<RecordData | null> {
+    return this.callAgent('getSingleRelatedData', async () => {
+      const projectedFields = Array.from(
+        new Set([...relatedSchema.primaryKeyFields, ...(fields ?? [])]),
```
Scra3: issue (blocking): projecting the PK *and* the reference field on one relation produces `fields[<relation>]=id,reference`, which the agent's `parseProjection` can't split into two sub-fields — it 400s with `Invalid projection: The '<relation>.id,name' field was not found`; project a single sub-field instead (the linkage `id` is returned by the JSON:API serializer regardless of projection).

Scra3 (commented):
Found a blocking bug in the xToOne relation projection (`getSingleRelatedData`).

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
Scra3: Composite/computed keys are supported: the related PK comes from `relatedSchema.primaryKeyFields` (a `string[]`), and a record id is `recordId: Array<string | number>` (composite ids arrive packed as `a|b` in the linkage and are split back in the adapter). The single-string `relatedPrimaryKey`/`computedKey` field on `RelationTarget` that prompted this — which indeed could not represent a composite key — has been removed.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
Scra3: Moot now: the `relatedPrimaryKey`/`computedKey` field on `RelationTarget` was removed (it was dead — the related PK comes from `relatedSchema.primaryKeyFields`), so there is no name left to change.

packages/workflow-executor/src/types/step-execution-data.ts
Scra3: Done — the magic-number comment was removed during the refactor; `LoadRelatedRecordPendingData.availableRecordIds` no longer carries it.

Scra3 (approved):

Scra3 (changes requested):
wait i saw a bug

Scra3 (approved):

hercemer42 (approved):

PR #1612, opened 2026-05-29
fix(workflow-executor): coerce update-record value to field type, support boolean and arrays

Tonours:
PRD-408

### General

### Security

### DIFF
--- a/packages/workflow-executor/src/http/pending-data-validators.ts
+++ b/packages/workflow-executor/src/http/pending-data-validators.ts
@@ -7,7 +7,11 @@ import { z } from 'zod';
 const updateRecordPatchSchema = z
   .object({
     userConfirmed: z.boolean(),
-    value: z.union([z.string(), z.number()]).optional(), // user may override the AI-proposed value
+    // User may override the AI-proposed value with any field type (boolean, array, etc.).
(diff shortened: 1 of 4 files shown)


packages/workflow-executor/src/http/pending-data-validators.ts
```diff
@@ -7,7 +7,11 @@ import { z } from 'zod';
 const updateRecordPatchSchema = z
   .object({
     userConfirmed: z.boolean(),
-    value: z.union([z.string(), z.number()]).optional(), // user may override the AI-proposed value
+    // User may override the AI-proposed value with any field type (boolean, array, etc.).
+    // The business type is validated/coerced field-aware in the update-record step executor
+    // (buildZodSchemaForField), so the HTTP schema stays a shape guard only — aligned with
+    // `actionResult` below.
+    value: z.unknown().optional(),
```
EnkiP: your frontend PR ensure only string or number is sent

EnkiP: I'm ok to leave the executor more permissive, it will allow future changes more easily. But you can remove the coms

Tonours: done

EnkiP (approved):

PR #1615, opened 2026-06-01
fix(workflow-executor): skip activity-log update when server returns no id (PRD-428)

christophebrun-forest:
## Problem

When an MCP step completes, the executor logs showed a burst of retry warnings ending in an error:

```
warn  "activity log mark-as-completed" failed, retrying attempt=1 status=404 error="Activity log not found"
warn  "activity log mark-as-completed" failed, retrying attempt=2 status=404 error="Activity log not found"
warn  "activity log mark-as-completed" failed, retrying attempt=3 status=404 error="Activity log not found"
error activity log mark-as-completed failed handleId=null error="Activity log not found"
```

## Root cause (confirmed in `forestadmin-server`)

The Forest server's `ActivityLogCreator.create()` returns `null` when the request carries no collection (authorization check), and the create route then responds **HTTP 200 with `data: { id: null }`** by design. `createPending` returned that handle verbatim, so the subsequent mark-as-completed/failed `PATCH /{index}/null/status` hit a **permanent 404** — retried 3× (404 is in `extraRetryStatuses`) before a swallowed error, polluting logs and burning background latency on every affected step.

## Fix (client robustness)

- `createPending` warns once when the server returns no id.
- `markSucceeded` / `markFailed` early-return when the handle has no id → kills the 404 storm at the source.
- 404 stays retriable for **valid** ids (transient read-path propagation) — unchanged.

This is defense-in-depth; the functional fix that makes the log actually persist (passing `collectionId`) already landed in #1608.

## Tests
- `createPending` returns a non-persisted handle and warns on 200 + `id: null`.
- `markSucceeded` / `markFailed` do not call `updateActivityLogStatus` when the handle has no id.
- Existing valid-id behavior (incl. 404 retry) unchanged. Full suite: 836/836 green, lint clean, build OK.

fixes PRD-428

> [!NOTE]
> ### Skip activity-log updates in `ForestadminClientActivityLogPort` when server returns no id
> - `createPending` now checks the server response id and returns `{ id: null }` with a warning log instead of proceeding when the server declines to persist the activity log.
> - `markSucceeded` and `markFailed` both short-circuit immediately when given a `{ id: null }` handle, skipping the drainer registration and status PATCH calls.
> - `ActivityLogHandle` in [activity-log-port.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1615/files#diff-c22bb4587a52235a0745b2e37acaae46ca961f3124a5e19f3a5aa7654d431edb) is updated to a union type to reflect the nullable id case.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized de9f473.</sup>
> 
>

### DIFF
--- a/packages/workflow-executor/src/adapters/forestadmin-client-activity-log-port.ts
+++ b/packages/workflow-executor/src/adapters/forestadmin-client-activity-log-port.ts
@@ -39,6 +39,16 @@ export default class ForestadminClientActivityLogPort implements ActivityLogPort
         { logger: this.logger },
       );
 
+      // 200 with id:null → server declined to persist; skip the later status PATCH (PRD-428).
+      if (!response.id) {
+        this.logger.warn('activity log not persisted by server — skipping status update', {
+          action: args.action,
+          collectionId: args.collectionId,
+        });
+
+        return { id: null };
+      }
+
       return { id: response.id, index: response.attributes.index };
     } catch (cause) {
       this.logger.error('activity log create failed', {
@@ -52,6 +62,11 @@ export default class ForestadminClientActivityLogPort implements ActivityLogPort
   }
 
   async markSucceeded(handle: ActivityLogHandle): Promise<void> {
+    // Not persisted (see createPending) → nothing to patch; skip to avoid a permanent-404 storm.
+    if (!('index' in handle)) return undefined;
+
+    const { id, index } = handle;
+
     return this.drainer.track(async () => {
       try {
         await withRetry(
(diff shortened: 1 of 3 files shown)


packages/workflow-executor/src/adapters/forestadmin-client-activity-log-port.ts
```diff
@@ -53,43 +64,53 @@ export default class ForestadminClientActivityLogPort implements ActivityLogPort
   }
 
   async markSucceeded(handle: ActivityLogHandle): Promise<void> {
+    // Not persisted (see createPending) → no log to update; skip the 404-retrying status call.
+    if (!('index' in handle)) return undefined;
```
hercemer42, line 66: **[claude-opus-4-8] Preferential** — Narrowing via `!('index' in handle)` is correct and compiles, but the union's natural discriminant is `id` (`{ id: null }` vs `{ id: string; index: string }`). Keying on `handle.id === null` would align the check with the documented discriminant and the "not persisted" comments, making intent more obvious. Same pattern at line 95 (`markFailed`).

hercemer42 (approved):

christophebrun-forest:
Closing: no longer relevant. The mark-as-completed 404 was fixed at the source on the server side (Elasticsearch GET fix + passing collectionId via #1608). Retested OK (run #309): no more 404s or retries. The client-side hardening in this PR is no longer needed.

PR #1621, opened 2026-06-03
feat(workflow-executor): make collection schema cache TTL configurable [PRD-430]

matthv:
## What

The collection-schema cache duration in the workflow-executor was fixed (~10 min) with no way to configure it. Schema changes on the orchestrator side (added/renamed field or relation, display-field change) were therefore only reflected after that fixed window, with no operator control.

This wires a `schemaCacheTtlMs` executor option through the CLI via the `SCHEMA_CACHE_TTL_MS` env var — exactly like the other executor tunables (`POLLING_INTERVAL_MS`, `STEP_TIMEOUT_MS`, `MAX_CHAIN_DEPTH`).

## Changes

- `defaults.ts`: `DEFAULT_SCHEMA_CACHE_TTL_MS = 10 * 60_000`
- `build-workflow-executor.ts`: new `ExecutorOptions.schemaCacheTtlMs`, passed to `new SchemaCache(...)` (the cache already accepted a TTL in its constructor)
- `cli-core.ts`: parse `SCHEMA_CACHE_TTL_MS` + document it in `--help`
- `cli.test.ts`: parse / unset-default / non-numeric-throws coverage

## Notes

- Default behaviour is **unchanged**: 10 min when `SCHEMA_CACHE_TTL_MS` is unset.
- Base branch is `feat/prd-214-server-step-mapper` (workflow-executor is not on `main`).

fixes PRD-430

> [!NOTE]
> ### Make collection schema cache TTL configurable in workflow-executor
> - Adds `schemaCacheTtlMs` to `ExecutorOptions` in [`build-workflow-executor.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1621/files#diff-9a7b00e34df678f7ad856e6420bb042e4f6e01afc81fdc706055f462da01d493), passing it to `SchemaCache` construction with a positive/finite guard that falls back to `DEFAULT_SCHEMA_CACHE_TTL_MS` (10 minutes).
> - Exposes `SCHEMA_CACHE_TTL_MS` as an environment variable in the CLI, parsed via `parsePositiveIntEnv`; invalid or missing values fall back to the default.
> - Moves the default TTL constant to [`defaults.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1621/files#diff-1766dc2654e9359f823371f8f5f5fb6d2d0af2f79d9b9f26542364d967287f95) so it is shared across `SchemaCache` and the CLI help text.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4f082d6.</sup>
> 
>


packages/workflow-executor/src/defaults.ts
```diff
@@ -4,3 +4,4 @@ export const DEFAULT_POLLING_INTERVAL_MS = 30_000;
 export const DEFAULT_STEP_TIMEOUT_MS = 5 * 60_000;
 export const DEFAULT_STOP_TIMEOUT_MS = 30_000;
 export const DEFAULT_MAX_CHAIN_DEPTH = 50;
+export const DEFAULT_SCHEMA_CACHE_TTL_MS = 10 * 60_000;
```
hercemer42, line 8: **[Claude Opus 4.8]** [Should fix] This duplicates `DEFAULT_TTL_MS = 10 * 60 * 1000` in `schema-cache.ts:3`. Both are 600 000 ms today, but they're unlinked: the build path always passes `DEFAULT_SCHEMA_CACHE_TTL_MS`, so the constructor default becomes dead code in production and only `new SchemaCache()` callers (tests) hit it. If either constant is tuned later, the `--help` text and the constructor fallback silently diverge. Suggest having `schema-cache.ts` import this constant as its constructor default and deleting `DEFAULT_TTL_MS`.

packages/workflow-executor/src/build-workflow-executor.ts
```diff
@@ -83,7 +86,7 @@ function buildCommonDependencies(options: ExecutorOptions) {
     aiModelPort = new ServerAiAdapter({ forestServerUrl, envSecret: options.envSecret });
   }
 
-  const schemaCache = new SchemaCache();
+  const schemaCache = new SchemaCache(options.schemaCacheTtlMs ?? DEFAULT_SCHEMA_CACHE_TTL_MS);
```
hercemer42: **[Claude Opus 4.8]** [Should fix] This wiring is untested. The sibling `stepTimeoutMs` has a dedicated pair in `build-workflow-executor.test.ts:165/173` (default applied / caller value respected), but nothing asserts the TTL reaching `SchemaCache` — the existing `expect.any(SchemaCache)` passes even if the `?? DEFAULT_SCHEMA_CACHE_TTL_MS` fallback is dropped (the constructor's own default would silently mask it). Mirroring the `stepTimeoutMs` pair with `jest.mock('../src/schema-cache')` + `expect(SchemaCache).toHaveBeenCalledWith(...)` would pin it down.

packages/workflow-executor/src/cli-core.ts
```diff
@@ -195,6 +197,7 @@ Optional environment variables:
   STOP_TIMEOUT_MS        Default: ${DEFAULT_STOP_TIMEOUT_MS}
   STEP_TIMEOUT_MS        Max duration of a step in ms (default: ${DEFAULT_STEP_TIMEOUT_MS})
   MAX_CHAIN_DEPTH        Max steps auto-executed per run before yielding (default: ${DEFAULT_MAX_CHAIN_DEPTH})
+  SCHEMA_CACHE_TTL_MS    Collection schema cache TTL in ms (default: ${DEFAULT_SCHEMA_CACHE_TTL_MS})
```
hercemer42, line 203: **[Claude Opus 4.8]** [Should fix] The test `printHelp prints every default value from defaults.ts` (cli.test.ts:299) asserts the rendered default for every other env var but wasn't updated for this new line — its name no longer holds, and a broken interpolation here would go uncaught. One-liner: `expect(output).toContain(`default: ${DEFAULT_SCHEMA_CACHE_TTL_MS}`);`.

hercemer42 (changes requested):

hercemer42:
As discussed on Slack - mitigating a cache issue by reducing a refresh delay only reduces the probability of data staleness, it doesn't avoid it completely. How about instead removing TTL entirely and have the orchestrator include a schemaVersion/hash in poll responses ? That way the executor can clear its schema cache when it detects a change which will automatically trigger a new fetch.

matthv:
good idea, the TTL only lowers the probability of staleness, it doesn’t remove it.
One thing though: the schema comes from the agent, not the orchestrator. So the version/hash would need to track the agent’s schema, otherwise we’d miss changes or evict for nothing. IMO we can probably reuse the schemaFileHash that’s already computed at publish time instead of a new hash.
I’d keep this PR as-is and open a follow-up to design the hash-based invalidation properly. WDYT?

hercemer42 (approved):

PR #1622, opened 2026-06-03
fix(agent-client): serialize fields[] projection as comma-separated string (PRD-437)

christophebrun-forest:
## Problem

A `getData` / `read-record` step run through the **workflow-executor** against a **Ruby agent** returns only **the last requested field**. Reproduced: reading `first_name, last_name, email` returns only `email`. Works in **browser** mode; breaks only in **Orchestrator** mode (through the executor). TS agents are unaffected.

## Root cause

`QuerySerializer.formatFields` returns an **array** for the sparse fieldset, e.g. `{ 'fields[Customer]': ['first_name','last_name','email'] }`. superagent serializes arrays via `qs.stringify(obj, { indices: false })` as repeated keys without brackets:

```
fields[Customer]=first_name&fields[Customer]=last_name&fields[Customer]=email
```

Rack-based (Ruby) agents collapse duplicate scalar keys to the **last value**, so `parse_projection` (`fields.split(',')`, expecting the JSON:API comma-separated string) only ever sees the last field → only `email` + the PK come back.

- **Browser OK**: the frontend already sends `fields[Customer]=first_name,last_name,email`.
- **TS agent OK by luck**: it does `fields.toString().split(',')`, and `Array.toString()` rejoins with commas. Ruby does `fields.split(',')` → no tolerance for an array.

## Fix

`formatFields` now joins each projection into a single **comma-separated string** (`fields[T]=a,b,c`), the JSON:API sparse-fieldset standard, parsed correctly by both Ruby and Node agents. Also fixes the same class of issue for `update-record` / `getRelatedData`, which share this serialization.

Sibling of PRD-273 #8 (filter operator casing). Distinct from PRD-432 (JWT casing).

## Test plan

- `agent-client`: 322/322 tests pass, including a new dedicated regression test asserting `fields[users]` is a comma-separated string, not an array.
- Round-trip verified: `fields[Customer]=first_name,last_name,email` → `"...".split(",")` = `["first_name","last_name","email"]`.

fixes PRD-437

> [!NOTE]
> ### Serialize `fields[]` query params as comma-separated strings in agent-client
> Ruby's Rack middleware collapses repeated query params, so `fields[users]=id&fields[users]=name` is not handled correctly. `QuerySerializer.formatFields` in [query-serializer.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1622/files#diff-c9087a46acb037e536ea20624c746a833e1111f77ec399817cfa5855403fdf2e) now joins projection arrays into a single comma-separated string (e.g. `fields[users]=id,name`). Tests in both [query-serializer.test.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1622/files#diff-05db598106d97f1d6e6298cc6195dd6e288222475358381fc7b4e8f515ef74b1) and [server.test.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1622/files#diff-5d4a69d062bd737c23410a18e7a6ed09bff5e28f1ea07b290ae397fefc805fc5) are updated to assert string values instead of arrays.
>
> 
> #### Changes since #1622 opened
>
> - Removed ticket reference from inline comment in `QuerySerializer.serialize` test [ff9c180]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f73aeb3.</sup>
> 
>
>

### DIFF
--- a/packages/agent-client/test/query-serializer.test.ts
+++ b/packages/agent-client/test/query-serializer.test.ts
@@ -81,13 +81,23 @@ describe('QuerySerializer', () => {
         expect(result.search).toBe('john');
         expect(result['page[size]']).toBe(15);
         expect(result['page[number]']).toBe(3);
-        expect(result['fields[users]']).toEqual(['id', 'name']);
+        expect(result['fields[users]']).toBe('id,name');
       });
     });
 
     it('should serialize fields with escaped collection name', () => {
       const result = QuerySerializer.serialize({ fields: ['id', 'name', 'email'] }, 'users');
-      expect(result['fields[users]']).toEqual(['id', 'name', 'email']);
+      expect(result['fields[users]']).toBe('id,name,email');
+    });
+
+    it('should serialize fields as a single comma-separated value, not an array', () => {
+      // Regression (PRD-437): an array serializes as repeated params that Ruby agents collapse to one.
+      const result = QuerySerializer.serialize(
+        { fields: ['first_name', 'last_name', 'email'] },
+        'users',
+      );
+      expect(result['fields[users]']).toBe('first_name,last_name,email');
+      expect(Array.isArray(result['fields[users]'])).toBe(false);
     });
 
     it('should serialize ascending sort', () => {
@@ -188,7 +198,7 @@ describe('QuerySerializer', () => {
       expect(result.search).toBe('john');
       expect(result['page[size]']).toBe(20);
       expect(result['page[number]']).toBe(1);
-      expect(result['fields[users]']).toEqual(['id', 'name']);
+      expect(result['fields[users]']).toBe('id,name');
       expect(result.sort).toBe('-createdAt');
       const parsed = JSON.parse(result.filters as string);
       expect(parsed.conditions[0].operator).toBe('equal');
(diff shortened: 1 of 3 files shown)


packages/agent-client/test/query-serializer.test.ts
```diff
@@ -81,13 +81,23 @@ describe('QuerySerializer', () => {
...
     });
 
     it('should serialize fields with escaped collection name', () => {
       const result = QuerySerializer.serialize({ fields: ['id', 'name', 'email'] }, 'users');
-      expect(result['fields[users]']).toEqual(['id', 'name', 'email']);
+      expect(result['fields[users]']).toBe('id,name,email');
+    });
+
+    it('should serialize fields as a single comma-separated value, not an array', () => {
+      // Regression (PRD-437): an array serializes as repeated params that Ruby agents collapse to one.
```
hercemer42: **Claude Opus 4.8** — _[Violates conventions]_ Comments must not reference PRD/Linear ticket numbers, including in tests. Drop the `(PRD-437)` token, e.g.:

```
// Regression: an array serializes as repeated params that Ruby agents collapse to one.
```

The source comment in `query-serializer.ts` correctly avoids the reference; this is the only ticket reference in the changed files.

hercemer42 (changes requested):

packages/agent-client/test/query-serializer.test.ts
christophebrun-forest: Done

hercemer42 (approved):

PR #1623, opened 2026-06-03
fix(workflow-executor): drop dead-branch steps from context (PRD-433)

hercemer42:
fixes PRD-433

## Problem

Revising an earlier workflow step did not discard the steps that originally ran after it. The orchestrator marks them correctly (`revised` on the pivot card, `cancelled` on the downstream branch, still-valid clones re-appended with `originalStepIndex`), but the executor read the run history branch-blind in two places:

1. **`toPreviousSteps`** filtered only on `done && stepIndex < current` → dead-branch entries polluted the AI context for every step type.
2. **The record pool** (`getAvailableRecordRefs`) was built from the raw local RunStore → records loaded by superseded steps were still offered as sources (revising "Load store" proposed loading from the dead branch's *owner* instead of reloading the store), while live clones (new `stepIndex`) couldn't be matched to their stored records.

## Fix (executor-only)

- **`toPreviousSteps`** now mirrors the orchestrator's own live-path filter — `!revised && !cancelled` — so dead-branch steps never reach the AI context or the record pool.
- **`originalStepIndex`** (already on the wire in the orchestrator's history) is declared in `ServerStepHistory` and forwarded onto each mapped `Step`. It is set on a revision clone — a still-valid step the orchestrator re-injects — and points at the step the clone copies, where that step's record lives.
- **`BaseStepExecutor.resolveStepExecution`** resolves a step's stored execution **own-index-first**: a step the executor ran has its execution at its own `stepIndex`; a clone (which never ran) falls back to `originalStepIndex`. Own-index-first is what stops a re-executed step — which has its own entry — from resurfacing the superseded original's record. Both the record pool and the previous-steps AI summary resolve through it.

This mirrors the frontend's `carryForwardExecutorDataForCopiedSteps` (which copies executor data from `originalStepIndex` to a clone's new index): the executor just resolves the record on read instead of copying it.

**Why not `stepName`:** LinkTo loops are a product feature — the same step name can legitimately appear twice on the live path — so the step index is the only safe key.

## Tests

Full package suite green (925/925); the changed behavior is covered:
- cancelled downstream steps excluded from `previousSteps` and the record pool
- a revision clone resolves to its copied step's record via `originalStepIndex`, while a re-executed step uses its own execution — never the superseded original's (the reported bug: a revised load that yields no record shows no stale record downstream)
- entry-point revision collapses the pool to the base record
- double revision resolves the freshest generation; the stale one is absent from the select-record enum
- LinkTo loop: two live instances of the same `stepName` keep their own records

> [!NOTE]
> ### Drop dead-branch steps from workflow executor context on revision
> - `toPreviousSteps` in [run-to-available-step-mapper.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1623/files#diff-d8e6d94d08cd810252fadde0e0f846eca9f05c1b39b880ed4bb1d2357ca64abf) now filters out `revised` and `cancelled` history entries so only live-path steps are surfaced to executors.
> - A new `toLineageStepIndexes` utility computes each step's revision lineage (freshest-first), stored as `lineageStepIndexes` on the `Step` type.
> - `BaseStepExecutor.resolveLineageExecution` in [base-step-executor.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1623/files#diff-f2629617b13363027748d444be5374dd404d26c2f031b0b1a967cceb01d6c6d9) uses `lineageStepIndexes` to find the most recent valid execution for a step when building previous-step summaries.
> - `RecordStepExecutor.getAvailableRecordRefs` in [record-step-executor.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1623/files#diff-8a51c158f5f5c6df99cb9483d4d66d54f51671a37f114fafce10ea8b1e51e29c) now sources related records from live `previousSteps` only, resolved via lineage, instead of scanning all RunStore executions.
> - Behavioral Change: after a workflow revision, dead-branch step outputs (records, summaries) are no longer visible to downstream steps; the record pool collapses to the base record plus live load-related steps only.
>
> 
> #### Changes since #1623 opened
>
> - Replaced `resolveLineageExecution` with `resolveStepExecution` in `BaseStepExecutor.buildPreviousStepsMessages` and `BaseStepExecutor.resolveStepExecution` static utility, and in `RecordStepExecutor.getAvailableRecordRefs`, changing execution resolution from iterating multiple lineage generation candidates to finding by the step's own `stepIndex` first, then falling back to `originalStepIndex` when present, ensuring re-executed steps never resurface superseded originals while clones inherit execution data from their copied step [03e2b08]
> - Removed `lineageStepIndexes` property and added `originalStepIndex` property to `validated.execution.StepSchema` in the `workflow-executor` package, where `originalStepIndex` is an optional number that points to the step a clone copies, with updated documentation describing clone inheritance semantics and warnings against keying on `stepName` [03e2b08]
> - Changed `tryMapStep` signature in `run-to-available-step-mapper` to remove `lineageStepIndexes` parameter and conditionally include `originalStepIndex` in returned `Step` objects when present on server step history, removed `toLineageStepIndexes` function that computed lineage arrays, and updated `toPreviousSteps` to call the modified `tryMapStep` without passing lineage data [03e2b08]
> - Updated test expectations across `run-to-available-step-mapper.test.ts`, `base-step-executor.test.ts`, `load-related-record-step-executor.test.ts`, `read-record-step-executor.test.ts`, `trigger-record-action-step-executor.test.ts`, and `update-record-step-executor.test.ts` to assert presence or absence of `originalStepIndex` instead of `lineageStepIndexes`, verify own-index-first resolution behavior, confirm clones inherit from originals, and ensure re-executed steps prefer their own execution [03e2b08]
> - Updated inline comment in `RecordStepExecutor.getAvailableRecordRefs` method describing load-related step record resolution logic [e17962e]
> - Clarified documentation for `originalStepIndex` field across `ServerStepHistory` interface, `BaseStepExecutor.resolveStepExecution` method, and `StepSchema` constant [76f58fb]
> - Converted `BaseStepExecutor.resolveStepExecution` from a static method to an instance method in the `workflow-executor` package [f2d6780]
> - Modified `BaseStepExecutor.resolveStepExecution` to prevent re-executed clone steps with error status from inheriting execution data from their original step [7979556]
> - Added test coverage in `read-record-step-executor.test` verifying that errored clone steps do not inherit execution data from their original step [7979556]
> - Reworded inline comment in `BaseStepExecutor.resolveStepExecution` method [7273844]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 584f656.</sup>
> 
>
>


packages/workflow-executor/test/executors/load-related-record-step-executor.test.ts
```diff
@@ -1904,4 +1935,51 @@ describe('LoadRelatedRecordStepExecutor', () => {
...
+        previousSteps: [],
+        stepDefinition: makeStep({ executionType: StepExecutionMode.FullyAutomated }),
+      });
+      const executor = new LoadRelatedRecordStepExecutor(context);
+
+      // When
+      const result = await executor.execute();
+
+      // Then: the pool collapsed to the base record — no select-record round; the relation
+      // was selected and loaded directly from the base record (the ticket's bug offered the
```
hercemer42: **[Claude Opus 4.8] [Violates conventions]** This comment references the user story ("the ticket's bug offered…"). Conventions bar referencing tickets/user stories in comments, including tests. Reword to describe the behavior itself, e.g. "…loaded directly from the base record (previously the dead branch's owner was offered as a source instead)". Note: not linter-enforced — generic wording like "the ticket" can't be caught by the existing no-PRD-reference greps; flagging per review instructions.

hercemer42: **[Claude Opus 4.8]** Fixed in b608f35b — reworded to "previously the dead branch's owner was offered as a source instead", no story reference.

hercemer42:
**[Claude Opus 4.8] — Review resolution on #1623**

**1. Inline conventions comment (test comment referencing the ticket)** — agreed and fixed in b608f35b.

**2. qlty: "Function with many returns (count = 4): getAvailableRecordRefs"** — pushing back, not fixing. The four returns are two guard-style early exits inside a `flatMap` callback plus the final pool return; the project's own code review principles explicitly prescribe this shape ("Return early — guard clauses at the top, avoid deep nesting"). Collapsing them into a single-exit ternary would also fight TypeScript's type narrowing on the discriminated `executionResult` union. 20-line function, single responsibility — restructuring would reduce readability to satisfy a counter, with no behavioral benefit. Flagging the conflict between the qlty `many-returns` rule and the project's return-early convention for a maintainer decision on the rule threshold.

**3. CI failure (Linting & Testing workflow-executor)** — root cause was not the review findings: the base branch moved under the PR (#1610 reworked the load-related executor; BelongsTo loads now go through `getSingleRelatedData` instead of `getRelatedData`). CI tests the merge result, so the revision repro test asserted a port call the merged flow no longer makes. Resolved by rebasing onto the current base and adapting the assertion to the sibling tests' shape (b608f35b). Full suite on the rebased branch: 898/898 green.

packages/workflow-executor/src/adapters/server-types.ts
```diff
@@ -169,6 +169,8 @@ export interface ServerStepHistory {
   done: boolean;
   revised?: boolean;
   cancelled?: boolean;
+  // Set on revision clones and re-executions; chains to the FIRST original across revisions.
```
EnkiP: no re-executions, originalStepIndex is used only for copied steps. The revised one is a completely new one that has no originalStepIndex

packages/workflow-executor/src/executors/record-step-executor.ts
```diff
@@ -46,15 +47,22 @@ export default abstract class RecordStepExecutor<
...
   }
 
+  // Candidate sources for the AI: the base record plus the record each live prior
+  // load-related step resolved (via lineage — clones key their execution under the original
+  // stepIndex).
   protected async getAvailableRecordRefs(): Promise<RecordRef[]> {
     const stepExecutions = await this.context.runStore.getStepExecutions(this.context.runId);
-    const relatedRecords = stepExecutions.flatMap(e => {
+    const relatedRecords = this.context.previousSteps.flatMap(step => {
+      if (step.stepDefinition.type !== StepType.LoadRelatedRecord) return [];
```
EnkiP, line 56: we should filter cancelled and revised steps here

packages/workflow-executor/src/types/validated/execution.ts
```diff
@@ -25,6 +25,10 @@ export const StepSchema = z
   .object({
     stepDefinition: StepDefinitionSchema,
     stepOutcome: StepOutcomeSchema,
+    // RunStore execution lookup candidates, own stepIndex first then earlier same-step
+    // generations descending. Revision clones run under a new stepIndex while their execution
+    // data stays keyed under the original one. Absent (legacy producers) means "own index only".
+    lineageStepIndexes: z.array(z.number().int().nonnegative()).nonempty().optional(),
```
EnkiP: to remove: we shouldn't pollute executor data with something used only for computing the real data

hercemer42:
Comment from Alban :

Found bug :

1. load a related dvd
2. display dvd name
3. revise step 1, load fails, next
4. display dvd name => it displays the dvd loaded in step 1, but it was revised.

hercemer42:
Comments from Enki :

I think we should actually copy the data for copied steps. It would make the code simpler for a negligible memory cost - the data you fetch with the resolveLineageExecution method

we should fetch & copy the data at the adapter level, so that executors don’t care about steps that are copied or not

packages/workflow-executor/src/adapters/server-types.ts
hercemer42: I thought the orchestrator set this ?

packages/workflow-executor/src/executors/record-step-executor.ts
hercemer42: There's a filter in the adaptor, it should be enough ?

packages/workflow-executor/src/adapters/server-types.ts
hercemer42: Never mind, I'm going to mirror what the front does so it's not relevant

packages/workflow-executor/src/executors/base-step-executor.ts
```diff
@@ -283,16 +284,34 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
...
       .join('\n\n');
 
     return [new SystemMessage(summary)];
   }
 
+  // A step the executor ran has its execution at its own stepIndex. A revision clone never ran,
+  // so it inherits the record of the step it copied (originalStepIndex). Own-index-first is what
+  // stops a re-executed step (which has its own entry) from resurfacing the superseded
+  // original's record.
+  protected static resolveStepExecution(
```
Scra3: remove static keyword, and use `this.` to call it. RecordStepExecutor extends BaseStepExecutor

Scra3 (approved):

PR #1628, opened 2026-06-04
fix(workflow-executor): operation activity log targets the acted record (PRD-442 #1)

Scra3:
## Context
`fixes PRD-442`

A workflow step's operation activity-log entry (audit trail) pointed at the run's **trigger** record, even when the step acted on a record loaded earlier in the run — possibly in **another collection**. Root cause: the audit args were built *before* `doExecute()` (so the acted record wasn't resolved yet) and used `context.collectionId` + `baseRecordRef`.

## Changes

**Audit moved into a dedicated `AgentWithLog` wrapper.** Activity-log responsibility left `BaseStepExecutor` / `OperationStepExecutor` (deleted) and now lives in `AgentWithLog`, which wraps `AgentPort` and emits **one audit entry per data-access / write call** (`pending → succeeded/failed`). The target is derived from the call itself — `recordId = query.id`, `collectionId` resolved from the call's collection — so each entry points at the **acted** record and its own collection, and fires **only when the operation actually happens**. Covers read / update / trigger-action / load-related; the MCP step uses the generic `logged()` (its side effect is a tool invocation, audited against the run's base record).

**Acted collection's numeric id.** `CollectionSchema` gains a **required** `collectionId`, resolved from the live schema via the new `SchemaCache.getOrLoad(name, loader)` (lazy cache-or-fetch, decoupled from `WorkflowPort`). The audit is not optional, so a missing id fails loud at the boundary (`DomainValidationError`).

**Fail-loud schema resolution.** `AgentClientAgentPort.resolveSchema` now throws `SchemaNotCachedError` instead of returning a synthetic fallback schema (which had a wrong `primaryKeyFields` / `collectionId`). Unreachable in the normal flow — `AgentWithLog` resolves (and caches) the schema right before delegating — so it's a true invariant guard.

**Idempotency unchanged, ordering preserved.** Idempotency stays entirely in the executors. The `executing` write-ahead marker is saved in a `beforeCall` thunk that `AgentWithLog` runs **after `createPending`, before the side effect** — so an audit-creation failure never leaves an orphan `executing` marker, and `AgentWithLog` stays ignorant of run state.

**Activity-log failures** surface as user-facing step errors; `markFailed` carries a **local-only** diagnostic message (the server status endpoint accepts just `{ status }`).

## ⚠️ Coordinated deploy (cross-repo)
The orchestrator must include `collectionId` in **every** `getCollectionSchema` response. Until it does, steps on that collection fail at schema parse — so this must ship in tandem with the orchestrator change.

## Scope
Bug #1 (wrong target), **plus** the confirmation-flow **premature / duplicate** symptom of #2: logging at the operation point means a confirmation step logs **nothing while awaiting input**, **nothing on reject**, and **exactly once on confirm** (covered by tests). Remaining #2 work: exactly-once across network retries (orchestrator-side dedup of identical `(runId, stepIndex)` outcomes).

## Verification
- `yarn workspace @forestadmin/workflow-executor build` ✅
- **922 tests** — incl. a cross-collection repro (trigger on `customers`, act on `orders#99` → audit targets `orders`, not the trigger), per-executor target assertions, `AgentWithLog` unit tests (beforeCall ordering, createPending-throws, `userMessage` vs `'Unexpected error'`), reject-emits-no-log, and `getOrLoad` hit/miss ✅
- lint: 0 errors ✅

> [!NOTE]
> ### Target the acted record in operation activity logs for workflow executor steps
> - Introduces `AgentWithLog` ([agent-with-log.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1628/files#diff-53c86f1b5b0728996228d51fd536d05b4217c4ce42696c5c99b29b6a18f1145b)), a wrapper around `AgentPort` that emits per-operation activity log entries (pending → succeeded/failed) with `collectionId` resolved from the collection schema and `recordId` from the query target.
> - Removes the step-level `buildActivityLogArgs()`/`runWithActivityLog()` hooks from `BaseStepExecutor`; activity logging now happens at the individual data-access call level inside each executor.
> - Adds `SchemaResolver` ([schema-resolver.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1628/files#diff-4d8636e9b6f7439f7dcfb6bf618eea0b8a3883021a40642eb08c605ffa6c380e)) to lazily fetch and cache collection schemas per run, and extends `CollectionSchemaSchema` to require a non-empty `collectionId` field.
> - `StepExecutorFactory.buildContext` now constructs both `SchemaResolver` and `AgentWithLog` and injects them into the execution context, replacing the `schemaCache` and direct `agentPort` access previously used by executors.
> - `markFailed` on `ActivityLogPort` no longer accepts an `errorMessage` parameter; write operations pass a `beforeCall` thunk to persist idempotency state immediately before the side effect.
> - Risk: `collectionId` is now a required field on collection schemas — any schema missing it will fail Zod validation.
>
> 
> #### Changes since #1628 opened
>
> - Removed `agentPort`, `workflowPort`, and `activityLogPort` properties from `ExecutionContext` interface and removed the cached `agent` field from `BaseStepExecutor` class [a7f536a]
> - Updated all step executor implementations to access agent through `this.context.agent` instead of `this.agent` [a7f536a]
> - Updated test helpers and test cases to remove port properties from mock contexts and accept port overrides as separate parameters [a7f536a]
> - Documented that `AgentWithLog.getActionFormInfo` is an unaudited passthrough method for read-only form information [6efee0f]
> - Changed `AgentWithLog` class methods to pass the complete `WriteOptions` object through the call chain to the `audit` method instead of extracting and passing only the `beforeCall` callback function [e4e4186]
> - Introduced `ActivityLogger` class to encapsulate activity-log emission around async operations [37f7048]
> - Refactored `AgentWithLog` class to delegate activity-log management to injected `ActivityLogger` [37f7048]
> - Updated `ExecutionContext` interface to include `activityLogger` property and modified `StepExecutorFactory.buildContext` to instantiate and inject `ActivityLogger` [37f7048]
> - Replaced `AgentWithLog.logged` calls with `activityLogger.run` in `McpStepExecutor.executeToolAndPersist` method [37f7048]
> - Updated all test suite context helpers to construct `ActivityLogger` instances and pass them to `AgentWithLog` instead of `ActivityLogPort` [37f7048]
> - Added comprehensive test coverage for `ActivityLogger` class and updated `AgentWithLog` test suite to use `ActivityLogger` [37f7048]
> - Renamed `ActivityLogger` to `ActivityLog` and refactored its primary method from `run(target, operation, opts?)` to `track(target, { operation, beforeCall? })` in the `workflow-executor` package [e41c488]
> - Updated `AgentWithLog`, `McpStepExecutor`, `StepExecutorFactory`, and the `ExecutionContext` interface to use `ActivityLog` instead of `ActivityLogger` and invoke `track` with an options object [e41c488]
> - Updated all test files in the `workflow-executor` package to use `ActivityLog` instead of `ActivityLogger` and updated test invocations to call `track` with an options object [e41c488]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ec949b6.</sup>
> 
>
>


packages/workflow-executor/src/adapters/agent-client-agent-port.ts
```diff
@@ -301,6 +301,7 @@ export default class AgentClientAgentPort implements AgentPort {
     return (
       cached ?? {
         collectionName,
+        collectionId: collectionName,
```
Scra3: not works

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
```diff
@@ -291,21 +292,9 @@ export default class AgentClientAgentPort implements AgentPort {
     const cached = this.schemaCache.get(collectionName);
 
     if (!cached) {
-      // eslint-disable-next-line no-console
```
Scra3: should never occur because schema is already loaded at this point

packages/workflow-executor/src/schema-cache.ts
```diff
@@ -30,6 +30,21 @@ export default class SchemaCache {
     this.store.set(collectionName, { schema, fetchedAt: this.now() });
   }
 
+  // Returns the cached schema, or loads it via `load`, caches it, and returns it.
+  // `load` is injected so the cache stays decoupled from the orchestrator port.
+  async getOrLoad(
+    collectionName: string,
+    load: () => Promise<CollectionSchema>,
```
nbouliol: why do we need to accept a load arg ? is it not always the same logic for getting the schema ?

This means you could access a cached value that was not loaded from the same loader as the one you provide

packages/workflow-executor/src/executors/agent-with-log.ts
```diff
@@ -0,0 +1,141 @@
...
+    try {
+      if (beforeCall) await beforeCall();
+      const result = await run();
+      void this.activityLogPort.markSucceeded(handle);
+
+      return result;
+    } catch (err) {
+      const errorMessage =
+        err instanceof WorkflowExecutorError ? err.userMessage : 'Unexpected error';
+      void this.activityLogPort.markFailed(handle, errorMessage);
```
nbouliol: If beforeCall throws, we markFailed though the side effect never ran — the audit then shows a failed write on a record that was never touched. Intended, or should we only markFailed around run()?

packages/workflow-executor/src/executors/agent-with-log.ts
```diff
@@ -0,0 +1,141 @@
...
+      throw err;
+    }
+  }
+
+  private async resolveCollectionId(collectionName: string): Promise<string> {
+    const schema = await this.schemaCache.getOrLoad(collectionName, () =>
+      this.workflowPort.getCollectionSchema(collectionName, this.runId),
```
nbouliol: Uncovered (qlty): a test with a real SchemaCache miss would exercise the workflowPort loader path here.

packages/workflow-executor/src/schema-cache.ts
Scra3: I introduce a schemaResolver to avoid to pass this ugly getOrLoad

nbouliol (commented):
Code LGTM, needs manual testing

packages/workflow-executor/src/executors/step-executor-factory.ts
```diff
@@ -138,7 +138,6 @@ export default class StepExecutorFactory {
       agentPort: cfg.agentPort,
       workflowPort: cfg.workflowPort,
       runStore: cfg.runStore,
-      schemaCache: cfg.schemaCache,
       schemaResolver: new SchemaResolver(cfg.schemaCache, cfg.workflowPort, step.runId),
```
Scra3: schemaCache is only used by schema resolver. No need to inject schema cache

packages/workflow-executor/src/executors/base-step-executor.ts
```diff
@@ -35,11 +33,8 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
 {
   protected readonly context: ExecutionContext<TStep>;
 
-  protected readonly agentPort: AgentPort;
-
   constructor(context: ExecutionContext<TStep>) {
     this.context = context;
-    this.agentPort = context.agentPort;
```
Scra3, line 42: useless

packages/workflow-executor/src/types/execution-context.ts
```diff
@@ -32,15 +30,13 @@ export interface ExecutionContext<TStep extends StepDefinition = StepDefinition>
   readonly baseRecordRef: RecordRef;
   readonly stepDefinition: TStep;
   readonly model: BaseChatModel;
-  readonly agentPort: AgentPort;
```
Scra3, line 33: unused

packages/workflow-executor/src/types/execution-context.ts
```diff
@@ -32,15 +30,13 @@ export interface ExecutionContext<TStep extends StepDefinition = StepDefinition>
...
-  readonly workflowPort: WorkflowPort;
```
Scra3, line 34: unused

packages/workflow-executor/src/types/execution-context.ts
```diff
@@ -32,15 +30,13 @@ export interface ExecutionContext<TStep extends StepDefinition = StepDefinition>
...
   readonly runStore: RunStore;
   readonly user: StepUser;
-  readonly schemaCache: SchemaCache;
+  readonly schemaResolver: SchemaResolver;
   readonly previousSteps: ReadonlyArray<Readonly<Step>>;
   readonly logger: Logger;
   readonly incomingPendingData?: unknown;
   readonly stepTimeoutMs?: number;
   readonly aiInvokeTimeoutMs?: number;
-  readonly activityLogPort: ActivityLogPort;
```
Scra3, line 43: unused

nbouliol (approved):

PR #1631, opened 2026-06-05
feat(workflow-executor): align datasource activity-log labels with the browser engine [PRD-449]

nbouliol:
## What

Aligns the datasource activity-log labels emitted by the workflow-executor with the legacy browser engine (ISO), fixing a production audit-trail crash on label-less entries.

fixes PRD-449

## Context

Since the activity-log refactor (PRD-442), logs are emitted per datasource call in `AgentWithLog.audit()`, not per step. This dissolves the original dynamic-label problem (the name is now known at the call site, after AI resolution) — no executor or base-class refactor needed. The fix is just injecting the ISO `label` into each audit target.

## Changes (`agent-with-log.ts`)

| Operation | Label | Source |
|---|---|---|
| `update` | `updated` | static |
| `action` | `triggered the action "<technical name>"` | `query.action` |
| `listRelatedData` | `list relation "<displayName>"` | resolved from source schema, falls back to technical name |
| `index` (read) | none | ISO; `getRecord` is shared across step types, a blanket label would mislabel update/trigger/load reads |

- `resolveCollectionId` -> `resolveSchema` (returns the full `CollectionSchema`) so the relation displayName can be looked up.
- `action` label only fires in FullyAutomated mode (preserved naturally: `executeAction` is only called in that branch; otherwise the front executes and logs).

## Tests

- `agent-with-log.test.ts`: exact-label assertions for update / action / relation displayName + technical-name fallback.
- Updated existing `createPending` assertions in update / trigger / load-related executor tests.

## Validation

- `yarn workspace @forestadmin/workflow-executor test` -> 925/925
- `yarn workspace @forestadmin/workflow-executor lint` -> 0 errors
- `yarn workspace @forestadmin/workflow-executor build` -> ok

## Out of scope

- MCP executor (already labelled)
- The N-logs-per-step behavior of load-related (pre-existing since PRD-442)
- Rich `\"...in workflow Y\"` labels (requires orchestrator contract change)

> [!NOTE]
> ### Add labels to datasource activity-log entries in `AgentWithLog`
> - `getRelatedData` and `getSingleRelatedData` now include a `label` of `'list relation "<displayName>"'` (falling back to the technical relation name) in their activity log payloads.
> - `updateRecord` adds a static `'updated'` label; `executeAction` adds `'triggered the action "<action>"'` using the technical action name.
> - A new `resolveSchema` helper replaces the removed `resolveCollectionId` helper, returning the full `CollectionSchema` so call sites can access both the `collectionId` and field metadata needed for label computation.
>
> #### 🖇️ Linked Issues
> Relates to [PRD-449](ticket:linear/PRD-449).
>
> 
> #### Changes since #1631 opened
>
> - Modified `AgentWithLog.relationLabel` method in `workflow-executor` package to filter schema fields by relationship status before resolving display names [4dad2f2]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c7d7642.</sup>
> 
>
>

### DIFF
--- a/packages/workflow-executor/src/executors/agent-with-log.ts
+++ b/packages/workflow-executor/src/executors/agent-with-log.ts
@@ -10,7 +10,7 @@ import type {
 } from '../ports/agent-port';
 import type SchemaResolver from '../schema-resolver';
 import type { StepUser } from '../types/execution-context';
-import type { RecordData } from '../types/validated/collection';
+import type { CollectionSchema, RecordData } from '../types/validated/collection';
 
 type WriteOptions = { beforeCall: () => Promise<void> };
 
@@ -42,7 +42,7 @@ export default class AgentWithLog {
   }
 
   async getRecord(query: GetRecordQuery): Promise<RecordData> {
-    const collectionId = await this.resolveCollectionId(query.collection);
+    const { collectionId } = await this.resolveSchema(query.collection);
 
     return this.activityLog.track(
       { action: 'index', type: 'read', collectionId, recordId: query.id },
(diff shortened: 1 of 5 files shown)


packages/workflow-executor/src/executors/agent-with-log.ts
```diff
@@ -98,9 +116,16 @@ export default class AgentWithLog {
     return this.agentPort.getActionFormInfo(query, this.user);
   }
 
-  private async resolveCollectionId(collectionName: string): Promise<string> {
-    const schema = await this.schemaResolver.resolve(collectionName);
+  // ISO with the browser engine: `list relation "<displayName>"`. The query carries the technical
+  // relation name; resolve its displayName from the source schema, falling back to the technical
+  // name when the field is absent (resilient to orchestrator schema drift).
+  private relationLabel(schema: CollectionSchema, relation: string): string {
+    const displayName = schema.fields.find(f => f.fieldName === relation)?.displayName ?? relation;
```
Scra3: can we filter if the field is a relation before to get the name ? Maybe it's useless

Scra3 (approved):

PR #1634, opened 2026-06-08
fix(workflow-executor): reject duplicate triggers of in-flight runs (PRD-468)

Scra3:
## What

A duplicate trigger for a run already executing **on this instance** was silently accepted (`200`) after a wasted orchestrator round-trip — the `inFlightRuns` check sat *after* `getAvailableRun` and only `return`ed.

`triggerPoll` now checks `inFlightRuns` **up front** (`assertRunNotInFlight`), before the orchestrator round-trip, and throws the boundary `RunAlreadyInFlightError` → `handleTrigger` maps it to `400` so the front knows the trigger was rejected, not silently accepted.

## Important: this is an optimization, not a concurrency guard

`inFlightRuns` is **per-instance** and a deployment can run **several executors**, so an executor-side check can never be the duplicate-execution guard. Genuine concurrency dedup is the **orchestrator's** job: it atomically claims a `pending` run (`UPDATE … WHERE runState='pending' … FOR UPDATE SKIP LOCKED`), so concurrent triggers — on this executor or another — get nothing back.

The up-front check is purely a best-effort local short-circuit to skip a useless orchestrator round-trip when *this* instance already knows it's running the run.

## Changes
- `assertRunNotInFlight(runId)` early in `triggerPoll` (+ explanatory comment on the intent)
- new boundary `RunAlreadyInFlightError`, mapped to `400` in `handleTrigger`
- tests: the concurrent-trigger test asserts a single execution **because the orchestrator denies the second dispatch** (the dedup is orchestrator-driven by design)

fixes PRD-468

> [!NOTE]
> ### Reject duplicate triggers of in-flight workflow runs with a 400 error
> - `Runner.triggerPoll` previously silently skipped duplicate triggers; it now throws `RunAlreadyInFlightError` via a new `assertRunNotInFlight` helper when the run ID is already being processed.
> - The `POST /runs/:runId/trigger` endpoint in [`executor-http-server.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1634/files#diff-f24eaa83e4d72cc710e9cb4cea88a2047543ee3c17a0b68627f6af587546b614) catches `RunAlreadyInFlightError` and returns a 400 response with the error message.
> - Behavioral Change: callers that previously received a silent no-op on duplicate trigger requests will now receive a 400 error response.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1e0f685.</sup>
> 
>

### DIFF
--- a/packages/workflow-executor/src/errors.ts
+++ b/packages/workflow-executor/src/errors.ts
@@ -328,6 +328,13 @@ export class UserMismatchError extends Error {
   }
 }
 
+export class RunAlreadyInFlightError extends Error {
+  constructor(runId: string) {
+    super(`Run "${runId}" is already being processed`);
+    this.name = 'RunAlreadyInFlightError';
+  }
+}
+
 export class PendingDataNotFoundError extends Error {
   constructor(runId: string, stepIndex: number) {
     super(`Step ${stepIndex} in run "${runId}" not found or has no pending data`);
--- a/packages/workflow-executor/src/http/executor-http-server.ts
+++ b/packages/workflow-executor/src/http/executor-http-server.ts
@@ -13,6 +13,7 @@ import koaJwt from 'koa-jwt';
 import serializeStepForWire from './step-serializer';
 import ConsoleLogger from '../adapters/console-logger';
 import {
+  RunAlreadyInFlightError,
   RunNotFoundError,
   UserMismatchError,
   WorkflowExecutorError,
@@ -187,6 +188,13 @@ export default class ExecutorHttpServer {
       if (err instanceof RunNotFoundError) {
         ctx.status = 404;
         ctx.body = { error: 'Run not found or unavailable' };
+
+        return;
+      }
+
+      if (err instanceof RunAlreadyInFlightError) {
+        ctx.status = 400;
+        ctx.body = { error: err.message };
 
         return;
       }
--- a/packages/workflow-executor/src/runner.ts
+++ b/packages/workflow-executor/src/runner.ts
@@ -14,6 +14,7 @@ import ConsoleLogger from './adapters/console-logger';
 import { DEFAULT_MAX_CHAIN_DEPTH, DEFAULT_STOP_TIMEOUT_MS } from './defaults';
 import {
   MalformedRunError,
+  RunAlreadyInFlightError,
   RunNotFoundError,
   UserMismatchError,
   causeMessage,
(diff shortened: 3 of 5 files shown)


matthv:
Nice work — the boundary error + the `assertRunNotInFlight` helper + the `400` mapping are clean, and the early short-circuit (no orchestrator round-trip on a re-trigger) is a real improvement. One thing I'd like to settle before merging though, around the concurrency guarantee.

**Description ↔ code mismatch**

The description says the in-flight check is done **twice** (early + late) and that *"the late check is the correctness guard … removing it in favor of the early one alone would reintroduce double execution"*. But the diff actually **removes** the late check (`if (inFlightRuns.has(step.runId)) return`) and keeps only the early `assertRunNotInFlight(runId)`. Macroscope's summary also describes a single check. So either the description is stale, or the late check was dropped unintentionally.

**Why this matters — `getAvailableRun` is only atomic for `pending` runs**

I checked the orchestrator side. `GET /api/workflow-orchestrator/available-run/:id` only claims atomically when the run is `pending`:

```
if (runState === Pending)  → claimPendingRunById()   // atomic, see below
if (runState === Started)  → no claim, just hydrate + return the run
```

`claimPendingRunById` is solid for the `pending` case:
```sql
UPDATE "workflowRuns" SET "lockedAt"=NOW(), "runState"='loading'
WHERE id IN ( SELECT wr.id ... WHERE wr."runState"='pending' AND wr.id=:runId ...
              FOR UPDATE SKIP LOCKED )
RETURNING *
```
→ two concurrent calls: one claims, the other gets `null`. 👍

But for a **`started`** run there's **no claim** — two concurrent `getAvailableRun` both return the step. So:

- **pending** run → orchestrator dedups → early-check-only is enough ✅
- **started** run + *truly simultaneous* triggers → both pass the early check (before either registers in `inFlightRuns`) **and** the orchestrator returns the step to both → **double execution**. This is exactly the case the removed late check covered.

**The test gives false confidence here**: `getAvailableRun.mockResolvedValueOnce(step).mockResolvedValueOnce(null)` reproduces the *pending-claim* behavior. A `started` run would resolve `step` **both** times, so the current test wouldn't catch the regression.

**Suggested resolution (either):**
1. **Restore the late check** (cheap, closes the `started` race), and reconcile the description; or
2. **Prove the trigger path never hits a `started` run concurrently** (always goes through `pending`) — then early-check-only is correct, but please fix the description and add a test that exercises the `started` case (`getAvailableRun` returning the step twice).

Minor (non-blocking): `400` vs `409` — "already being processed" reads more like a `409 Conflict`, though I see the ticket explicitly asked for `400`, so up to you.

Scra3:
Thanks for the deep read — you're right that the commit message was stale (it described a late check that the diff removed), I've fixed the description + commit message.

On the substance: I'm **intentionally not restoring the late check**. `inFlightRuns` is per-instance and we run **multiple executors**, so an executor-side check can't be the concurrency guard for the `started`-run race you describe — it would only close it same-instance while giving false confidence about the cross-executor case. Genuine duplicate-execution prevention has to live in the **orchestrator's atomic claim**, which is the real guarantee.

So the up-front `assertRunNotInFlight` is scoped as what it is: a best-effort local short-circuit to avoid a useless orchestrator round-trip when this instance already knows it's running the run. I added a comment to that effect, and the concurrent test now reads as "one execution because the orchestrator denies the second dispatch" (dedup is orchestrator-driven).

Re 400 vs 409: agreed 409 reads better, keeping 400 per the ticket.

matthv (approved):

PR #1640, opened 2026-06-09
feat(workflow-executor): handle polymorphic relations in load-related (skip + follow) [PRD-493]

matthv:
## What

Makes the *load-related-record* step handle polymorphic relations end-to-end:

1. **Skip unfollowable ones** — a polymorphic relation the orchestrator can't resolve is never offered nor picked (no crash on `getCollectionSchema("<association>")`).
2. **Follow resolvable ones** — a polymorphic relation exposed with a discriminator + candidate models (`polymorphicReferencedModels`) is followable; the target collection is resolved **per record**.

## How the target is resolved (no UI-exposed field)

The discriminator column (e.g. `imageable_type`) must **not** be a Forest field (it would appear in the UI), so it isn't projectable. The target type is read from the **raw JSON:API relationship linkage** instead — `data.relationships.<relation>.data.type` — which the agent-client deserializer drops:

- `AgentPort.resolvePolymorphicType`: raw fetch of the `<relation>@@@id` projection, returns the linkage `{ type, id }` (reuses the run's JWT → passes the agent's permission layer).
- `resolveTargetCollection` maps that type to one of `polymorphicReferencedModels` (centralised; used by `buildTarget` and the Branch A re-entry path).
- `isFollowableRelation` = has `relatedCollectionName` **or** `polymorphicTypeField`.
- Clean errors when the linkage has no target or the type isn't among the candidates.

## Tests

74/74 in `load-related-record-step-executor.test.ts`, incl. per-record resolution into the right collection, unknown discriminator, missing target. Build + lint clean.

## Validated end-to-end (forest_liana v1)

Same `imageable` relation, target resolved per record:
- Image #4 → **User** #2
- Image #2 → **Product** #53

No UI-exposed discriminator field.

## Dependencies (cross-repo)

- **Orchestrator** exposes `polymorphicTypeField` + `polymorphicReferencedModels` (forestadmin-server #8286).
- **Agent** marks the relation polymorphic via `foreign_key_type_field` (forest-rails `feat/polymorphic-foreign-key-type-field`).

fixes PRD-493

### DIFF
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -5,6 +5,7 @@ import type {
   GetRecordQuery,
   GetRelatedDataQuery,
   GetSingleRelatedDataQuery,
+  ResolvePolymorphicTypeQuery,
   UpdateRecordQuery,
 } from '../ports/agent-port';
 import type SchemaCache from '../schema-cache';
@@ -211,9 +212,9 @@ export default class AgentClientAgentPort implements AgentPort {
     }
   }
 
-  private createClient(user: StepUser) {
+  private mintToken(user: StepUser): string {
     // snake_case aliases: Ruby/Python agents splat JWT claims into Caller.new (snake_case kwargs).
-    const token = jsonwebtoken.sign(
+    return jsonwebtoken.sign(
       {
         ...user,
         first_name: user.firstName,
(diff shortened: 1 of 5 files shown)


matthv:
Superseded by #1647 — rebases this work onto main (post feat/prd-214 merge + RelationCandidate refactor), with additional tests. Validated locally (executor boots, 141 tests green). Closed in favour of #1647.

PR #1643, opened 2026-06-10
fix(workflow-executor): align @forestadmin/* deps with the workspace

hercemer42:
## Summary

`@forestadmin/workflow-executor:build` is currently failing in CI (and on `main`) with a wall of `TS2305`/`TS2339`/`TS2614` errors:

```
'@forestadmin/ai-proxy' has no exported member 'AiClient' / 'BaseChatModel' / 'BaseMessage' / 'DynamicStructuredTool' / …
'@forestadmin/forestadmin-client' has no exported member 'ServerUtils'
Property 'getOne' does not exist on type 'Collection'   (@forestadmin/agent-client)
```

Root cause: `workflow-executor` pins its `@forestadmin/*` deps to **stale exact versions** that lag the workspace and don't export the APIs its source uses. Every other package in the monorepo references the current workspace versions and builds fine — `workflow-executor` is the lone laggard. (It's a chicken-and-egg: semantic-release bumps these on release, but the broken build blocks the release that would fix them.)

## Change

Bump the three pinned deps to the current workspace versions:

| dep | from | to |
|---|---|---|
| `@forestadmin/agent-client` | 1.5.10 | 1.6.0 |
| `@forestadmin/ai-proxy` | 1.10.1 | 1.11.0 |
| `@forestadmin/forestadmin-client` | 1.39.9 | 1.40.0 |

(plus the corresponding `yarn.lock` update).

## Also: exclude `claude-fable-5` from tool-supporting models (ai-proxy)

Fixing the build let the full test matrix run for the first time in a while, which unmasked a pre-existing failure in the required `LLM Integration Tests (ai-proxy)` job: Anthropic's new `claude-fable-5` fails the tool-support sweep. It has **always-on thinking** — it rejects the `thinking: {type: "disabled"}` param that LangChain sends by default, and it requires thinking blocks to be replayed verbatim on multi-turn tool calls, which the proxy's OpenAI-format message conversion drops. Adding it to `ANTHROPIC_UNSUPPORTED_MODELS` gives users a clear `AIModelNotSupportedError` upfront instead of provider 400s (same pattern as the o1/o3/o4 exclusions on the OpenAI side). Proper Fable support would need thinking-block passthrough in the wire format — a separate feature.

This has to land here rather than in its own PR because the integration job can't go green on any branch until this PR's build fix is on it.

## Verification

`lerna run build --scope @forestadmin/workflow-executor --include-dependencies` now succeeds (`tsc` clean) where it previously failed. ai-proxy unit tests (404) and lint pass; the previously failing `datasource-mongo` and OpenAI-503 failures cleared on re-run (flaky/transient).

This is independent of #1641 (PRD-496) — it unblocks the `workflow-executor` build on `main` for everyone.

> [!NOTE]
> ### Align `@forestadmin/*` dependencies in `workflow-executor` with workspace versions
> Bumps three direct dependencies in [package.json](https://github.com/ForestAdmin/agent-nodejs/pull/1643/files#diff-39959c5d53211b31724a713f0f4021345ce98a0dd752aadcde43c1729bdff151) to match the rest of the workspace: `@forestadmin/agent-client` to 1.6.0, `@forestadmin/ai-proxy` to 1.11.0, and `@forestadmin/forestadmin-client` to 1.40.0. The lockfile is updated accordingly.
>
> 
> #### Changes since #1643 opened
>
> - Added 'claude-fable-5' to the `ANTHROPIC_UNSUPPORTED_MODELS` constant in the `ai-proxy` package with documentation explaining the incompatibility is due to always-on thinking requirements and mandatory verbatim replay of thinking blocks across tool calls, and added test cases verifying that `isModelSupportingTools` correctly identifies 'claude-fable-5' with provider 'anthropic' as unsupported while confirming 'claude-opus-4-8' remains supported [655833a]
> - Reformatted comment documentation for the `ANTHROPIC_UNSUPPORTED_MODELS` constant [2e56cd2]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 13493e0.</sup>
> 
>
>

### DIFF
--- a/packages/ai-proxy/src/supported-models.ts
+++ b/packages/ai-proxy/src/supported-models.ts
@@ -77,6 +77,10 @@ const ANTHROPIC_UNSUPPORTED_MODELS = [
   'claude-3-7-sonnet-20250219', // EOL 2026-02-19
   'claude-opus-4-20250514', // Requires streaming (non-streaming times out)
   'claude-opus-4-1-20250805', // Requires streaming (non-streaming times out)
+  // Always-on thinking: rejects the `thinking: {type: 'disabled'}` LangChain sends by default,
+  // and requires thinking blocks to be replayed verbatim on multi-turn tool calls — our
+  // OpenAI-format message conversion drops them, so follow-up turns would be rejected too.
+  'claude-fable-5',
 ];
 
 function isAnthropicModelSupported(model: string): boolean {
--- a/packages/ai-proxy/src/supported-models.ts
+++ b/packages/ai-proxy/src/supported-models.ts
@@ -77,10 +77,7 @@ const ANTHROPIC_UNSUPPORTED_MODELS = [
   'claude-3-7-sonnet-20250219', // EOL 2026-02-19
   'claude-opus-4-20250514', // Requires streaming (non-streaming times out)
   'claude-opus-4-1-20250805', // Requires streaming (non-streaming times out)
-  // Always-on thinking: rejects the `thinking: {type: 'disabled'}` LangChain sends by default,
-  // and requires thinking blocks to be replayed verbatim on multi-turn tool calls — our
-  // OpenAI-format message conversion drops them, so follow-up turns would be rejected too.
-  'claude-fable-5',
+  'claude-fable-5', // Always-on thinking (rejects 'disabled'; proxy drops required thinking blocks)
 ];
 
 function isAnthropicModelSupported(model: string): boolean {
--- a/packages/ai-proxy/test/supported-models.test.ts
+++ b/packages/ai-proxy/test/supported-models.test.ts
@@ -12,4 +12,12 @@ describe('isModelSupportingTools', () => {
   it('should return false for a blacklisted model', () => {
     expect(isModelSupportingTools('gpt-4')).toBe(false);
   });
+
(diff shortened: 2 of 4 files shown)


Scra3 (approved):

packages/ai-proxy/src/supported-models.ts
```diff
@@ -77,6 +77,10 @@ const ANTHROPIC_UNSUPPORTED_MODELS = [
   'claude-3-7-sonnet-20250219', // EOL 2026-02-19
   'claude-opus-4-20250514', // Requires streaming (non-streaming times out)
   'claude-opus-4-1-20250805', // Requires streaming (non-streaming times out)
+  // Always-on thinking: rejects the `thinking: {type: 'disabled'}` LangChain sends by default,
+  // and requires thinking blocks to be replayed verbatim on multi-turn tool calls — our
+  // OpenAI-format message conversion drops them, so follow-up turns would be rejected too.
```
hercemer42: Shorten the comment

hercemer42: Done in 2e56cd21 — collapsed to a single trailing comment.

PR #1647, opened 2026-06-10
feat(workflow-executor): handle polymorphic relations in load-related (skip + follow) [PRD-493]

christophebrun-forest:
## What

Handle polymorphic `BelongsTo` relations (e.g. Rails `commentable` → `Author` / `Library` / …) in the *load-related-record* step. These relations carry no single static target, so they were previously unfollowable. The step can now both **skip** them when unresolvable and **follow** them by resolving the concrete target per record.

Rebased onto `main` — supersedes #1640, which targeted the now-merged `feat/prd-214-server-step-mapper` and predated `main`'s `RelationCandidate` refactor of the relation-selection model.

## Why

A polymorphic relation has no `relatedCollectionName` (the orchestrator omits it). Two things were needed:

- **Skip**: never offer an unfollowable relation to the AI or in the awaiting-input card.
- **Follow**: for a relation that *is* resolvable, pick the concrete target collection per record.

The discriminator column (e.g. `commentable_type`) is **not** exposed as a Forest field, so it can't be read via a projection. The target type is instead read from the **raw JSON:API linkage** (`data.relationships.<rel>.data = { type, id }`), which the agent-client deserializer drops — then mapped to one of `polymorphicReferencedModels` (exact, then case-insensitive, which aligns with Rails model-name = Forest collection-name).

## Changes

- `collection.ts` — schema: optional `polymorphicTypeField` + `polymorphicReferencedModels`.
- `agent-port.ts` / `agent-client-agent-port.ts` — new `resolvePolymorphicType` (raw linkage fetch); `mintToken` extracted for reuse.
- `agent-with-log.ts` — unaudited passthrough (metadata probe; the actual load is audited separately).
- `load-related-record-step-executor.ts` — `isFollowableRelation` + `resolveTargetCollection` (static + polymorphic), wired into `buildRelationCandidates` (resolves per record), `buildTarget`, and the Branch A re-entry path.

## Tests

- Adapter: linkage read, composite id, no-linkage → null, non-2xx → error.
- `AgentWithLog`: unaudited passthrough.
- Executor: follow resolved target, case-insensitive mapping, no linkage → error, discriminator not among candidates → error, unfollowable relation not offered.

998 tests / 45 suites passing, lint clean.

## Dependency

Requires the orchestrator exposing `polymorphicTypeField` / `polymorphicReferencedModels` (forestadmin-server PRD-493).

fixes PRD-493

> [!NOTE]
> ### Handle polymorphic relations in load-related step executor skip and follow logic
> - Adds `resolvePolymorphicType` to the `AgentPort` interface and implements it in `AgentClientAgentPort` by fetching JSON:API linkage for a relation from the agent's by-id route, returning `{ type, id }` or null.
> - Updates `LoadRelatedRecordStepExecutor` to resolve the concrete target collection for polymorphic relations per record using the new method, mapping the discriminator type to one of `polymorphicReferencedModels` case-insensitively.
> - Filters available relations in the UI prompt to only those with a static or resolvable polymorphic target; skips records with no linkage and errors when a discriminator doesn't match any candidate model.
> - Extends the `FieldSchemaSchema` zod schema with optional `polymorphicTypeField` and `polymorphicReferencedModels` fields to carry polymorphic relation metadata.
> - Behavioral Change: the step now throws a `StepStateError` instead of proceeding when a relation target cannot be determined.
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 8cd54ab. 5 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -5,6 +5,7 @@ import type {
   GetRecordQuery,
   GetRelatedDataQuery,
   GetSingleRelatedDataQuery,
+  ResolvePolymorphicTypeQuery,
   UpdateRecordQuery,
 } from '../ports/agent-port';
 import type SchemaCache from '../schema-cache';
@@ -222,9 +223,9 @@ export default class AgentClientAgentPort implements AgentPort {
     }
   }
 
-  private createClient(user: StepUser) {
+  private mintToken(user: StepUser): string {
     // snake_case aliases: Ruby/Python agents splat JWT claims into Caller.new (snake_case kwargs).
-    const token = jsonwebtoken.sign(
+    return jsonwebtoken.sign(
       {
         ...user,
         first_name: user.firstName,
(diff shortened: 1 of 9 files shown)


matthv (approved):
ok for me

PR #1659, opened 2026-06-11
feat(workflow-executor): validate bearer claims via a zod middleware [PRD-508]

Scra3:
## What

Validates the bearer JWT payload **as early as possible** via a zod middleware, instead of the ad-hoc `Number.isFinite` check in `handleTrigger`. Follow-up to PRD-477.

`koa-jwt` only validates a token's **signature + expiry** — never the **payload shape**. So `ctx.state.user.id` was not guaranteed; `hasRunAccessMiddleware` even did `as StepUser` without validating (latent gap).

## Changes

- **`src/http/bearer-claims.ts`** — `BearerClaimsSchema = z.object({ id: z.number() })`. **Non-strict on purpose**: `jsonwebtoken` adds `iat`/`exp` to the decoded payload and Forest may send extra claims — a strict schema would reject every real token. Only `id` is consumed downstream (`handleTrigger` + `hasRunAccess` → `?userId=`); `StepUserSchema` (9 required fields) is a different contract and isn't reusable.
- **Middleware right after `koa-jwt`** — validates `ctx.state.user` once; on failure throws `UnauthorizedHttpError`. Both routes now receive a user with a guaranteed numeric `id`.
- **`handleTrigger`** — drops the `Number.isFinite` branch.

## ⚠️ Behaviour change

A well-signed token without a valid numeric `id` now returns **401 Unauthorized** (invalid credentials, consistent with koa-jwt's own 401s) instead of 400.

## Out of scope (decided)

`pendingData` is **not** boundary-validatable — its schema is step-type-dependent (`patchBodySchemas[execution.type]`), known only after `getAvailableRun`. It stays validated downstream in `base-step-executor`. A boundary `z.union`/`z.discriminatedUnion` was considered and rejected (accepts the wrong step type; `.transform()` footgun; no discriminator in the front-sent payload).

## Tests

- `test/http/bearer-claims.test.ts`: accepts `{ id }` and `{ id, iat, exp, … }` (the non-strict trap), rejects missing/non-numeric id.
- `test/http/executor-http-server.test.ts`: invalid claims → 401; a token with extra claims → 200 (end-to-end non-strict proof).
- Full suite: 1025 passing.

## Note

Stacked on `feature/prd-477-…` (base of this PR) — it builds on the typed HTTP errors / middleware from PRD-477. Retarget to `main` once PRD-477 merges.

fixes PRD-508

> [!NOTE]
> ### Validate bearer JWT claims via Zod middleware in workflow-executor HTTP server
> - Adds a Zod-based middleware in [`executor-http-server.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1659/files#diff-f24eaa83e4d72cc710e9cb4cea88a2047543ee3c17a0b68627f6af587546b614) that validates `ctx.state.user` against `BearerClaimsSchema` (integer `id` required); invalid payloads receive 401 before reaching any route handler.
> - Introduces [`http-errors.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1659/files#diff-52c694f90bcb7a90ab609520d7e5c04d6461f0ea999fb4d85801f745a5597eba) with typed HTTP error classes and a `toHttpError` utility that maps domain error categories to HTTP statuses: 404 for `NotFoundError`, 403 for `AccessDeniedError`, 503 for `UnavailableError`, 400 for uncategorized `WorkflowExecutorError`, and 401 for koa-jwt errors.
> - Refactors domain errors in [`errors.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1659/files#diff-0109f91a733affdd89e2fae234f7010a44b4b923e1d56be5c84f8014b219cef0) into abstract category base classes (`NotFoundError`, `AccessDeniedError`, `UnavailableError`) so HTTP translation is driven by error category rather than per-handler branching.
> - Removes manual user-id extraction and per-error branching from route handlers; all error responses are now handled by a centralized error-translation middleware.
> - `UserMismatchError` now requires both `bearerUserId` and `ownerUserId` in its constructor, including both in the technical log message while keeping a generic user-facing message.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bde84af.</sup>
> 
>
>

### DIFF
--- a/packages/workflow-executor/src/adapters/forest-server-workflow-port.ts
+++ b/packages/workflow-executor/src/adapters/forest-server-workflow-port.ts
@@ -6,7 +6,6 @@ import type {
   MalformedRunInfo,
   WorkflowPort,
 } from '../ports/workflow-port';
-import type { StepUser } from '../types/execution-context';
 import type { CollectionSchema } from '../types/validated/collection';
 import type { StepOutcome } from '../types/validated/step-outcome';
 import type { ToolConfig } from '@forestadmin/ai-proxy';
@@ -226,7 +225,7 @@ export default class ForestServerWorkflowPort implements WorkflowPort {
     );
   }
 
-  async hasRunAccess(runId: string, user: StepUser): Promise<boolean> {
+  async hasRunAccess(runId: string, user: { id: number }): Promise<boolean> {
     return this.callPort('hasRunAccess', async () => {
       const { hasAccess } = await ServerUtils.query<{ hasAccess: boolean }>(
         this.options,
--- a/packages/workflow-executor/src/http/bearer-claims.ts
+++ b/packages/workflow-executor/src/http/bearer-claims.ts
@@ -1,0 +1,8 @@
+import { z } from 'zod';
+
+// Claims we require from the decoded bearer JWT payload (ctx.state.user). Non-strict on purpose:
+// jsonwebtoken adds standard claims (iat/exp) and Forest may send extra ones — strict would reject
+// every real token. Only `id` is consumed downstream (handleTrigger + hasRunAccess → ?userId=).
+export const BearerClaimsSchema = z.object({ id: z.number() });
+
+export type BearerClaims = z.infer<typeof BearerClaimsSchema>;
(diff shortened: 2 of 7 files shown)


matthv (approved):
LGTM

PR #1662, opened 2026-06-15
feat(workflow-executor): report executor version to orchestrator at startup

nbouliol:
## What

At startup, the workflow-executor reports its running version to the orchestrator, which stores it per environment.

## Changes

- `WorkflowPort.reportExecutorVersion(version)` + adapter implementation → `POST /api/workflow-orchestrator/executor-version` with `{ version }` (auth via existing `forest-secret-key`/`envSecret`).
- Version read from `package.json`; pushed **fire-and-forget** in `createWorkflowExecutor.start()` after the runner and HTTP server are up. Failures are logged (`Warn`) and never block or crash startup.

Companion PRs: backend adds the column + route (forestadmin-server), frontend exposes the field (forestadmin).

## Tests

- Adapter: asserts the call hits the right path/body.
- Startup: asserts the version is reported, and that `start()` still resolves when reporting fails.
- All 1054 workflow-executor tests pass; lint clean.

> [!NOTE]
> ### Report executor version to the workflow orchestrator on startup
> - Adds `WorkflowPort.reportExecutorMetadata` and implements it in `ForestServerWorkflowPort` as a POST to `/api/workflow-orchestrator/executor-metadata` with the executor version read from `package.json`.
> - `Runner.start` fires this call asynchronously after transitioning to running state; startup success is unaffected if the call fails.
> - Failures are logged at Warn level with version and error details via `Runner.pushMetadataToOrchestrator`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c61f3b9.</sup>
> 
>

### DIFF
--- a/packages/workflow-executor/src/build-workflow-executor.ts
+++ b/packages/workflow-executor/src/build-workflow-executor.ts
@@ -1,5 +1,4 @@
 import type { Logger, LoggerLevel } from './ports/logger-port';
-import type { WorkflowPort } from './ports/workflow-port';
 import type { RunnerState } from './runner';
 import type { AiConfiguration } from '@forestadmin/ai-proxy';
 import type { Options as SequelizeOptions } from 'sequelize';
@@ -29,9 +28,6 @@ import DatabaseStore from './stores/database-store';
 import InMemoryStore from './stores/in-memory-store';
 
 const FORCE_EXIT_DELAY_S = 5;
-
-// eslint-disable-next-line @typescript-eslint/no-var-requires, import/no-dynamic-require, global-require
-const { version: EXECUTOR_VERSION } = require('../package.json') as { version: string };
 
 export interface WorkflowExecutor {
   start(): Promise<void>;
@@ -141,7 +137,6 @@ function createWorkflowExecutor(
   runner: Runner,
   server: ExecutorHttpServer,
   logger: Logger,
-  workflowPort: WorkflowPort,
 ): WorkflowExecutor {
   let shutdownPromise: Promise<void> | null = null;
 
@@ -188,15 +183,6 @@ function createWorkflowExecutor(
       await runner.start();
       await server.start();
 
-      // Fire-and-forget: report our version to the orchestrator (last-write-wins per env).
-      // Must never block or crash startup, so swallow failures after logging.
-      workflowPort.reportExecutorMetadata(EXECUTOR_VERSION).catch(error => {
-        logger('Warn', 'Failed to report executor version to orchestrator', {
-          version: EXECUTOR_VERSION,
-          error: error instanceof Error ? error.message : String(error),
-        });
-      });
-
       process.on('SIGTERM', onSignal);
       process.on('SIGINT', onSignal);
     },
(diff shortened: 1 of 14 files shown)


packages/workflow-executor/src/build-workflow-executor.ts
```diff
@@ -183,6 +188,15 @@ function createWorkflowExecutor(
       await runner.start();
       await server.start();
 
+      // Fire-and-forget: report our version to the orchestrator (last-write-wins per env).
+      // Must never block or crash startup, so swallow failures after logging.
+      workflowPort.reportExecutorMetadata(EXECUTOR_VERSION).catch(error => {
```
Scra3: maybe we can move this to runner.start method :) Like this, the factory is agnostic to this behavior.

packages/workflow-executor/test/build-workflow-executor.test.ts
```diff
@@ -374,6 +379,23 @@ describe('WorkflowExecutor lifecycle', () => {
...
+
+  it('start() resolves even when reporting the version fails', async () => {
+    MockedPort.prototype.reportExecutorMetadata = jest
+      .fn()
+      .mockRejectedValue(new Error('orchestrator down'));
+
+    const exec = buildInMemoryExecutor(BASE_OPTIONS);
+
+    await expect(exec.start()).resolves.toBeUndefined();
+    expect(MockedRunner.prototype.start).toHaveBeenCalled();
```
hercemer42: **Claude Opus 4.8** `claude-opus-4-8[1m]` — **[Should fix]**

This test executes the `.catch` handler in `build-workflow-executor.ts:194` (the `logger('Warn', …)` call) but never asserts it ran — the line is covered without an expectation, which the project's testing rule explicitly calls fake coverage. Assert the warning is emitted, e.g. inject a mock `logger` via `BASE_OPTIONS` and `expect(logger).toHaveBeenCalledWith('Warn', 'Failed to report executor version to orchestrator', expect.objectContaining({ version: EXECUTOR_VERSION, error: 'orchestrator down' }))`. Since `start()` doesn't await the report, flush microtasks first (`await new Promise(r => setImmediate(r))`) so the assertion isn't timing-dependent. A non-`Error` rejection case would also cover the `String(error)` branch at line 196.

packages/workflow-executor/src/build-workflow-executor.ts
hercemer42: **Claude Opus 4.8** `claude-opus-4-8[1m]` — **[Preferential]**

The documented contract here is that version reporting must never block startup, but no test pins it. The failure test uses a *rejected* promise, which settles immediately — a slow/hanging orchestrator call wouldn't be caught. Consider a test with a never-resolving mock (`jest.fn().mockReturnValue(new Promise(() => {}))`) asserting `start()` still resolves, so a future stray `await` on this line can't silently start blocking startup.

packages/workflow-executor/src/ports/workflow-port.ts
```diff
@@ -36,6 +36,8 @@ export interface WorkflowPort {
   ): Promise<AvailableRunDispatch | null>;
   getCollectionSchema(collectionName: string, runId: string): Promise<CollectionSchema>;
   getMcpServerConfigs(): Promise<Record<string, ToolConfig>>;
+  // Pushes the running executor's version to the orchestrator (last-write-wins, stored per env).
```
hercemer42: I don't think the comment is needed because if you follow the function call then the function explains itself well

hercemer42 (changes requested):

hercemer42 (approved):

PR #1663, opened 2026-06-15
fix(workflow-executor): dedup record-level update via atomic step claim

christophebrun-forest:
## Context

PRD-451 — **layer 2** (executor). An `UpdateData` step with `executionType: AutomatedWithConfirmation` + `automaticCompletion: true` could be executed twice when two consumers fetched the same `Started` run (confirmation POST + polling executor, or two replicas), emitting a duplicate record-level `update` activity log and mutating the record twice.

Layer 1 (orchestrator, single dispatch via atomic claim) is fixed in [forestadmin-server#8303](https://github.com/ForestAdmin/forestadmin-server/pull/8303). This PR closes the executor side so the executor is idempotent on its own for the residual windows the orchestrator delegates to layer 2 (lock-expiry re-dispatch, multiple replicas, auto-chain retry).

## Root cause

Two ordering problems made the executor non-idempotent:

1. **TOCTOU** — `checkIdempotency()` read the phase at the start of `execute()`, but the `executing` write-ahead marker was written much later (inside the `beforeCall` thunk, after a network round-trip). Two concurrent executions both passed the check before either claimed.
2. **Log before claim** — `ActivityLog.track()` called `createPending()` (emits the `update` log) *before* running `beforeCall` (the marker). The claim could therefore never gate the log.

## What this PR does

- Adds `RunStore.claimStepExecution`: an **atomic compare-and-set** that transitions a step into `idempotencyPhase: 'executing'` only if currently unclaimed.
  - `DatabaseStore`: `SELECT … FOR UPDATE` (PG/MySQL) for a pre-existing row + plain `INSERT` guarded by the unique `(run_id, step_index)` index for the insert race (`UniqueConstraintError` → re-read). A delete+insert would have erased a rival's row and let both win — fixed with separate `updateData`/`insertData`.
  - `InMemoryStore`: atomic read-modify-write (single-threaded).
- `BaseStepExecutor.claimForExecution()` + shared `idempotencyOutcome()` mapper (`won`→proceed, `done`→replay, `executing`→`StepStateError`).
- The 3 mutating executors (`update-record`, `trigger-action`, `mcp`) claim **before** the side effect and **before** `createPending`. Removes the `beforeCall` thunk from `AgentWithLog`/`ActivityLog`.
- **Cleanup**: the 3 `checkIdempotency` overrides now delegate to `idempotencyOutcome()` instead of re-inlining the same `done`/`executing` mapping (drops the now-unused `StepStateError` import from the mcp executor).

Resulting order: **claim → createPending → mutation → done**. The log and the mutation fire at most once.

## What is load-bearing vs defense-in-depth

- **Load-bearing (independent of concurrency):** the reordering (write-ahead marker **before** `createPending`) and the marker itself fix the duplicate log on sequential re-dispatch / crash recovery.
- **Defense-in-depth:** the *atomicity* of the claim (`FOR UPDATE` / unique-index race) only matters under truly concurrent dispatch — which the orchestrator now prevents on the happy path, but which still occurs in the windows it delegates here (60s lock-expiry while a slow update runs, multi-replica, auto-chain `POST /update-step` retry).

## Tradeoff

Claim-first means a failure between the claim and success leaves an orphan `executing` marker → manual retry. This is the correct fail-safe direction (never double-applies); the activity-log port already retries transient failures internally. Documented in `CLAUDE.md`.

## Tests

- Concurrency regression (PRD-451): two concurrent `execute()` on the same `(runId, stepIndex)` confirmation → `agentPort.updateRecord` **and** `createPending` called **exactly once**.
- Store-level claim tests (won / executing / done / pre-existing unclaimed row / sequential).
- Updated the 3 executors' idempotency tests; removed obsolete `beforeCall` tests.

✅ 1005 tests pass · ✅ lint 0 errors · ✅ build OK.

> [!NOTE]
> Not addressed here: record-level idempotency at the **agent** level (`agentPort.updateRecord`), 3rd point of the ticket's dev note. The layer-2 claim prevents the double call *from the executor*; agent-side dedup can be a follow-up.

fixes PRD-451


christophebrun-forest:
Closing: no longer relevant.

PR #1664, opened 2026-06-16
feat(datasource-demo-fintech): in-memory fintech demo data source [PRD-533]

Gawtier:
## What

Adds `@forestadmin/datasource-demo-fintech` — a zero-config, **database-free** data source that powers `forest create demo`, bootstrapping a project with realistic sample data themed around a **fintech / AML-KYC back office** (instead of the generic `datasource-dummy` books).

## Why

`forest create demo` (PRD-531) needs a curated, interesting demo dataset that showcases Forest Admin out of the box. This package is that dataset — and ships its own business behaviour, so the generated project needs **zero customization code**.

## What's inside

- **8 related collections**: `customers`, `cards`, `aml_alerts`, `kyc_cases`, `kyc_documents`, `chargebacks`, `refund_requests`, `sar_reports` (FK relationships wired).
- **22 Smart Actions, 2 computed fields, 4 segments** packaged in the data source — actions mutate the in-memory records through the Forest collection API (e.g. block a card, clear an AML alert, create/file a SAR, approve KYC, open a chargeback from a refund).
- **Rolling dates** (`src/dates.ts`): records are dated relative to load time, so the demo always looks fresh.
- **Deterministic, dependency-free generator** (`scripts/generate.ts`, `yarn generate`) — dev-only, **excluded from the published package** (`files` ships `dist` only).

## Usage

```ts
import { createDemoFintechDataSource } from '@forestadmin/datasource-demo-fintech';

agent.addDataSource(createDemoFintechDataSource());
```

## Validation

- `yarn build` / `yarn lint` / `yarn test` green.
- Verified end-to-end on a live standalone agent + Forest project (schema pushed, actions execute and mutate state).
- Data invariants enforced by the generator: unique emails, phone dial-code matches residence, severity correlated to risk rating, sanctioned customers never approved, SAR↔alert status coherence.

## Note

The consuming change in the CLI (toolbelt `create:demo` dumper → this package) lives in the `toolbelt` repo and lands separately. Until this is published to npm, generated demo projects can use a local tarball (`npm pack`).

> [!NOTE]
> ### Add in-memory fintech demo data source package `@forestadmin/datasource-demo-fintech`
> - Adds a new package that seeds a fresh SQLite database on boot with 150 customers, cards, KYC cases/documents, AML alerts, SAR reports, refund requests, and chargebacks using `@faker-js/faker`.
> - Exposes `createDemoFintechDataSource`, a factory that wires the seeded SQL data source with relations, computed fields, segments, and Smart Actions across all eight collections (customers, cards, aml_alerts, kyc_cases, kyc_documents, chargebacks, refund_requests, sar_reports).
> - Smart Actions cover the full compliance/ops lifecycle: block/unblock/replace cards, approve/reject/escalate KYC, clear/escalate/file AML alerts and SARs, file/dismiss/win/lose chargebacks, and decline/escalate refunds to chargebacks.
> - Categorical fields are re-typed as `Enum` at runtime by patching the introspected schema via `applyEnums`.
> - Risk: the SQLite database is recreated from scratch on every boot, so no state persists between restarts.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 85c4e7d.</sup>
> 
>


matthv:
## Code review — `@forestadmin/datasource-demo-fintech` (PRD-533)

Overall this is solid and faithful to the repo conventions (collection structure and the `DataSourceCustomizer` pattern match `datasource-dummy`; packaging and the `multi-semantic-release` publish path are correctly set up). No blocking bugs. Below are the actionable items — the only two worth doing before merge are **S1** and **S2**; the rest is optional polish.

### 🟡 Should-fix

**S1 — Test coverage is far too thin.** `test/` only checks "it instantiates", "8 collections", "doesn't throw". None of the risky logic is exercised: the 22 Smart Actions (including cross-collection writes — `Create SAR`, `Open Chargeback`, `File SAR`, `Block Linked Card`), the 4 segments, the 2 computed fields, and the CRUD/filter logic in `base.ts`. Suggested minimum:
- `base.ts`: a `list` test with `conditionTree` / `sort` / `page`, and the `id` auto-assignment in `create`.
- One happy-path per collection with an action (e.g. `Clear Alert` sets `status: CLEARED`; `Open Chargeback` creates a linked chargeback and flips the refund to `ESCALATED`).
- One segment test (also covers S2).
- The invariants advertised in the PR description (unique emails, dial-code ↔ residence, sanctioned never `APPROVED`, SAR↔alert coherence) are **not** tested, even though they're sold as guaranteed.

**S2 — Confirm the segments actually return rows (operator emulation).** `base.ts.supportedOperators` does **not** declare `Before` or `Missing`, yet the `chargebacks` segments use them (`Overdue` → `Before`, `Awaiting evidence` → `Missing`). In theory the `DataSourceCustomizer` emulates `Before`→`LessThan` and `Missing`→`Equal null`/`Blank` (all present in the set), so it *should* work — but a 5-line segment test would confirm it rather than assume it.

### 🟢 Nice-to-have

**N1 — Unescaped HTML interpolation** in `resultBuilder.success(.., { html })` (`kyc_cases.ts`, `kyc_documents.ts`): `c.first_name`, `c.email`, etc. are interpolated raw into the HTML. Data is 100% synthetic, so **no real XSS here** — but since this package is also a `forest-code` showcase, raw HTML interpolation is a pattern that will get copied. Consider a small shared `esc()` helper.

**N2 — `Open Chargeback` (`refund_requests.ts`): partial dedup guard.** The "a chargeback already exists" check only runs when `refund.status === 'ESCALATED'`. On a non-escalated refund it creates without checking. In practice the action flips the refund to `ESCALATED` right after, so a 2nd call hits the guard — functionally OK, but it would read more clearly if the check were unconditional.

**N3 — Two sources of `id` allocation.** `base.create` already auto-assigns an `id` (max+1), but `Create SAR` and `Open Chargeback` recompute a `nextId` in the action and pass it explicitly. Redundant (the explicit `id` is honored, so no bug), but it's two ID strategies to maintain — could be simplified.

**N4 — Fields left unset on action-created rows.** `Open Chargeback` creates a chargeback without `merchant_category` / `mcc_code` / `processor_dispute_id` (which exist in the schema and are populated by the generator). Action-created rows will look "emptier" than seeded ones. Cosmetic for a demo.

**N5 — `Verify Document` fails randomly 20% of the time** (`Math.random() < 0.8`). For a demo meant to impress, a random failure may confuse the evaluator — product decision, worth confirming.

**N6 — File-naming inconsistency**: collections are kebab-case (`aml-alerts.ts`), customizations are snake_case (`aml_alerts.ts`). Cosmetic (snake_case mirrors the collection *name*), but noticeable.

**N7 — In-memory state is non-persistent**: action mutations live in a single instance's `this.records` → reset on restart, not shared across server instances. This is the intended demo behaviour and is documented — just keep it in mind for the onboarding messaging.

### Minor packaging note
`scripts/generate.ts` is neither type-checked at build (`tsconfig.include = src/**`) nor linted (`lint` only covers `src test`), and runs with `TS_NODE_TRANSPILE_ONLY=1`. The determinism guarantee rests on it, so a type error there would slip through CI. Acceptable for a dev-only tool, just flagging.

**N8 — A few cryptic abbreviations in `scripts/generate.ts`.** Class, file, and most variable/helper names are clear and self-documenting. The only nit: a handful of terse names in the generator that aren't obvious out of context — `ATYPES` (alert types), `cor` (country of residence), `onb` (onboarding status), and the `CB_*` prefixes (chargeback). Understandable inside the loop, but `ALERT_TYPES` / `countryOfResidence` / `onboardingStatus` would read better. Generator-only, so very low priority.

matthv (approved):

PR #1619, opened 2026-06-02
feat(workflow-executor): add OAuth credential store + deposit endpoint (PRD-367 PR1)

hercemer42:
## What

Executor-side persistence + intake for OAuth-protected MCP credentials (PRD-367, PR1 of the story):

- `ai_mcp_oauth_credentials` table via a new Umzug migration (`002`) + `McpOAuthCredentialsStore`
- `CredentialEncryption`: HKDF (read lazily from `FOREST_EXECUTOR_ENCRYPTION_KEY`, fixed context label) + AES-256-GCM, fail-closed
- `POST`/`DELETE /mcp-oauth-credentials` on the existing `koaJwt`-authed HTTP server (`user_id` from the token), encrypts + upserts; returns a typed `executor_encryption_key_missing` (503) when the key is unset

## Notes

- **Additive and dormant** — nothing reads the table or calls the endpoint yet, so it merges with no production impact.
- The encryption key is read lazily; an executor without OAuth in use boots unaffected.
- No `enc_key_version` column: managed key rotation isn't built, so `decrypt` is single-generation and rotation is a **destructive re-consent** — a key change makes existing rows undecryptable and they fall to the re-consent path (fail-closed, no silent breakage).

## Tests

- Unit — `CredentialEncryption`: round-trip, AES-GCM tamper/cross-key fail-closed, lazy/missing-key.
- Store (sqlite) — upsert/get/delete, `UNIQUE(user_id, mcp_server_id)`, nullable public-client fields, isolation, migration.
- Route — JWT auth, `user_id`-from-token (never body), encrypt-before-persist, key-missing 503, body validation.

Refs PRD-367

🤖 Generated with [Claude Code](https://claude.com/claude-code)

> [!NOTE]
> ### Add MCP OAuth credential store and deposit/delete endpoints to workflow executor
> - Adds `POST /mcp-oauth-credentials` and `DELETE /mcp-oauth-credentials/:mcpServerId` routes to `ExecutorHttpServer`, with JWT-authenticated user identity extraction, Zod body validation, and AES-256-GCM encryption before persistence.
> - Introduces `CredentialEncryption` (AES-256-GCM via HKDF-SHA256) keyed from `FOREST_EXECUTOR_ENCRYPTION_KEY`, and `ExecutorEncryptionKeyMissingError` thrown when the key is absent.
[...]

PR #1665, opened 2026-06-16
feat(workflow-executor): use stored OAuth credentials for MCP steps [PRD-367]

hercemer42:
## What & why
Executor-side runtime use of stored OAuth credentials for OAuth2-protected MCP servers (PRD-367, PR2). At an `oauth2` MCP step the executor looks up the stored credential by `(user, server)`, runs a refresh-token grant against the stored `token_endpoint` behind an expiry-skew cache, injects the Bearer token before connecting, retries once on an upstream 401 (covering both list-tools and the tool call), and pauses the step `awaiting-input` with `awaitingInputReason: needs-oauth-reauth` when there is no usable credential or the refresh is rejected. Bearer/none steps are unchanged.

## Stacked PR / deploy ordering
- Based on `feat/prd-367-pr1-executor-oauth-credentials` (PR #1619), not `main`: PR2 depends on PR1's credential store / encryption / deposit endpoint, which is not yet merged. Review the diff against that branch.
- Behaviour stays dormant until the Forest server serves `authType` + accepts the typed `awaitingInputReason` (PR1.5) and the frontend ships (PR3) — deploy the orchestrator first. Safe to deploy alone: the oauth2 path only activates once `authType=oauth2` is served; bearer/none is unchanged.

## Notable choices (large-PR annotation)
- Serialization (option B): in-process mutex + DB re-read + single retry on `invalid_grant`, not a Postgres row lock — the executor and token endpoints can sit behind a client VPN, so no DB lock is held across the refresh HTTP call. Row lock is the documented hardening path if a strict reuse-detection provider plus real cross-process contention appears.
- Executor isolation: the executor gains only a re-loadable tool source (`reloadWithFreshAuth`) plus a typed `OAuthReauthRequiredError -> awaiting-input` mapping; all token/credential/HTTP logic stays behind `RemoteToolFetcher` and the new OAuth token service.
- Shared `@forestadmin/ai-proxy` change (consumed by the agent too) is additive-only (new `loadToolsWithFailures` / `loadRemoteToolsWithFailures` + `mcp-auth-error` export) plus behaviour-preserving delegation; `authType` is stripped in the `McpClient` constructor like `id`.
- In-memory executor (dev-only) raises a `ConfigurationError` for oauth2 steps (no credential store wired).

## Tests
294 tests across the touched suites (ai-proxy 59 + workflow-executor 235); build, lint, tsc clean; >=90% line/stmt coverage on changed files.

## Known limitation
A re-auth at tool-call time leaves the step at `idempotencyPhase=executing`, so it needs a manual retry after re-auth; the common list-tools-time re-auth (pre-executor) resumes cleanly.

Part of PRD-367.

> [!NOTE]
> ### Add stored OAuth credential support for MCP steps in workflow executor
> - Introduces `OAuthTokenService` to retrieve, cache, and refresh OAuth access tokens per (userId, mcpServerId), with per-key serialization via a new `KeyedMutex` and rotation-aware retry logic.
> - Adds `refreshAccessToken` and `assertSafeTokenEndpoint` to perform the OAuth2 refresh-token grant and validate token endpoints against SSRF risks before any network call.
> - Updates `RemoteToolFetcher.fetch` to inject Bearer tokens for OAuth2 MCP servers, detect auth failures, and retry once with a forced refresh before surfacing `OAuthReauthRequiredError`.
> - Updates `McpStepExecutor` and `StepExecutorFactory` so auth failures during tool invocation produce an `awaiting-input` outcome with reason `needs-oauth-reauth` instead of failing the step.
> - Adds a `GET /list-mcp-tools` HTTP endpoint for design-time tool listing, with typed 409/503 responses for reauth and missing encryption key cases.
> - `POST /mcp-oauth-credentials` now validates `tokenEndpoint` safety at deposit time; `DELETE` now evicts cached tokens immediately.
> - Risk: `Runner` and `ExecutorHttpServer` constructor signatures have changed to require OAuth-related dependencies.
>
> 
> #### Changes since #1665 opened
>
> - Prevented OAuth refresh token flow from following redirects and restricted token endpoint URL validation to enforce HTTPS and reject loopback addresses in non-development/non-test environments [5f1c6f3]
> - Evicted cached OAuth access tokens when credentials are upserted for a user and MCP server pair [5f1c6f3]
> - Modified `McpStepExecutor.doExecute` to clear the write-ahead idempotency phase marker when an `OAuthReauthRequiredError` is caught by calling a new `clearReauthPauseState` method that either re-saves the step execution with the phase cleared (if `pendingData` exists) or deletes the step execution entirely (if no `pendingData` exists) [9d48e40]
> - Changed `OAuthTokenService.persistRotatedRefreshToken` to use `store.updateIfPresent` instead of `store.upsert` when writing back rotated refresh tokens [9d48e40]
> - Added `updateIfPresent` method to `McpOAuthCredentialsStore` interface and implemented it in `DatabaseMcpOAuthCredentialsStore` and `InMemoryMcpOAuthCredentialsStore` [9d48e40]
> - Added `deleteStepExecution` method to `RunStore` interface and implemented it in `DatabaseStore` and `InMemoryStore` [9d48e40]
> - Added comprehensive test suite 'McpStepExecutor — re-auth pause hardening' in `mcp-step-executor.test.ts` validating re-auth pause behavior [9d48e40]
> - Updated all test mock factories across executor and service test files to include `deleteStepExecution` and `updateIfPresent` mocks [9d48e40]
> - Added tests for `updateIfPresent` in `database-mcp-oauth-credentials-store.test.ts` and `in-memory-mcp-oauth-credentials-store.test.ts` verifying in-place updates, no-insert behavior when absent, and no-update behavior when the row was re-created with a new id [9d48e40]
> - Added tests for `deleteStepExecution` in `database-store.test.ts` and `in-memory-store.test.ts` ensuring step deletion, preservation of other steps, and no-op behavior when the target step does not exist [9d48e40]
> - Updated token-service tests to assert `updateIfPresent` is called instead of `upsert` for refresh token rotation and added a test ensuring credentials deleted between read and write-back are not resurrected [9d48e40]
> - Added subsection 'OAuth2-protected MCP' to `workflow-executor` README describing how to exercise the executor's OAuth2-protected MCP path using `mcp-oauth-test-server` [9d48e40]
> 
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 9d48e40. 9 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>
>

### DIFF
--- a/packages/ai-proxy/src/mcp-auth-error.ts
+++ b/packages/ai-proxy/src/mcp-auth-error.ts
@@ -42,10 +42,11 @@ function errorChain(error: unknown): unknown[] {
 export function isMcpAuthError(error: unknown): boolean {
   return errorChain(error).some(link => {
     const status = statusOf(link);
+    // An explicit status is authoritative — a 403 is not a refreshable auth error even if its
+    // message says "unauthorized". Fall back to the message only when no status is present.
+    if (status !== undefined) return AUTH_STATUSES.has(status);
 
-    return (
-      (status !== undefined && AUTH_STATUSES.has(status)) || AUTH_PATTERN.test(messageOf(link))
-    );
+    return AUTH_PATTERN.test(messageOf(link));
   });
 }
 
--- a/packages/ai-proxy/src/mcp-auth-error.ts
+++ b/packages/ai-proxy/src/mcp-auth-error.ts
@@ -1,8 +1,10 @@
-// Classifies errors surfaced while connecting to or calling an MCP server. The MCP SDK / HTTP
+// Classifies errors surfaced while connecting to or calling an MCP server. Only 401 (the token was
+// rejected) is a refreshable auth failure; 403 is a permission/scope problem a token refresh or
+// re-consent cannot resolve, so it is left to surface as an ordinary failure. The MCP SDK / HTTP
 // transport reports failures in several shapes (a numeric status field, or only a message string),
 // so the checks walk the cause chain and inspect both structured status and the message text.
-const AUTH_STATUSES = new Set([401, 403]);
-const AUTH_PATTERN = /\b40[13]\b|unauthorized|forbidden/i;
+const AUTH_STATUSES = new Set([401]);
+const AUTH_PATTERN = /\b401\b|unauthorized/i;
 const CONNECTION_PATTERN =
   /econnrefused|econnreset|etimedout|enotfound|eai_again|fetch failed|network|socket|timeout|connect/i;
 
--- a/packages/ai-proxy/src/mcp-auth-error.ts
+++ b/packages/ai-proxy/src/mcp-auth-error.ts
@@ -1,0 +1,58 @@
+// Classifies errors surfaced while connecting to or calling an MCP server. The MCP SDK / HTTP
+// transport reports failures in several shapes (a numeric status field, or only a message string),
+// so the checks walk the cause chain and inspect both structured status and the message text.
+const AUTH_STATUSES = new Set([401, 403]);
+const AUTH_PATTERN = /\b40[13]\b|unauthorized|forbidden/i;
+const CONNECTION_PATTERN =
+  /econnrefused|econnreset|etimedout|enotfound|eai_again|fetch failed|network|socket|timeout|connect/i;
+
+export type McpLoadFailureKind = 'auth' | 'connection' | 'unknown';
+
+function statusOf(value: unknown): number | undefined {
+  const candidate = value as { code?: unknown; status?: unknown; statusCode?: unknown };
+
+  for (const field of [candidate?.code, candidate?.status, candidate?.statusCode]) {
+    if (typeof field === 'number') return field;
+  }
+
+  return undefined;
+}
+
+function messageOf(value: unknown): string {
+  if (value instanceof Error) return value.message;
+  if (typeof value === 'string') return value;
+
+  return '';
+}
+
+function errorChain(error: unknown): unknown[] {
+  const links: unknown[] = [];
+  let current: unknown = error;
+
+  while (current && links.length < 10 && !links.includes(current)) {
+    links.push(current);
+    current = (current as { cause?: unknown }).cause;
+  }
+
+  return links;
+}
+
+export function isMcpAuthError(error: unknown): boolean {
+  return errorChain(error).some(link => {
+    const status = statusOf(link);
+
+    return (
+      (status !== undefined && AUTH_STATUSES.has(status)) || AUTH_PATTERN.test(messageOf(link))
+    );
+  });
+}
+
+export function classifyMcpLoadError(error: unknown): McpLoadFailureKind {
+  if (isMcpAuthError(error)) return 'auth';
+
+  const isConnectionFailure = errorChain(error).some(link =>
+    CONNECTION_PATTERN.test(messageOf(link)),
+  );
+
+  return isConnectionFailure ? 'connection' : 'unknown';
+}
--- a/packages/workflow-executor/src/adapters/always-error-ai-model-port.ts
+++ b/packages/workflow-executor/src/adapters/always-error-ai-model-port.ts
@@ -1,5 +1,10 @@
 import type { AiModelPort } from '../ports/ai-model-port';
-import type { BaseChatModel, RemoteTool, ToolConfig } from '@forestadmin/ai-proxy';
+import type {
+  BaseChatModel,
+  McpServerLoadFailure,
+  RemoteTool,
+  ToolConfig,
+} from '@forestadmin/ai-proxy';
 
 import { AiModelPortError } from '../errors';
 
@@ -20,6 +25,13 @@ export default class AlwaysErrorAiModelPort implements AiModelPort {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   loadRemoteTools(_configs: Record<string, ToolConfig>): Promise<RemoteTool[]> {
     return Promise.resolve([]);
+  }
+
+  loadRemoteToolsWithFailures(
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    _configs: Record<string, ToolConfig>,
+  ): Promise<{ tools: RemoteTool[]; failures: McpServerLoadFailure[] }> {
+    return Promise.resolve({ tools: [], failures: [] });
   }
 
   closeConnections(): Promise<void> {
--- a/packages/workflow-executor/src/adapters/step-outcome-to-update-step-mapper.ts
+++ b/packages/workflow-executor/src/adapters/step-outcome-to-update-step-mapper.ts
@@ -32,6 +32,10 @@ export default function toUpdateStepRequest(
     context.selectedOption = outcome.selectedOption;
   }
 
+  if (outcome.type === 'mcp' && outcome.awaitingInputReason !== undefined) {
+    context.awaitingInputReason = outcome.awaitingInputReason;
+  }
+
   const attributes: ServerStepHistoryUpdate = {
     done: outcome.status !== 'awaiting-input',
     context,
--- a/packages/workflow-executor/src/executors/mcp-step-executor.ts
+++ b/packages/workflow-executor/src/executors/mcp-step-executor.ts
@@ -1,16 +1,22 @@
 import type { ExecutionContext, StepExecutionResult } from '../types/execution-context';
 import type { McpStepExecutionData, McpToolCall } from '../types/step-execution-data';
 import type { McpStepDefinition } from '../types/validated/step-definition';
-import type { RecordStepStatus } from '../types/validated/step-outcome';
+import type { AwaitingInputReason, RecordStepStatus } from '../types/validated/step-outcome';
 import type { RemoteTool } from '@forestadmin/ai-proxy';
 
-import { DynamicStructuredTool, HumanMessage, SystemMessage } from '@forestadmin/ai-proxy';
+import {
+  DynamicStructuredTool,
+  HumanMessage,
+  SystemMessage,
+  isMcpAuthError,
+} from '@forestadmin/ai-proxy';
 import { z } from 'zod';
 
 import {
   McpToolInvocationError,
   McpToolNotFoundError,
   NoMcpToolsError,
+  OAuthReauthRequiredError,
   StepStateError,
 } from '../errors';
 import BaseStepExecutor from './base-step-executor';
@@ -28,14 +34,18 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
 
   private readonly mcpServerName?: string;
 
+  private readonly reloadWithFreshAuth?: () => Promise<RemoteTool[]>;
+
   constructor(
     context: ExecutionContext<McpStepDefinition>,
     remoteTools: readonly RemoteTool[],
     mcpServerName?: string,
+    reloadWithFreshAuth?: () => Promise<RemoteTool[]>,
   ) {
     super(context);
     this.remoteTools = remoteTools;
     this.mcpServerName = mcpServerName;
+    this.reloadWithFreshAuth = reloadWithFreshAuth;
   }
 
   protected override getExtraLogContext(): Record<string, unknown> {
@@ -48,6 +58,7 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
   protected buildOutcomeResult(outcome: {
     status: RecordStepStatus;
     error?: string;
+    awaitingInputReason?: AwaitingInputReason;
   }): StepExecutionResult {
     return {
       stepOutcome: {
@@ -74,6 +85,22 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
   }
 
   protected async doExecute(): Promise<StepExecutionResult> {
+    try {
+      return await this.runStep();
+    } catch (error) {
+      // An unrefreshable OAuth credential pauses the step for re-authentication rather than failing it.
+      if (error instanceof OAuthReauthRequiredError) {
+        return this.buildOutcomeResult({
+          status: 'awaiting-input',
+          awaitingInputReason: error.awaitingInputReason,
+        });
+      }
+
+      throw error;
+    }
+  }
+
+  private async runStep(): Promise<StepExecutionResult> {
     // Branch A -- Re-entry after pending execution found in RunStore
     const pending = await this.patchAndReloadPendingData<McpStepExecutionData>(
       this.context.incomingPendingData,
@@ -124,13 +151,7 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
         recordId: this.context.baseRecordRef.recordId,
       },
       {
-        operation: async () => {
-          try {
-            return await tool.base.invoke(target.input);
-          } catch (cause) {
-            throw new McpToolInvocationError(target.name, cause);
-          }
-        },
+        operation: () => this.invokeWithReauthRetry(tool, target),
         beforeCall: () =>
           this.context.runStore.saveStepExecution(this.context.runId, {
             ...existingExecution,
@@ -193,6 +214,35 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
     }
 
     return this.buildOutcomeResult({ status: 'success' });
+  }
+
+  // No-op for bearer/none steps (no reloadWithFreshAuth). For an OAuth2 step, a 401 on the call means
+  // the token was rejected after listing tools succeeded: force one refresh, rebuild the tool, retry
+  // once. A second 401 pauses the step for re-authentication.
+  private async invokeWithReauthRetry(tool: RemoteTool, target: McpToolCall): Promise<unknown> {
+    try {
+      return await tool.base.invoke(target.input);
+    } catch (cause) {
+      if (!this.reloadWithFreshAuth || !isMcpAuthError(cause)) {
+        throw new McpToolInvocationError(target.name, cause);
+      }
+
+      const refreshedTools = await this.reloadWithFreshAuth();
+      const refreshedTool = refreshedTools.find(
+        t => t.base.name === target.name && t.sourceId === target.sourceId,
+      );
+      if (!refreshedTool) throw new McpToolNotFoundError(target.name);
+
+      try {
+        return await refreshedTool.base.invoke(target.input);
+      } catch (retryCause) {
+        if (isMcpAuthError(retryCause)) {
+          throw new OAuthReauthRequiredError(this.context.stepDefinition.mcpServerId);
+        }
+
+        throw new McpToolInvocationError(target.name, retryCause);
+      }
+    }
   }
 
   private async formatToolResult(tool: McpToolCall, toolResult: unknown): Promise<string | null> {
--- a/packages/workflow-executor/src/executors/mcp-step-executor.ts
+++ b/packages/workflow-executor/src/executors/mcp-step-executor.ts
@@ -101,7 +101,6 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
   }
 
   private async runStep(): Promise<StepExecutionResult> {
-    // Branch A -- Re-entry after pending execution found in RunStore
     const pending = await this.patchAndReloadPendingData<McpStepExecutionData>(
       this.context.incomingPendingData,
     );
@@ -112,7 +111,6 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
       );
     }
 
-    // Branches B & C -- First call
     const tools = this.requireTools();
     const { toolName, args } = await this.selectTool(tools);
     const selectedTool = tools.find(t => t.base.name === toolName);
@@ -120,11 +118,9 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
     const target: McpToolCall = { name: toolName, sourceId: selectedTool.sourceId, input: args };
 
     if (this.context.stepDefinition.executionType === StepExecutionMode.FullyAutomated) {
-      // Branch B -- direct execution
       return this.executeToolAndPersist(target);
     }
 
-    // Branch C -- Awaiting confirmation
     await this.context.runStore.saveStepExecution(this.context.runId, {
       type: 'mcp',
       stepIndex: this.context.stepIndex,
--- a/packages/workflow-executor/src/executors/mcp-step-executor.ts
+++ b/packages/workflow-executor/src/executors/mcp-step-executor.ts
@@ -88,8 +88,11 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
     try {
       return await this.runStep();
     } catch (error) {
-      // An unrefreshable OAuth credential pauses the step for re-authentication rather than failing it.
+      // An unrefreshable OAuth credential pauses the step for re-authentication rather than failing
+      // it. Clear the write-ahead marker so the resumed step is not rejected as interrupted.
       if (error instanceof OAuthReauthRequiredError) {
+        await this.clearReauthPauseState();
+
         return this.buildOutcomeResult({
           status: 'awaiting-input',
           awaitingInputReason: error.awaitingInputReason,
@@ -97,6 +100,22 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
       }
 
       throw error;
+    }
+  }
+
+  // Keep a confirmation-flow record's approved pendingData (clear only the marker) so resume replays
+  // it; delete a pendingData-less record, which would otherwise mis-route resume into confirmation.
+  private async clearReauthPauseState(): Promise<void> {
+    const existing = await this.findPendingExecution<McpStepExecutionData>('mcp');
+    if (!existing) return;
+
+    if (existing.pendingData) {
+      await this.context.runStore.saveStepExecution(this.context.runId, {
+        ...existing,
+        idempotencyPhase: undefined,
+      });
+    } else {
+      await this.context.runStore.deleteStepExecution(this.context.runId, this.context.stepIndex);
     }
   }
 
--- a/packages/workflow-executor/src/http/executor-http-server.ts
+++ b/packages/workflow-executor/src/http/executor-http-server.ts
@@ -1,4 +1,5 @@
 import type CredentialEncryption from '../crypto/credential-encryption';
+import type OAuthTokenService from '../oauth/token-service';
 import type { Logger } from '../ports/logger-port';
 import type { McpOAuthCredentialsStore } from '../ports/mcp-oauth-credentials-store';
 import type { WorkflowPort } from '../ports/workflow-port';
@@ -44,6 +45,9 @@ export interface ExecutorHttpServerOptions {
   mcpOAuthCredentialsStore: McpOAuthCredentialsStore;
   credentialEncryption: CredentialEncryption;
   remoteToolFetcher: RemoteToolFetcher;
+  // The runtime always provides this (build-workflow-executor); optional so tests that don't
+  // exercise credential deletion don't all have to construct one.
+  oauthTokenService?: OAuthTokenService;
 }
 
 export default class ExecutorHttpServer {
@@ -297,6 +301,9 @@ export default class ExecutorHttpServer {
     const userId = (ctx.state.user as BearerClaims).id;
 
     await store.delete(userId, ctx.params.mcpServerId);
+    // Evict any in-process cached access token so the disconnect is immediate, not deferred until
+    // the cached token expires (the row is gone, but the runtime would otherwise still serve it).
+    this.options.oauthTokenService?.evict(userId, ctx.params.mcpServerId);
     ctx.status = 204;
   }
 
--- a/packages/workflow-executor/src/http/executor-http-server.ts
+++ b/packages/workflow-executor/src/http/executor-http-server.ts
@@ -290,6 +290,9 @@ export default class ExecutorHttpServer {
       throw err;
     }
 
+    // Evict any cached access token so a reconnect/re-deposit takes effect immediately, instead of
+    // serving the token minted from the previous credential until it expires (mirrors delete).
+    this.options.oauthTokenService?.evict(userId, parsed.data.mcpServerId);
     ctx.status = 200;
     ctx.body = { stored: true };
   }
--- a/packages/workflow-executor/src/http/executor-http-server.ts
+++ b/packages/workflow-executor/src/http/executor-http-server.ts
@@ -2,6 +2,7 @@ import type CredentialEncryption from '../crypto/credential-encryption';
 import type { Logger } from '../ports/logger-port';
 import type { McpOAuthCredentialsStore } from '../ports/mcp-oauth-credentials-store';
 import type { WorkflowPort } from '../ports/workflow-port';
+import type RemoteToolFetcher from '../remote-tool-fetcher';
 import type Runner from '../runner';
 import type { Server } from 'http';
 
@@ -25,7 +26,11 @@ import {
 } from './mcp-oauth-credentials';
 import serializeStepForWire from './step-serializer';
 import createConsoleLogger from '../adapters/console-logger';
-import { ExecutorEncryptionKeyMissingError, extractErrorMessage } from '../errors';
+import {
+  ExecutorEncryptionKeyMissingError,
+  OAuthReauthRequiredError,
+  extractErrorMessage,
+} from '../errors';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires, import/no-dynamic-require, global-require
 const { version } = require('../../package.json') as { version: string };
@@ -38,6 +43,7 @@ export interface ExecutorHttpServerOptions {
   logger?: Logger;
   mcpOAuthCredentialsStore: McpOAuthCredentialsStore;
   credentialEncryption: CredentialEncryption;
+  remoteToolFetcher: RemoteToolFetcher;
 }
 
 export default class ExecutorHttpServer {
@@ -153,7 +159,15 @@ export default class ExecutorHttpServer {
     );
     router.post('/runs/:runId/trigger', this.handleTrigger.bind(this));
 
-    const { mcpOAuthCredentialsStore: credentialsStore, credentialEncryption } = this.options;
+    const {
+      mcpOAuthCredentialsStore: credentialsStore,
+      credentialEncryption,
+      remoteToolFetcher,
+    } = this.options;
+
+    // Design-time tool listing for the MCP-server details page: resolve the caller's vault
+    // credential, refresh, and list the oauth2 server's tools — no workflow run involved.
+    router.get('/list-mcp-tools', ctx => this.handleListMcpTools(ctx, remoteToolFetcher));
 
     router.post('/mcp-oauth-credentials', ctx =>
       this.handleDepositCredentials(ctx, credentialsStore, credentialEncryption),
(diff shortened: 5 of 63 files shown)


packages/workflow-executor/src/oauth/token-service.ts
```diff
@@ -0,0 +1,178 @@
...
+    const result = await this.runGrantWithRotationRetry(credential, userId, mcpServerId);
+
+    this.cache.set(key, {
+      accessToken: result.accessToken,
+      expiresAtMs:
+        result.expiresInS !== undefined ? this.now() + result.expiresInS * 1000 : undefined,
+    });
+
+    if (result.refreshToken) {
+      await this.persistRotatedRefreshToken(credential, result.refreshToken);
```
hercemer42: **Claude Opus 4.8 — [Should fix]** On the invalid_grant retry path, `runGrantWithRotationRetry` re-reads `latest` and grants with it, but this write-back passes the pre-retry `credential` for the non-token fields (clientId / clientSecret / tokenEndpoint / scopes). `refreshTokenEnc` and `encKeyVersion` are correct (from the fresh encrypt), so this is not token/key corruption — but if a peer concurrently re-deposited new config, those fields get reverted: a narrow lost-update on the exact multi-instance path this logic guards. Thread the credential actually used for the successful grant through to the write-back.

packages/ai-proxy/src/mcp-auth-error.ts
```diff
@@ -0,0 +1,58 @@
+// Classifies errors surfaced while connecting to or calling an MCP server. The MCP SDK / HTTP
+// transport reports failures in several shapes (a numeric status field, or only a message string),
+// so the checks walk the cause chain and inspect both structured status and the message text.
+const AUTH_STATUSES = new Set([401, 403]);
```
hercemer42: **Claude Opus 4.8 — [Should fix]** 403 is treated like 401, so a 403 on the tool call forces a refresh and (on a second 403) pauses with `needs-oauth-reauth`. The spec specifies retry on **401**; a 403 is usually insufficient-scope, which re-consent often cannot fix, so the user re-auths back into the same 403. Confirm 403 should drive the reauth flow, or narrow the refresh/reauth trigger to 401 and surface 403 as a normal failure. (Classifying 403 as `kind: auth` is fine — this is only about the reauth trigger.)

packages/ai-proxy/src/mcp-auth-error.ts
```diff
@@ -0,0 +1,58 @@
...
+const AUTH_PATTERN = /\b40[13]\b|unauthorized|forbidden/i;
```
hercemer42: **Claude Opus 4.8 — [Preferential]** This message-text fallback can match a bare `401`/`403` appearing for non-auth reasons in an error body (IDs, counts, timestamps). It only runs when no structured status field is present (those are checked first), so the blast radius is small — flagging as a known fuzzy edge, not a blocker.

packages/workflow-executor/src/oauth/token-service.ts
hercemer42: **Claude Opus 4.8 — resolved (2b5e5e5b).** Agreed. `runGrantWithRotationRetry` now returns the credential whose token produced the grant, and `refreshAndCache` writes the rotated token back onto that (current) row — so a concurrent re-deposit is no longer partially reverted. Added a regression test asserting the write-back uses the re-read credential fields.

packages/ai-proxy/src/mcp-auth-error.ts
hercemer42: **Claude Opus 4.8 — resolved (8d441d20).** Narrowed to 401 per the spec: 403 is no longer treated as a refreshable auth error, so it surfaces as an ordinary failure instead of forcing a refresh + reauth loop. Treating a scope-403 as a re-consent opportunity is a separate, out-of-scope enhancement. `isMcpAuthError` / `classifyMcpLoadError` now match 401/unauthorized only; tests updated.

packages/ai-proxy/src/mcp-auth-error.ts
hercemer42: **Claude Opus 4.8 — acknowledged, keeping as-is.** Deliberate last-resort fallback, used only when no structured status field is present (code/status/statusCode are checked first), so the misclassification window is narrow and low-probability. Tightening the message regex risks false negatives on real 401s that carry the status only in text. Leaving it as the documented fuzzy edge.

hercemer42:
**Agent code review — Claude Opus 4.8 (2026-06-18)**

Reviewed the folded-in commits `7bf9357b` (key-rotation re-consent) and `842d4680` (`GET /list-mcp-tools`) via the pr-review-toolkit code-reviewer; the runtime/max-run path was covered in the earlier review pass. **No issues found.**

Verified:
- **list-mcp-tools** — `user_id` is taken from the validated JWT (never the request); the `mcpServerId` query is validated (400 if absent). Responses leak no secrets (200 → `{name, description}` per tool; 409 → `{awaitingInputReason, mcpServerId}`). The typed `needs-oauth-reauth` 409 is set on `ctx` directly so it survives the error middleware (which would otherwise remap `OAuthReauthRequiredError` → 400).
- **Key-rotation classification** (`toGrantParams`) — the try wraps only the decrypt, so a missing key stays terminal (`ExecutorEncryptionKeyMissingError` rethrown) and a key-present decrypt failure → `needs-oauth-reauth`; the grant call sits outside the try, so `invalid_grant`/transient/5xx errors cannot be misclassified into a re-consent loop.
- Domain error classes (no raw `Error`), no PRD refs in comments, tool serialization type-safe.

packages/workflow-executor/src/oauth/refresh-grant.ts
```diff
@@ -0,0 +1,108 @@
...
+        typeof payload.error_description === 'string' ? payload.error_description : undefined,
+      );
+    }
+
+    throw new OAuthRefreshError(
+      typeof payload.error === 'string'
+        ? payload.error
+        : `the token endpoint returned ${response.status}`,
+    );
+  }
```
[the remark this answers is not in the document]
hercemer42: **Claude Opus 4.8 — agreed, fixed (11aba95b).** A literal `null` (or other non-object) body overwrote the `{}` parse default, so `payload.error` / `payload.access_token` threw a TypeError instead of the typed `OAuthRefreshError`. Non-object bodies now keep the `{}` default so the status checks surface `OAuthRefreshError` as intended. Added a regression test (literal-null body → OAuthRefreshError).

packages/workflow-executor/src/oauth/token-service.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    const { result, credential: grantedCredential } = await this.runGrantWithRotationRetry(
+      credential,
+      userId,
+      mcpServerId,
+    );
+
+    this.cache.set(key, {
+      accessToken: result.accessToken,
+      expiresAtMs:
```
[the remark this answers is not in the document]
hercemer42: **Claude Opus 4.8 — keeping as-is; the "unbounded growth" is overstated.** The cache is keyed by `(userId, mcpServerId)` and `cache.set` overwrites per key, so a no-expiry entry doesn't accumulate — it's one slot, replaced on the next refresh. `readCache` returns `undefined` when `expiresAtMs === undefined`, so a no-expiry token is intentionally never served from cache and is always re-fetched (the safe choice for a token of unknown lifetime; reviewed + accepted in the earlier pass). The write is a harmless per-key no-op; the map is bounded by distinct user×server pairs, like any in-process token cache — not unbounded.

packages/workflow-executor/src/adapters/always-error-ai-model-port.ts
```diff
@@ -22,6 +27,13 @@ export default class AlwaysErrorAiModelPort implements AiModelPort {
     return Promise.resolve([]);
   }
 
+  loadRemoteToolsWithFailures(
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
hercemer42, line 31: Why the disable ?

packages/workflow-executor/src/oauth/token-service.ts
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  private async persistRotatedRefreshToken(
+    credential: StoredMcpOAuthCredential,
+    refreshToken: string,
+  ): Promise<void> {
+    const encrypted = this.encryption.encrypt(refreshToken);
+
+    try {
+      await this.store.upsert({
```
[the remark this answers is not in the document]
hercemer42: Fixed in a27a5c4f — moved the `encrypt` call inside the best-effort `try`, so an encryption failure is logged and swallowed like an upsert failure rather than failing an otherwise-valid `getAccessToken`. New test: encrypt throws during write-back → token still returned, no upsert attempted.

packages/workflow-executor/src/http/executor-http-server.ts
```diff
@@ -285,4 +299,37 @@ export default class ExecutorHttpServer {
...
+      throw new BadRequestHttpError('Missing required query parameter "mcpServerId"');
+    }
+
+    try {
+      const { tools } = await fetcher.fetch(mcpServerId, userId);
+      ctx.status = 200;
+      ctx.body = {
+        tools: tools.map(tool => ({ name: tool.base.name, description: tool.base.description })),
+      };
+    } catch (err) {
```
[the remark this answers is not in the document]
hercemer42: Fixed in a27a5c4f — `handleListMcpTools` now maps `ExecutorEncryptionKeyMissingError` to the same typed `503 { code: 'executor_encryption_key_missing' }` the deposit endpoint returns, so the details page shows the operator-misconfig message instead of a generic error. New route test asserts the 503 + body.

packages/ai-proxy/src/mcp-auth-error.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export function isMcpAuthError(error: unknown): boolean {
+  return errorChain(error).some(link => {
+    const status = statusOf(link);
+
+    return (
+      (status !== undefined && AUTH_STATUSES.has(status)) || AUTH_PATTERN.test(messageOf(link))
+    );
+  });
+}
```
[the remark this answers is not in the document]
hercemer42: Fixed in a27a5c4f — an explicit status is now authoritative: a 403 stays non-auth even when its message contains "unauthorized". The message fallback runs only when no structured status is present. New test: `Object.assign(new Error('Unauthorized scope'), { status: 403 })` → `false`.

packages/workflow-executor/src/oauth/token-service.ts
hercemer42: Addressed earlier in this stack — the invalid_grant retry path writes back onto the re-read (`latest`) credential, not the pre-retry snapshot, so the rotated token lands on the current row's non-token fields. Covered by the test "persists the rotated token onto the re-read credential fields, not the stale ones".

packages/ai-proxy/src/mcp-auth-error.ts
hercemer42: Addressed — `AUTH_STATUSES` is now `{401}` only; a 403 (insufficient scope) no longer forces a refresh/re-auth and surfaces as an ordinary failure, matching the spec's 401-only retry.

packages/ai-proxy/src/mcp-auth-error.ts
hercemer42: Mitigated in a27a5c4f — the message-text check now runs only when the link carries no structured status, so a bare `401`/`403` appearing in a body for non-auth reasons can't trigger it when a real status is present. Keeping the message fallback for the status-less transport shapes.

packages/workflow-executor/src/oauth/token-service.ts
```diff
@@ -0,0 +1,203 @@
...
+        mcpServerId: credential.mcpServerId,
+        refreshTokenEnc: encrypted.ciphertext,
+        clientId: credential.clientId,
+        clientSecretEnc: credential.clientSecretEnc,
+        clientSecretExpiresAt: credential.clientSecretExpiresAt,
+        tokenEndpoint: credential.tokenEndpoint,
+        tokenEndpointAuthMethod: credential.tokenEndpointAuthMethod,
+        scopes: credential.scopes,
+      });
+    } catch (error) {
```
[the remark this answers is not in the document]
hercemer42: Intentional, with one caveat now corrected. Swallowing is deliberate: the access token just obtained is valid for the current call, so failing `getAccessToken` over a transient write error would be strictly worse. You're right that the re-read path does **not** silently recover this case — the stored token is unchanged, so the next refresh sees no rotation and forces re-authentication, which is the safe fallback. The misleading "recover via the re-read path" comment is corrected in 6bc22e64.

packages/workflow-executor/src/oauth/token-service.ts
```diff
@@ -0,0 +1,203 @@
...
+
+  async getAccessToken(
+    userId: number,
+    mcpServerId: string,
+    options?: { forceRefresh?: boolean },
+  ): Promise<string> {
+    const key = `${userId}:${mcpServerId}`;
+    const forceRefresh = options?.forceRefresh ?? false;
+
+    if (!forceRefresh) {
```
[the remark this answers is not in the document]
hercemer42: Delete is handled: `handleDeleteCredentials` calls `oauthTokenService.evict(userId, mcpServerId)`, dropping the cached access token immediately (test: "evicts the in-process cached access token so the disconnect takes effect immediately"). Update (re-deposit) intentionally leaves the cached access token valid until it expires — it still authenticates; only the refresh token changed, and the next refresh reads the new row. So no stale-credential exposure.

packages/workflow-executor/src/adapters/step-outcome-to-update-step-mapper.ts
```diff
@@ -32,6 +32,10 @@ export default function toUpdateStepRequest(
     context.selectedOption = outcome.selectedOption;
   }
 
+  if (outcome.type === 'mcp' && outcome.awaitingInputReason !== undefined) {
```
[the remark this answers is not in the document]
hercemer42: Not reachable in practice. The reverse mapper's `toStepOutcome` runs only for `done` steps (`toPreviousSteps` filters `s.done`), but `awaitingInputReason` is written only when `done` is false (`done: status !== 'awaiting-input'`). So a record carrying the reason is always the non-done pending step, mapped via `stepDefinition` — never through `toStepOutcome`. The frontend reads the reason from the orchestrator's persisted step context directly. The asymmetry is real but inert; happy to add the read-back for round-trip symmetry if you'd prefer, but it changes no behavior.

packages/workflow-executor/src/executors/mcp-step-executor.ts
```diff
@@ -74,6 +85,22 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
   }
 
   protected async doExecute(): Promise<StepExecutionResult> {
```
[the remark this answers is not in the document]
hercemer42: Valid — confirmed. `beforeCall` persists `idempotencyPhase: 'executing'`; when `invokeWithReauthRetry` throws `OAuthReauthRequiredError`, `doExecute` converts it to `awaiting-input` but never clears the phase, so the resume hits `checkIdempotency` and throws "Step execution was interrupted" (this is the runId-442 symptom).

Fix needs a way to clear the phase on the awaiting-input path — there is no delete/reset on the `RunStore` port today, so it's a port addition + both store impls + tests. Tracking as a follow-up rather than rushing it into this PR; it is also dormant until PR3 wires the reconnect UX. Leaving open. cc @hercemer42

packages/workflow-executor/src/executors/mcp-step-executor.ts
```diff
@@ -195,6 +216,35 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
     return this.buildOutcomeResult({ status: 'success' });
   }
 
+  // No-op for bearer/none steps (no reloadWithFreshAuth). For an OAuth2 step, a 401 on the call means
+  // the token was rejected after listing tools succeeded: force one refresh, rebuild the tool, retry
+  // once. A second 401 pauses the step for re-authentication.
+  private async invokeWithReauthRetry(tool: RemoteTool, target: McpToolCall): Promise<unknown> {
```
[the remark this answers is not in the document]
hercemer42: Valid, and the same root cause as the idempotency item on the executor: the `OAuthReauthRequiredError` thrown inside `activityLog.track` is recorded as a failed activity before `doExecute` reclassifies it as `awaiting-input`, so the audit trail shows a spurious failure for a normal re-auth pause. The fix belongs with the reauth-pause cleanup (record the pause as a non-failure outcome). Deferring together with the idempotency follow-up; leaving open.

packages/workflow-executor/src/oauth/token-service.ts
```diff
@@ -0,0 +1,209 @@
...
+        scopes: credential.scopes,
+      };
+    } catch (error) {
+      if (error instanceof ExecutorEncryptionKeyMissingError) throw error;
+
+      throw new OAuthReauthRequiredError(credential.mcpServerId);
+    }
+  }
+
+  private async persistRotatedRefreshToken(
```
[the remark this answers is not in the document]
hercemer42: Valid narrow race: if a delete lands between `refreshAndCache`'s read and `persistRotatedRefreshToken`'s upsert, the write-back resurrects the just-deleted row. Low probability (needs a concurrent disconnect mid-refresh) and self-corrects on the user's next connect, but real. Fix is to guard the write-back — re-check existence, or scope read+write under the same per-key mutex the grant already holds. Deferring as a follow-up; leaving open.

packages/workflow-executor/src/adapters/always-error-ai-model-port.ts
hercemer42: `_configs` is required by the `AiModelPort` signature but unused in this always-error stub. This repo's `@typescript-eslint/no-unused-vars` is `['error']` with no `argsIgnorePattern`, so the `_` prefix alone doesn't exempt it — hence the explicit disable (same as `loadRemoteToolsWithFailures` just below). If you'd rather, I can add `argsIgnorePattern: '^_'` to the root `.eslintrc.js` and drop both disables, but that's a repo-wide lint change.

packages/workflow-executor/src/executors/mcp-step-executor.ts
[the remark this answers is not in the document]
hercemer42: Tracked in **[PRD-692](https://linear.app/forestadmin/issue/PRD-692)** — the hardening PR that lands before enable (it blocks PRD-627 activation). The fix (clear the idempotency phase on the awaiting-input path, via a new `RunStore` method + both store impls) is captured there. Resolving here.

packages/workflow-executor/src/executors/mcp-step-executor.ts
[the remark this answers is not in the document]
hercemer42: Tracked in **[PRD-692](https://linear.app/forestadmin/issue/PRD-692)** — same root cause as the idempotency item, fixed together (record the reauth pause as a non-failure outcome). Resolving here.

packages/workflow-executor/src/oauth/token-service.ts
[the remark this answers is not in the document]
hercemer42: Tracked in **[PRD-692](https://linear.app/forestadmin/issue/PRD-692)** — write-back to be guarded against a concurrent delete (re-check existence / update-only, or shared per-(user,server) mutex). Resolving here.

packages/workflow-executor/src/adapters/always-error-ai-model-port.ts
hercemer42: Keeping the disable as-is — it's required by this repo's `@typescript-eslint/no-unused-vars` (`['error']`, no `argsIgnorePattern`), so the `_` prefix alone doesn't exempt the interface-required param. Resolving; say the word if you'd rather I add `argsIgnorePattern: '^_'` to the root `.eslintrc.js` and drop both disables (repo-wide lint change).

packages/workflow-executor/src/oauth/token-service.ts
```diff
@@ -0,0 +1,210 @@
...
+        if (cached) return cached;
+      }
+
+      return this.refreshAndCache(userId, mcpServerId, key);
+    });
+  }
+
+  // Drop the cached access token for a (user, server). Called on credential delete so a disconnect
+  // takes effect immediately, instead of the executor serving the cached token until it expires.
+  evict(userId: number, mcpServerId: string): void {
```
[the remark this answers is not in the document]
hercemer42: Tracked in **[PRD-692](https://linear.app/forestadmin/issue/PRD-692)** (finding #4) — same concurrent-delete-vs-refresh family as the write-back race, fixed together: eviction needs to synchronize with / abort the in-flight refresh (aborted flag checked before caching, under the per-(user,server) mutex). Lands in the hardening PR before enable. Resolving here.

packages/workflow-executor/src/oauth/refresh-grant.ts
```diff
@@ -0,0 +1,111 @@
...
+    body.set('client_id', clientId);
+  }
+
+  return { headers, body };
+}
+
+// Runs the OAuth2 refresh-token grant (RFC 6749 §6) against the stored token endpoint. Distinguishes
+// a non-retryable invalid_grant (revoked / rotated) from a transient failure so the caller can decide
+// between forcing re-auth and retrying.
+export default async function refreshAccessToken(
```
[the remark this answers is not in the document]
hercemer42: Real — confirmed. `tokenEndpoint` is validated only as `z.string().min(1).max(2048)` (no scheme/host check), and `refreshAccessToken` POSTs the grant (with `client_secret`) to it, so an authenticated caller who can configure an MCP server can point the executor at an internal URL (SSRF from inside the customer network, incl. cloud metadata).

Not auto-resolving because the obvious mitigation collides with the design: **private token endpoints are an intended, documented feature** (private providers refresh from inside the customer network — see PRD-367 "Public-or-private servers"), and the **dev sim runs on `localhost`**, so a blanket private-IP/loopback block is wrong. Proposed mitigation that doesn't break either: enforce `https` (allow `http` only when explicitly non-prod), and in production block link-local + cloud-metadata ranges (169.254.0.0/16, fd00:ec2::/etc.) — legit private providers don't live there. Holding for @hercemer42's call on disposition (fix in PR2 vs fold into the PRD-692 pre-enable hardening PR).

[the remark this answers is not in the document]
hercemer42: Fixed in 5b0ddb7e. Added `assertSafeTokenEndpoint`, enforced at deposit (400, per-field message) and again before the refresh POST (defence in depth for rows predating the validation):

- scheme must be `http(s)`; **https required in production** (http stays allowed off-prod for the local OAuth sim);
- **link-local / cloud-metadata** (`169.254.0.0/16`, `fe80::/10`, incl. `169.254.169.254`) blocked on every environment;
- **loopback** blocked in production;
- **RFC1918 private ranges stay allowed** — private OAuth providers refresh from inside the customer network by design (PRD-367), so a blanket private-IP block would break the feature.

Covered by `token-endpoint-url.test.ts` (incl. the prod/non-prod matrix and the RFC1918-allowed assertion), a deposit 400 case, and a refresh-grant "blocks link-local without fetching" case. Residual: IP-literal hosts only — DNS-rebinding (a hostname that resolves to an internal IP at fetch time) isn't covered, since a resolve-then-fetch check is racy; noted as a known limitation. Resolving.

packages/workflow-executor/src/oauth/token-endpoint-url.ts
```diff
@@ -0,0 +1,76 @@
...
+function classify(host: string): { loopback: boolean; linkLocal: boolean } {
+  const kind = net.isIP(host);
+
+  if (kind === 4) {
+    const [a, b] = host.split('.').map(Number);
+
+    return { loopback: a === 127, linkLocal: a === 169 && b === 254 };
+  }
+
+  if (kind === 6) {
```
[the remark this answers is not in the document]
hercemer42: Good catch — fixed in 0da92984. `classify()` now extracts the embedded IPv4 from an `::ffff:` address (the URL parser normalizes `::ffff:127.0.0.1` to the hex form `::ffff:7f00:1`, both handled) and runs it through the IPv4 loopback/link-local checks. Tests added: `::ffff:127.0.0.1` → blocked as loopback in prod, `::ffff:169.254.169.254` → blocked as link-local everywhere. Resolving.

packages/workflow-executor/src/oauth/token-endpoint-url.ts
```diff
@@ -0,0 +1,76 @@
+import net from 'net';
+
+import { InvalidTokenEndpointError } from '../errors';
+
+// Constrains the OAuth token endpoint without breaking intended private-provider support:
+//   - scheme must be http(s); in production it must be https (http stays allowed off-prod, since the
+//     local OAuth sim runs on http);
```
hercemer42: Don't reference a local test tool in a production comment

packages/workflow-executor/src/oauth/token-endpoint-url.ts
```diff
@@ -0,0 +1,76 @@
...
+//   - link-local (incl. cloud metadata, e.g. 169.254.169.254) is never a valid endpoint — blocked
+//     on every environment;
+//   - loopback / the executor host itself is blocked in production (allowed off-prod for the sim);
+//   - RFC1918 private ranges stay allowed — a private OAuth provider refreshes from inside the
+//     customer network by design (see PRD-367).
```
hercemer42: Don't mention a PRD in a comment as per conventions.  This comment could be improved overall.  Explain the why, not the how.

packages/workflow-executor/src/oauth/token-endpoint-url.ts
hercemer42: Done in 110ffb81 — dropped the local-dev-tool reference; the comment now just says http/loopback are relaxed off-production for local development, no specific tool.

packages/workflow-executor/src/oauth/token-endpoint-url.ts
hercemer42: Done in 110ffb81 — removed the PRD reference and rewrote the block to explain the why (it's an SSRF vector; RFC1918 is deliberately allowed because reaching a private provider over the internal network is a supported deployment; DNS isn't resolved because it'd be racy) instead of enumerating the rules.

packages/workflow-executor/src/oauth/token-endpoint-url.ts
```diff
@@ -0,0 +1,108 @@
...
+    return {
+      loopback: normalized === '::1' || normalized === '0:0:0:0:0:0:0:1',
+      linkLocal: /^fe[89ab]/.test(firstHextet), // fe80::/10
+      unspecified: normalized === '::' || normalized === '0:0:0:0:0:0:0:0',
+    };
+  }
+
+  // Not an IP literal — only the obvious local alias is treated as loopback; other hostnames are
+  // left to the scheme rule (not resolved).
+  return { loopback: host.toLowerCase() === 'localhost', linkLocal: false, unspecified: false };
```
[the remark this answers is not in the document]
hercemer42: Fixed in 8de88881. `classify()` now treats any host whose last label is `localhost` as loopback — bare `localhost`, the trailing-dot FQDN `localhost.`, and the `.localhost` TLD (RFC 6761) — while not over-blocking real domains (e.g. `auth.localhost.example.com` stays allowed; test added).

Worth flagging: string-level matching of loopback-resolving *names* is inherently incomplete (any /etc/hosts entry or DNS record can point an arbitrary name at 127.0.0.1). The complete fix for that class is a socket-level post-DNS filtering agent, tracked as an optional follow-up in PRD-692 (#5). This commit closes the known reserved-name cases. Resolving.

packages/workflow-executor/src/oauth/token-endpoint-url.ts
```diff
@@ -0,0 +1,114 @@
...
+  if (!hex) return null;
+
+  const hi = parseInt(hex[1], 16); // high 16 bits → first two octets
+  const lo = parseInt(hex[2], 16); // low 16 bits → last two octets
+
+  return `${Math.floor(hi / 256)}.${hi % 256}.${Math.floor(lo / 256)}.${lo % 256}`;
+}
+
+function classify(host: string): HostClass {
+  const kind = net.isIP(host);
```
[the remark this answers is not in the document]
hercemer42: This one's a false positive for the actual call path. The finding looks at `classify()`/`net.isIP()` in isolation, but `assertSafeTokenEndpoint` runs `new URL()` first, and the WHATWG IPv4 parser strips a trailing dot — so `classify()` receives `127.0.0.1` (isIP=4 → loopback → REJECT), never `127.0.0.1.`. Verified:

- `https://127.0.0.1./t` → hostname `127.0.0.1` → REJECT
- `https://169.254.169.254./t` → `169.254.169.254` → REJECT
- `https://localhost./t` → matched by the localhost regex → REJECT
- `https://[::1]./t` → invalid URL → REJECT

Added a regression test in 81b56071 locking the trailing-dot IP cases. (General note: any analysis of `classify()` in isolation will over-report, since the effective input is always the `new URL()`-normalized hostname.) Resolving.

packages/workflow-executor/src/oauth/refresh-grant.ts
```diff
@@ -0,0 +1,117 @@
...
+  // validation could still carry one — re-check before the outbound POST. Throws (terminal) rather
+  // than reaching the network.
+  assertSafeTokenEndpoint(params.tokenEndpoint);
+
+  const { headers, body } = buildRequest(params);
+
+  let response: Awaited<ReturnType<typeof fetch>>;
+
+  try {
+    response = await fetch(params.tokenEndpoint, { method: 'POST', headers, body });
```
Scra3: issue (blocking): `fetch()` follows redirects by default, so a deposited token endpoint that passes `assertSafeTokenEndpoint` but 3xx-redirects to `169.254.169.254`/loopback re-sends the POST body to the internal host — leaking the `refresh_token` (and the `client_secret` in `client_secret_post` mode) and parsing the internal response as a token; pass `redirect: 'manual'` and treat any 3xx as an `OAuthRefreshError`.

packages/workflow-executor/src/http/executor-http-server.ts
```diff
@@ -283,6 +301,51 @@ export default class ExecutorHttpServer {
     const userId = (ctx.state.user as BearerClaims).id;
 
     await store.delete(userId, ctx.params.mcpServerId);
+    // Evict any in-process cached access token so the disconnect is immediate, not deferred until
+    // the cached token expires (the row is gone, but the runtime would otherwise still serve it).
+    this.options.oauthTokenService?.evict(userId, ctx.params.mcpServerId);
```
Scra3, line 309: issue (non-blocking): the POST deposit handler has no symmetric `evict()`, so re-depositing a credential (reconnect without a prior DELETE) keeps serving the previously cached access token until it expires — mirror this evict on the deposit path.

packages/workflow-executor/src/oauth/token-service.ts
```diff
@@ -0,0 +1,210 @@
...
+    const credential = await this.store.get(userId, mcpServerId);
+    if (!credential) throw new OAuthReauthRequiredError(mcpServerId);
+
+    const { result, credential: grantedCredential } = await this.runGrantWithRotationRetry(
+      credential,
+      userId,
+      mcpServerId,
+    );
+
+    this.cache.set(key, {
```
Scra3, line 105: issue (non-blocking): `evict()` runs outside the `KeyedMutex`, so a disconnect landing during an in-flight refresh no-ops on the empty cache and this `cache.set` then re-populates a just-deleted credential (served until expiry); it is also process-local, so on a multi-instance deployment other instances never drop the token — consider an eviction epoch checked before `cache.set`.

packages/workflow-executor/src/oauth/token-endpoint-url.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  if (unspecified) {
+    throw new InvalidTokenEndpointError('it points at the unspecified address (0.0.0.0 / ::)');
+  }
+
+  if (linkLocal) {
+    throw new InvalidTokenEndpointError('it points at a link-local / metadata address');
+  }
+
+  if (isProduction()) {
```
Scra3: suggestion: the loopback block and the https requirement are gated on `NODE_ENV === 'production'`, so a real deployment with `NODE_ENV` unset or misspelled silently allows loopback targets and cleartext http — prefer failing closed and opting into the relaxation for dev/test explicitly.

Scra3 (commented):
Review of the OAuth2 MCP executor runtime (PRD-367 PR2). One blocking security issue (SSRF via redirect on the token endpoint) plus a few non-blocking hardening points. Note: the tool-call re-auth idempotency-phase issue is already fixed in the stacked #1724, so it's not raised here.

packages/workflow-executor/src/oauth/token-service.ts
```diff
@@ -0,0 +1,210 @@
...
+        refreshToken: this.encryption.decrypt(credential.refreshTokenEnc),
+        clientId: credential.clientId,
+        clientSecret: credential.clientSecretEnc
+          ? this.encryption.decrypt(credential.clientSecretEnc)
+          : null,
+        tokenEndpointAuthMethod: credential.tokenEndpointAuthMethod,
+        scopes: credential.scopes,
+      };
+    } catch (error) {
+      if (error instanceof ExecutorEncryptionKeyMissingError) throw error;
```
Scra3, line 176: suggestion: rethrowing the raw `ExecutorEncryptionKeyMissingError` here surfaces as a generic "An unexpected error occurred" in the run history (it's a boundary `Error` with no `userMessage`, so the base executor's catch-all handles it) — wrap it in a `WorkflowExecutorError` carrying an admin-actionable `userMessage` so the runtime key-missing case is legible in the UI, not only in the executor logs.

Scra3 (commented):
Runtime missing-key error surfaces too generically in run history.

packages/workflow-executor/src/oauth/refresh-grant.ts
hercemer42: Good catch — fixed in 5f1c6f36. The refresh POST now uses `redirect: 'manual'`, and any 3xx (`response.type === 'opaqueredirect'` or a 300–399 status) is turned into an `OAuthRefreshError` **before** the body is read. So a validated endpoint that 3xx-redirects to an internal host can no longer re-send the grant body (refresh token / client secret in `client_secret_post`) there or have its reply parsed as a token. Test asserts the redirect is rejected without `json()` ever being called, and that `redirect:'manual'` is set.

packages/workflow-executor/src/http/executor-http-server.ts
hercemer42: Fixed in 5f1c6f36. `handleDepositCredentials` now calls `oauthTokenService.evict(userId, mcpServerId)` after a successful upsert, mirroring the delete path — a reconnect / re-deposit drops the cached access token immediately instead of serving the prior one until expiry. Test asserts `evict(7, 'mcp-server-1')` on deposit.

packages/workflow-executor/src/oauth/token-service.ts
hercemer42: Agreed — and already tracked as **PRD-692 #4** (evict runs outside the `KeyedMutex`, so a disconnect during an in-flight refresh gets re-populated by `cache.set`) plus the process-local point. The fix (an eviction epoch checked before `cache.set`, and/or scoping evict + refresh under the same per-key mutex) lands in that hardening PR rather than here — it's in-flight there, and it also covers the deposit-evict just added. Leaving this open for your call; happy to pull it into this PR instead if you'd rather it not be split.

packages/workflow-executor/src/oauth/token-endpoint-url.ts
hercemer42: Fixed in 5f1c6f36 — you're right that a security control shouldn't fail open on env misconfig. It now **fails closed**: the strict `https` + no-loopback rules apply unless `NODE_ENV` explicitly opts into the relaxation (`development` / `test`); an unset or misspelled `NODE_ENV` stays strict. Tests cover unset → strict, a misspelled value → strict, and explicit dev/test → relaxed. (Consequence: a dev stack must now set `NODE_ENV=development` to use `localhost` / cleartext http.)

packages/workflow-executor/src/oauth/token-service.ts
hercemer42: Pushing back a little on this one — it's an intentional decision in the PRD-367 spec, not an oversight. The primary key-missing signal is the deposit-time typed `503`; the runtime residual — a key removed *after* a credential was already stored — is deliberately left as an ordinary terminal error with no bespoke runtime channel (executor logs / Observability are the admin signal), and it's a low-probability path.

There's also a concrete constraint: wrapping it in `toGrantParams` would break the typed `503` that `list-mcp-tools` depends on (that handler catches the raw `ExecutorEncryptionKeyMissingError`), so a clean version would need run-path-only handling. Happy to add an admin-actionable `userMessage` on the run path if you'd like it, but leaving as-is per the spec for now — flagging for a conscious call.

PR #1666, opened 2026-06-16
refactor(agent): forward workflow executor routes generically [PRD-567]

matthv:
## What

The agent proxies the workflow executor with **one explicitly declared route per executor route** (`GET /_internal/workflow-executions/:runId`, `POST .../trigger`). Each new executor route therefore requires a matching agent route **and an agent release** — and the agent is customer-deployed, so a new executor capability is unreachable until the parc is upgraded.

This replaces the two hand-written routes with a **single catch-all** that forwards any sub-path and any verb to the executor's `/runs/*` namespace. Adding an executor route no longer requires any agent change.

fixes PRD-567

## Design (decisions agreed with the ticket author)

- **Catch-all** `router.all('/_internal/workflow-executions/:path(.*)')` → `/runs/<path>`, all verbs.
- **Request forwarded verbatim**: all client headers (minus hop-by-hop / `Host` / `Content-Length`), **raw body** via `ctx.request.rawBody` (no reshaping), query string and method as-is.
- **Security boundary**: the wildcard can only ever map into `/runs` — path-traversal (`..`), encoded dots (`%2e`), absolute paths and null bytes throw `NotFoundError`, so executor routes outside `/runs` (e.g. `mcp-oauth-credentials`) stay **unreachable** through the proxy.
- Single request timeout, executor errors passed through verbatim.

## Tests

`workflow-executor-proxy.test.ts` → **14/14** (real HTTP executor server), including 5 path-traversal security cases (the "internal route not exposed" guard). ESLint clean.

## Scope

Phase 2 of 3 — this is the **Node `@forestadmin/agent`**. Ruby v2 (`forest_admin_agent`) is done in agent-ruby#319; the Ruby v1 liana (`forest-rails`) carries the identical coupling and gets the same generic proxy next.

> [!NOTE]
> ### Refactor `WorkflowExecutorProxyRoute` to forward all HTTP verbs and headers generically
> - Replaces explicit GET/POST route registrations with a single `router.all('/_internal/workflow-executions/:path(.*)')` catch-all that handles any HTTP verb and sub-path.
> - Forwards all client request headers to the executor except hop-by-hop, `host`, and `content-length`; forwards all executor response headers back to the client except hop-by-hop ones.
> - Passes raw request body verbatim for non-GET requests instead of re-serializing JSON.
> - Adds a 120s timeout to executor requests; adds path traversal protection via the new `executorPath` method that rejects unsafe wildcards with `NotFoundError`.
> - Behavioral Change: previously only `Authorization` and `Cookie` were forwarded; now all safe client headers are forwarded, and executor response headers are propagated to the caller.
>
> #### 🖇️ Linked Issues
> Relates to [PRD-567](ticket:linear/PRD-567).
>
> 
> #### Changes since #1666 opened
>
> - Made the workflow executor request timeout configurable via instance property [e17bd12]
> - Refactored the internal request forwarding method signature [e17bd12]
> - Added test coverage for edge cases in workflow executor proxying [e17bd12]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2fbfce1.</sup>
> 
>
>

### DIFF
--- a/packages/agent/src/routes/workflow/workflow-executor-proxy.ts
+++ b/packages/agent/src/routes/workflow/workflow-executor-proxy.ts
@@ -1,7 +1,7 @@
 import type { ForestAdminHttpDriverServices } from '../../services';
 import type { AgentOptionsWithDefaults } from '../../types';
 import type KoaRouter from '@koa/router';
-import type { OutgoingHttpHeaders } from 'http';
+import type { IncomingHttpHeaders, OutgoingHttpHeaders } from 'http';
 import type { Context } from 'koa';
 
 import { NotFoundError } from '@forestadmin/datasource-toolkit';
@@ -59,6 +59,14 @@ export default class WorkflowExecutorProxyRoute extends BaseRoute {
     );
 
     context.response.status = response.status;
+
+    // Forward response headers so the version gate (X-Forest-Executor-Version) survives the proxy.
+    for (const [name, value] of Object.entries(response.headers)) {
+      if (value !== undefined && !SKIPPED_HEADERS.has(name.toLowerCase())) {
+        context.response.set(name, value);
+      }
+    }
+
     context.response.body = response.body;
   }
 
@@ -100,7 +108,7 @@ export default class WorkflowExecutorProxyRoute extends BaseRoute {
     url: URL,
     body: string | undefined,
     headers: OutgoingHttpHeaders,
-  ): Promise<{ status: number; body: unknown }> {
+  ): Promise<{ status: number; body: unknown; headers: IncomingHttpHeaders }> {
     const requestFn = url.protocol === 'https:' ? httpsRequest : httpRequest;
 
     return new Promise((resolve, reject) => {
@@ -117,7 +125,11 @@ export default class WorkflowExecutorProxyRoute extends BaseRoute {
             parsed = raw;
           }
 
-          resolve({ status: res.statusCode ?? HttpCode.InternalServerError, body: parsed });
+          resolve({
+            status: res.statusCode ?? HttpCode.InternalServerError,
+            body: parsed,
+            headers: res.headers,
+          });
         });
         res.on('error', reject);
       });
--- a/packages/agent/src/routes/workflow/workflow-executor-proxy.ts
+++ b/packages/agent/src/routes/workflow/workflow-executor-proxy.ts
@@ -13,7 +13,8 @@ import BaseRoute from '../base-route';
 
 const AGENT_PREFIX = '/_internal/workflow-executions';
 const EXECUTOR_PREFIX = '/runs';
-// Hop-by-hop headers + those the HTTP client recomputes — never forwarded.
+// Never forwarded: hop-by-hop headers, Host (Node derives the executor's from the target URL),
+// and length/encoding the HTTP client recomputes.
 const SKIPPED_HEADERS = new Set([
   'connection',
   'keep-alive',
@@ -34,6 +35,8 @@ const REQUEST_TIMEOUT_MS = 120_000;
 export default class WorkflowExecutorProxyRoute extends BaseRoute {
   readonly type = RouteType.PrivateRoute;
   private readonly executorUrl: URL;
+  // Overridable so tests can exercise the timeout branch without waiting the full default.
+  protected readonly requestTimeoutMs: number = REQUEST_TIMEOUT_MS;
 
   constructor(services: ForestAdminHttpDriverServices, options: AgentOptionsWithDefaults) {
     super(services, options);
@@ -50,13 +53,13 @@ export default class WorkflowExecutorProxyRoute extends BaseRoute {
   private async handleProxy(context: Context): Promise<void> {
     const targetUrl = this.buildTargetUrl(context);
 
-    const response = await this.forwardRequest(
-      context.method,
-      targetUrl,
+    const response = await this.forwardRequest({
+      method: context.method,
+      url: targetUrl,
       // Raw body forwarded verbatim (set by @koa/bodyparser); undefined for GET.
-      context.method === 'GET' ? undefined : context.request.rawBody,
-      this.forwardedHeaders(context),
-    );
+      body: context.method === 'GET' ? undefined : context.request.rawBody,
+      headers: this.forwardedHeaders(context),
+    });
 
     context.response.status = response.status;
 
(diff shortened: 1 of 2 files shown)


packages/agent/src/routes/workflow/workflow-executor-proxy.ts
```diff
@@ -80,16 +126,19 @@ export default class WorkflowExecutorProxyRoute extends BaseRoute {
...
+            status: res.statusCode ?? HttpCode.InternalServerError,
+            body: parsed,
+            headers: res.headers,
+          });
         });
         res.on('error', reject);
       });
 
       req.on('error', reject);
+      req.on('timeout', () => req.destroy(new Error('Workflow executor request timed out')));
```
Scra3, line 143: **suggestion (non-blocking):** the timeout path is correct (`req.destroy(err)` rejects via the `error` handler registered above), but it has zero test coverage — the "unreachable" test hits `localhost:1`, which is `ECONNREFUSED` (the `error` path), not a socket timeout. Since `REQUEST_TIMEOUT_MS` is a hardcoded 120s, this branch can't easily be exercised. Consider making the timeout injectable so a test can stand up a non-responding executor and assert the timeout rejection. A hung executor leaving the promise pending forever is exactly the regression this guards against.

packages/agent/src/routes/workflow/workflow-executor-proxy.ts
```diff
@@ -80,16 +126,19 @@ export default class WorkflowExecutorProxyRoute extends BaseRoute {
...
 
-      if (body && method !== 'GET') {
-        req.write(JSON.stringify(body));
-      }
+      if (body !== undefined && method !== 'GET') req.write(body);
```
Scra3, line 145: **suggestion (non-blocking):** on a non-GET where `@koa/bodyparser` didn't populate `rawBody` (a content-type it ignores, or a bodyless POST/DELETE), `body` is `undefined` and the payload is silently dropped — no body forwarded, no error. Low real impact since this proxy sees JSON workflow calls, but the DELETE test never asserts `receivedBody`. Worth a test pinning a bodyless non-GET forwards cleanly (empty body, no hang) so a future guard change (`if (body)` vs `if (body !== undefined)`) doesn't shift behavior unnoticed.

packages/agent/src/routes/workflow/workflow-executor-proxy.ts
```diff
@@ -80,16 +126,19 @@ export default class WorkflowExecutorProxyRoute extends BaseRoute {
             parsed = raw;
           }
 
-          resolve({ status: res.statusCode ?? HttpCode.InternalServerError, body: parsed });
+          resolve({
+            status: res.statusCode ?? HttpCode.InternalServerError,
+            body: parsed,
+            headers: res.headers,
+          });
```
Scra3, line 137: **suggestion (non-blocking):** the `JSON.parse` → `catch` → raw fallback is the branch that makes a non-JSON or empty executor response transparent (e.g. a future `text/plain` route, or a `204` with empty body where `JSON.parse('')` throws → falls back to `''`). All tests return valid JSON, so this branch is uncovered. A passthrough test (a `204`/empty body and a plain-text body) locks the "new executor route needs no agent change" contract.

packages/agent/src/routes/workflow/workflow-executor-proxy.ts
```diff
@@ -1,18 +1,35 @@
...
 import { HttpCode, RouteType } from '../../types';
 import BaseRoute from '../base-route';
 
-type ForwardedHeaders = {
-  authorization?: string;
-  cookie?: string;
-};
+const AGENT_PREFIX = '/_internal/workflow-executions';
+const EXECUTOR_PREFIX = '/runs';
+// Hop-by-hop headers + those the HTTP client recomputes — never forwarded.
```
Scra3: **nit:** `host` isn't strictly hop-by-hop (RFC 7230) — it's dropped so the client-facing Host isn't sent to the executor (Node derives the correct Host from the target URL). The taxonomy in the comment doesn't quite cover it; minor reword so nobody concludes `host` is safe to remove from the set, e.g. "hop-by-hop, Host (rewritten to the executor's), and length/encoding recomputed by the client".

packages/agent/src/routes/workflow/workflow-executor-proxy.ts
```diff
@@ -24,50 +41,79 @@ export default class WorkflowExecutorProxyRoute extends BaseRoute {
...
+      wildcard === '' ||
+      wildcard.startsWith('/') ||
+      UNSAFE_PATH_FRAGMENTS.some(fragment => wildcard.includes(fragment));
+
+    if (unsafe) throw new NotFoundError('Invalid workflow executor path');
+
+    return `${EXECUTOR_PREFIX}/${wildcard}`;
+  }
+
+  // Forwards every client header except hop-by-hop / client-recomputed ones (Q4).
```
Scra3: **nit:** `(Q4)` is an opaque reference — unlike `PRD-567` it points to nothing resolvable from the code and will be meaningless in a few months. Drop it or link the decision.

Scra3 (commented):
Reviewed the transparent-proxy refactor (catch-all route + forward-all headers + path sanitization). The design is sound and serves the PRD-567 contract well — `router.all('.*')`, the hop-by-hop skip list, and the traversal guard are intentional, not over-engineering, so I'm not flagging them.

**No blocking issues, no correctness bug, no real security hole** — verified:
- `JSON.parse`-fallback doesn't mask errors (status comes from `res.statusCode` on a separate field; 4xx/5xx pass through intact).
- timeout `req.destroy(err)` rejects correctly (error handler registered first — no hang/swallow).
- the path boundary holds: escaping `/runs/` requires `..`, which is blocked in raw and encoded (`%2e`/`%2E`) form. (An encoded slash `%2f` alone stays under `/runs/`, so it's not a real escape — adding it would be consistency-only, optional.)

The comments above are **all non-blocking**: 3 test gaps on new-but-correct code paths (timeout, bodyless non-GET, non-JSON/204 response) that would let a future regression slip silently, plus 2 doc nits. Mergeable as-is; the tests are hardening, not a prerequisite.

One separable item (not inlined): no failure on this agent→executor hop is logged — timeout/ECONNREFUSED/DNS all collapse into a generic 500 with no context. Worth a follow-up ticket if observability matters here.

Scra3 (approved):

PR #1674, opened 2026-06-18
fix(workflow): save ai reasoning in loadRelatedRecord & updateRecord steps

EnkiP:
### General

### Security

> [!NOTE]
> ### Save AI reasoning in `loadRelatedRecord` and `updateRecord` workflow steps
> - AI reasoning is now captured from tool calls in the `LoadRelatedRecordStepExecutor`, `UpdateRecordStepExecutor`, and `McpStepExecutor`, and persisted in both `pendingData` (awaiting confirmation) and `executionResult` (on success).
> - Reasoning is only retained in the final result if the user accepts the AI suggestion unchanged; overriding the suggested value or relation drops the reasoning from the stored result.
> - `BelongsToMany` relation resolution switches from taking the first candidate to an AI ranking pass over up to 50 candidates.
> - Tool schemas for field-selection and record-selection now include a required `reasoning` field; the MCP executor strips reasoning from args before invoking the real tool and stores it separately in `McpStepExecutionData.toolSelectionReasoning`.
> - Behavioral Change: `UpdateRecordStepExecutor.selectFieldAndValue` now throws `MalformedToolCallError` when the model returns no field name, and accepts flattened field selections without an `input` wrapper.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9ec9a71.</sup>
> 
>


PMerlet:
Superseded by #1914, which rebuilds this work on current `main`.

The three commits here sat on a June `main` that has since moved by ~690 commits: the load-related-record executor was rewritten around execution modes, AI degradation and ambiguity handling, so replaying them was a re-implementation rather than a rebase. #1914 carries the same intent — persist the AI reasoning of load-related-record, update-record and MCP steps — with the same field names, so a front reading this contract is unaffected.

It also answers a review pass this branch never had: the justification is now compared by value rather than by reference, kept out of later steps' prompts, and written before the mutation rather than after it.

PR #1670, opened 2026-06-17
feat(audit-trail): create compliant audit trail

ShohanRahman:
### General

- [ ] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)
- [ ] Test manually the implemented changes
- [ ] Validate the code quality (indentation, syntax, style, simplicity, readability)

### Security

- [ ] Consider the security impact of the changes made

> [!NOTE]
> ### Add compliant audit trail plugin and agent endpoint for record change history
> - Introduces a new `plugin-audit-trail` package ([audit-trail.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1670/files#diff-97f75eb4df58933a8d7e728798addb85ba800373a685fc5ff9ff073f6bafa5ec)) that registers CRUD hooks on every collection to emit structured audit records containing operation, timestamp, userId, correlationKey, and previous/new values with optional field redaction.
> - Adds an [in-memory store](https://github.com/ForestAdmin/agent-nodejs/pull/1670/files#diff-4a734a7e77536f1ca8cab996b3d777ea69ed8bb1ad32496c98bc7466c75d4373) and a [SQL sink](https://github.com/ForestAdmin/agent-nodejs/pull/1670/files#diff-3dd5815a70cf4f35382c23f73d6dc72df725ccef3d3d3d21f34bebaeee68a77c) (Sequelize-backed) for persisting audit records, with automatic schema/table creation on supported dialects.
> - Exposes a new `GET /_audit-trail/{collectionSlug}/:id` route in the agent ([audit-trail.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1670/files#diff-bf29a683356a0984639d6233e206b57d96b2c3d113f738fb7bc519ed6dac3558)) that returns a record's change history after asserting read permission; only mounted when `auditTrail.store` is configured.
> - Adds a `correlationIdMiddleware` and `getRequestId` helper ([correlation-id.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1670/files#diff-a7814adf2a3138d69672fe8e25292f36cc594028320471958eb098d374481971)) so a stable UUID is memoized per request, reused by `parseCaller`, and echoed in the `x-forest-correlation-id` response header.
> - Behavioral Change: `parseCaller` now derives `requestId` from Koa context state instead of generating a new UUID on each call, so multiple invocations within the same request share the same id.
>
> 
>
[...]

PR #1686, opened 2026-06-19
feat(agent): built-in audit trail — history, state-at-timestamp, and smart-action capture

ShohanRahman:
### General

### Security

## Summary

Alternative to #1670: builds the compliant audit trail **directly into the agent** instead of shipping a separate plugin package. A single `auditTrail: { connectionString }` agent option enables it; the agent constructs the SQL store, instruments collections, and mounts the routes on its own. See [`packages/agent/src/audit-trail/README.md`](https://github.com/ForestAdmin/agent-nodejs/blob/feat/agent-native-audit-trail/packages/agent/src/audit-trail/README.md) for the full write-up.

### Capture

- Registers CRUD hooks on every collection using a **pending/confirm** write protocol: a `pending` row is inserted with whatever is already known *before* the write runs, then confirmed to `done` with the final values once it resolves. Recording *after* the fact meant a broken audit store failed the request at the one moment that's useless — the write had already committed. A new `critical` option (default `false`) controls what happens when recording the pending row itself fails: logged and swallowed (matching the old behavior), or `true` to refuse the write outright before it runs.
- Every row denormalizes the caller's identity (`userFirstName`/`userLastName`/`userEmail`) and, for actions, the `actionName` — captured **at write time**, not looked up live, so a renamed/deleted/reassigned account doesn't rewrite history.
- Nested JSON/array values are diffed structurally (only the changed leaf is recorded) and are safe for `BigInt` and BSON-like values. A key present on only one side of a diff is simply **omitted** from the other side rather than encoded with a sentinel (the original `ABSENT` marker embedded a NUL byte that Postgres's `::jsonb` cast rejects).
- A collection with no primary key is skipped at install time (logged) instead of throwing on its first write. A bulk update/delete's before-write snapshot is capped at 1000 records — exceeding it logs a warning under `critical: false`, or refuses the whole operation under `critical: true`.
- Writes initiated from inside a smart action (`context.collection.update`/`create`/`delete`) are audited too, under the same correlation key as the action, via the `internalHook` decorator layer — traced to already run before any user-registered hook in both customizer stacks, so the `prepend` ordering option this PR originally added turned out to be unnecessary and was removed.

### Smart-action capture

- Records the action invocation itself — the same pending/confirm split as CRUD writes — one row per targeted record (narrowed to the caller's authorized selection, resolved *before* `execute()` runs so a delete or a filter-matching change doesn't lose the association), or one row attached to no record for a global/select-all run. `actionName` records which action ran.
- `previousValues` holds the submitted form (redacted); `newValues` holds a safe summary of the action's answer (`type` plus whichever of `message`/`name`/`mimeType`/`method`/`url`/`path` apply) — `html`, `invalidated`, a Webhook's `headers`/`body`, and a File result's contents are dropped. A Webhook/Redirect's `url`/`path` is sanitized (userinfo, query, fragment stripped) since either can carry credentials or a signed token.
- `action_failed` covers both a thrown exception (no answer to summarize) and a resolved `Error`-type result (kept as the failure summary). Under `critical: false` (default), a failure resolving the audited record ids or recording the pending entry is logged and swallowed, same as before; under `critical: true`, the same failure refuses the action before `execute()` runs instead of letting it run under-audited.

### Storage & routes

- Persists to the caller-provided database via a built-in SQL sink (Sequelize-backed), auto-creating the audit schema/table through a single Umzug migration — squashed from the several this PR shipped incrementally, since the feature hadn't released yet — with a Postgres advisory lock so several agents booting at once apply it one after another. Migration state is tracked in a `{tableName}_migration` table; a same-process collision between one store's own table and another's derived tracking-table name is now caught with a clear, actionable error instead of the confusing generic one that surfaced before. No in-memory store is exposed in the public API — persistence is mandatory.
- `GET /_audit-trail/{collection}/:id` — per-record history, JSON:API paginated/sorted, with user/date-range/`fields`/**free-text `search`** filters. `search` matches the action name, the denormalized user identity, and the keys/scalar values of the recorded JSON at any depth (as a SQL `WHERE` clause, so it composes with pagination and `meta.count`) — never a redacted value, since the real one was never stored to find. `meta.availableUsers` returns the distinct authors matching the active filters, on the first fetch only.
- `GET /_audit-trail/{collection}/:id/state?at=…` — the record as it stood at a past instant, rebuilt by undoing everything recorded since (action rows are excluded from this replay, since their value objects hold a form/reply rather than column values; a `pending` row is excluded too, since its write may not have landed).
- `GET`/`POST /_audit-trail/correlation(s)` — every entry recorded under one (or several) correlation key(s) for a record.
- All routes assert read permission and are scope-aware: once a record is genuinely deleted, its history stays visible regardless of a caller's record-level scope (a scope can't be evaluated against a record that no longer exists) — but a `create`/`delete` row's captured values are withheld if they themselves would have failed that scope, and an `update` row's values (only a partial diff, which can't be scope-checked reliably) are withheld unconditionally in that situation. Only a still-existing, out-of-scope id is refused outright.
- `agentCapabilities.canUseAuditTrail` on `/_internal/capabilities` reports whether the audit trail is configured, so the frontend can gate the History tab on that instead of inferring it from the `x-forest-correlation-id` header, which is emitted on every response regardless.
- Adds `correlationIdMiddleware` and `getRequestId` so a stable UUID is memoized per request, reused by `parseCaller`, and echoed in the `x-forest-correlation-id` response header (exposed through CORS).

### What this PR does *not* do

- It does **not** add a `@forestadmin/plugin-audit-trail` package — the audit trail is agent-native here.
- The `/state` route's scope-withholding does not yet match the history route's: a genuinely-deleted, out-of-scope record's reconstructed value at a past instant is not currently redacted. Closing that consistently needs a product decision on what a single-value response should return in that case, tracked as a follow-up.

## Test plan

- [ ] `yarn workspace @forestadmin/agent test` — 1432 passing
- [ ] `yarn workspace @forestadmin/datasource-customizer test` — 852 passing
- [ ] `yarn workspace @forestadmin/agent lint` — 0 errors
- [ ] Manually configure an agent with `auditTrail: { connectionString: process.env.AUDIT_TRAIL_DATABASE_URL }` against a Postgres database; verify the schema/table are auto-created and that the `/_audit-trail` history, `/state`, and correlation routes all return the expected data.
- [ ] Trigger a smart action and confirm the audit row's `previousValues`/`newValues` hold the submitted form and the result summary, and that a Webhook/Redirect result's URL is sanitized.
- [ ] Verify `x-forest-correlation-id` round-trips through the CORS preflight.
- [ ] Manually trigger a create/update/delete/action with `auditTrail.critical: true` against a deliberately broken store (e.g. wrong connection) and confirm the operation is refused before any write occurs; repeat with `critical: false` (default) to confirm it proceeds unaudited.
- [ ] Manually exercise `search` against a MySQL-backed store to confirm the `~`-escaped `LIKE` clause runs without a syntax error.

> [!NOTE]
> ### Add built-in audit trail with SQL storage, collection hooks, and action capture to the agent
> - Installs collection-level `Before`/`After` hooks on create, update, and delete operations to capture JSON diffs (previous/new values) as `AuditRecord` entries, stored in a SQL table via Sequelize and managed by Umzug migrations.
> - Adds `captureAction` to record smart action invocations (including failures) per targeted record, with form-value redaction and safe result summarization that strips credentials, HTML, and request bodies.
> - Exposes new HTTP endpoints `GET /_audit-trail/<collection>/:id` (paginated history with filters) and `GET /_audit-trail/<collection>/:id/state` (record state at a given timestamp via `revertRecord`), plus correlation-key batch endpoints, all enforcing read authorization and scope-aware 404s.
> - Adds `correlationIdMiddleware` and `getRequestId` to stamp every request with a stable `x-forest-correlation-id` header exposed via CORS, used to group audit entries from the same request.
> - Introduces an `internalHook` decorator layer in the customizer stack so writes emitted from inside smart actions also traverse audit hooks.
> - Risk: Sequelize and Umzug are now production dependencies; the audit table migration runs automatically on agent start when `auditTrail.connectionString` is configured, requiring DB write access at startup.
>
> 
> #### Changes since #1686 opened
>
> - Implemented two-phase audit protocol with pending insertion before operations and confirmation after completion [1dda297]
> - Added denormalized user identity fields and action name to audit records and storage schema [1dda297]
> - Added critical mode option to control operation refusal when pending audit insertion fails [1dda297]
> - Removed ABSENT sentinel and changed diff encoding to omit absent keys from previous and next values [1dda297]
> - Added distinct user listing capability to audit store and history metadata [1dda297]
> - Updated audit storage schema with millisecond timestamp precision, nullable TEXT recordId, status column, and idempotent index management [1dda297]
> - Removed hook prepending functionality from collection customizer and hook decorator [1dda297]
> - Added logic to skip audit instrumentation for collections without primary keys with warning [1dda297]
> - Withheld `previousValues` on delete operation audit entries when the deleted record falls outside the caller's scope while the record is genuinely deleted [e912f53]
> - Added logging of truncation warnings when bulk update or delete operations match at least 1000 records in audit trail instrumentation [e912f53]
> - Refactored record visibility checking from returning a boolean to returning a structured `RecordVisibility` type containing `visible` and `goneEntirely` properties [e912f53]
> - Added `canUseAuditTrail` flag to the capabilities endpoint response indicating whether audit-trail routes are mounted [e912f53]
> - Added documentation clarifying bulk operation snapshot limits, delete row visibility behavior, and correlation header emission [e912f53]
> - Introduced critical mode for audit trail operations [12f4c2e]
> - Implemented out-of-scope value withholding in audit trail history responses [12f4c2e]
> - Excluded pending status entries from state-at-timestamp reconstruction [12f4c2e]
> - Fixed audit trail connection closure during agent startup failures [12f4c2e]
> - Modified audit trail migrations binary implementation [12f4c2e]
> - Added test coverage for critical mode, out-of-scope withholding, pending exclusion, and startup error handling [12f4c2e]
> - Added free-text search capability to audit trail history queries [f767720]
> - Implemented dialect-specific JSON column text casting utility [f767720]
> - Refactored history query WHERE clause builder to compose multiple filter conditions [f767720]
> - Added comprehensive test coverage for search functionality [f767720]
> - Updated documentation for audit trail search filter [f767720]
> - Changed the `searchCondition` function in the SQL audit trail store to use tilde as the LIKE escape character instead of backslash, updating the escape regex to prefix occurrences of '~', '%', and '_' with '~' and generating SQL with `ESCAPE '~'` and `LOWER(...) LIKE LOWER(...)` clauses to address MySQL/MariaDB string-literal escaping behavior with backslashes [a53882a]
> - Added and updated tests in the SQL audit trail store test suite to verify tilde-escaped pattern behavior, including validation that '~' characters are properly escaped in patterns and that no backslashes appear in generated MySQL SQL [a53882a]
> - Changed snapshot fetching in `instrumentCollection` Before:Update and Before:Delete hooks to request `MAX_SNAPSHOT_RECORDS + 1` records using `PaginatedFilter` and `Page` instances, detect truncation by checking if results exceed `MAX_SNAPSHOT_RECORDS`, and only enforce cap warnings or refusals when matches exceed the cap rather than when exactly at the cap [5737b87]
> - Clarified documentation in `AuditHistoryQuery.search` to explain that search matching operates on serialized JSON text, which affects how punctuation-only terms and escaped characters behave [5737b87]
> - Clarified semantics in audit trail README regarding the `critical` option and pending row interpretation [5737b87]
> - Modified `instrumentCollection` utility to confirm audit trail entries for update operations that result in no actual value changes [a9348f5]
> - Updated audit trail instrumentation tests to validate that no-op update operations produce confirmations with empty value diffs [a9348f5]
> - Wrapped `Recorder.confirm` method emit call in try/catch to log and swallow sink rejections instead of propagating errors [ef756ce]
> - Documented audit trail behavior for no-op write operations [94d1ccd]
> - Added documentation clarifying audit trail behavior when a record's writable primary key is renamed [1cb15e6]
> - Modified database migrations in the `agent` package to support audit trail functionality [77a33e6]
> - Implemented selection limit enforcement for smart actions in `ActionRoute.auditedRecordIds` method [9c973a0]
> - Exported `MAX_SNAPSHOT_RECORDS` constant from `@.../audit-trail` module [9c973a0]
> - Documented smart action selection limit behavior in audit trail README [9c973a0]
> - Resolved live-query segment context variables in `ActionRoute.auditedRecordIds` before enumerating authorized records for auditing [db55532]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 37ad243.</sup>
> 
>


bexchauveto:
Addressed in 444dfa681 — going through each finding:

- 🔴 **Critical** (`access/audit-trail.ts:41`, record-level scope leak): fixed. `handleHistory`/`handleStateAt` now require the target id to currently exist and match the caller's scope when one is configured; a scoped caller can no longer read the history/reconstructed state of an out-of-scope or deleted record.
- 🟠 **High** (`sql-store.ts:100`, mssql throws on `fields`): fixed, added a `JSON_PATH_EXISTS`-based branch (SQL Server 2022+).
- 🟠 **High** (`routes/index.ts:183`, correlation route shadowing): fixed, the correlation route now registers before the per-collection routes.
- 🟠 **High** (`sql-store.ts:26`, `recordId` VARCHAR(255) truncation): fixed, widened to TEXT via a new migration (003), index rebuilt with a length prefix so it stays valid on MySQL/MariaDB.
- 🟠 **High** (`sql-store.ts:160` + `migrations.ts:84`, migration table not scoped by `tableName`): fixed, migration state (table *and* Sequelize model name) is now scoped per `tableName`.
- 🟠 **High** (`migrations.ts:144`, Postgres migrations outside the advisory-lock transaction): fixed, the transaction is threaded through to every migration's DDL calls.
- 🟠 **High** (`migrations.ts:132`, non-Postgres concurrency guard): partially addressed — migrations 001-003 are now idempotent (existence-checked before creating), which closes most of the race window on a truly simultaneous boot. I did not implement a real distributed lock for sqlite/mysql/mssql (no live instance here to validate one against), so concurrent boot on those dialects is safer but not airtight the way Postgres's advisory lock is.
- 🟡 **Medium** (`options-validator.ts:93`, discarded `close()`): fixed, `close` is now part of the runtime `auditTrail` option and called from `agent.stop()`.
- 🟡 **Medium** (`revert.ts:11` + `revert.ts:60`, revert logic bugs): both fixed — an `ABSENT` sentinel now distinguishes "didn't exist" from a real `null` so a newly-added key/index is removed on revert instead of left as a bogus `null`, and `revertRecord` no longer stops permanently at the first create/delete (it restores that snapshot and keeps walking older entries).
- 🟡 **Medium** (`instrument.ts:183`, logging the requested patch instead of the persisted value): fixed, the After-Update hook re-reads the record and diffs against that.
- 🟡 **Medium** (`instrument.ts:126`, `snapshotsByFilter` WeakMap collision): fixed, snapshots are now queued (FIFO) per filter instead of a single slot.
- 🟡 **Medium** (`instrument.ts:16`, empty `recordId` on no-PK collections): fixed, now throws instead of silently merging every such collection's history under `recordId: ''`.
- 🟡 **Medium** (`instrument.ts:197`, recordId from the pre-update snapshot on a PK change): fixed as a side effect of the persisted-value re-read — the entry is now filed under the post-update id.
- 🟡 **Medium** (`sql-store.ts:80`, JSON path breaks on a dotted field name): fixed, the field is quoted as a literal path segment.
- 🟡 **Medium** (`correlation-id.ts:7`, trusting a pre-existing `context.state.requestId`): fixed, moved to a namespaced `forestRequestId` state key.
- 🟢 **Low** (`sql-store.ts:162`, `id` tiebreaker ignoring `order`): fixed.

Also rolled the qlty complexity/many-params flag on `emit()` into this pass (now takes one options object instead of 5 positional params). Left the `diff()` complexity flag alone — it's an inherent structural-diff recursion; splitting it further looked more likely to hurt readability than help.

1308/1308 tests pass, lint clean.

bexchauveto:
Addressed in 6a658b9a9:

- 🔴 **Critical** (`README.md:101`, correlation routes bypassing scope): fixed for real, not just documented — `AuditTrailCorrelationRoute` now applies the same record-level scope check as the per-collection route (extracted into a shared `audit-trail/scope.ts` used by both).
- 🟠 **High** (`agent.ts:127`, `stop()` hangs if `close()` rejects): fixed, wrapped in try/catch with a logged error so framework shutdown always proceeds.
- 🟡 **Medium** (`correlation-id.ts:19`, header missing on error responses): fixed, `correlationIdMiddleware` now sets the header in a `finally` block.
- 🟡 **Medium** (`revert.ts:14`, `ABSENT` sentinel collision risk): the sentinel now embeds a NUL byte, making collision with any real audited value effectively impossible.

Also bumped coverage per the qlty report — `migrations.ts` was the big gap (56.3% → ~96%): added tests for all three migrations' `down()` paths (via a newly-exported `buildUmzug`), `ensureSchema`'s schema-creation/tolerance branches, and `isSchemaAlreadyExistsError`. Filled in the smaller gaps too: `agent.ts`'s `stop()` error path, `options-validator`'s `auditTrail` branch, `revert.ts`'s remaining edge cases, `sql-store`'s retry-after-failed-init path, and the scope-check branches in both routes.

One honest gap left: the Postgres-specific advisory-lock transaction branch in `runAuditMigrations` (~7 lines) is still untested — I don't have a real Postgres instance here to validate a test against, and faking the whole Sequelize/Umzug/SequelizeStorage stack well enough to exercise it meaningfully looked more likely to produce a misleading test than a useful one.

1334/1334 tests pass, lint clean.

bexchauveto:
Addressed in 237314881:

- 🟠 **High** (`audit-trail.ts:97`, off-by-one on the state-at-timestamp boundary): fixed — entries timestamped exactly at the requested `at` are now filtered out before the revert walk, so they're kept rather than undone.
- 🟠 **High** (`agent.ts:451`, `generateSchemaOnly()` connecting to/migrating the audit DB): fixed — `buildNoCodeDataSource()` now takes an `installAuditTrail` flag, off for `generateSchemaOnly()`, so it no longer touches the audit database at all for a schema-only run.

1337/1337 tests pass, lint clean.

bexchauveto:
Addressed in 3442a18b9: reordered `stop()` so the audit-trail connection closes only after `super.stop()` (framework drain) resolves, instead of before — a create/update/delete still in flight at shutdown can now still reach its audit after-hook and append its entry, rather than failing against an already-closed store. 1338/1338 tests pass, lint clean.

bexchauveto:
Addressed in 1c45f2adc: `start()`'s catch block now closes the audit-trail connection (best-effort, logged on failure) before rethrowing the original startup error, so a failed start no longer leaks the pool opened by `installAuditTrailHooks` -> `store.init()` during `buildRouterAndSendSchema()`. 1340/1340 tests pass, lint clean.

packages/agent/src/audit-trail/scope.ts
```diff
@@ -0,0 +1,56 @@
...
+  collection: Collection,
+  packedId: string,
+  context: Context,
+): Promise<boolean> {
+  const scope = await services.authorization.getScope(collection, context);
+
+  if (!scope) return true;
+  if (await recordExists(collection, packedId, context, scope)) return true;
+
+  return !(await recordExists(collection, packedId, context, null));
```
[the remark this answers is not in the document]
bexchauveto: This is intentional, not a bug — please do not apply the suggested `return false`.

The fail-open-on-delete policy is deliberate: once a record scope can no longer be evaluated (the record is genuinely gone), showing its history/state to anyone who can read the collection at all is the point of an audit trail, matching this feature's Ruby implementation. Only a still-existing, out-of-scope record is refused — see the comment directly above this line and the audit-trail README's "HTTP routes" section for the full rationale. Reverting to `return false` would silently undo that behavior.

packages/agent/src/audit-trail/action-capture.ts
```diff
@@ -0,0 +1,62 @@
...
+// values, and whether it raised. Which action it was lives in the Forest activity logs; `correlationKey`
+// is the join.
+export default async function captureAction(
+  sink: AuditSink,
+  redactedFields: string[],
+  params: ActionCaptureParams,
+): Promise<void> {
+  const timestamp = new Date().toISOString();
+  const newValues = redactValues(params.formValues ?? {}, redactedFields);
+  const recordIds = params.recordIds.length > 0 ? params.recordIds : [NO_RECORD];
```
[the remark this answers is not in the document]
bexchauveto: Confirmed and left as-is for now: this matches Ruby's identical design (same `NO_RECORD` sentinel, same per-record-scoped routes), so it's an inherited characteristic rather than a Node-specific regression. These rows remain available via direct SQL/compliance access. Adding a collection+correlationKey browse route (without a record id) would need to land on both Node and Ruby together — tracked separately, not blocking this PR.

bexchauveto:
Addressed the latest round of Macroscope findings (verified each against the code rather than applying suggestions blindly):

- **Fixed**: an update that changed a field its own filter matched on (including a writable primary key) silently disappeared from the audit trail — the After-hook now re-reads by the record's own identity instead of the stale filter.
- **Fixed**: a failed update/delete could leave its snapshot to be wrongly picked up by a later, unrelated operation reusing the same `Filter` object — Update now correlates by patch content instead of queue position (Delete has no equivalent secondary key and keeps a narrower, documented limitation).
- **Fixed**: an id excluded by the caller's scope still got an `action` audit entry for an action that never touched it — explicit selections are now narrowed to the authorized subset.
- **Fixed**: a resolved `{ type: 'Error' }` action result (HTTP 400) was only audited as a successful invocation — now counted as `action_failed` too (mirrored on the Ruby side).
- **Docs**: noted that `auditTrail`'s connection string needs its own Sequelize driver installed (e.g. `pg`/`pg-hstore`), since `@forestadmin/agent` doesn't bundle one.
- **Rejected** (Critical, `scope.ts:55`): the suggested fix would revert this session's deliberate fail-open-on-delete policy — replied on the thread with the rationale, no code change.
- **Deferred**: global/select-all action rows (`recordId: ''`) aren't independently browsable via the HTTP API — this matches Ruby's existing design exactly, not a Node-specific gap. Left as-is; would need a joint Node+Ruby change to add.

1367/1367 tests passing, lint clean.

bexchauveto:
Addressed the two new Macroscope findings from this round:

- **Fixed** (High): `deepEquals`/`JSON.stringify` threw `TypeError: Do not know how to serialize a BigInt` when an update touched a writable bigint column — the write had already committed, so the After hook crashing lost the audit entry. Values are now normalized to strings before comparison and before being stored (a raw BigInt would have also failed later at the SQL-storage or HTTP-response serialization layer, not just the comparison).
- **Fixed** (Medium): `auditedRecordIds` re-queried the caller's authorized selection *after* `execute()` had already resolved — an action that deleted its targets (or moved them out of their own selection filter) found nothing on that re-query, losing the id/audit-entry association. Ids are now resolved before `execute()` runs.

1370/1370 tests passing, lint clean.

bexchauveto:
Fixed: \`toJsonSafe\` only checked whether the value being stored *was itself* a BigInt, missing a BigInt nested inside a JSON column's object/array value. Two real gaps:

- \`pickColumns\` (Create/Delete) hands it a whole column value that never goes through \`diff\`'s own recursion.
- \`diff\` itself keeps a primitive array "whole" as a single leaf rather than walking its elements, so a BigInt array element also slipped through.

\`toJsonSafe\` now recurses through arrays and plain objects (excluding \`Date\`, matching \`isPlainObject\`'s existing exclusion) so a nested BigInt is caught wherever it appears. Added regression tests for a BigInt nested in a JSON object, in a primitive array, and on create.

1373/1373 tests passing, lint clean.

bexchauveto:
Fixed: the previous recursion walked \`Object.entries\`/\`Object.fromEntries\` on any non-Date object, including BSON-like values. That's fine when \`toJSON\` is an own property (the existing test double), but a real BSON wrapper typically defines \`toJSON\` on its prototype with the actual data in unrelated (and sometimes non-enumerable) own fields — verified empirically that reconstructing such a value drops the prototype method entirely, e.g. \`{ toJSON: () => 'abc' }\`-shaped data backed by a private field serializes as \`{"hex":"abc"}\` instead of \`"abc"\`.

Now detects a callable \`toJSON\` (own or inherited) and leaves that value untouched — the eventual JSON serialization already invokes it correctly, same as \`deepEquals\` already relies on elsewhere in this file. Added a regression test using a class with a prototype \`toJSON\` and a private field, confirmed it fails without the fix (reproduced the exact corruption) and passes with it.

1374/1374 tests passing, lint clean.

bexchauveto:
Fixed both new findings:

- **Fixed** (High): the Webhook result's raw `url` was persisted verbatim into `newValues`, which can carry credentials (userinfo) or a signed/one-time token in the query string — permanently written to the audit database. Now strips userinfo, query, and fragment, keeping only origin+path; falls back to stripping at the first `?`/`#` for a relative or otherwise unparseable URL.
- **Fixed** (High): resolving the caller's authorized record ids (added last round, ahead of `execute()`) was awaited outside any try/catch — a transient failure there (e.g. a dropped connection) escaped uncaught and blocked the action from running at all, turning an otherwise-executable action into a failed request just because audit trail was enabled. Now best-effort: logged and falls back to no audited ids, matching the same principle already applied to `captureAction`'s own failures.

Both verified with a regression test that fails without the fix. 1385/1385 tests passing, lint clean.

bexchauveto:
Fixed both:

- **Fixed** (High): `sanitizeUrl`'s fallback (used when `new URL` rejects an otherwise-mostly-valid target, e.g. a space in the host) only stripped query/fragment, leaving real credentials sitting in the userinfo — verified empirically that `https://user:pass@exa mple.com/...` throws in `new URL` while the credentials remain genuine. Now strips userinfo in the fallback path too.
- **Fixed** (High): a Redirect result's `path` was never sanitized at all, despite being equally free-form action-supplied data that can carry a signed/one-time token in its own query string (e.g. `/callback?token=secret`). Now goes through the same `sanitizeUrl`.

Both verified with a regression test confirmed to fail without the fix. 1387/1387 tests passing, lint clean.

PMerlet (commented):
Adversarial review pass, read together with agent-ruby#320 and the front PR. Most of the comments below are seams between the three rather than problems visible from this diff alone.

For context, the architecture we settled on while reviewing:

1. **Capability, not feature flag** — the agent advertises `canUseAuditTrail` on `POST /_internal/capabilities`; the History tab exists only when the customer's agent has the trail configured.
2. **Two tabs, two sources, no cross-hydration** — Activity keeps reading activity logs, History reads the audit trail and nothing else.
3. **The audit row is self-sufficient** — it must carry the actor identity and the action name, because the activity log is written client-side after the commit, in one unretried attempt, with failures swallowed. It is structurally less complete than the trail.
4. **Failure policy, identical in both agents: pending before, confirm after** — the row is written in a `pending` state *before* the write and confirmed after. A failed pre-write insert refuses the operation without having touched the data. No compensating write, ever.
5. **Pagination and filters belong to `GET /_audit-trail/{collection}/:id`** — the per-record route this PR already implements, which the front should call instead of `/correlations`.

Six things I'd want settled before this lands, in rough order of consequence: the `ABSENT` sentinel (NUL bytes vs PostgreSQL), the failure policy, the unbounded before-snapshot, `record_id` width parity with ruby, the missing action name and actor, and the capability gating on the front side.

The code itself is careful and the comments anticipate most of the obvious objections — none of what follows is about sloppiness.

packages/agent/src/audit-trail/types.ts
```diff
@@ -0,0 +1,87 @@
+export type AuditOperation = 'create' | 'update' | 'delete' | 'action' | 'action_failed';
+
+// A sentinel distinguishable from a genuine `null` value: `diff` (instrument.ts) uses it to mark a
+// changed leaf that did not exist before/after, so `revertRecord` (revert.ts) can remove the
+// key/index on revert instead of writing back a bogus `null`. Must be JSON-safe (unlike a `Symbol`
+// or `undefined`) since previous/new values round-trip through a SQL JSON column. The embedded NUL
+// (written as an explicit escape, not a raw byte) makes collision with a real audited value
+// effectively impossible: legitimate text/JSON submitted through an HTTP API essentially never
+// contains a raw control character.
+export const ABSENT = '\u0000forest-audit-trail-absent-field\u0000';
```
PMerlet: **Blocker — the sentinel embeds NUL bytes, which PostgreSQL refuses to convert to text.**

The comment argues collision with real data is impossible, and that's right — but this picked the one character PostgreSQL treats specially. `jsonb` rejects `\u0000` outright; `json` accepts it on input and throws on *any* conversion to text, including the `::jsonb` cast in `fieldsChangedCondition`. The column is `DataTypes.JSON` → PG `json`, so the insert lands and then the `?fields=` filter fails for the whole query, not just the offending row:

```
psql -c "select '{\"a\":\"\u0000x\"}'::json::jsonb"
-- ERROR: unsupported Unicode escape sequence
--   DETAIL: \u0000 cannot be converted to text.
```

Triggered by any update that adds or removes a key inside a JSON column, so not an exotic path. It also reaches the front, which knows nothing about the sentinel and renders it through `formatValue` as a raw control-character string.

agent-ruby#320 solved the same problem without a sentinel in the database: `Diff::ABSENT = Object.new`, and the key is simply left out of that side's hash. "Key added" and "key holding null" stay distinguishable, nothing special round-trips through JSON, and it's the shape the front already handles. Worth aligning on that.

I could not run PostgreSQL locally, so this rests on documented `json`/`jsonb` behaviour — worth confirming with the one-liner above before acting on it.

packages/agent/src/audit-trail/instrument.ts
```diff
@@ -0,0 +1,349 @@
...
+  const primaryKeys = SchemaUtils.getPrimaryKeys(schema);
+  // PKs may be read-only, but they are required to build the packed recordId; the diff itself
+  // stays restricted to writableColumns.
+  const readProjection = [...new Set([...primaryKeys, ...writableColumns])];
+  const { name } = collection;
+
+  const emit = (
+    caller: Caller,
+    entry: Pick<AuditRecord, 'operation' | 'recordId' | 'previousValues' | 'newValues'>,
+  ): Promise<void> | void =>
```
PMerlet: **Blocker — a failed audit write fails the user's request, after the write committed.**

`emit()` is awaited inside the `After` hooks and nothing catches it. If the store is unreachable the record has already been created/updated/deleted and the operator gets a 500 — fail-closed at the one moment where it's useless, since there is nothing left to prevent.

Worth noting this is the opposite of what the action path in this same PR does, with reasoning I agree with: *"a broken audit store must not also fail the request, which would report a successful run as an error and could prompt a client retry that reruns the action."* The CRUD path doesn't get that treatment, and the PR description ("a failure … writing the entry itself, is logged and swallowed") only holds for actions. agent-ruby#320 makes the opposite, documented choice everywhere via `audit_safely`.

Same shape, sharper trigger: `toPackedRecordId` throws `'Cannot audit a collection with no primary key'`, and hooks are installed on every collection unconditionally — so enabling `auditTrail` turns every write to a PK-less collection into a post-commit 500. Better detected at install time and skipped with a log line.

The policy we landed on is **pending before, confirm after**: write the row in a `pending` state before the write, confirm it after. A failed pre-write insert then refuses the operation without having touched the data, so there is nothing to repair. Three things it needs: a status column (a new migration here), a decision for `create` — the record id doesn't exist yet, so the pending row carries none and is completed after the write — and an honest line in the README about what the invariant buys: *no unaudited write*, not *every row holds exact after-values*. There's already a precedent for the pattern in this monorepo: `packages/mcp-server` does `createPendingActivityLog` then `updateActivityLogStatus`.

packages/agent/src/audit-trail/instrument.ts
```diff
@@ -0,0 +1,349 @@
...
+            previousValues: {},
+            newValues: pickColumns(record, writableColumns),
+          }),
+        ),
+      );
+    },
+    { prepend: true },
+  );
+
+  collection.addInternalHook('Before', 'Update', async (context: HookBeforeUpdateContext) => {
```
PMerlet, line 417: **Blocker — the before-snapshot lists every matched record, unbounded.**

This `list(context.filter, readProjection)` has no limit, and the result is held until the `After` hook. A "delete all" (same pattern in the Before/Delete hook below) materialises every matched row in the agent process before the write runs. The After hook then re-lists the whole set a second time and issues one `model.create()` per record through `Promise.all` with no concurrency bound.

What makes this worth a blocker rather than a note: it silently changes the cost of a write path customers already use, behind a config flag they may well enable in production without re-testing bulk operations. And under the pending-then-confirm policy the same volume gets written twice.

Suggestion: batch the inserts (`bulkCreate`), bound the snapshot, and make truncation explicit and logged — "N records audited, M skipped" beats an OOM, and beats silence.

packages/agent/src/audit-trail/action-capture.ts
```diff
@@ -0,0 +1,104 @@
...
+  } catch {
+    return url.split(/[?#]/)[0].replace(/^([a-z][a-z\d+.-]*:\/\/)[^/]*@/i, '$1');
+  }
+};
+
+// Only a safe summary of the answer is kept: `html` (Success/Error), `invalidated` (Success),
+// `headers`/`body` (Webhook, which routinely carry credentials) and a File result's `stream` (file
+// bytes don't belong in an audit table) are all dropped. Absent when `execute()` threw rather than
+// resolving — there is no answer to summarize.
+const summarizeActionResult = (result?: ActionResult): Record<string, unknown> => {
```
PMerlet, line 48: **Contract addition — record the action name here.**

The PR deliberately leaves it out and defers to the activity log ("Which action ran is not recorded here either; the Forest activity log has that, joined by `correlation_key`"). That reasoning inverts once History reads the audit trail as its only source, because the log can be missing: it is written client-side, after the commit, in a single unretried `activity.save()` whose failure is swallowed (`trySaveActivityLog` on the front). Closed tab, dropped connection, expired session — no log, and then an `action` row renders as "Action — user 7" with the name nowhere to be found.

`this.actionName` is in scope at the `captureAction` call site in `routes/modification/action/action.ts`, so this is cheap.

Same reasoning applies to the actor: `userId` alone can't render a name, and resolving ids at read time is the wrong shape for an audit record — a user can be renamed or deleted later, and the row should say who acted *then*. Denormalising name/email onto the row also drops any read-time dependency on the Forest backend.

On the sanitising just above: `sanitizeUrl` is the right call and agent-ruby#320 is currently missing it — flagged there.

packages/agent/src/audit-trail/migrations.ts
```diff
@@ -0,0 +1,259 @@
...
+// an existing one (already-applied migrations are skipped).
+const migrations = [
+  {
+    name: '001-create-audit-logs',
+    up: async ({ context }: { context: MigrationContext }) => {
+      const table = { tableName: context.tableName, schema: context.schema };
+
+      // Idempotent: lets a process that loses a concurrent-boot race retry cleanly instead of
+      // failing on a "relation already exists" error from the winner's DDL.
+      if (await context.queryInterface.tableExists(table, { transaction: context.transaction })) {
```
PMerlet: **The existence check silently adopts a pre-existing table.**

Idempotency here is right for concurrent boots, and the comment says so. The side effect is that if the customer already has a table by this name, the migration no-ops and the store starts writing into *theirs*. Outside PostgreSQL there is no schema namespace at all (`DIALECTS_WITH_SCHEMAS` is `postgres`/`mssql` only), so on MySQL the default target is a bare `audit_logs` — and the README explicitly invites pointing the connection string at "the database already used by the agent".

Suggestion: after the existence check, verify the columns match what the model expects, and fail loudly naming the collision and the `tableName` option, rather than adopting whatever is there.

Separately, on the `timestamp` column below: `DataTypes.DATE` maps to MySQL `DATETIME` with *zero* fractional-second precision, so milliseconds are silently dropped. That defeats the inclusive `:59.999` end boundary, multiplies ordering ties, and undermines the `entry.timestamp !== at` string comparison in `handleStateAt` (which agent-ruby does in SQL as `timestamp > ?` — the right layer). `DATE(3)` or higher in both the migration and the model.

packages/agent/src/audit-trail/scope.ts
```diff
@@ -0,0 +1,56 @@
...
+  return records.length > 0;
+}
+
+// A record-level scope can't be evaluated against a record that no longer exists, so a caller whose
+// access is scoped down is only denied when the id currently exists and fails that scope — once it's
+// genuinely gone there is nothing left to scope against, and showing what it was (including its
+// delete snapshot) is much of the point of an audit trail. Anyone who can read the collection at all
+// can therefore see the history of a deleted record, scope aside; only a still-existing, out-of-scope
+// record is refused.
+export default async function isRecordVisible(
```
PMerlet: **Worth an explicit product sign-off rather than only a code comment.**

The rule — a record-level scope is enforced only while the id still exists — is reasoned carefully and the comment is honest about the trade-off. The adversarial reading: record ids are guessable, so this is an enumerable endpoint that returns the full prior contents of deleted records (the `delete` row holds every writable column) to users who were never allowed to see them while they existed. For a role scoped to one team's records, deleting a record *widens* who can read it.

This gets more salient now that History becomes the primary tab for customers who enable the trail. If the rule stands, one middle option: withhold the `delete` row's `previousValues` from out-of-scope callers while still surfacing that a deletion happened, by whom and when — which is the part the comment argues is "much of the point".

Either way I'd want it in the ticket and the customer-facing docs, not only here. Same rule and same question in agent-ruby#320.

packages/datasource-customizer/src/decorators/hook/hook.ts
```diff
@@ -19,9 +19,15 @@ export default class Hooks<B extends HookContext, A extends HookContext> {
     }
   }
 
-  addHandler(position: HookPosition, handler: HookHandler<A> | HookHandler<B>) {
+  // `prepend`: `execute{Before,After}` stops at the first exception, so a handler that must not be
+  // preempted by another one throwing (e.g. the audit trail's after-hooks, which record a write
+  // that already happened) can ask to run first instead of being appended.
+  addHandler(position: HookPosition, handler: HookHandler<A> | HookHandler<B>, prepend = false) {
```
PMerlet: **This doesn't do what the PR description says it does, and it adds public API for it.**

The stated purpose is that the audit after-hooks "run first and can't be preempted by another hook throwing". But the audit hooks are registered on `internalHook`, a *separate, inner* decorator: in `decorators-stack.ts`, `internalHook` is created before `hook`, so `hook` ends up outermost and user after-hooks already run **after** the audit ones. And nothing except the audit trail registers on `internalHook`, so prepending among its own handlers orders a list of one.

So the guarantee is already provided by the stack's construction. The cost is signature churn across six files plus two new public methods on `CollectionCustomizer` — `addHook(…, {prepend})` and `addInternalHook`, the latter exposing an internal decorator layer to userland.

Suggestion: drop `prepend` here and keep `addInternalHook` internal. Note that agent-ruby#320's `prepend:` **is** load-bearing — one `hook` decorator shared with user hooks — so the two agents legitimately differ on this one; worth a comment saying so rather than mirroring for symmetry.

packages/agent/src/utils/correlation-id.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function getRequestId(context: Context): string {
+  if (!context.state[STATE_KEY]) {
+    context.state[STATE_KEY] = uuidv4();
+  }
+
+  return context.state[STATE_KEY] as string;
+}
+
+export async function correlationIdMiddleware(context: Context, next: Next): Promise<void> {
```
PMerlet, line 18: **Mounted unconditionally while the routes are conditional — which breaks the front's only way to detect the feature.**

`router.use(correlationIdMiddleware)` and the CORS `exposeHeaders` entry in `agent.ts` are unconditional, while `getAuditTrailRoutes()` returns `[]` when `options.auditTrail` is null. So an agent on this version *without* the trail configured still emits `x-forest-correlation-id`.

The front takes that as evidence the feature exists: the adapter captures the header, the activity log persists a `correlationId`, the MFE sees a non-empty key list, calls `/_audit-trail/correlations`, gets a 404, and renders a full-page error *instead of* the timeline. Records whose logs predate the upgrade have no key and render fine, which makes the failure look intermittent rather than structural.

The fix we agreed on is on the other side — advertise `canUseAuditTrail` on `/_internal/capabilities` and gate the tab on it — so this middleware can stay unconditional. Flagging it here because the coupling isn't visible from either PR alone.

Minor, same file: the header is set in a `finally`, and Koa's default `onerror` strips response headers before writing an error, so error responses never carry it. That mattered for joining `action_failed` rows; it stops mattering once the front no longer needs the key, but the header is still useful for support and tracing.

packages/agent/src/audit-trail/types.ts
bexchauveto: Fixed — dropped the sentinel entirely. `diff()` now returns `{ previous, next, previousAbsent, nextAbsent }`, and every caller (`diff`'s own recursive assembly, `changedValues`, `revert.ts`) omits the key/index on whichever side is marked absent instead of writing a placeholder. `revert.ts` walks the union of both sides' keys so an added key (present only in `next`) still gets deleted on revert, and a removed key (present only in `previous`) falls through to the existing leaf branch and gets restored. Verified `select '{"a":"^@x"}'::json::jsonb` throws as you described before committing to this — thanks for the reproducible one-liner.

packages/agent/src/audit-trail/instrument.ts
bexchauveto: Fixed with the pending/confirm protocol you described. A `pending` row is now inserted before the write runs (identity, before-values, `record_id` where it exists yet) and confirmed to `done` with the final values after — including for `create`, via a new Before-Create hook keyed off `context.data`'s array reference. Added a `critical` option: `false` (default) logs and swallows a failed pending insert, matching today's behavior; `true` refuses the operation before anything runs. Also added the status column via migration, and the PK-less-collection case — `installAuditTrailHooks` now checks `SchemaUtils.getPrimaryKeys(schema).length === 0` and skips instrumenting with a Warn log line, instead of installing hooks that throw on first write. README now states the invariant your way: no unaudited write, not every row holding exact after-values.

packages/agent/src/audit-trail/instrument.ts
bexchauveto: Fixed — the snapshot read is now capped at 1000 records (`MAX_SNAPSHOT_RECORDS`), and hitting the cap logs an explicit warning naming the collection and operation rather than silently auditing only part of the write. Pending inserts for a bulk Create/Update/Delete now batch through `bulkCreate` (verified cross-dialect: Postgres/MSSQL return ids natively, MySQL/MariaDB's inferred-sequential-id approach is safe under the default autoincrement lock mode, sqlite returns them directly — checked empirically). Confirm calls are similarly batched via `Promise.all` over the correlated pending ids rather than one-by-one with no bound.

packages/agent/src/audit-trail/action-capture.ts
bexchauveto: Fixed — every audit row now carries `actionName` (action/action_failed rows only) plus denormalized `userFirstName`/`userLastName`/`userEmail`, captured from the caller at write time rather than resolved live, for exactly the reason you gave: a renamed/deleted/reassigned account shouldn't rewrite what the row says. `captureActionPending`/`captureActionConfirm` (split from the old single-shot `captureAction` to fit the pending/confirm protocol) thread `actionName` through from `action.ts`'s `this.actionName`, already in scope at the call site as you noted.

packages/agent/src/audit-trail/migrations.ts
bexchauveto: Fixed both. Existence check now calls `assertOwnsTable`, which verifies the pre-existing table has every column the migration creates before treating it as ours, throwing a clear error naming the collision and the `tableName` escape hatch rather than silently adopting it. (Side note: the 8 incremental migrations from this PR ended up squashed into one before merge, since the feature hadn't shipped yet — so `assertOwnsTable` now checks the full column set rather than an original-migration subset, but the fix is the same shape you suggested.) `timestamp` is now `DATE(3)` in both the migration and the model.

packages/agent/src/audit-trail/scope.ts
bexchauveto: Implemented the middle option — didn't get a separate product sign-off, flagging that below. `isRecordVisible` (renamed `checkRecordVisibility`) now also reports `goneEntirely`; the per-record history route re-evaluates the caller's scope `ConditionTree` (via `.match()`) against a `delete` row's `previousValues` in that case, and withholds it (`{}`) when it would have failed the scope — while the row itself (that a deletion happened, by whom and when) stays visible.

This is scoped to the history route only, matching your suggestion literally. It does **not** extend to `/state`'s reconstructed value for the same genuinely-deleted-and-would-be-out-of-scope case — that route returns a single reconstructed record rather than a list of rows with an envelope to preserve, so the equivalent fix isn't a drop-in, and I didn't want to invent a new response shape (e.g. return `null` with 200 instead of the value) without the sign-off you asked for. Flagging that gap explicitly rather than presenting this as fully closed — worth a ticket.

packages/agent/src/utils/correlation-id.ts
bexchauveto: Fixed with the approach you suggested — kept the middleware unconditional and instead added `agentCapabilities.canUseAuditTrail` to `POST /_internal/capabilities` (true iff `options.auditTrail` is configured). Frontend can gate the History tab on that instead of inferring the feature from the header's presence. README now says explicitly that the header is emitted regardless of whether the trail is enabled.

packages/datasource-customizer/src/decorators/hook/hook.ts
bexchauveto: Agreed and fixed — traced the same thing you did: `internalHook` is built before `hook` in both `decorators-stack.ts` and `decorators-stack-no-code.ts` (the NoCode stack `installAuditTrailHooks` actually runs against has no separate `hook` layer at all), so the guarantee was already free. Dropped `prepend` entirely — `addHandler`/`addHook`/`addInternalHook` all lost the param, `addInternalHook` stays internal (not on `CollectionCustomizer`'s public surface), and added a comment next to `internalHook`'s construction in `decorators-stack.ts` explaining the ordering guarantee so a future reader doesn't reintroduce `prepend` for the same reason. Kept your note about agent-ruby's `prepend:` being genuinely load-bearing there (shared hook decorator) — the two agents differ on this one intentionally.

packages/agent/src/audit-trail/migrations.ts
```diff
@@ -0,0 +1,258 @@
...
+
+// Arbitrary but stable pair, only needs to be unique across the advisory locks the application
+// takes — the audit-trail migration is the only critical section we hold this lock for.
+const ADVISORY_LOCK: readonly [number, number] = [0x464f, 0x5254];
+
+// Scoped by `tableName` (not a single fixed name) so two stores configured with different table
+// names in the same schema don't share migration state — otherwise Umzug would see the second
+// store's migrations as already applied and never create its table.
+function migrationsTableName(tableName: string): string {
+  return `${tableName}_migration`;
```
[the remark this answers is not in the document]
bexchauveto: Fixed — this was the same collision you described. Added a per-process registry (`claimTableName` in migrations.ts) that claims both a store's own table name and its derived `{tableName}_migration` tracking name at `buildUmzug()` time, throwing a clear error naming the actual collision instead of leaving it to the confusing `assertOwnsTable` abort deep inside migration 001. Went with the same-process registry over full validation since cross-process collisions (two separate agent deployments sharing a database) can't be detected without coordination the agent doesn't have — `assertOwnsTable` still catches those, just less clearly; noted as a residual limitation in the commit.

packages/agent/src/agent.ts
```diff
@@ -110,6 +112,20 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
     } catch (error) {
       const { message } = error as Error;
       this.options.logger('Error', `Forest Admin agent startup failure: ${message}`);
+
+      // buildRouterAndSendSchema() may already have opened the audit-trail connection even
+      // though a later startup step failed; close it so a failed start doesn't leak the pool
+      // (e.g. across restart retries).
+      try {
+        await this.options.auditTrail?.close();
```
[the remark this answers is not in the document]
bexchauveto: Fixed — tracked whether `mount()` has succeeded and only close the audit-trail connection in the catch block when the failure happened before it. Once mounted, the host framework is already serving requests against that connection, so closing it there (because a later step, the embedded executor, failed) would break those requests instead of freeing an unused pool.

packages/agent/src/routes/modification/action/action.ts
```diff
@@ -143,6 +147,112 @@ export default class ActionRoute extends CollectionRoute {
...
+    try {
+      return await this.auditedRecordIds(context, caller, filter);
+    } catch (error) {
+      this.options.logger(
+        'Error',
+        `[ForestAdmin] Unable to resolve audited record ids, continuing: ${error}`,
+      );
+
+      return [];
+    }
```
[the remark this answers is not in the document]
bexchauveto: Fixed exactly as suggested — `resolveAuditedRecordIds` now rethrows when `this.options.auditTrail.critical` is set, refusing the action before `execute()` runs instead of falling back to a no-record entry. `critical: false` keeps the existing swallow-and-continue.

packages/agent/src/routes/access/audit-trail.ts
```diff
@@ -0,0 +1,334 @@
...
+
+  private withholdOutOfScopeDeletes(
+    entries: AuditRecord[],
+    scope: ConditionTree,
+    context: Context,
+  ): AuditRecord[] {
+    const { timezone } = QueryStringParser.parseCaller(context, { defaultTimezone: 'UTC' });
+
+    return entries.map(entry => {
+      if (entry.operation !== 'delete') return entry;
```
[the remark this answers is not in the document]
bexchauveto: Good catch — you're right, my delete-only fix left this open. `create`'s newValues and `delete`'s previousValues both capture every writable column (same `pickColumns` call in instrument.ts), so both are now checked against the scope the same way. `update`'s previousValues/newValues are only a partial diff of the changed columns, so a scope condition referencing an untouched column can't be evaluated reliably against it — withheld unconditionally there rather than risk a false negative that would've reproduced the exact leak you found, just for update rows instead of delete.

packages/agent/src/routes/access/audit-trail.ts
```diff
@@ -0,0 +1,334 @@
...
+
+    // `startTimestamp` is an inclusive lower bound, so an entry timestamped exactly `at` comes
+    // back too — but the record already reflects that entry's change at instant `at`, so it must
+    // be kept rather than reverted (which would wrongly return the state just *before* it).
+    //
+    // Action rows are excluded outright: their previousValues/newValues hold a submitted form and a
+    // summary of the action's answer, not column values, so replaying them as a diff would corrupt
+    // the reconstructed record. An action's effect on the data still shows up through the
+    // create/update/delete rows it produced under the same correlation key.
+    const entries = fetched.filter(
```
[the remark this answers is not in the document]
bexchauveto: Fixed — added `entry.status !== 'pending'` to the existing filter alongside action/action_failed.

packages/agent/src/audit-trail/README.md
```diff
@@ -0,0 +1,312 @@
...
+
+That's what the `critical` option controls:
+
+- **`critical: false`** (default) — a failure inserting the pending row is logged and swallowed; the
+  write proceeds exactly as it does today, just without an audit entry for it.
+- **`critical: true`** — the same failure instead refuses the operation: nothing is written, no
+  compensating write is issued.
+
+What this buys is **no unaudited write, not every row holding exact after-values**. A row left
+`pending` means the write may or may not have landed — the confirm step that would have flipped it
```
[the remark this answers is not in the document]
bexchauveto: Went with making the guarantee actually true rather than just softening the wording — hitting the cap now throws under `critical: true`, refusing the whole bulk operation instead of letting it proceed partially audited. `critical: false` keeps today's behavior (logs and proceeds with only the first 1000 audited), and the README now states that distinction explicitly instead of the blanket claim.

packages/agent/src/audit-trail/sql-store.ts
```diff
@@ -0,0 +1,394 @@
...
+    'user_first_name',
+    'user_last_name',
+    'user_email',
+    jsonColumnAsText(sequelize, 'previous_values'),
+    jsonColumnAsText(sequelize, 'new_values'),
+  ];
+
+  return Sequelize.literal(
+    `(${columns
+      .map(column => `LOWER(${column}) LIKE LOWER(${pattern}) ESCAPE '\\'`)
```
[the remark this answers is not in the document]
bexchauveto: Fixed exactly as suggested — you're right, `ESCAPE '\\'` is itself malformed on MySQL/MariaDB under the default sql_mode, since backslash escapes the closing quote there instead of terminating the literal. Switched to `~` for both the wildcard-escaping and the ESCAPE clause — it has no special meaning in any of the four dialects' string literals, so it sidesteps the whole backslash-escaping interaction. Added a regression test asserting the produced condition never contains a bare backslash.

packages/agent/src/audit-trail/instrument.ts
```diff
@@ -0,0 +1,524 @@
...
+    ...(filter as Record<string, unknown>),
+    page: { skip: 0, limit: MAX_SNAPSHOT_RECORDS },
+  });
+
+  // Hitting the cap means the operation may have matched more records than were actually audited.
+  // Under `critical: false`, silently dropping the excess would look identical to "everything got
+  // audited" — logged instead. Under `critical: true`, that gap is exactly what the option exists
+  // to refuse: no unaudited write, so the whole operation is refused rather than let through
+  // partially covered.
+  const enforceSnapshotCap = (operation: 'update' | 'delete', before: RecordData[]): void => {
```
PMerlet: **Off-by-one: an operation matching *exactly* `MAX_SNAPSHOT_RECORDS` is treated as truncated.**

`withLimit` asks the datasource for `limit: MAX_SNAPSHOT_RECORDS`, so `before.length` saturates at 1000 whether the filter matched 1000 records or 100 000. The check below then fires on `>= 1000`, which means a bulk update or delete touching exactly 1000 records is reported as "matched at least 1000 — only the first 1000 can be audited" even though every one of them *was* audited. Under `critical: true` that operation is **refused**.

agent-ruby#320 handles the same case correctly, and the idiom is worth copying:

```ruby
filter.override(page: Page.new(offset: 0, limit: cap + 1))
…
return records.map { … } if records.size <= cap
```

Fetching `cap + 1` is what makes "exactly cap" and "more than cap" distinguishable. Same change here: request `MAX_SNAPSHOT_RECORDS + 1`, keep the first `MAX_SNAPSHOT_RECORDS`, and treat only `length > MAX_SNAPSHOT_RECORDS` as overflow.

packages/agent/src/audit-trail/instrument.ts
```diff
@@ -0,0 +1,524 @@
...
+    userLastName: caller.lastName,
+    userEmail: caller.email,
+    actionName: null,
+    correlationKey: caller.requestId,
+    ...entry,
+    previousValues: redactValues(entry.previousValues, redactedFields),
+    newValues: redactValues(entry.newValues, redactedFields),
+  });
+
+  const withLimit = (filter: unknown) => ({
```
PMerlet: **This spreads a `Filter` instance into a plain object, losing its prototype.**

`{ ...(filter as Record<string, unknown>), page: … }` copies own enumerable properties only, so the result is a bare object where `context.filter` was a `Filter`. It's then cast `as never` and handed to `context.collection.list`. Before this change the real instance was passed through, so any decorator below that calls a method on the filter — `override`, `nest`, and friends — worked; now it would throw on a plain object.

It evidently works against the datasources exercised in tests, but it depends on nothing in the chain below `internalHook` treating the filter as more than a bag of properties, which isn't a property the toolkit guarantees.

agent-ruby#320 does this the safe way, and the equivalent exists here:

```ruby
filter.override(page: Page.new(offset: 0, limit: cap + 1))
```

Constructing a `PaginatedFilter` (or using the filter's own override) keeps the instance intact and costs nothing.

packages/agent/src/audit-trail/instrument.ts
```diff
@@ -0,0 +1,524 @@
...
+    );
+  }
+
+  return value;
+};
+
+// Records touched by a bulk operation, bounded so one write can't hold an unbounded result set in
+// memory (or fan out an unbounded number of pending-row inserts). Truncation is logged explicitly
+// rather than silently auditing only part of the operation.
+const MAX_SNAPSHOT_RECORDS = 1000;
```
PMerlet: **This cap is 1000 here and 500 in agent-ruby#320 (`MAX_RECORDS_PER_OPERATION`), and the two agents behave differently on overflow.**

Same feature, same configuration surface, two thresholds: a bulk operation on 700 records is fully audited by a Node agent and over the cap on a Ruby one. And past the cap the behaviour diverges too — Ruby records the run as a single row attached to no record, Node logs a warning (or refuses the whole operation under `critical`).

Since a customer can move between agents and the front renders both identically, this is worth making one number and one documented behaviour rather than two independent judgement calls. No opinion on which value; 500 and 1000 are both defensible.

packages/agent/src/audit-trail/instrument.ts
```diff
@@ -0,0 +1,524 @@
...
+    );
+  });
+}
+
+export default function installAuditTrailHooks(
+  dataSourceCustomizer: DataSourceCustomizer,
+  _collectionCustomizer: CollectionCustomizer | null,
+  options?: AuditTrailInstrumentOptions,
+): Promise<void> | void {
+  const { sink, store, redact, critical = false, logger } = options ?? {};
```
PMerlet, line 535: **`critical` defaults to `false`, so the guarantee the protocol exists to provide is off unless a second flag is found.**

The two-phase protocol is what makes "no unaudited write" achievable, and this default means a customer who has configured `auditTrail` — and therefore believes they have an audit trail — still gets a best-effort one: a failed pending insert is logged and the write proceeds unaudited. Configuring the audit database is already an explicit opt-in; requiring a second one to get the invariant inverts where the surprise lands.

Ruby defaults the same way (`options[:critical] == true`), so this is a shared decision rather than a Node oversight — flagging it in both places because it is a product call, not an implementation detail.

Whichever way it goes, the README should say plainly that without `critical` a write can succeed with no audit row. Right now it documents the protocol without documenting that its guarantee is opt-in.

packages/agent/src/audit-trail/sql-store.ts
```diff
@@ -0,0 +1,396 @@
...
+  return column;
+}
+
+// A free-text search against the *serialized* JSON text naturally covers "keys and scalar values
+// at any depth" without walking the structure by hand: both a key and a value appear as a quoted
+// literal substring of that text. This runs as a SQL WHERE clause (not an in-memory scan of
+// fetched rows), so it composes with pagination and COUNT the same way every other filter does. A
+// redacted value can't match a search for the real value — the real value already isn't in this
+// text, `redactValues` replaced it before the row was ever written.
+export function searchCondition(sequelize: Sequelize, term: string) {
```
PMerlet, line 168: **The search runs against serialized JSON, so it matches JSON syntax as well as content.**

The escaping is right — `~` sidesteps the MySQL backslash interaction, and the two wildcards plus the escape character itself are escaped, so a term containing `%` or `_` is matched literally. `sequelize.escape` covers the injection side. Two consequences of searching the serialized text are worth knowing about, though:

- A punctuation-only term matches almost everything. Searching `,` or `:` or `{` hits every row whose diff has more than one key, because those characters are structure rather than content.
- A value containing a double quote or a backslash cannot be found. `5"` is stored in the JSON text as `5\"`, so a search for `5"` never matches the row that holds it.

Neither is severe and both are inherent to the approach. Bounding the predicate to one record's history via the indexed `collection`/`record_id` columns is the right call and keeps this cheap. But the doc comment on `AuditHistoryQuery.search` promises "the keys and scalar values of previousValues/newValues" — worth adding that the match is against the serialized form, so structure participates.

packages/agent/src/audit-trail/types.ts
```diff
@@ -0,0 +1,138 @@
...
+export type AuditOperation = 'create' | 'update' | 'delete' | 'action' | 'action_failed';
+
+export type AuditStatus = 'pending' | 'done';
+
+export type AuditRecord = {
+  id: number;
+  timestamp: string;
+  operation: AuditOperation;
+  collection: string;
+  /** Null for a pending create — the record's primary key isn't assigned yet. */
```
PMerlet, line 12: **New divergence: Ruby now follows a record across a primary-key change, this agent doesn't.**

agent-ruby#320 added a `previous_record_id` column and chains history across a rename, so a record whose primary key was updated keeps a reachable history. `previousRecordId` appears nowhere here.

Both agents file the update row under the record's *new* packed id (which is right, and was the first round's D3), but only Ruby can then walk back to what came before. Since the front queries by the record's current id, the same renamed record shows a complete chain on a Ruby agent and a history that starts at the rename on a Node one.

Either port the column, or agree explicitly that rename-chaining is Ruby-only and say so in both READMEs — but the current state is an accidental difference rather than a decision.

PMerlet (commented):
Second pass, against a53882a. Almost everything from the first round landed, and the pending/confirm protocol in particular is a real implementation rather than a gesture: the `Delta` rewrite removes the sentinel without losing the absent-vs-null distinction, PK-less collections are skipped at install with a warning instead of throwing per write, and the snapshot is bounded, batched and logged. `prepend` is gone. `canUseAuditTrail` is exposed. Identity, action name and status are on the row.

Six things below. The first one is not in any inline comment because it *cannot* be — which is the finding.

---

### `packages/agent/src/audit-trail/migrations.ts` contains two raw NUL bytes, so git treats it as binary

GitHub renders it as `Binary files a/… and b/… differ`, and so does `git diff` locally. The file's ~12 KB of migration and table-claim logic is **unreviewable in any diff view** — it is the one file in this PR I could not read as a diff, which is how I found this.

Both occurrences are a Map-key separator written as a literal NUL rather than an escape:

```
// … keyed by `schema<NUL>name` — both its own data table and the migration table Umzug derives…
const key = `${schema ?? ''}<NUL>${name}`;
```

Worth noting the irony: this PR just removed a NUL-bearing sentinel from `types.ts` for adjacent reasons.

The fix is to write the separator as an escape (`'\u0000'`) so the source stays ASCII, or pick a separator that cannot occur in a schema/table name. Either way the file stops being binary and becomes reviewable. Beyond diffs, a raw NUL also makes the file invisible to `grep`/`rg` by default and can upset formatters.

For what it's worth, having read the file through the API rather than the diff: the collision logic itself is sound — `claimedBy` is the `tableName`, so two stores sharing a table name claim the same key with the same owner and don't collide, and the module-level Map is bounded by distinct configurations. No issue there.

packages/agent/src/audit-trail/instrument.ts
bexchauveto: Fixed — `withLimit` now fetches `MAX_SNAPSHOT_RECORDS + 1` (1001) and `enforceSnapshotCap` checks `fetched.length > MAX_SNAPSHOT_RECORDS` instead of checking the fetched length against the cap directly, so exactly 1000 matches is no longer misreported as truncated. Added a regression test seeding exactly 1000 records asserting no warning, plus updated the truncation/critical tests to seed 1001. See `instrument.ts`/`instrument.test.ts` in 5737b87b484a44086b314d517f5cca95bc07c250.

packages/agent/src/audit-trail/instrument.ts
bexchauveto: Fixed — `withLimit` now builds a real `new PaginatedFilter({ ...filter, page: new Page(0, MAX_SNAPSHOT_RECORDS + 1) })` instead of spreading `page` onto a plain object. `context.filter` is typed as the plain `TFilter` shape for customer ergonomics but is actually a frozen `Filter` instance at runtime, so I cast `context.filter as unknown as Filter` before spreading it — same cast the hook decorator itself performs internally to produce that value. Tests now assert `collection.list` is called with an actual `PaginatedFilter` instance.

packages/agent/src/audit-trail/instrument.ts
bexchauveto: No opinion from me either, so I haven't changed anything here yet — this needs a product call rather than a unilateral pick on my end. Options as I see them: (a) leave Node at 1000, Ruby at 500, and document the divergence explicitly in both READMEs; (b) drop Node to 500 to match Ruby; (c) raise Ruby to 1000. Separately, the overflow *behavior* also differs (Node logs-a-warning-or-refuses-under-critical vs Ruby recording a single no-record row) — happy to port either direction once there's a preference, but didn't want to guess at a schema-adjacent behavior change.

packages/agent/src/audit-trail/instrument.ts
bexchauveto: Fixed — extended the JSDoc on `AuditHistoryQuery.search` in `types.ts` and added a matching paragraph to the README's Searching section, stating explicitly that the guarantee only holds under `critical: true`, and by default a write can succeed with no audit row at all.

packages/agent/src/audit-trail/sql-store.ts
bexchauveto: Fixed — extended the JSDoc on `AuditHistoryQuery.search` and the README's Searching section to state explicitly that the match runs against serialized JSON text, not a structural walk: a punctuation-only term (`,`, `:`, `{`) can match unrelated rows, and a literal quote/backslash in a value can't be searched for as typed. Neither is severe but both are worth documenting rather than leaving as a surprise.

packages/agent/src/audit-trail/types.ts
bexchauveto: Not ported yet — this is a bigger question than the others since it's a schema change (a new `previous_record_id` column) plus read-side rename-chaining logic, and I don't have verified visibility into Ruby's exact semantics to port it faithfully rather than guess. Would rather get a decision alongside #3 above (Node/Ruby parity) than silently skip it or implement something that diverges from Ruby's actual behavior. If it's wanted, I can go look at agent-ruby#320 in detail and come back with a concrete port.

packages/agent/src/audit-trail/instrument.ts
```diff
@@ -0,0 +1,544 @@
...
+    await Promise.all(
+      before.map((record, index) => {
+        const updated = afterById.get(toPackedRecordId(expectedRecords[index], primaryKeys));
+        const pendingId = pendingIds[index] ?? null;
+
+        if (!updated) return undefined;
+
+        const { previousValues, newValues } = changedValues(record, updated, writableColumns);
+
+        if (Object.keys(newValues).length === 0) return undefined;
```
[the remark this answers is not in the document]
bexchauveto: Confirmed real, fixed in a9348f5f980c23b18e0d14f83830fa2139c29434 — the early return skipped `recorder.confirm` on a no-diff update, leaving the pending row stuck forever (and leaking an entry in the sink-based recorder's in-memory map indefinitely). `confirm` now always runs, with empty `previousValues`/`newValues` when nothing actually changed, so the row correctly ends up `status: 'done'` instead of permanently `pending`. Updated all the "equal value/object/array/date/BigInt/BSON" tests that previously asserted `sink` was never called for a no-op update to now assert it's called with empty diffs.

packages/agent/src/audit-trail/instrument.ts
```diff
@@ -0,0 +1,544 @@
...
+    const pendingIds = await recorder.insertPendingBatch(pendingRecords);
+
+    pendingIdsByData.set(context.data, pendingIds);
+  });
+
+  collection.addInternalHook('After', 'Create', async (context: HookAfterCreateContext) => {
+    const pendingIds = pendingIdsByData.get(context.data) ?? [];
+    pendingIdsByData.delete(context.data);
+
+    await Promise.all(
```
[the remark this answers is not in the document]
bexchauveto: Confirmed real, fixed in ef756ce01169398e2c445d294612382aac2532fe — the sink-backed recorder's `confirm` had no try/catch around `emit()`, unlike the store-backed path, so a rejecting sink propagated through the After-hook's `Promise.all` and would have failed the response for an already-committed write, bypassing `critical: false`'s log-and-continue contract. Wrapped it the same way the store path already is, and added a test asserting the write resolves and the error is logged rather than thrown.

packages/agent/src/audit-trail/README.md
```diff
@@ -0,0 +1,346 @@
...
+  `forest` schema is created (and committed) first, before the lock, since the migration runner
+  opens its own connection and would not see an uncommitted `CREATE SCHEMA`; that step is made
+  idempotent (existence check + tolerating a concurrent "already exists").
+
+**Evolving the table (maintainers):** append a new migration to the array built in
+`packages/agent/src/audit-trail/migrations.ts` and update the model in `sql-store.ts` to match.
+Never edit, reorder or delete an existing migration once it has shipped, and keep changes
+additive/backward-compatible — the connection string may point at the customer's own database.
+
+## Notes
```
PMerlet, line 348: **The rename divergence is documented on the Ruby side only.**

agent-ruby#320's `AUDIT_TRAIL.md` now states it plainly: *"A record that was renamed keeps one timeline — on this agent. The Node agent files an update under the record's new id too, but has no `previous_record_id` and so cannot walk back past the rename: the same record shows a complete chain here and a history beginning at the rename there, until that column is ported."*

That's the honest half of the ask; this README says nothing about it. So a customer on a Node agent has no way to learn that a renamed record's history starts at the rename — they'll read the per-record history section, query by the record's id, and get a truncated story with nothing indicating it's truncated.

Either port `previous_record_id` (and then both docs can drop the caveat), or add the mirror of Ruby's paragraph here — a couple of lines in this **Notes** section is enough. The point isn't the wording, it's that the behaviour is currently discoverable in one repo and invisible in the other.

PMerlet (commented):
Third pass, against 94d1ccd. Everything from round 2 is addressed except the NUL bytes in `migrations.ts` — raised separately as a top-level comment, since a binary file has no diff position to anchor to and a review body evidently isn't the place for it.

The cap fix is right (fetch `MAX + 1`, treat only `> MAX` as overflow, slice to `MAX`), the filter now arrives as a real `PaginatedFilter`, the caps are aligned with Ruby at 1000 with matching over-cap behaviour, and the `critical` opt-in is documented exactly where someone would look for the guarantee. The search caveat is in too.

Also worth saying: the two bugs found here since round 2 are both ones I missed. A no-op update leaving its pending row stuck at `pending` is *false evidence* — a row that reads as "a write may not have landed" for a write that landed fine — and I read that early-return in round 2 without connecting it to the new protocol. And catching a rejecting sink at confirm time closes the same post-commit failure the store path already guarded. Good catches.

One small thing left, below.

PMerlet:
### `packages/agent/src/audit-trail/migrations.ts` contains two raw NUL bytes, so git treats it as binary

Raising this at the top level rather than inline or in a review body: a binary file has no diff position to anchor an inline comment to, and this was in my round-2 review body where it understandably got missed. It is the only item from two review rounds still open.

**What happens.** GitHub renders the file as `Binary files a/... and b/... differ`, and so does `git diff` locally. Its ~12 KB of migration and table-claim logic is unreviewable in any diff view — nobody reviewing this PR can see what changed in it. That is how I found this: it was the one file in the PR I could not read as a diff and had to fetch through the contents API instead.

**Where.** Two occurrences, both a Map-key separator written as a literal NUL rather than an escape — one in code, one in the comment documenting it:

```
// Every table name claimed by an audit-trail store configured in this process, keyed by
// `schema<NUL>name` - both its own data table and the migration table Umzug derives from it.
const key = `${schema ?? ''}<NUL>${name}`;
```

**Why it is worth a change rather than a shrug.** Beyond the diff, a raw NUL makes the file invisible to `grep`/`rg` by default (they classify it as binary and skip the content), and can upset formatters and editors. There is also an irony worth noting: this PR removed a NUL-bearing sentinel from `types.ts` for adjacent reasons in commit `5c03191`.

**Fix.** Write the separator as an escape so the source stays ASCII:

```ts
const SEPARATOR = '\u0000';
const key = `${schema ?? ''}${SEPARATOR}${name}`;
```

or pick a separator that cannot occur in a schema or table name (`\u001f`, `::`, anything). Either way the file stops being binary and becomes reviewable, and the runtime behaviour is identical.

**For the record**, since I read the file through the API rather than the diff: the collision logic itself is sound. `claimedBy` is the `tableName`, so two stores sharing a table name claim the same key with the same owner and correctly do not collide; the module-level `Map` is bounded by distinct configurations rather than growing per agent. No issue there — it just deserves to be reviewable by someone other than me.

packages/agent/src/audit-trail/README.md
bexchauveto: Fair — didn't want to guess-port a schema-adjacent feature (previous_record_id + rename-chaining) without verified Ruby semantics, per my earlier reply on the cap/overflow-parity comment, but you're right that leaving the asymmetry undocumented on this side is a separate problem from not porting it. Added the mirror of Ruby's AUDIT_TRAIL.md paragraph to this README's Notes section in 1cb15e6d541b77fba6a7722dbd0a3167d272a100, and to the public docs (get-started/control/audit-trail.mdx) as well. Still open: whether to actually port previous_record_id — that's the same product decision as the cap-value/overflow-behavior parity question, not resolved yet.

bexchauveto:
Confirmed real, fixed in 77a33e642a2a1f46bc35a87a5bcc2dcbdccced23 — both NUL bytes (the `schema<NUL>name` separator in `claimTableName` and the same character in the comment describing it) are now written as the `\0` escape instead of a literal byte, so the file is plain ASCII text again and diffs normally going forward. Same runtime key/behavior, verified via the migrations test suite (18/18) and the full agent suite (1435/1435). Thanks for tracking this down through the contents API — appreciate the persistence across two rounds.

packages/agent/src/audit-trail/instrument.ts
bexchauveto: Update: this got resolved on the Ruby side yesterday (agent-ruby@87b05e9c) — cap set to 1000 in both agents (was 500 in Ruby), and both agents now refuse an over-cap run under `critical: true` rather than disagreeing on truncate-vs-refuse. That commit also caught that Node's smart-action path had no cap at all (only bulk update/delete did), so I ported the same fix here in 9c973a06023bb2f369138d4aea5470a8dc0d059c: a smart action's audited selection is now capped at the same `MAX_SNAPSHOT_RECORDS` (exported and reused, not a second constant), truncated selections are recorded as one entry attached to no record under `critical: false`, and refused before `execute()` under `critical: true`. Cap-value/overflow parity is closed; `previous_record_id` is the one still open.

**Edit:** that port had its own bug — it passed `filterForCaller` straight to `collection.list` without resolving a live-query segment's `$contextVariable` placeholders first, unlike `list.ts`/`count.ts`, which run the filter through `segmentQueryHandler.handleLiveQuerySegmentFilter` before their own `collection.list` call. A selection scoped by a live-query segment would reach the datasource with unresolved placeholders — failing under `critical: true`, or silently downgrading to a no-record entry under `critical: false`. Fixed in db55532beccfffc0eef481248cb0f890cbb5bdb0, with a regression test that fails without the fix.

PMerlet (approved):
LGTM

PR #1687, opened 2026-06-19
feat(workflow-executor): PRD-551 apply deterministic Load Related Record config (revise-safe)

Scra3:
## Context
Runtime half of the Deterministic Load Related Record epic (PRD-471). The executor must consume the `preRecordedArgs` the orchestrator now sends (PRD-548) and resolve the source record + relation deterministically.

## What
- **Forward `preRecordedArgs`** from the orchestrator wire: it was silently dropped by `step-definition-mapper` (only `prompt`/`executionType`/`title` were mapped). Added to `ServerWorkflowTaskLoadRelatedRecord` + the mapper.
- **Revise-safe source resolution**: `resolveSourceRecordRef` matches a previous load-related step on its own `stepIndex` **or** `originalStepIndex` (mirrors `resolveStepExecution` / `getAvailableRecordRefs`), plus the workflow-start record. The previous strict own-index match on the flattened pool reported a valid chained source as "no source record" after a revise.
- **Removed `selectedRecordIndex`** (index-based record pinning) from schema/executor/tests: brittle, not revise-safe, unused here. The final record stays AI-suggested + user-confirmed.

## Scope
Source + relation are pinned; the final record choice remains AI-suggested + user-confirmed (Full AI pinning is PRD-148). Error on deleted relation/collection reuses existing `InvalidPreRecordedArgsError` / `RelationNotFoundError` (dual technical/user message).

## Tests
80/80 in the load-related suite (added a revise-safety test that fails on the old strict match), 1081/1081 across the package.

fixes PRD-551

🤖 Generated with [Claude Code](https://claude.com/claude-code)

> [!NOTE]
> ### Replace index-based source record pinning with stable step ID resolution in `LoadRelatedRecordStepExecutor`
> - `preRecordedArgs` for Load Related Record steps now uses `selectedRecordStepId` (a string step ID or the `workflow-start` sentinel) instead of `selectedRecordStepIndex` to identify the source record, making references survive workflow revisions that shift step order.
> - A new `resolveSourceRecordRef` method looks up the matching prior step execution in `runStore` by step ID; if the step ran but loaded no record, it throws the new `SourceRecordMissingError` with a distinct user-facing message.
> - Index-based pinning of the final related record (`selectedRecordIndex`) is removed; AI ranking is now always used for final record selection.
> - Behavioral Change: workflows with existing `preRecordedArgs` using `selectedRecordStepIndex` or `selectedRecordIndex` will fail schema validation after this change.
>
> 
>
[...]

PR #1688, opened 2026-06-20
test(workflow-executor): PRD-552 legacy and deterministic Load Related Record coexist

Scra3:
## Context
Joint backward-compat guard for the Deterministic Load Related Record epic (PRD-471), stacked on #1687 (PRD-551).

## What
An explicit regression test: within one run the same executor takes the **AI path** for a legacy prompt-only step (no `preRecordedArgs`) and the **deterministic no-AI path** for a step with source + relation pinned. Coexistence was already implied by the green suite; this names it as a guard.

## Where the rest of the coexistence is covered
- **Orchestrator** (#8312): step with no `forest:preRecordedArgs` → forwarded unchanged (parser test).
- **Run panel** (#9773): Handle Manually present on legacy, absent on deterministic; field unlocked on legacy.
- **Editor** (#9770): legacy steps render the untouched `guidance` panel (gating).

## Tests
82/82 in the load-related suite (run locally). 

fixes PRD-552

> [!NOTE]
> ### Add coexistence test for legacy and deterministic `LoadRelatedRecordStepExecutor` paths
> Adds a test suite in [load-related-record-step-executor.test.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1688/files#diff-ac728230ca190f664862205507ae80cd2c51bc30da32dfa3e0ccbd36e13f3345) verifying that legacy prompt-only `FullyAutomated` steps call `bindTools` for AI relation selection, while deterministic steps with `preRecordedArgs` and prior execution data skip `bindTools` entirely and load related data directly.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 7412b6a.</sup>
> 
>

### DIFF
--- a/packages/workflow-executor/test/executors/load-related-record-step-executor.test.ts
+++ b/packages/workflow-executor/test/executors/load-related-record-step-executor.test.ts
@@ -3603,8 +3603,6 @@ describe('LoadRelatedRecordStepExecutor', () => {
     });
   });
 
-  // PRD-552: legacy prompt-only steps and new deterministic steps must coexist — the same executor
-  // takes the AI path for one and the deterministic (no-AI) path for the other within one run.
   describe('backward compatibility (PRD-552): legacy and deterministic coexist', () => {
     it('runs the AI path for a legacy step and the deterministic path for a configured step', async () => {
       // Legacy prompt-only step → AI selects the relation.
--- a/packages/workflow-executor/test/executors/load-related-record-step-executor.test.ts
+++ b/packages/workflow-executor/test/executors/load-related-record-step-executor.test.ts
@@ -3635,7 +3635,10 @@ describe('LoadRelatedRecordStepExecutor', () => {
           agentPort: makeMockAgentPort([
             makeRelatedRecordData({ collectionName: 'customers', recordId: [7], values: {} }),
           ]),
-          workflowPort: makeMockWorkflowPort({ customers: makeCollectionSchema(), orders: ordersSchema }),
+          workflowPort: makeMockWorkflowPort({
+            customers: makeCollectionSchema(),
+            orders: ordersSchema,
+          }),
           previousSteps: [makeLoadRelatedPreviousStep(1)],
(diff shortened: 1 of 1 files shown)


packages/workflow-executor/test/executors/load-related-record-step-executor.test.ts
```diff
@@ -3603,6 +3603,73 @@ describe('LoadRelatedRecordStepExecutor', () => {
...
+            makeRelatedRecordData({ collectionName: 'customers', recordId: [7], values: {} }),
+          ]),
+          workflowPort: makeMockWorkflowPort({
+            customers: makeCollectionSchema(),
+            orders: ordersSchema,
+          }),
+          previousSteps: [makeLoadRelatedPreviousStep(1)],
+          stepDefinition: makeStep({
+            executionType: StepExecutionMode.FullyAutomated,
+            preRecordedArgs: { selectedRecordStepId: 'load-1', relationName: 'customer' },
```
Scra3, line 3661: **non-blocking:** `selectedRecordStepId` is not a recognized `preRecordedArgs` key (the schema/executor only accept `selectedRecordStepIndex` / `selectedRecordIndex` / `relationName`), so the source-record pinning this test sets up is silently ignored — selection falls back to all available records. The test passes only because `relationName: 'customer'` still pins the relation; it doesn't validate the step-index pinning its fixture implies. Use `selectedRecordStepIndex: 1` (as in the existing deterministic test) to actually exercise it.

packages/workflow-executor/test/executors/load-related-record-step-executor.test.ts
```diff
@@ -3603,6 +3603,73 @@ describe('LoadRelatedRecordStepExecutor', () => {
     });
   });
 
+  // PRD-552: legacy prompt-only steps and new deterministic steps must coexist — the same executor
```
Scra3: remove this comment

PR #1698, opened 2026-06-22
fix(workflow-executor): serialize cold-start migrations behind a pg advisory lock

PMerlet:
## Summary

Fixes the lack of concurrency control in the `@forestadmin/workflow-executor` migration runner. `DatabaseStore.init()` called `umzug.up()` with no advisory lock, so when several executor instances boot at the same time — any customer running **≥2 replicas** (k8s Deployment, multiple containers, ASG min ≥2) — they all run the **same pending migrations** against the shared database concurrently. This happens on first deploy and on **every release that ships a new migration**.

- **Today:** the lone migration's `createTable` fails fast for the losers → boot crash / crash-looping pods.
- **Future:** a non-idempotent data backfill or multi-statement migration run by N instances can leave **partial/duplicated schema or data** that does not self-heal.

## Fix

Serialize the migration runner behind a Postgres advisory lock (`withMigrationLock`): the first instance migrates while the others block, then see the migrations applied and no-op. Cross-instance serialization works because `pg_advisory_xact_lock(KEY)` is **database-global** — the only requirement is a shared database (which the bug already establishes), not a shared connection.

Three things make it production-safe for the common HA setup (RDS, often behind a pooler):

1. **Transaction-scoped lock** (`pg_advisory_xact_lock`, not session `pg_advisory_lock`). It lives for exactly the transaction a connection pooler keeps pinned and auto-releases on commit/rollback/disconnect — so it's safe behind **RDS Proxy / PgBouncer transaction mode** and can never leave a dangling lock. A session lock would leak there.
2. **Idle-timeout guard.** The lock transaction sits idle while the migration runs on other pooled connections, so it first issues `SET LOCAL idle_in_transaction_session_timeout = 0` — otherwise a client-configured timeout could terminate it mid-migration, drop the lock, and let a sibling in.
3. **Transactional + idempotent migration.** Migration `001` wraps its DDL in a transaction and short-circuits via `tableExists`. The lock guarantees one writer; this guarantees no half-applied state and no `"already exists"` crash-loop if a prior run committed but failed before umzug recorded it in `SequelizeMeta`.

`init()` runs two locked transactions in sequence: the first creates the schema (committed so the second can see it — umzug runs on other pooled connections), the second runs `umzug.up()`.

- Gated on `dialect === 'postgres'`; SQLite (test suite) is single-process and skips locking entirely.
- Lock key (`MIGRATION_ADVISORY_LOCK_KEY`) is a fixed constant and must never change across releases.
- Requires `pool.max >= 2` (lock connection + migration connection); Sequelize's default of 5 is fine.

## Tests

**Unit** (`database-store-schema.test.ts`, mocked): two-phase locked ordering (`SET LOCAL` → lock → schema, then `SET LOCAL` → lock → migrate), the `pg_advisory_xact_lock` bind, SQLite takes no lock, the idempotent `tableExists` skip, and a serialization test proving concurrent `init()` applies a non-idempotent migration exactly once.

**Real Postgres** (`database-store.pg.test.ts`, gated on `WORKFLOW_EXECUTOR_TEST_DATABASE_URL`, skipped in CI): 5 concurrent `init()` on an empty schema migrate exactly once; no advisory lock lingers after `init()`; `init()` blocks while another session holds the lock then proceeds on release; and the guarded transaction keeps its lock across an idle window longer than a tight 200 ms `idle_in_transaction_session_timeout`.

Full suite: 1087 passing, lint clean. (Postgres in CI, explicit unsupported-dialect handling, and the theoretical advisory-key collision are noted as follow-ups.)

fixes PRD-593

> [!NOTE]
> ## Summary
>
> Fixes the lack of concurrency control in the `@forestadmin/workflow-executor` migration runner. `DatabaseStore.init()` called `umzug.up()` with no advisory lock, so when several executor instances boot at the same time — any customer running **≥2 replicas** (k8s Deployment, multiple containers, ASG min ≥2) — they all run the **same pending migrations** against the shared database concurrently. This happens on first deploy and on **every release that ships a new migration**.
>
> - **Today:** the lone migration's `createTable` fails fast for the losers → boot crash / crash-looping pods.
> - **Future:** a non-idempotent data backfill or multi-statement migration run by N instances can leave **partial/duplicated schema or data** that does not self-heal.
>
> ## Fix
>
> Serialize the migration runner behind a Postgres advisory lock (`withMigrationLock`): the first instance migrates while the others block, then see the migrations applied and no-op. Cross-instance serialization works because `pg_advisory_xact_lock(KEY)` is **database-global** — the only requirement is a shared database (which the bug already establishes), not a shared connection.
>
> Three things make it production-safe for the common HA setup (RDS, often behind a pooler):
>
> 1. **Transaction-scoped lock** (`pg_advisory_xact_lock`, not session `pg_advisory_lock`). It lives for exactly the transaction a connection pooler keeps pinned and auto-releases on commit/rollback/disconnect — so it's safe behind **RDS Proxy / PgBouncer transaction mode** and can never leave a dangling lock. A session lock would leak there.
> 2. **Idle-timeout guard.** The lock transaction sits idle while the migration runs on other pooled connections, so it first issues `SET LOCAL idle_in_transaction_session_timeout = 0` — otherwise a client-configured timeout could terminate it mid-migration, drop the lock, and let a sibling in.
> 3. **Transactional + idempotent migration.** Migration `001` wraps its DDL in a transaction and short-circuits via `tableExists`. The lock guarantees one writer; this guarantees no half-applied state and no `"already exists"` crash-loop if a prior run committed but failed before umzug recorded it in `SequelizeMeta`.
>
> `init()` runs two locked transactions in sequence: the first creates the schema (committed so the second can see it — umzug runs on other pooled connections), the second runs `umzug.up()`.
>
> - Gated on `dialect === 'postgres'`; SQLite (test suite) is single-process and skips locking entirely.
> - Lock key (`MIGRATION_ADVISORY_LOCK_KEY`) is a fixed constant and must never change across releases.
> - Requires `pool.max >= 2` (lock connection + migration connection); Sequelize's default of 5 is fine.
>
> ## Tests
>
> **Unit** (`database-store-schema.test.ts`, mocked): two-phase locked ordering (`SET LOCAL` → lock → schema, then `SET LOCAL` → lock → migrate), the `pg_advisory_xact_lock` bind, SQLite takes no lock, the idempotent `tableExists` skip, and a serialization test proving concurrent `init()` applies a non-idempotent migration exactly once.
>
> **Real Postgres** (`database-store.pg.test.ts`, gated on `WORKFLOW_EXECUTOR_TEST_DATABASE_URL`, skipped in CI): 5 concurrent `init()` on an empty schema migrate exactly once; no advisory lock lingers after `init()`; `init()` blocks while another session holds the lock then proceeds on release; and the guarded transaction keeps its lock across an idle window longer than a tight 200 ms `idle_in_transaction_session_timeout`.
>
> Full suite: 1087 passing, lint clean. (Postgres in CI, explicit unsupported-dialect handling, and the theoretical advisory-key collision are noted as follow-ups.)
>
> fixes PRD-593
>
> #### Changes since #1698 opened
>
> - Added connection pool size validation in `DatabaseStore.init` method [04c1843]
> - Changed `DatabaseStore.init` method to read maximum pool size from `connectionManager.pool.maxSize` instead of `sequelize.options.pool.max` [73e0cea]
> - Added test validation for Postgres connection pool minimum size requirement [a5d17c0]
> - Extended Sequelize test mocks to expose connection pool configuration [a5d17c0]
> - Reformatted TypeScript cast assignment in test case [96d61eb]
> 
>


packages/workflow-executor/src/stores/database-store.ts
```diff
@@ -103,11 +117,20 @@ export default class DatabaseStore implements RunStore {
 
     return this.callPort('init', async () => {
       try {
+        if (this.sequelize.getDialect() !== 'postgres') {
```
Scra3, line 120: question: Non-sqlite dialects still resolve `schema` to `forest`, but this early return skips the `CREATE SCHEMA IF NOT EXISTS` the old path ran — is any non-postgres, non-sqlite dialect (e.g. MySQL) supported, since the DDL would then target an uncreated schema?

packages/workflow-executor/src/stores/database-store.ts
```diff
@@ -103,11 +117,20 @@ export default class DatabaseStore implements RunStore {
...
+        // Schema in its own committed transaction so umzug (on other connections) sees it.
         if (schema) {
-          await this.sequelize.query(`CREATE SCHEMA IF NOT EXISTS "${schema}"`);
+          await this.withMigrationLock(transaction =>
+            this.sequelize.query(`CREATE SCHEMA IF NOT EXISTS "${schema}"`, { transaction }),
+          );
         }
 
-        await umzug.up();
+        await this.withMigrationLock(() => umzug.up());
```
Scra3, line 145: issue (non-blocking): The lock-holding transaction occupies one pool connection while `umzug.up()` acquires a separate one, so a Sequelize pool configured with `max: 1` self-deadlocks here.

packages/workflow-executor/src/stores/database-store.ts
```diff
@@ -11,6 +11,9 @@ import { RunStorePortError, WorkflowExecutorError, extractErrorMessage } from '.
 const TABLE_NAME = 'workflow_step_executions';
 const DEFAULT_SCHEMA = 'forest';
 
+// Must stay constant across releases, or an old and a new deploy could migrate concurrently.
+const MIGRATION_ADVISORY_LOCK_KEY = 6_438_071_259_157;
```
Scra3, line 15: suggestion: `pg_advisory_xact_lock` keys are database-global, so two executors sharing one Postgres database under different schemas serialize on this single key — folding the schema into the key (e.g. `hashtext(schema)`) avoids cross-tenant contention.

packages/workflow-executor/src/stores/database-store.ts
```diff
@@ -117,6 +140,25 @@ export default class DatabaseStore implements RunStore {
...
+  private async withMigrationLock(
+    run: (transaction: Transaction) => Promise<unknown>,
+  ): Promise<void> {
+    await this.sequelize.transaction(async transaction => {
+      // Stop a client idle-in-transaction timeout from killing this idle txn mid-migration,
+      // which would drop the lock.
+      await this.sequelize.query('SET LOCAL idle_in_transaction_session_timeout = 0', {
+        transaction,
+      });
+      await this.sequelize.query('SELECT pg_advisory_xact_lock($1)', {
```
Scra3, line 166: praise: Transaction-scoped advisory lock + `SET LOCAL idle_in_transaction_session_timeout = 0` + committing the schema in its own lock before umzug is a correct, pooler-safe serialization.

Scra3 (commented):
Solid concurrency fix. A few non-blocking points: a potential non-postgres schema regression, a pool-size deadlock edge, and cross-tenant lock contention. Details inline.

packages/workflow-executor/src/stores/database-store.ts
PMerlet: Pre-existing behaviour, not a regression from this PR. Before this change, non-postgres dialects already landed on `umzug.up()` directly without any `CREATE SCHEMA` call. MySQL does not have schemas in the Postgres sense — in Sequelize, passing `{ schema }` to `createTable` on MySQL targets the current database, which already exists. No DDL change needed there.

packages/workflow-executor/src/stores/database-store.ts
PMerlet: Fixed with a runtime guard: if `pool.max < 2` on Postgres, `init()` now throws immediately with a clear message rather than hanging. Added in the Postgres branch, before `withMigrationLock` is ever called.

packages/workflow-executor/src/stores/database-store.ts
PMerlet: Valid point. In practice, two executors sharing the same Postgres instance but different schemas is an unlikely deployment (each customer runs their own executor pointing at their own DB). The contention is harmless — it just serializes the lock acquisition unnecessarily. I am leaving the key as-is to keep the constant stable; changing it mid-deployment would allow concurrent migrations during the upgrade window.

packages/workflow-executor/src/stores/database-store.ts
PMerlet: Thanks. For context: Metabase faces the same problem and uses Liquibase's `DATABASECHANGELOGLOCK` table (session-scoped row lock). Their lock can get stuck if the process dies mid-migration, requiring manual `migrate release-locks` — issues [#10651](https://github.com/metabase/metabase/issues/10651) and [#37866](https://github.com/metabase/metabase/issues/37866) document this failure mode. The transaction-scoped advisory lock here avoids it entirely.

Scra3 (approved):

PR #1701, opened 2026-06-23
feat(workflow-executor): attribute AI usage to the triggering user [PRD-619]

Scra3:
## Problem

Executor-run workflow AI calls hit `POST /liana/v1/ai-proxy`, which is authenticated by **env secret only** — no user context. So AI usage from orchestrator-engine runs couldn't be attributed to the user who triggered the run.

The server-side fix (PRD-619, `forestadmin-server`) tried to resolve the user via a `workflow-run-id` header lookup — but **the executor never sent any identifying header** (the proxy `fetch` set only `forest-secret-key`), so attribution was always empty.

## Fix

The executor already has `step.user.id` (it comes from the orchestrator's run). Forward it directly as a `user-id` header on the ai-proxy call — no DB lookup, no `runId` string→number mismatch.

- `getModel(aiConfigName?, userId?)` — userId threaded from `step.user.id` at the factory call site.
- The model is built **per call** so the userId is captured in that model's own `fetch` closure. The adapter is shared across concurrent runs, so a mutable field would race; a fresh `AiClient` per call is cheap and race-free.

## Trust

The route is already env-secret authed (client infra = existing trust boundary). `user-id` is used only for usage tracking (billing/analytics), not authorization.

## Tests

`server-ai-adapter.test.ts`: `user-id` header set when userId given, omitted otherwise.

## Pairs with

`forestadmin-server` PR (reads the `user-id` header, drops the dead runId lookup).

> [!NOTE]
> ### Attribute AI usage to the triggering user in workflow executor
> - Introduces a `GetModelOptions` interface in [`ai-model-port.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1701/files#diff-3b48820bd99c28dccf8be78e8d352be9fa22b68de22521bcbb666e0d2dcdaad6) with optional `aiConfigName` and `userId` fields, replacing the previous string parameter on `AiModelPort.getModel`.
> - Updates [`ServerAiAdapter.getModel`](https://github.com/ForestAdmin/agent-nodejs/pull/1701/files#diff-c956f4e501a2d4f6236f5e7ed03a9ad79a955818f220f62955023fd87bb9b045) to construct a per-call `AiClient` using the provided `userId`, injecting a `user-id` header into downstream proxy requests when present.
> - Updates [`StepExecutorFactory.buildContext`](https://github.com/ForestAdmin/agent-nodejs/pull/1701/files#diff-8bd043179aba0b7fb86303fd2d3cd99f76e8f762d374c0e24be5389107c50b4d) to pass `step.user.id` as `userId` when calling `getModel`, so each workflow step attributes AI usage to the triggering user.
> - Behavioral Change: `AiModelPort.getModel` now expects an options object instead of a plain string; all implementations and callers must be updated.
>
> 
> #### Changes since #1701 opened
>
> - Added test case verifying AI usage attribution to triggering user [1734bdb]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8cbee6e.</sup>
> 
>
>

### DIFF
--- a/packages/workflow-executor/src/executors/step-executor-factory.ts
+++ b/packages/workflow-executor/src/executors/step-executor-factory.ts
@@ -152,7 +152,10 @@ export default class StepExecutorFactory {
       stepDefinition: step.stepDefinition,
       previousSteps: step.previousSteps,
       user: step.user,
-      model: cfg.aiModelPort.getModel(step.stepDefinition.aiConfigName),
+      model: cfg.aiModelPort.getModel({
+        aiConfigName: step.stepDefinition.aiConfigName,
+        userId: step.user.id,
+      }),
       agent: new AgentWithLog({
         agentPort: cfg.agentPort,
         schemaResolver,
--- a/packages/workflow-executor/src/adapters/ai-client-adapter.ts
+++ b/packages/workflow-executor/src/adapters/ai-client-adapter.ts
@@ -1,4 +1,4 @@
-import type { AiModelPort } from '../ports/ai-model-port';
+import type { AiModelPort, GetModelOptions } from '../ports/ai-model-port';
 import type { AiConfiguration, BaseChatModel, RemoteTool, ToolConfig } from '@forestadmin/ai-proxy';
 
 import { AiClient } from '@forestadmin/ai-proxy';
@@ -13,7 +13,7 @@ export default class AiClientAdapter implements AiModelPort {
     this.aiClient = new AiClient({ aiConfigurations: withRetries as AiConfiguration[] });
   }
 
-  getModel(aiConfigName?: string): BaseChatModel {
+  getModel({ aiConfigName }: GetModelOptions = {}): BaseChatModel {
     try {
       return this.aiClient.getModel(aiConfigName);
     } catch (cause) {
--- a/packages/workflow-executor/src/adapters/server-ai-adapter.ts
+++ b/packages/workflow-executor/src/adapters/server-ai-adapter.ts
@@ -1,4 +1,4 @@
-import type { AiModelPort } from '../ports/ai-model-port';
+import type { AiModelPort, GetModelOptions } from '../ports/ai-model-port';
 import type { AiConfiguration, BaseChatModel, RemoteTool, ToolConfig } from '@forestadmin/ai-proxy';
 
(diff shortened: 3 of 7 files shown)


packages/workflow-executor/src/executors/step-executor-factory.ts
```diff
@@ -152,7 +152,10 @@ export default class StepExecutorFactory {
       stepDefinition: step.stepDefinition,
       previousSteps: step.previousSteps,
       user: step.user,
-      model: cfg.aiModelPort.getModel(step.stepDefinition.aiConfigName),
+      model: cfg.aiModelPort.getModel({
+        aiConfigName: step.stepDefinition.aiConfigName,
+        userId: step.user.id,
```
hercemer42, line 157: **Opus 4.8 (1M context)** — [Should fix]

This call site is the PR's core behaviour — forwarding `step.user.id` for per-user usage attribution — but no test asserts it. The adapter test pins `getModel({ userId: 42 })` → `user-id: 42` header, yet nothing verifies the factory actually passes `step.user.id` here: the `getModel` mocks in `runner.test.ts` and `workflow-execution.test.ts` only `mockReturnValue`/`mockImplementationOnce`, never `toHaveBeenCalledWith`. A regression dropping `userId` from this call would keep the line at 100% coverage and pass every test.

Per CLAUDE.md ("verify exact arguments, not just that a function was called"; "coverage without assertions is fake coverage"), add one assertion to an existing runner happy-path test, e.g.:

```ts
expect(aiClient.getModel).toHaveBeenCalledWith({ aiConfigName: <expected>, userId: 1 });
```

(`makePendingStep` already sets `user.id = 1`, so the value flows — only the assertion is missing.)

hercemer42 (changes requested):

packages/workflow-executor/src/executors/step-executor-factory.ts
Scra3: Addressed in 1734bdb: added a factory test asserting `getModel` is called with `{ aiConfigName: step.stepDefinition.aiConfigName, userId: step.user.id }` (`makePendingStep` sets `user.id = 1`, so a regression dropping `userId` now fails this test).

hercemer42 (approved):

PR #1704, opened 2026-06-24
feat(agent): detect action slug collisions when unsafe endpoint is on

Tonours:
## Problem

Turn on `useUnsafeActionEndpoint` and Smart Action routes drop the positional index:`/_actions/{collection}/{slug}` instead of `/_actions/{collection}/{index}/{slug}`.

Problem: two action names can slugify to the same thing. `"Send email"` and `"SEND EMAIL"` both become `send-email`. 

Same route. Koa keeps one, throws nothing, and triggering one action runs the other. Silent, wrong action.

In the default index-based mode this never bites — the index keeps the two routes distinct. It only shows up once you drop the index.

## Root cause

`SchemaGeneratorActions.getActionSlug(name)` lowercases and strips everything that isn't `[a-z0-9-]`, so different names collapse to the same slug. 
Nothing ever checked slug uniqueness. The index was the only thing keeping the collision invisible, and `useUnsafeActionEndpoint` removes it.

## Fix

A fail check at schema generation (runs on agent `start()`), **only when `useUnsafeActionEndpoint` is true**. 
Per collection: if two actions resolve to the same route slug, the agent refuses to start and tells you exactly what collides:

```
Actions "SEND EMAIL" and "Send email" on collection "User" resolve to the same endpoint slug "send-email". Rename one of them or disable useUnsafeActionEndpoint.
```

It lives in `SchemaGeneratorCollection.buildActions` — the only place that sees all of a collection's actions at once and reuses the exact same `getActionSlug` route registration uses. So the check can't drift from the real routes.

## Scope / safety

- Default mode is untouched. Flag off (the default) → check doesn't run, the index already prevents the collision.
- Scoped per collection. Same slug across two different collections is fine (distinct routes), doesn't throw.
- No change to route registration, endpoint format, or the index path. One private method + its call site.
- Security: turns a silent wrong-action execution into a startup failure. Strictly safer, fail-closed. No new surface, no new input handling.

## How to test

Manual (on your local agent):

**Before:** with `useUnsafeActionEndpoint: true` and two actions slugifying alike (`"Send an email"` + `"SEND AN EMAIL"`) on `User`, both produced `/forest/_actions/User/send-an-email` — one shadowed the other, no error.

**After:** the agent refuses to start with the message above. Rename one action, or set `useUnsafeActionEndpoint: false`, and it starts fine.

Automated:

```
yarn workspace @forestadmin/agent jest test/utils/forest-schema/generator-collection.test.ts
```

Covers 2-way and 3-way collision (throws), distinct slugs and cross-collection reuse (no throw), and asserts the real endpoints in both modes — index-free when unsafe, index-disambiguated when safe.

### General

### Security

### Documentation

- [ ] Add documentation for `useUnsafeActionEndpoint` in https://docs.forest.app/

### DIFF
--- a/packages/agent/src/utils/forest-schema/generator-collection.ts
+++ b/packages/agent/src/utils/forest-schema/generator-collection.ts
@@ -15,9 +15,11 @@ import SchemaGeneratorSegments from './generator-segments';
 
 export default class SchemaGeneratorCollection {
   private readonly schemaGeneratorActions: SchemaGeneratorActions;
+  private readonly useUnsafeActionEndpoint: boolean;
 
   constructor(options: AgentOptionsWithDefaults) {
     this.schemaGeneratorActions = new SchemaGeneratorActions(options);
+    this.useUnsafeActionEndpoint = options.useUnsafeActionEndpoint;
   }
 
   /** Build forest-server schema for a collection */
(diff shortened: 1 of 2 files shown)


arnaud-moncel (approved):
Nice 👌

PR #1706, opened 2026-06-24
feat(workflow-executor): add production Docker image and GHCR CI publish

PMerlet:
fixes PRD-238

## Summary

Distributes `@forestadmin/workflow-executor` as a production Docker image on GHCR (`ghcr.io/forestadmin/workflow-executor`), as an alternative to the npm package.

**Image (~412 MB, 3-stage Dockerfile, build context = monorepo root)**
- Reproducible, isolated dependencies: a generator (`docker/build-deps-manifest.js`) merges the external (non-`@forestadmin`) runtime deps of the executor and its 5 workspace dependencies plus the OpenTelemetry packages into a single manifest; a committed `docker/deps/yarn.lock` pins every transitive version. The build regenerates the manifest and runs `yarn install --frozen-lockfile`, so a dependency bump the lock doesn't cover fails the build instead of silently drifting.
- Base image pinned by digest; OCI labels; `HEALTHCHECK` on `GET /health`; non-root `USER node`.
- `pg` moved from `devDependencies` to `dependencies` (Sequelize loads the Postgres driver at runtime).
- Entrypoint runs the executor on flags/no-args and `exec`s any other command verbatim (`node`, `sh`), keeping the image debuggable.

**Observability — OpenTelemetry (vendor-neutral, Docker-only)**
- `src/tracing.ts` is loaded via `--require`, a no-op unless `OTEL_EXPORTER_OTLP_ENDPOINT` is set; uses the OTel Node SDK + auto-instrumentations + OTLP HTTP exporter, so it works with any OTLP backend (Datadog, Tempo, Jaeger…).
- Spans are flushed on SIGTERM/SIGINT without pre-empting the executor's graceful drain.
- The OTel packages live only in the Docker image, never shipped to npm consumers of the CLI.

**Configuration**
- `DATABASE_SSL` now defaults to `true` — managed Postgres (RDS, Supabase, Railway…) requires TLS; set `DATABASE_SSL=false` for a local DB.
- `docker-compose.yml` + `.env.example`: turnkey single-instance setup. `env_file` forwards every documented knob into the container; `extra_hosts` lets `AGENT_URL`/`DATABASE_URL` reach host services on Linux.

**CI (`docker-publish.yml`)**
- Multi-arch on native runners (amd64 on `ubuntu-latest`, arm64 on `ubuntu-24.04-arm`): each arch is built and loaded locally, **smoke-tested and Trivy-scanned, then pushed by digest only on success** — nothing reaches the registry before both gates pass. A `merge` job assembles the multi-arch manifest via `imagetools create`.
- Smoke test (`docker/smoke-test.sh`) runs the image (`--version`/`--help` + an `--in-memory` boot) to catch a broken entrypoint, a missing module, or a startup crash — things a build-only step never exercises. Shared by the PR and release jobs.
- `check-deps-closure.js` recomputes the real `@forestadmin` dependency closure and fails CI if `WORKSPACE_PACKAGES` or the Dockerfile `COPY` lines drift from it.
- PR builds validate the Dockerfile (build + smoke + scan, no push), triggered by changes to the executor or any of its 5 dependency package manifests.
- Release: SBOM attestation; mutable tags (`:latest`/`:major`/`:minor`) move only for the highest stable version, never for a prerelease or a `workflow_dispatch` rebuild of an older version.

**Docs**
- README leads with Docker (recommended), then npm, then test-only modes; documents the agent-side `workflowExecutorUrl` / `WORKFLOW_EXECUTOR_URL` wiring (required, otherwise the agent 404s the forward routes).
- `docker/README.md`: how to update the dedicated lockfile.

## Test plan

- [x] `docker build` succeeds locally (~412 MB image)
- [x] Entrypoint: `node --version`, `node -e`, and executor flags (`--version`, `--help`, `--in-memory`) all work
- [x] Boot with valid secrets: JSON logs, agent probe, OTel SDK init when the endpoint is set
- [x] Drift detection: an uncovered dependency fails `--frozen-lockfile`; closure check fails on a missing `COPY`
- [x] Smoke test asserts startup, no missing module, OTel present
- [x] **End-to-end validated on Beanstalk staging**: the standalone image runs against the real agent + RDS (TLS), and OTel traces land in Datadog tagged `env:staging`
- [x] `docker compose config` valid; CI green

> [!NOTE]
> ### Add production Docker image and GHCR CI publish pipeline for `workflow-executor`
> - Adds a multi-stage [Dockerfile](https://github.com/ForestAdmin/agent-nodejs/pull/1706/files#diff-d7484fdcdb83aeab152a6c03a53f6a75ea7d34e6a5d6ce41374fe9f3cd1ff9b5) that installs isolated runtime dependencies (separate from the monorepo's hoisted `node_modules`) using a committed [lockfile](https://github.com/ForestAdmin/agent-nodejs/pull/1706/files#diff-557c5a183d1efad614210139cc00b0d57a1b78f2c7d18053fab0c52ef9e5937b), copies only built workspace packages, and includes a healthcheck and OCI labels.
> - Adds [docker-entrypoint.sh](https://github.com/ForestAdmin/agent-nodejs/pull/1706/files#diff-f1edda19e64dae22eb047607519b6712015771b7f839cc888fcce2b0e155094d) that preloads [tracing.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1706/files#diff-221b75ff629357990edc4a83569011571fd8fff98b78787159797730adcab6f5) via `--require` on startup; tracing initialises the OpenTelemetry SDK when `OTEL_EXPORTER_OTLP_ENDPOINT` is set.
> - Adds a [GitHub Actions workflow](https://github.com/ForestAdmin/agent-nodejs/pull/1706/files#diff-5b21991be47c2922383bdc0b6bf00b65af7db51b82d049dd8d6ad03e3d37ac98) that builds per-arch images, runs a smoke test and Trivy scan, then merges digests into a multi-arch manifest published to GHCR with immutable version tags and optional mutable `:minor`, `:major`, `:latest` tags.
> - Changes `DATABASE_SSL` to default to `true` in [cli-core.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1706/files#diff-1605eabb559b21939e3b8d15b4a7a4c933da2a7cd5190022502db6ecaf93b396); set `DATABASE_SSL=false` to connect to a local DB without TLS.
> - Risk: existing deployments that relied on `DATABASE_SSL` defaulting to `false` will now require `DATABASE_SSL=false` to avoid TLS connection errors.
>
> 
> #### Changes since #1706 opened
>
> - Removed the `package.json` manifest that defined Docker dependencies for the `workflow-executor` package [e4484a8]
> - Expanded Trivy vulnerability scanning in CI/CD workflow to fail on HIGH severity issues in addition to CRITICAL [d286c77]
> - Refactored healthcheck configuration to use image-defined HEALTHCHECK instead of docker-compose override [d286c77]
> - Updated lockfile maintenance documentation to commit only yarn.lock and generate package.json directly to temporary path [d286c77]
> - Modified Trivy vulnerability scanning in Docker workflow to be non-blocking with SARIF reporting [68c8e9e]
> - Replaced SARIF-based vulnerability scanning in `docker-publish` workflow with two-phase Trivy scanning strategy [d63369c]
> - Added custom vulnerability gating script for library dependency scanning [d63369c]
> - Added `yarn.lock` to trigger paths in `docker-publish` workflow [37c569e]
> - Updated Docker README to document Dockerfile regeneration requirement for OTEL dependency changes [37c569e]
> - Added conditional inclusion of `packageManager` field from repository root `package.json` to generated dependency manifest [c1bd2a7]
> - Reworked dependency version conflict resolution in `generate` function within `workflow-executor` package [ff24db7]
> - Modified lockfile contents in `workflow-executor` Docker dependencies [ff24db7]
> - Reordered Dockerfile operations in the builder stage to copy package manifests and workspace `package.json` files using the `--parents` flag before running `yarn install`, followed by copying the full source code [60b53bf]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1daefd6.</sup>
> 
>
>


.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,197 @@
...
+          MAJOR: ${{ needs.extract-version.outputs.major }}
+          PRERELEASE: ${{ needs.extract-version.outputs.prerelease }}
+        run: |
+          # Always publish the immutable full-version tag. The mutable tags
+          # (minor, major, latest) only move for stable releases, so a prerelease
+          # like 1.8.0-beta.1 never overwrites the stable :latest / :1 images.
+          TAGS="-t $IMAGE:$FULL"
+          if [ "$PRERELEASE" != "true" ]; then
+            TAGS="$TAGS -t $IMAGE:$MINOR -t $IMAGE:$MAJOR -t $IMAGE:latest"
+          fi
```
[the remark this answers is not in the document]
PMerlet: Fixed in 55c5d33c. Gated the mutable tags (:latest/:major/:minor) on this being the highest **stable** version. Note: I did not use the suggested `releases/latest` API call — in this monorepo that endpoint returns whichever package shipped last (e.g. `@forestadmin/agent@…`), so it would never match and `:latest` would stop moving entirely. Instead I compare against this package's own git tags (`@forestadmin/workflow-executor@*`, stable only, `sort -V`).

packages/workflow-executor/package.json
```diff
@@ -37,6 +37,7 @@
     "jsonwebtoken": "^9.0.3",
     "koa": "^3.0.1",
     "koa-jwt": "^4.0.4",
+    "pg": "^8.8.0",
```
Scra3, line 40: issue (non-blocking): the executor is DB-agnostic — Sequelize derives the dialect from `DATABASE_URL`, so by this logic `mysql2`/`mariadb`/`tedious` would equally qualify; promoting `pg` to `dependencies` forces a Postgres driver on every npm consumer (incl. in-memory ones). Keep drivers in `devDependencies` (or `peerDependenciesMeta.optional`, the Sequelize model) and add `pg` to a Docker-only hardcoded set in `build-deps-manifest.js`, next to `OTEL_DEPENDENCIES`.

Scra3 (commented):
Review focused on dependency placement and Docker packaging redundancy. Main point: the executor is DB-agnostic (Sequelize derives the dialect from DATABASE_URL), so no DB driver should be a hard runtime dependency — it belongs in the Docker-only set.

packages/workflow-executor/docker/deps/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "workflow-executor-docker-deps",
+  "private": true,
+  "dependencies": {
```
Scra3: suggestion: this is a generated artifact — the Docker build regenerates it via `build-deps-manifest.js` and the README's lockfile-update step regenerates it before use, so nothing consumes the committed copy; commit only `deps/yarn.lock` to remove a drift source.

packages/workflow-executor/docker-compose.yml
```diff
@@ -0,0 +1,35 @@
...
+    # defaults — no whitelist to keep in sync.
+    env_file:
+      - .env
+    # Lets AGENT_URL / DATABASE_URL reach host services via host.docker.internal
+    # on Linux too (resolves natively on Docker Desktop, no-op there).
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    ports:
+      - "${HTTP_PORT:-3400}:${HTTP_PORT:-3400}"
+    healthcheck:
```
Scra3: suggestion: this healthcheck duplicates the `HEALTHCHECK` already baked into the image (Dockerfile) — Compose inherits the image's healthcheck, so the whole block can be dropped.

packages/workflow-executor/package.json
Scra3: Resolving: decision is Postgres-first — the migration concurrency lock (pg_advisory_xact_lock) is Postgres-only, so Postgres is the supported runtime DB. `pg` as a hard dependency is correct and is the simplest option (no Docker-only set to maintain).

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,249 @@
...
+      # The build only proves the image compiles. Run it to catch entrypoint
+      # breakage, a missing module (e.g. an un-copied @forestadmin package or OTel
+      # dep), or a startup crash — none of which a build-only step would surface.
+      - name: Smoke test (entrypoint + module graph + boot)
+        run: sh packages/workflow-executor/docker/smoke-test.sh workflow-executor:pr
+
+      - name: Scan image for vulnerabilities
+        uses: aquasecurity/trivy-action@v0.36.0
+        with:
+          image-ref: workflow-executor:pr
```
nbouliol, line 79: Could we scan `CRITICAL,HIGH`? With `ignore-unfixed: true`, fixable HIGH-severity vulns would otherwise ship silently on a production image. Same applies to the build-job scan below (line ~173).

packages/workflow-executor/docker/deps/package.json
PMerlet: Done in d286c778 / e4484a83 — stopped committing `docker/deps/package.json`; only `deps/yarn.lock` is committed now, and the Docker build + the README's lockfile-update flow both regenerate the manifest into a temp dir. README updated.

packages/workflow-executor/docker-compose.yml
PMerlet: Done in d286c778 — dropped the compose healthcheck block; Compose inherits the image's HEALTHCHECK (now identical).

.github/workflows/docker-publish.yml
PMerlet: Addressed in d63369cc, with a refinement. A blanket CRITICAL,HIGH `exit-code: 1` on the image would block the publish on (a) base-image OS CVEs and (b) npm deps already shipped via the package — desyncing GHCR from npm without reducing real risk. Instead we gate by **origin**: OS packages (Trivy `os` scan, blocking) and Docker-only `@opentelemetry/*` deps (scan-gate.js, blocking) fail the publish since they can only be fixed here; the executor's npm deps are scanned CRITICAL,HIGH and **reported** (logs + step summary), remediated at the source. So a fixable HIGH is never silent, and the image never desyncs from npm.

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,249 @@
...
+    # manifest — a dep bump in one of the 5 source packages can drift the
+    # image's dedicated lockfile (see packages/workflow-executor/docker/).
+    paths:
+      - 'packages/workflow-executor/**'
+      - 'packages/agent-client/package.json'
+      - 'packages/ai-proxy/package.json'
+      - 'packages/forestadmin-client/package.json'
+      - 'packages/datasource-toolkit/package.json'
+      - 'packages/agent-toolkit/package.json'
+      - '.github/workflows/docker-publish.yml'
```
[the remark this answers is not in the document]
PMerlet: Done in 37c569e5 — added `yarn.lock` to the validate `paths` filter (the builder installs from the repo root with `--frozen-lockfile`, which build.yml's plain `yarn` install wouldn't catch). Also updated docker/README to say `deps/yarn.lock` must be regenerated when the OTel versions in build-deps-manifest.js are bumped.

packages/workflow-executor/docker/README.md
```diff
@@ -0,0 +1,35 @@
...
+Docker build regenerates it too), so there is no stale `package.json` to drift.
+Run this whenever a runtime dependency of one of the 6 workspace packages
+changes **or when the `OTEL_DEPENDENCIES` versions in `build-deps-manifest.js`
+are bumped** (the build will fail with `--frozen-lockfile` until you do):
+
+```bash
+# from the monorepo root
+TMP=$(mktemp -d)
+node packages/workflow-executor/docker/build-deps-manifest.js packages "$TMP/package.json"
+( cd "$TMP" && yarn install --ignore-scripts )
```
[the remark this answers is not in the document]
PMerlet: Done in c1bd2a7c — rather than the cp (which the generator would overwrite right after), the generator now emits the repo's pinned `packageManager` (yarn@1.22.19) into the manifest, so with Corepack enabled the refresh uses the same Yarn as the Docker build. It's resolved from the repo root and gracefully omitted in the Docker build (root package.json isn't copied there; the image uses its bundled Yarn 1.x). README updated to note `corepack enable`.

packages/workflow-executor/docker/build-deps-manifest.js
```diff
@@ -0,0 +1,90 @@
...
+];
+
+// Pinned to exact versions — OTel ships only in the Docker image.
+const OTEL_DEPENDENCIES = {
+  '@opentelemetry/sdk-node': '0.219.0',
+  '@opentelemetry/auto-instrumentations-node': '0.77.0',
+  '@opentelemetry/exporter-trace-otlp-http': '0.219.0',
+};
+
+function generate(packagesDir, outFile) {
```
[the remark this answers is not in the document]
PMerlet: Fixed in ff24db7c. The merge is now deterministic: ranges are collected per dep, and on conflict the executor's own pin is authoritative (the image IS the executor → its tested version wins; 4.3.6 still satisfies ai-proxy's ^4.3.5). Any conflict the executor doesn't declare fails the build with a clear message instead of silently collapsing. Only zod conflicted; deps/yarn.lock regenerated to pin zod 4.3.6 (diff is zod-only, no transitive churn). Verified: full image build + smoke test pass.

Scra3 (commented):
Docker audit (optimize + compose + security). Image itself is production-grade (digest-pinned base, non-root, clean multi-stage ~70% smaller, HEALTHCHECK, no baked secrets, solid .dockerignore). Findings are mostly compose runtime-hardening + one builder cache-bust. No blockers.

packages/workflow-executor/Dockerfile
```diff
@@ -1,37 +1,95 @@
...
+FROM base AS builder
 WORKDIR /app
 
-# --ignore-scripts skips husky, native (node-gyp) builds and binary downloads
-# we don't need here: the runtime path is Postgres and `pg` is pure JS. The only
-# native dep (sqlite3) is dev-only and unused at runtime.
+# --frozen-lockfile guarantees the resolved tree matches yarn.lock exactly.
+# --ignore-scripts skips husky, native (node-gyp) builds, and binary downloads
+# irrelevant here (sqlite3 is dev-only and never loaded at runtime).
 COPY . .
```
Scra3: suggestion: the builder `COPY . .` precedes `yarn install`, so any source edit busts the install layer — copy the root + workspace `package.json`s and `yarn.lock` first, then `COPY . .` (CI's GHA cache amortizes this, but local/incremental builds reinstall the whole monorepo each time).

packages/workflow-executor/Dockerfile
```diff
@@ -1,37 +1,95 @@
...
+COPY packages/workflow-executor/docker/deps/yarn.lock ./yarn.lock
+COPY packages/workflow-executor/package.json  ./packages/workflow-executor/package.json
+COPY packages/agent-client/package.json       ./packages/agent-client/package.json
+COPY packages/ai-proxy/package.json           ./packages/ai-proxy/package.json
+COPY packages/forestadmin-client/package.json ./packages/forestadmin-client/package.json
+COPY packages/datasource-toolkit/package.json ./packages/datasource-toolkit/package.json
+COPY packages/agent-toolkit/package.json      ./packages/agent-toolkit/package.json
+
+RUN node build-deps-manifest.js packages package.json
+RUN yarn install --frozen-lockfile --ignore-scripts
```
Scra3, line 57: suggestion: `# syntax=docker/dockerfile:1` is already set, so a BuildKit cache mount (`RUN --mount=type=cache,target=/usr/local/share/.cache/yarn yarn install …`) would speed up repeated installs in both this stage and the builder.

packages/workflow-executor/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+# Minimal single-instance setup for the workflow executor.
+#
+# 1. Copy .env.example to .env and fill in your secrets.
+# 2. docker compose up
+#
+# Point your Forest Admin agent at: workflowExecutorUrl: "http://localhost:3400"
+
+services:
+  workflow-executor:
+    image: ghcr.io/forestadmin/workflow-executor:latest
```
Scra3, line 10: suggestion: `:latest` is non-deterministic for deployment — use `ghcr.io/forestadmin/workflow-executor:${IMAGE_TAG:-latest}` so a fixed version can be pinned.

packages/workflow-executor/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+    restart: unless-stopped
```
Scra3, line 11: issue (non-blocking): no runtime hardening — the container keeps all Linux capabilities and a writable root fs; add `cap_drop: [ALL]`, `security_opt: ["no-new-privileges:true"]`, and `read_only: true` + `tmpfs: ["/tmp"]` (port 3400 > 1024 needs no privileged cap).

packages/workflow-executor/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+    # Forward the whole .env into the container (see .env.example for the full
+    # list of knobs). Unset optional vars fall back to the image's built-in
+    # defaults — no whitelist to keep in sync.
+    env_file:
+      - .env
+    # Lets AGENT_URL / DATABASE_URL reach host services via host.docker.internal
+    # on Linux too (resolves natively on Docker Desktop, no-op there).
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    ports:
```
Scra3, line 21: suggestion: no resource limits or log rotation — add `deploy.resources.limits` (memory/cpus) and a `logging` json-file driver with `max-size`/`max-file` so a runaway run can't OOM the host or fill the disk.

packages/workflow-executor/Dockerfile
PMerlet: Done in 60b53bfd — manifests + yarn.lock are now COPYed (via `--parents`) and installed before `COPY . .`, so a source-only change keeps the install layer cached. This carries to CI through the type=gha layer cache.

packages/workflow-executor/Dockerfile
PMerlet: Decided not to add the BuildKit cache mount: CI runs on ephemeral shared runners where the cache mount starts empty every run (it isn't exported to the gha cache), so it gives no cross-run benefit there — only a local-dev speedup. Keeping the Dockerfile lean. The layer reorder above is the change that actually helps CI.

packages/workflow-executor/docker-compose.yml
PMerlet: This docker-compose is a local/quick-start convenience, never a production deployment artifact — real deployments (our Beanstalk, clients' k8s/ECS) pin the image tag at the platform level. Keeping `:latest` here for a frictionless local run.

packages/workflow-executor/docker-compose.yml
PMerlet: Same rationale: this compose is for local use only. Runtime hardening (cap_drop, no-new-privileges, read_only/tmpfs) belongs to the production deployment platform, not this quick-start file — adding it here only complicates the local DX for no prod benefit.

packages/workflow-executor/docker-compose.yml
PMerlet: Same: resource limits and log rotation are production-deployment concerns set at the orchestrator level (Beanstalk/ECS/k8s). This compose is a local convenience, so we're intentionally keeping it minimal.

Scra3 (approved):

PR #1403, opened 2026-01-06
feat(forest mcp): add smart action tools

VincentMolinie:
### General

- [ ] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)
- [ ] Test manually the implemented changes
- [ ] Validate the code quality (indentation, syntax, style, simplicity, readability)

### Security

- [ ] Consider the security impact of the changes made

PR #1708, opened 2026-06-25
feat(mcp-server): create approval request on action trigger when required by role

Scra3:
## Context

Some custom actions are gated behind **approval-by-role**: a user whose role can't execute the action triggers it → an approval request is created → another user (with an approving role) reviews it on the front. Until now this guard existed **only via the frontend** — triggering the same action through the **Forest MCP** executed it directly, bypassing approval (compliance hole for refunds, exports, deletions…).

This PR makes an action triggered via MCP **create an approval request** instead of executing, and reports *"approval requested"* to the LLM.

fixes PRD-288

## Approach — approval logic mutualized in `agent-client`

The approval handling lives in `agent-client` `Action.execute()` so every prod consumer (mcp-server, and later the workflow executor) gets it for free, with **no duplicated catch+create** per consumer:

- On a 403 `CustomActionRequiresApprovalError` (mapped to `ActionRequiresApprovalError` by the deterministic-action-steps base work), `execute()` calls an injected `ApprovalRequestCreator` and returns `{ approvalRequested: true }`.
- The creator is **data-injected** via `createRemoteAgentClient({ forestServer: { url, bearerToken } })` — no callback. Prod clients wire it; **`agent-testing` omits it**, so triggering an action in tests never creates an approval request (`execute()` rethrows as before). `agent-testing` is untouched.

### Changes
- **agent-client**: new `ApprovalRequestCreator` (SaaS call); `execute()` approval branch; creator threaded `RemoteAgentClient → Collection → Action`; `createRemoteAgentClient` accepts `forestServer`.
- **mcp-server**: `agent-caller` wires `forestServer`; `execute-action` returns the "approval requested" message (caught **inside** the activity-log operation → log stays `completed`, never `failed`).

## ⚠️ Pending before merge
1. **SaaS endpoint spec** (path / payload / auth) for creating an approval request — used by the front today. `ApprovalRequestCreator.create()` carries a `TODO(PRD-288)` placeholder until confirmed.
2. Based on `feature/workflow-deterministic-ai-steps` (retarget to `main` once that merges).

## Tests
- agent-client: creator wired → `create()` called with correct args + `{ approvalRequested: true }`; no creator → rethrows, nothing created.
- mcp-server: approval path → LLM "approval requested" message, not an error.
- `agent-client` 330 ✓ · `mcp-server` 545 ✓ · lint 0 error.

> [!NOTE]
> ### Create approval requests on action trigger when required by role in workflow executor
> - `TriggerRecordActionStepExecutor` now supports dynamic action forms: Manual mode pauses for user input with no AI prefill; AI-assisted mode pre-fills and pauses for review; Full AI mode iteratively fills required fields and executes, falling back to review if fields remain empty.
> - When the agent rejects execution with `CustomActionRequiresApprovalError`, the executor calls `makeCreateApprovalRequest` to POST to `/api/action-approvals` on the Forest server, returning `{ approvalRequested: true }` instead of throwing.
> - `AgentHttpError` replaces generic stringified errors throughout `agent-client`, carrying `status`, parsed `body`, and raw `responseText` for structured downstream handling.
> - `preRecordedArgs.selectedRecordStepId` (stable BPMN step id) replaces the previous index-based `selectedRecordStepIndex` in both `TriggerActionStepDefinitionSchema` and `LoadRelatedRecordStepDefinitionSchema`.
> - `StepExecutionFormatters` gains a `formatTriggerAction` path that distinguishes pending-approval from executed submissions and highlights AI prefill vs. user edits.
> - Risk: `TriggerActionStepDefinitionSchema` no longer coerces invalid `executionType` values to `AutomatedWithConfirmation`; existing configs with unrecognized values will now fail validation.
>
> 
> #### Changes since #1708 opened
>
> - Added `ApprovalRequestCreationError` class and wrapped approval request creation in `Action.execute` method with try/catch block to throw the new error type [106db88]
> - Configured `createRemoteAgentClient` to accept and pass `forestServer` connection details including url, token, and renderingId to wire approval request creation [106db88]
> - Added test coverage for `ApprovalRequestCreationError` throwing behavior and `forestServer` configuration wiring [106db88]
> - Added inline comment in `makeCreateApprovalRequest` factory explaining why status attribute is set to null on creation [106db88]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 23b2b0a.</sup>
> 
>
>

### DIFF
--- a/packages/agent-client/src/approval-request-creator.ts
+++ b/packages/agent-client/src/approval-request-creator.ts
@@ -29,6 +29,7 @@ export default function makeCreateApprovalRequest(options: {
         data: {
           type: 'action-approvals',
           attributes: {
+            // Required null on create: the Forest server assigns the pending status itself.
             status: null,
             action_name: payload.actionName,
             collection_name: payload.collectionName,
--- a/packages/agent-client/src/approval-request-creator.ts
+++ b/packages/agent-client/src/approval-request-creator.ts
@@ -32,7 +32,8 @@ export default function makeCreateApprovalRequest(options: {
             action_name: payload.actionName,
             collection_name: payload.collectionName,
             record_ids: payload.recordIds,
-            inputs: payload.values,
+            // The Forest server stores inputs as a list of { name, value } (not a values map).
+            inputs: Object.entries(payload.values).map(([name, value]) => ({ name, value })),
           },
         },
       },
--- a/packages/agent-client/src/approval-request-creator.ts
+++ b/packages/agent-client/src/approval-request-creator.ts
@@ -1,6 +1,6 @@
 import { ServerUtils } from '@forestadmin/forestadmin-client';
 
-export type ApprovalRequestInput = { name: string; type: string; value: unknown };
+export type ApprovalRequestInput = { name: string; type: string | string[]; value: unknown };
 
 export type ApprovalRequestPayload = {
   collectionName: string;
--- a/packages/agent-client/src/approval-request-creator.ts
+++ b/packages/agent-client/src/approval-request-creator.ts
@@ -1,10 +1,12 @@
 import { ServerUtils } from '@forestadmin/forestadmin-client';
 
+export type ApprovalRequestInput = { name: string; type: string; value: unknown };
+
 export type ApprovalRequestPayload = {
   collectionName: string;
   actionName: string;
   recordIds: (string | number)[];
-  values: Record<string, unknown>;
+  inputs: ApprovalRequestInput[];
 };
 
 export type CreateApprovalRequest = (payload: ApprovalRequestPayload) => Promise<void>;
@@ -32,8 +34,7 @@ export default function makeCreateApprovalRequest(options: {
             action_name: payload.actionName,
             collection_name: payload.collectionName,
             record_ids: payload.recordIds,
-            // The Forest server stores inputs as a list of { name, value } (not a values map).
-            inputs: Object.entries(payload.values).map(([name, value]) => ({ name, value })),
+            inputs: payload.inputs,
           },
         },
       },
--- a/packages/agent-client/src/approval-request-creator.ts
+++ b/packages/agent-client/src/approval-request-creator.ts
@@ -11,7 +11,6 @@ export type ApprovalRequestPayload = {
 
 export type CreateApprovalRequest = (payload: ApprovalRequestPayload) => Promise<void>;
 
-// forestadmin-server private-api: POST /api/action-approvals (status null = creation).
 const APPROVAL_REQUEST_PATH = '/api/action-approvals';
 
 export default function makeCreateApprovalRequest(options: {
--- a/packages/agent-client/src/approval-request-creator.ts
+++ b/packages/agent-client/src/approval-request-creator.ts
@@ -7,25 +7,23 @@ export type ApprovalRequestPayload = {
   values: Record<string, unknown>;
 };
 
+export type CreateApprovalRequest = (payload: ApprovalRequestPayload) => Promise<void>;
+
 // forestadmin-server private-api: POST /api/action-approvals (status null = creation).
 const APPROVAL_REQUEST_PATH = '/api/action-approvals';
 
-export default class ApprovalRequestCreator {
-  constructor(
-    private readonly options: {
-      forestServerUrl: string;
-      forestServerToken: string;
-      renderingId: number | string;
-    },
-  ) {}
-
-  async create(payload: ApprovalRequestPayload): Promise<void> {
+export default function makeCreateApprovalRequest(options: {
+  forestServerUrl: string;
+  forestServerToken: string;
+  renderingId: number | string;
+}): CreateApprovalRequest {
+  return async payload => {
     await ServerUtils.queryWithBearerToken({
-      forestServerUrl: this.options.forestServerUrl,
-      bearerToken: this.options.forestServerToken,
+      forestServerUrl: options.forestServerUrl,
+      bearerToken: options.forestServerToken,
       method: 'post',
       path: APPROVAL_REQUEST_PATH,
-      headers: { 'forest-rendering-id': String(this.options.renderingId) },
+      headers: { 'forest-rendering-id': String(options.renderingId) },
       body: {
         data: {
           type: 'action-approvals',
@@ -39,5 +37,5 @@ export default class ApprovalRequestCreator {
         },
       },
     });
-  }
+  };
 }
--- a/packages/agent-client/src/approval-request-creator.ts
+++ b/packages/agent-client/src/approval-request-creator.ts
@@ -5,31 +5,36 @@ export type ApprovalRequestPayload = {
   actionName: string;
   recordIds: (string | number)[];
   values: Record<string, unknown>;
-  roleIdsAllowedToApprove?: number[];
 };
 
-// TODO(PRD-288 / Brice): confirm the SaaS endpoint + payload used by the front to create an
-// approval request after a 403. The front already calls it; once known, replace this placeholder.
-const APPROVAL_REQUEST_PATH = '/liana/action-approval-requests';
+// forestadmin-server private-api: POST /api/action-approvals (status null = creation).
+const APPROVAL_REQUEST_PATH = '/api/action-approvals';
 
 export default class ApprovalRequestCreator {
-  constructor(private readonly options: { forestServerUrl: string; bearerToken: string }) {}
+  constructor(
+    private readonly options: {
+      forestServerUrl: string;
+      forestServerToken: string;
+      renderingId: number | string;
+    },
+  ) {}
 
   async create(payload: ApprovalRequestPayload): Promise<void> {
     await ServerUtils.queryWithBearerToken({
       forestServerUrl: this.options.forestServerUrl,
-      bearerToken: this.options.bearerToken,
+      bearerToken: this.options.forestServerToken,
       method: 'post',
       path: APPROVAL_REQUEST_PATH,
+      headers: { 'forest-rendering-id': String(this.options.renderingId) },
       body: {
         data: {
-          type: 'action-approval-requests',
+          type: 'action-approvals',
           attributes: {
-            collection_name: payload.collectionName,
+            status: null,
             action_name: payload.actionName,
-            ids: payload.recordIds,
-            values: payload.values,
-            role_ids_allowed_to_approve: payload.roleIdsAllowedToApprove,
+            collection_name: payload.collectionName,
+            record_ids: payload.recordIds,
+            inputs: payload.values,
           },
         },
       },
--- a/packages/agent-client/src/approval-request-creator.ts
+++ b/packages/agent-client/src/approval-request-creator.ts
@@ -1,0 +1,38 @@
+import { ServerUtils } from '@forestadmin/forestadmin-client';
+
+export type ApprovalRequestPayload = {
+  collectionName: string;
+  actionName: string;
+  recordIds: (string | number)[];
+  values: Record<string, unknown>;
+  roleIdsAllowedToApprove?: number[];
+};
+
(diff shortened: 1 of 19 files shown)


Scra3:
### Code review

Ran the standard passes plus a ponytail (over-engineering) pass and a skeptic pass. No blocking bugs.

**Skeptic note (flagged, then dismissed).** Multiple passes flagged `recordIds: this.ids` as possibly `undefined` for global actions, while `record_ids` is `Joi.array().required()` server-side. False positive: `Collection.action()` is the only production path and always builds `ids` as an array (`[]` at minimum), so `record_ids: []` is sent and validates.

https://github.com/ForestAdmin/agent-nodejs/blob/9f90e140caf1ff6d4ff1964c6c92f5620ff1a7c5/packages/agent-client/src/domains/action.ts#L121-L128

**Non-blocking:**

1. The Forest server URL is re-derived from `process.env.FOREST_SERVER_URL` here, duplicating `cli.ts` and bypassing a `forestServerUrl` configured via server options. For self-hosted hosts the approval POST could hit a different host than the activity-log calls.

https://github.com/ForestAdmin/agent-nodejs/blob/9f90e140caf1ff6d4ff1964c6c92f5620ff1a7c5/packages/mcp-server/src/utils/agent-caller.ts#L38-L44

2. `approvalRequestCreator.create()` is not wrapped, so a SaaS network/5xx failure escapes `execute()` as a raw error (outside the typed `ActionRequiresApprovalError` contract). Acceptable, but the failure path is undocumented.

https://github.com/ForestAdmin/agent-nodejs/blob/9f90e140caf1ff6d4ff1964c6c92f5620ff1a7c5/packages/agent-client/src/domains/action.ts#L120-L131

3. Layering: a prior review on these files ([#1403](https://github.com/ForestAdmin/agent-nodejs/pull/1403#discussion_r2673859908)) set the convention "agent-client throws, mcp-server shapes the result". Here `execute()` swallows the approval error and returns `{ approvalRequested: true }`, gated on whether a creator was wired. Deliberate (mutualizing for the future workflow-executor), but the dual contract for the same 403 is worth a short comment.

**Ponytail:** `ApprovalRequestCreator` is a class with a single method and a single caller; a plain `createApprovalRequest(options, payload)` function would do the same with less surface. Minor.

https://github.com/ForestAdmin/agent-nodejs/blob/9f90e140caf1ff6d4ff1964c6c92f5620ff1a7c5/packages/agent-client/src/approval-request-creator.ts#L13-L23

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/agent-client/src/domains/action.ts
```diff
@@ -108,7 +115,26 @@ export default class Action {
         body: requestBody,
       });
     } catch (error) {
-      throw toActionError(error);
+      const mapped = toActionError(error);
+
+      if (mapped instanceof ActionRequiresApprovalError && this.createApprovalRequest) {
+        const inputs = this.fieldsFormStates.getFields().map(field => ({
```
hercemer42, line 125: **Claude Opus 4.8 — [Should fix]** Approval `inputs` are built from `getFields()`, which returns *all* fields including ones whose value is `undefined`, whereas the executed path (line 103) uses `getFieldValues()` that drops `undefined`. Confirm the server's `/api/action-approvals` validation accepts entries for unfilled fields — otherwise an action with optional, unfilled fields could have its approval request rejected. Mirror `getFieldValues()` semantics if so.

packages/agent-client/src/domains/action.ts
```diff
@@ -108,7 +115,26 @@ export default class Action {
...
+          name: field.getName(),
+          type: field.getType(),
+          value: field.getValue(),
+        }));
+
+        await this.createApprovalRequest({
```
hercemer42: **Claude Opus 4.8 — [Preferential]** If this POST fails, the error propagates and the LLM sees only the transport error — the fact that the action is approval-gated is lost, and "couldn't file the approval request" becomes indistinguishable from "the action failed". Consider wrapping to disambiguate the two failure modes.

packages/agent-client/src/approval-request-creator.ts
```diff
@@ -0,0 +1,42 @@
...
+      forestServerUrl: options.forestServerUrl,
+      bearerToken: options.forestServerToken,
+      method: 'post',
+      path: APPROVAL_REQUEST_PATH,
+      headers: { 'forest-rendering-id': String(options.renderingId) },
+      body: {
+        data: {
+          type: 'action-approvals',
+          attributes: {
+            status: null,
```
hercemer42, line 33: **Claude Opus 4.8 — [Preferential]** `status: null` is an unexplained literal. A one-line note on the server contract (e.g. the server assigns the pending status; `null` is required on create) would save a future reader the "is this needed?" question.

packages/agent-client/test/index.test.ts
```diff
@@ -61,6 +61,20 @@ describe('createRemoteAgentClient', () => {
...
+      url: 'https://api.example.com',
+      token: 'test-token',
+      forestServer: {
+        serverUrl: 'https://api.forestadmin.com',
+        serverToken: 'server-token',
+        renderingId: 42,
+      },
+    });
+
+    expect(client).toBeInstanceOf(RemoteAgentClient);
```
hercemer42: **Claude Opus 4.8 — [Violates conventions]** This test ("build a client when forestServer is provided") asserts only `toBeInstanceOf(RemoteAgentClient)` — identical to the tests above that pass no `forestServer`, so it passes even if the `forestServer` → `createApprovalRequest` wiring is dropped entirely. CLAUDE.md: *test name = assertion / avoid weak assertions*. Assert the wiring is observable (e.g. spy `makeCreateApprovalRequest` and verify it's called with the forestServer fields). Semantic — not linter-catchable.

packages/mcp-server/test/utils/agent-caller.test.ts
```diff
@@ -264,4 +264,28 @@ describe('buildClientWithActions', () => {
...
+          renderingId: 456,
+          environmentApiEndpoint: 'http://localhost:3310',
+          forestServerToken: 'server-token',
+        },
+      },
+    } as unknown as RequestHandlerExtra<ServerRequest, ServerNotification>;
+
+    const result = await buildClientWithActions(request, mockForestServerClient);
+
+    expect(result.rpcClient).toBeDefined();
```
hercemer42: **Claude Opus 4.8 — [Violates conventions]** Asserts only that `rpcClient` is defined — the same as the tests without forest-server fields — so it doesn't verify `forestServer` was actually passed, and there is no negative test for the guard (`forestServerUrl && forestServerToken && renderingId != null`). Mock `createRemoteAgentClient` and assert it's called with `forestServer: { serverUrl, serverToken, renderingId }` when complete, and `undefined` when a field is missing. (CLAUDE.md weak-assertion rule; semantic, not linter-catchable.)

packages/mcp-server/test/tools/execute-action.test.ts
```diff
@@ -308,6 +309,31 @@ describe('declareExecuteActionTool', () => {
...
+      const result = await registeredToolHandler(
+        { collectionName: 'users', actionName: 'refund', recordIds: [1] },
+        mockExtra,
+      );
+
+      expect(result).toEqual({
+        content: [{ type: 'text', text: expect.stringContaining('Approval requested') }],
+      });
+      // It is not an error → the activity log is not marked failed.
+      expect((result as { isError?: boolean }).isError).toBeUndefined();
```
hercemer42: **Claude Opus 4.8 — [Preferential]** `isError` is never set by either branch of the handler, and `withActivityLog` is mocked in this suite, so this assertion (and the "activity log is not marked failed" comment) is tautological — it cannot fail. The `stringContaining('Approval requested')` assertion above is the meaningful one. Drop this line, or assert the activity-log status against the real `withActivityLog` if the intent is to prove the log stays completed.

hercemer42 (changes requested):

packages/agent-client/src/domains/action.ts
Scra3: Fixed in 106db8807: the approval POST is now wrapped in `ApprovalRequestCreationError` (carries the original `cause`), so an approval-gate filing failure is distinguishable from the action itself failing.

packages/agent-client/src/approval-request-creator.ts
Scra3: Fixed in 106db8807: added a comment on the create contract (the Forest server assigns the pending status itself; `null` is required on create).

packages/agent-client/test/index.test.ts
Scra3: Fixed in 106db8807: the test now mocks `makeCreateApprovalRequest` and asserts it is called with `{ forestServerUrl, forestServerToken, renderingId }` when `forestServer` is provided, plus a negative test asserting it is not called when omitted.

packages/mcp-server/test/utils/agent-caller.test.ts
Scra3: Fixed in 106db8807: now asserts `createRemoteAgentClient` is called with `forestServer: { serverUrl, serverToken, renderingId }` when complete, and a negative test asserting `forestServer: undefined` when `forestServerToken` is missing (the guard's silent-disable branch).

packages/mcp-server/test/tools/execute-action.test.ts
Scra3: Fixed in 106db8807: dropped the tautological `isError` assertion and its misleading comment; the `stringContaining('Approval requested')` assertion remains.

packages/agent-client/src/domains/action.ts
Scra3: Not changing: the server validator (forestadmin-server #8327) declares `value: Joi.any()` and explicitly tests "allow a list field type and a missing value", so unfilled fields are accepted (and `JSON.stringify` omits `value: undefined` on the wire anyway). Keeping `getFields()` also preserves the {name, type} of unfilled fields for the approver.

packages/agent-client/src/domains/collection.ts
```diff
@@ -45,11 +49,13 @@ export default class Collection extends CollectionChart {
...
     const action = new Action(
       this.name,
+      actionName,
       this.httpRequester,
       actionInfo.endpoint,
       fieldsFormStates,
       ids,
       actionInfo.id,
+      this.createApprovalRequest,
     );
```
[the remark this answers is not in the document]
Scra3: Not a reachable bug here: the endpoint map is built in schema-fetcher.ts:132 keyed by `action.name`, with `.name` set to the same `action.name` — so the lookup key always equals `actionInfo.name` and no alias divergence can occur (an alias key would fail getActionInfo). Leaving as-is.

packages/agent-client/src/domains/action.ts
```diff
@@ -108,7 +119,30 @@ export default class Action {
...
       });
     } catch (error) {
-      throw toActionError(error);
+      const mapped = toActionError(error);
+
+      if (mapped instanceof ActionRequiresApprovalError && this.createApprovalRequest) {
+        const inputs = this.fieldsFormStates.getFields().map(field => ({
+          name: field.getName(),
+          type: field.getType(),
+          value: field.getValue(),
```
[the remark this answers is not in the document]
Scra3: Not changing: the server validator (forestadmin-server #8327) declares `value: Joi.any()`, which accepts a missing/undefined value as well as null, and it has an explicit test for an entry with a missing value — so the serialized `{name, type}` entry is valid, not malformed. Coercing undefined → null would also change the semantics from "field not provided" to "field explicitly null", which is less faithful to the executed path (`getFieldValues` omits unfilled fields entirely).

hercemer42 (approved):

PR #1716, opened 2026-06-26
fix(agent): forward workflow executor responses as raw bytes

PMerlet:
## Problem

In production, opening a workflow run failed intermittently with `net::ERR_CONTENT_DECODING_FAILED` on `GET /forest/_internal/executor/runs/:id`, surfacing in the app as a silent `TypeError: Failed to fetch` (the workflow side panel never loaded).

Server-side everything returned `200`. The corruption was in the **agent's workflow-executor proxy**.

## Root cause

The chain is: `Browser → agent (this proxy) → nginx (Elastic Beanstalk) → executor (koa)`. The Beanstalk nginx gzips the executor's JSON response once it crosses the gzip size threshold — which is why only **large** runs broke (small runs aren't compressed).

`WorkflowExecutorProxyRoute.forwardRequest` did:

```ts
const raw = Buffer.concat(chunks).toString('utf-8'); // ← corrupts gzip bytes
let parsed; try { parsed = JSON.parse(raw); } catch { parsed = raw; }
resolve({ status, body: parsed, headers: res.headers }); // ← still forwards Content-Encoding: gzip
```

So a gzip-compressed body was decoded as UTF-8 (destroying the gzip stream), re-emitted, while the executor's `Content-Encoding: gzip` header was forwarded untouched. The browser tried to gunzip a non-gzip body → `ERR_CONTENT_DECODING_FAILED`. The `// transparent proxy` intent in the code was not actually met: it lossily round-tripped `bytes → utf-8 → JSON → re-serialize`.

## Fix

Forward the executor body as **raw bytes** (`Buffer.concat(chunks)`). Koa sends the `Buffer` verbatim and recomputes `Content-Length`; `Content-Type`/`Content-Encoding` are forwarded untouched, so a compressed body stays consistent with its encoding header. This also makes the proxy genuinely transparent for any content type, not just JSON.

## Testing

- Added a regression test for a gzip-compressed executor response. It **fails on the previous behavior** (`gunzip` throws — exactly the production `ERR_CONTENT_DECODING_FAILED`) and **passes** with this fix.
- Full `workflow-executor-proxy` suite: **25/25 green**, eslint clean.
- The existing body assertions were updated to decode the now-Buffer body (uncompressed JSON, 204, text/plain, POST trigger, error passthrough — all still covered).

## Notes

A temporary monkey-patch of the compiled file was deployed on `forestadmin-server` (Dockerfile `COPY`) to validate on staging; it should be removed once this ships and the `@forestadmin/agent` bump lands there.

> [!NOTE]
> ### Fix `WorkflowExecutorProxyRoute.forwardRequest` to forward response bodies as raw bytes
> Previously, `forwardRequest` decoded the response body as UTF-8 and attempted a JSON parse, which corrupted gzip-compressed responses that were still forwarded with `Content-Encoding: gzip`. The method now returns the raw concatenated `Buffer` instead, preserving binary and encoded bodies. Tests are updated to read body content from a `Buffer` and a new test verifies that gzip-compressed responses are forwarded intact and remain decodable by clients.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 52f34d7.</sup>
> 
>

### DIFF
--- a/packages/agent/src/routes/workflow/workflow-executor-proxy.ts
+++ b/packages/agent/src/routes/workflow/workflow-executor-proxy.ts
@@ -124,12 +124,8 @@ export default class WorkflowExecutorProxyRoute extends BaseRoute {
         const chunks: Uint8Array[] = [];
         res.on('data', chunk => chunks.push(chunk));
         res.on('end', () => {
-          // Forward the executor body as raw bytes. Decoding it as UTF-8 (and JSON-parsing)
-          // corrupted any compressed payload while still forwarding the executor's
-          // `Content-Encoding: gzip` header, so the browser failed with
-          // net::ERR_CONTENT_DECODING_FAILED. Koa sends the Buffer verbatim and recomputes
-          // Content-Length; Content-Type/Content-Encoding are forwarded untouched, keeping a
-          // compressed body and its encoding header consistent.
+          // Raw bytes: decoding as UTF-8 corrupted gzipped bodies while still forwarding
+          // Content-Encoding: gzip, breaking the browser with ERR_CONTENT_DECODING_FAILED.
           resolve({
             status: res.statusCode ?? HttpCode.InternalServerError,
             body: Buffer.concat(chunks),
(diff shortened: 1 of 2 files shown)


christophebrun-forest (approved):
Ok for me
OK with the note about Monkey Patch

PR #1724, opened 2026-06-29
fix(workflow-executor): harden OAuth reauth and token write-back [PRD-692]

hercemer42:
## What

Two correctness fixes on the executor OAuth2 MCP runtime, surfaced in the PR2 (#1665) review and deliberately deferred to land before the feature is enabled (PRD-627). Both touch the store-port surface, so they're bundled.

- **Reauth pause stranding `idempotencyPhase: 'executing'` (HIGH)** — a reauth pause left the write-ahead marker set, so the resumed step was rejected with "Step execution was interrupted". Added `deleteStepExecution(runId, stepIndex)` to the `RunStore` port (both store impls) and clear the marker on the `OAuthReauthRequiredError` path.
- **Refresh write-back resurrecting a deleted credential (MED)** — the rotated-token write-back now uses an atomic `updateIfPresent` (`UPDATE … WHERE`) on the credentials store instead of get-then-upsert, so a concurrent disconnect (DELETE) cannot be resurrected. `upsert` stays for the consent deposit path.

## Audit-log behavior (intentional)

A re-auth pause's MCP tool call genuinely 401'd, so it surfaces as a **`failed`** activity-log entry. The audit records the *call* (which failed); the run history records the *step* (which pauses with `needs-oauth-reauth` and resumes). PR2's original "spurious failure" finding was reconsidered and dropped: the failure is real and worth surfacing for observability, and the resumed run logs its own entry — so no audit special-casing. (The earlier `paused`-status follow-up, PRD-693, is cancelled for the same reason.)

## Stacking

Stacked on PR2 (#1665) — this PR targets `feat/prd-367-pr2-executor-oauth-runtime`, not `main`, because the OAuth runtime code it fixes only exists on that branch. Once PR2 merges, rebase and repoint this PR to `main`.

## Tests

- Behavioral tests: resume-after-reauth (idempotency marker cleared), a re-auth pause audited as `failed` while the step pauses, and the write-back race (a concurrently-deleted credential is not resurrected).
- `deleteStepExecution` covered in both run stores; `updateIfPresent` covered in both credential stores (updates in place / does not insert when absent).
- Impacted suites green; `tsc` + eslint clean. Postgres `*.pg.test.ts` left to CI.

fixes PRD-692

> [!NOTE]
> ### Harden OAuth re-auth pause state and rotated refresh token write-back in `McpStepExecutor`
> - When an OAuth re-auth pause is triggered, `McpStepExecutor.doExecute` now calls a new `clearReauthPauseState()` helper before returning `awaiting-input`, preventing later resumes from being rejected as interrupted due to a stale `executing` write-ahead marker.
> - `clearReauthPauseState` clears only the `idempotencyPhase` when `pendingData` exists (preserving confirmation-flow state), or deletes the step execution record entirely when there is no `pendingData`.
> - `OAuthTokenService.persistRotatedRefreshToken` now calls `store.updateIfPresent(id, ...)` instead of `store.upsert(...)`, making rotated token write-backs a no-op if the credential row was deleted or replaced since the token was read.
> - Both `McpOAuthCredentialsStore` and `RunStore` interfaces gain new methods (`updateIfPresent` and `deleteStepExecution`) implemented across the database and in-memory stores.
> - Risk: resumes that previously succeeded despite a stale write-ahead marker will now require the OAuth re-auth flow to complete before proceeding.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 50b58c5.</sup>
> 
>

### DIFF
--- a/packages/workflow-executor/src/executors/mcp-step-executor.ts
+++ b/packages/workflow-executor/src/executors/mcp-step-executor.ts
@@ -158,7 +158,6 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
             stepIndex: this.context.stepIndex,
             idempotencyPhase: 'executing',
           }),
-        isNonFailure: error => error instanceof OAuthReauthRequiredError,
       },
     );
 
--- a/packages/workflow-executor/src/executors/mcp-step-executor.ts
+++ b/packages/workflow-executor/src/executors/mcp-step-executor.ts
@@ -91,7 +91,7 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
       // An unrefreshable OAuth credential pauses the step for re-authentication rather than failing
       // it. Clear the write-ahead marker so the resumed step is not rejected as interrupted.
       if (error instanceof OAuthReauthRequiredError) {
-        await this.context.runStore.deleteStepExecution(this.context.runId, this.context.stepIndex);
+        await this.clearReauthPauseState();
 
         return this.buildOutcomeResult({
           status: 'awaiting-input',
@@ -100,6 +100,33 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
       }
 
       throw error;
+    }
+  }
+
+  // Drop the 'executing' write-ahead marker left by beforeCall so the resumed step is not rejected
+  // as interrupted. A confirmation-flow record carries the user-approved pendingData — preserve it
+  // (clear only the phase) so resume replays that exact call instead of re-selecting a tool; a
+  // record with no pendingData would mis-route the resumed step into the confirmation flow, so
+  // delete it. Best-effort: a store error here must not turn the pause into a hard failure.
+  private async clearReauthPauseState(): Promise<void> {
+    try {
+      const existing = await this.findPendingExecution<McpStepExecutionData>('mcp');
+      if (!existing) return;
+
+      if (existing.pendingData) {
+        await this.context.runStore.saveStepExecution(this.context.runId, {
+          ...existing,
+          idempotencyPhase: undefined,
+        });
+      } else {
+        await this.context.runStore.deleteStepExecution(this.context.runId, this.context.stepIndex);
+      }
+    } catch (cause) {
+      this.context.logger('Error', 'Failed to clear re-auth pause state; resume may need a retry', {
+        runId: this.context.runId,
+        stepIndex: this.context.stepIndex,
+        cause: cause instanceof Error ? cause.message : String(cause),
+      });
     }
   }
 
--- a/packages/workflow-executor/src/executors/mcp-step-executor.ts
+++ b/packages/workflow-executor/src/executors/mcp-step-executor.ts
@@ -106,24 +106,16 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
   // Keep a confirmation-flow record's approved pendingData (clear only the marker) so resume replays
   // it; delete a pendingData-less record, which would otherwise mis-route resume into confirmation.
   private async clearReauthPauseState(): Promise<void> {
-    try {
-      const existing = await this.findPendingExecution<McpStepExecutionData>('mcp');
-      if (!existing) return;
+    const existing = await this.findPendingExecution<McpStepExecutionData>('mcp');
+    if (!existing) return;
 
-      if (existing.pendingData) {
-        await this.context.runStore.saveStepExecution(this.context.runId, {
-          ...existing,
-          idempotencyPhase: undefined,
-        });
-      } else {
-        await this.context.runStore.deleteStepExecution(this.context.runId, this.context.stepIndex);
-      }
-    } catch (cause) {
-      this.context.logger('Error', 'Failed to clear re-auth pause state; resume may need a retry', {
-        runId: this.context.runId,
-        stepIndex: this.context.stepIndex,
-        cause: cause instanceof Error ? cause.message : String(cause),
+    if (existing.pendingData) {
+      await this.context.runStore.saveStepExecution(this.context.runId, {
+        ...existing,
+        idempotencyPhase: undefined,
       });
+    } else {
+      await this.context.runStore.deleteStepExecution(this.context.runId, this.context.stepIndex);
     }
   }
 
--- a/packages/workflow-executor/src/executors/mcp-step-executor.ts
+++ b/packages/workflow-executor/src/executors/mcp-step-executor.ts
@@ -88,8 +88,11 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
     try {
       return await this.runStep();
     } catch (error) {
-      // An unrefreshable OAuth credential pauses the step for re-authentication rather than failing it.
+      // An unrefreshable OAuth credential pauses the step for re-authentication rather than failing
+      // it. Clear the write-ahead marker so the resumed step is not rejected as interrupted.
       if (error instanceof OAuthReauthRequiredError) {
+        await this.context.runStore.deleteStepExecution(this.context.runId, this.context.stepIndex);
+
         return this.buildOutcomeResult({
           status: 'awaiting-input',
           awaitingInputReason: error.awaitingInputReason,
@@ -155,6 +158,7 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
             stepIndex: this.context.stepIndex,
             idempotencyPhase: 'executing',
           }),
+        isNonFailure: error => error instanceof OAuthReauthRequiredError,
       },
     );
 
--- a/packages/workflow-executor/src/executors/mcp-step-executor.ts
+++ b/packages/workflow-executor/src/executors/mcp-step-executor.ts
@@ -103,11 +103,8 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
     }
   }
 
-  // Drop the 'executing' write-ahead marker left by beforeCall so the resumed step is not rejected
-  // as interrupted. A confirmation-flow record carries the user-approved pendingData — preserve it
-  // (clear only the phase) so resume replays that exact call instead of re-selecting a tool; a
-  // record with no pendingData would mis-route the resumed step into the confirmation flow, so
-  // delete it. Best-effort: a store error here must not turn the pause into a hard failure.
+  // Keep a confirmation-flow record's approved pendingData (clear only the marker) so resume replays
+  // it; delete a pendingData-less record, which would otherwise mis-route resume into confirmation.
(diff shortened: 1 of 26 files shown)


packages/workflow-executor/test/executors/mcp-step-executor.test.ts
```diff
@@ -1247,3 +1249,108 @@ describe('McpStepExecutor — OAuth2 tool-call re-authentication', () => {
...
+        context,
+        [tool],
+        'srv',
+        reloadWithFreshAuth,
+      ).execute();
+
+      // THEN the activity entry is opened but a normal re-auth pause is not recorded as a failure.
+      expect(result.stepOutcome.status).toBe('awaiting-input');
+      expect(activityLogPort.createPending).toHaveBeenCalledTimes(1);
+      expect(activityLogPort.markFailed).not.toHaveBeenCalled();
```
hercemer42: **Claude Opus 4.8** — **[Should fix]** This test asserts `markFailed` was *not* called but never asserts `markSucceeded` *was* called, so a port that called neither would also pass (weak coverage per the meaningful-tests convention: verify exact outcomes, not just that something wasn't called). Behavior is covered by the dedicated `activity-log.test.ts` unit test, so this is non-blocking — but adding `expect(activityLogPort.markSucceeded).toHaveBeenCalledWith({ id: 'log-1', index: '0' })` here would make the test match its name.

packages/workflow-executor/test/executors/mcp-step-executor.test.ts
```diff
@@ -1247,3 +1249,108 @@ describe('McpStepExecutor — OAuth2 tool-call re-authentication', () => {
     expect(result.stepOutcome.status).toBe('error');
   });
 });
+
+// PRD-692 hardening — three correctness findings on the executor OAuth runtime that are NOT yet
+// fixed (this is the TDD step). Each test below is RED against the current code and must go green
```
hercemer42: **Claude Opus 4.8** — **[Violates conventions]** This comment narrates iteration history ('this is the TDD step', 'RED against the current code', 'must go green once the fix lands') — against the convention *avoid comments that reference iteration history* — and is now stale post-merge (the tests are green, and finding #3's test lives in token-service.test.ts). Suggest dropping the TDD framing and keeping only the per-scenario why. Not linter-catchable (semantic), so it won't be flagged by CI.

packages/workflow-executor/src/oauth/token-service.ts
```diff
@@ -186,6 +186,11 @@ export default class OAuthTokenService {
     try {
       const encrypted = this.encryption.encrypt(refreshToken);
 
+      // A disconnect (DELETE) may have landed after the grant read; re-check before writing so the
```
hercemer42: **Claude Opus 4.8** — **[Preferential]** The re-check narrows but doesn't fully close the race: a DELETE landing between this `get` and the `upsert` below can still resurrect the row (self-corrects on next connect). The function already documents its lossy write-back fallback just below — a half-sentence noting the residual window would match that bar. (Full closure would need an atomic update-only-if-present store method.)

packages/workflow-executor/test/executors/mcp-step-executor.test.ts
hercemer42: **Claude Opus 4.8** — Agreed, fixed in 92cde36f. Added `expect(activityLogPort.markSucceeded).toHaveBeenCalledTimes(1)` so the test asserts the pause is positively closed as succeeded, not merely not-failed — a no-op port no longer passes.

packages/workflow-executor/test/executors/mcp-step-executor.test.ts
hercemer42: **Claude Opus 4.8** — Agreed, fixed in 92cde36f. Rewrote the block comment to state only the behavioral intent (clear the marker + record the pause as a non-failure), dropping the TDD/RED framing. Also removed the `PRD-692` and `finding #N` references from the surrounding describe titles, since the same no-iteration/ticket-refs convention applies there.

packages/workflow-executor/src/oauth/token-service.ts
hercemer42: **Claude Opus 4.8** — Declining (preferential, intentional omission). The residual get→upsert window is a very-low-probability race — a DELETE landing in the microseconds between the existence re-check and the upsert — and it self-corrects on the next connect. It's already documented in the PR description and the deferred-work notes, and the inline comment is at the 2-line convention limit; a third line of caveat would read as over-documenting a corner that's tracked elsewhere. Full closure (an atomic update-only-if-present store method) is the deferred follow-up. Leaving the comment focused on why the re-check exists.

hercemer42: **Claude Opus 4.8** — Update: we went with the fully-atomic fix instead of the re-check. The write-back now uses an atomic `updateIfPresent` (`UPDATE … WHERE`) on the credentials store rather than get-then-upsert, so the residual `get→upsert` window this thread flagged is closed entirely — there's no window left to note. Done in 0905ddf4. The separate `completed`-vs-`paused` audit-log semantics question is tracked in PRD-693.

packages/workflow-executor/test/executors/mcp-step-executor.test.ts
hercemer42: **Claude Opus 4.8** — Moot now: finding #2 is reverted (commit 26e01588). On reflection the MCP call genuinely 401'd, so it should surface as a `failed` audit entry rather than be remapped to `completed` — the failure is real and useful for observability, and the step's recoverable pause is recorded separately in the run history (`needs-oauth-reauth`). The `isNonFailure`/`markSucceeded` special-casing is gone, so this assertion no longer applies; the test now asserts the call is audited as `failed` while the step pauses. PRD-693 (the `paused`-status follow-up) is cancelled.

packages/workflow-executor/src/executors/mcp-step-executor.ts
```diff
@@ -88,8 +88,11 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
     try {
       return await this.runStep();
     } catch (error) {
```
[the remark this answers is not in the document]
hercemer42: **Claude Opus 4.8** — Fixed in c7ac43a9, with one correction: clearing only the phase (or leaving the row intact) would break the FullyAutomated path — a record with no `pendingData` mis-routes the resumed step into `handleConfirmationFlow` ("no pending data"). So it's differentiated: a confirmation-flow record (has the approved `pendingData`) is preserved with only the marker cleared, so resume replays that exact approved call; a record with no `pendingData` (FullyAutomated) is deleted so resume re-runs cleanly. Added a test asserting the approved `pendingData` survives the pause.

packages/workflow-executor/src/executors/mcp-step-executor.ts
```diff
@@ -88,8 +88,11 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
...
-      // An unrefreshable OAuth credential pauses the step for re-authentication rather than failing it.
+      // An unrefreshable OAuth credential pauses the step for re-authentication rather than failing
+      // it. Clear the write-ahead marker so the resumed step is not rejected as interrupted.
       if (error instanceof OAuthReauthRequiredError) {
+        await this.context.runStore.deleteStepExecution(this.context.runId, this.context.stepIndex);
```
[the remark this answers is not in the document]
hercemer42: **Claude Opus 4.8** — Fixed in c7ac43a9. The cleanup is now best-effort: `clearReauthPauseState` wraps the store ops in try/catch and logs, so a store error still returns the `awaiting-input` outcome instead of throwing a generic step error. Added a test (delete rejects → the step still pauses).

packages/workflow-executor/src/stores/database-mcp-oauth-credentials-store.ts
```diff
@@ -196,6 +196,33 @@ export default class DatabaseMcpOAuthCredentialsStore implements McpOAuthCredent
     });
   }
 
+  async updateIfPresent(credential: McpOAuthCredentialInput): Promise<void> {
```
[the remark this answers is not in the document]
hercemer42: **Claude Opus 4.8** — Fixed in c7ac43a9. `updateIfPresent` now keys on the row `id` the caller read (`UPDATE … WHERE id = :id`), not `(user_id, mcp_server_id)`. A disconnect + re-authorize inserts a new row with a new id, so a stale in-flight write-back affects zero rows instead of clobbering the re-authorized credential. Added in-memory + DB store tests for the re-created-with-a-different-id case.

packages/workflow-executor/src/executors/mcp-step-executor.ts
```diff
@@ -100,6 +103,30 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
     }
   }
 
+  // Keep a confirmation-flow record's approved pendingData (clear only the marker) so resume replays
+  // it; delete a pendingData-less record, which would otherwise mis-route resume into confirmation.
+  private async clearReauthPauseState(): Promise<void> {
```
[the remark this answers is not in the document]
hercemer42: **Claude Opus 4.8** — Agreed, fixed in f0aefab9. Removed the best-effort try/catch from `clearReauthPauseState`: swallowing a cleanup failure left the `executing` marker in place, so the pause returned `awaiting-input` but could never resume (`checkIdempotency` rejects the stale marker). The store error now propagates to an ordinary step error — consistent with the executor's other store-failure paths. Note this **reverses the earlier finding on line 94**, which asked to swallow it; the two contradict, and propagating is the correct resolution.

packages/workflow-executor/src/executors/mcp-step-executor.ts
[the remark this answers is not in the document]
hercemer42: **Claude Opus 4.8** — Update: superseded by the line-108 finding and reversed in f0aefab9. The best-effort try/catch this finding asked for created a non-resumable pause — a swallowed cleanup failure leaves the `executing` marker, so resume is rejected with "interrupted". Propagating the store error as an ordinary step error (no try/catch) is the correct behavior.

Scra3 (commented):
praise: the `clearReauthPauseState` fix correctly clears the write-ahead 'executing' marker on the tool-call re-auth-pause path (delete for FullyAutomated, keep pendingData for confirmation flows), with resume-to-success and cleanup-failure coverage; the `updateIfPresent` id-scoped write-back also correctly stops a concurrent disconnect from resurrecting a deleted credential. No issues in this PR's diff. Heads-up: the SSRF blocker for this feature lives in the base branch (#1665, `oauth/refresh-grant.ts` — `fetch` follows redirects), and the evict/set race is also base-branch code.

PR #1728, opened 2026-06-30
feat(workflow-executor): add 3-way execution mode to the load related record step (PRD-148)

matthv:
## What

Execution logic for the 3-way **Execution** mode (Manual / AI-assisted / Full AI) on the workflow **Load Related Record** step. Part of PRD-148 — 3-PR set (executor + `ForestAdmin/forestadmin-server` orchestrator + `ForestAdmin/forestadmin` editor).

- **Manual** (`manual`): present the narrowed candidate list with **no AI** — no relation chooser, no record suggestion. Single-record relations (xToOne, or HasMany with 1 candidate) still pre-fill, since that's the only option, not an AI pick. The Zod schema drops `.catch(AutomatedWithConfirmation)` and adds `Manual`, so a `manual` value is no longer silently coerced back into AI prefill (same reasoning as TriggerAction).
- **AI-assisted** (`automated-with-confirmation`, default): AI suggests a record, user confirms (unchanged).
- **Full AI** (`fully-automated`): AI selects **and** loads. **Auto-skips** (persists `executionResult: { skipped: true }` + `success`) when there is no source record (`SourceRecordMissingError`, Full-AI only), an empty candidate list, or the AI judges none relevant (`select-record-by-content` returns `-1` via the new Full-AI-only `allowNone`).

`selectedRecordRef` is now optional on `LoadRelatedRecordStepExecutionData` (absent only on the no-source Full-AI skip); the wire serializer omits it instead of dereferencing `undefined`.

## Tests

97 tests in the executor suite. Added: Manual = no AI / no pre-selection (incl. relation switch), Manual single-record pre-fill, Full-AI skip on each of the 3 causes, await-mode no-source still errors (regression guard), serializer no-source round-trip.

## Validation

Validated end-to-end on the local stack (account → bills) across the 3 modes, plus a local multi-agent code review — **1 Critical found & fixed**: serializer crash (503) on the no-source Full-AI skip.

Relates to PRD-148.

> [!NOTE]
> ### Add Manual execution mode to `LoadRelatedRecordStepExecutor` with auto-skip for FullyAutomated
> - Adds `Manual` as a valid `executionType` for the load-related-record step; in Manual mode, AI is not called for relation selection or candidate suggestion — the first eligible relation is picked deterministically and no `suggestedRecord` is included.
> - In `FullyAutomated` mode, steps now auto-skip (via new `persistSkip`) when there is no source record, no linked xToOne record, or when the AI ranks all candidates as irrelevant (returns `-1`).
> - `selectBestRecordIndex` gains an `allowNone` option so AI can return `-1` only when the candidate list was not truncated, preventing forced selections.
> - `fetchRecordForRelation` and `fetchXToOneRecordRef` now return `null` on absence instead of throwing, with callers deciding skip vs. error behavior.
> - `selectedRecordRef` is made optional in `LoadRelatedRecordStepExecutionData` and the wire serializer skips the field when absent, fixing a serialization crash on auto-skipped steps.
> - Behavioral Change: `executionType` schema no longer coerces unknown values to `AutomatedWithConfirmation`; invalid values will now fail validation instead of being silently swallowed.
>
> 
> #### Changes since #1728 opened
>
> - Added 3-way execution mode allowing AI to decline record selection by returning -1 when no candidate satisfies the request [9ee1164]
> - Changed FullyAutomated execution mode to fall back to awaiting-input with `suggestNoRecord` flag instead of auto-skipping when no relevant record exists [9ee1164]
> - Added `suggestNoRecord` field to `LoadRelatedRecordPendingData` and updated `LoadRelatedRecordStepExecutionData` interface documentation [9ee1164]
> - Removed helper methods and unified record resolution logic through `collectCandidateIds` and `resolveAndLoadAutomatic` [9ee1164]
> - Updated test suite to validate new awaiting-input fallback behavior and AI -1 guidance [9ee1164]
> - Implemented AI failure degradation logic throughout LoadRelatedRecordStepExecutor to convert AI unavailability or errors into deterministic manual fallback execution [f90b868]
> - Implemented confidence-based record selection for fully automated mode where low-confidence AI picks trigger user confirmation instead of auto-loading [f90b868]
> - Modified LoadRelatedRecordStepExecutor.refreshCandidatesForField to rebuild pendingData from scratch when switching target relations and conditionally set suggestNoRecord only when AI was attempted and returned no suggestion [f90b868]
> - Updated select-record-by-content AI tool schema to accept and return a confident boolean parameter indicating AI confidence in record selection [f90b868]
> - Updated and expanded tests in load-related-record-step-executor.test.ts to verify AI degradation behavior, low-confidence handling, and stale state clearing [f90b868]
> - Created step-definition.test.ts to validate LoadRelatedRecordStepDefinitionSchema parsing of executionType enum values [f90b868]
> - Added test verifying `LoadRelatedRecordStepExecutor` degrades to manual candidate selection when AI ranking fails during field switch [71542cd]
> - Added test verifying `LoadRelatedRecordStepExecutor` propagates non-AI errors during field switch instead of degrading [71542cd]
> - Updated inline comment in `LoadRelatedRecordStepExecutor.resolveAndLoadAutomatic` method [44e23be]
> - Renamed test case for `LoadRelatedRecordStepExecutor` FullyAutomated BelongsTo functionality [44e23be]
> - Appended explanatory comment to ESLint directive [08c1949]
> - Revised documentation comments across `AiAssistUnavailableError` class and multiple methods in `LoadRelatedRecordStepExecutor` including `doExecute`, `refreshCandidatesForField`, `degradeToManualAwaitInput`, `selectBestFromRelatedData`, `withAiAssist`, and `selectBestRecordIndex` within the `workflow-executor` package [9fcdda2]
> - Updated inline test comments throughout the `load-related-record-step-executor.test.ts` file in the `workflow-executor` package [9fcdda2]
> - Modified the `workflow-executor` package's `LoadRelatedRecordStepExecutor` to consult AI for single-candidate scenarios during related record ranking [04b2be8]
> - Refactored AI wrapper placement in `workflow-executor` package's `LoadRelatedRecordStepExecutor` methods for relation selection and candidate refresh [04b2be8]
> - Updated tests in `workflow-executor` package's `load-related-record-step-executor.test.ts` to reflect AI consultation for single-candidate scenarios and narrowed AI-wrapping scope [04b2be8]
> - Extracted `AiAssistUnavailableError` class from `load-related-record-step-executor` into a shared error class in the `workflow-executor` package [53670e4]
> - Modified `LoadRelatedRecordStepExecutor` class in `workflow-executor` package to pre-select a single related record candidate in manual execution mode [2f50ee3]
> - Added test case verifying single candidate pre-selection behavior in manual execution mode for to-many relations [2f50ee3]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 4f8b398.</sup>
> 
>
>

### DIFF
--- a/packages/workflow-executor/src/executors/load-related-record-step-executor.ts
+++ b/packages/workflow-executor/src/executors/load-related-record-step-executor.ts
@@ -1,3 +1,4 @@
+/* eslint-disable max-classes-per-file */
 import type { StepExecutionResult } from '../types/execution-context';
 import type {
   LoadRelatedRecordCandidate,
@@ -22,6 +23,7 @@ import {
   RelatedRecordNotFoundError,
   RelationNotFoundError,
   StepStateError,
+  extractErrorMessage,
 } from '../errors';
 import RecordStepExecutor from './record-step-executor';
 import { StepExecutionMode } from '../types/validated/step-definition';
@@ -80,6 +82,19 @@ function isFollowableRelation(field: FieldSchema): boolean {
   return field.isRelationship && Boolean(field.relatedCollectionName || field.polymorphicTypeField);
 }
 
+// Internal marker: wraps any failure raised while calling the AI (timeout, provider outage,
+// missing/malformed tool call, out-of-range choice) so the step handlers can degrade to the
+// deterministic Manual path instead of erroring the run. Never leaves this module.
+class AiAssistUnavailableError extends Error {
+  readonly reason: unknown;
+
+  constructor(reason: unknown) {
+    super('AI assistance unavailable');
+    this.name = 'AiAssistUnavailableError';
+    this.reason = reason;
+  }
+}
+
 export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<LoadRelatedRecordStepDefinition> {
   protected async doExecute(): Promise<StepExecutionResult> {
     // Branch A -- Re-entry after pending execution found in RunStore
@@ -111,22 +126,37 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
       throw new StepStateError(`Step at index ${this.context.stepIndex} has no pending data`);
     }
 
-    const suggestViaAi = this.context.stepDefinition.executionType !== StepExecutionMode.Manual;
+    const useAi = this.context.stepDefinition.executionType !== StepExecutionMode.Manual;
     const schema = await this.getCollectionSchema(execution.selectedRecordRef.collectionName);
     const target = await this.buildTarget(schema, fieldName, execution.selectedRecordRef);
-    const { availableRecordIds, suggestedRecord } = await this.collectCandidateIds(
-      target,
-      suggestViaAi,
-    );
+
+    // A field switch must not error the run: an AI failure/timeout while re-listing degrades to the
+    // no-AI candidate path (no suggestion), like the first-call degrade.
+    let aiSuggested = useAi;
+    let candidates;
+
+    try {
+      candidates = await this.collectCandidateIds(target, useAi);
+    } catch (error) {
+      if (!(useAi && error instanceof AiAssistUnavailableError)) throw error;
+      this.logAiDegrade(error.reason);
+      aiSuggested = false;
+      candidates = await this.collectCandidateIds(target, false);
+    }
+
+    const { availableRecordIds, suggestedRecord } = candidates;
 
     await this.context.runStore.saveStepExecution(this.context.runId, {
       ...execution,
       userConfirmation: undefined,
+      // Rebuild pendingData for the new relation from scratch (retain only the immutable field list)
+      // so no stale suggestion state — suggestedRecord or suggestNoRecord — survives the field switch.
       pendingData: {
-        ...execution.pendingData,
+        availableFields: execution.pendingData.availableFields,
         suggestedField: { name: target.name, displayName: target.displayName },
         availableRecordIds,
         suggestedRecord,
+        ...(aiSuggested && !suggestedRecord && { suggestNoRecord: true }),
       },
     });
 
@@ -137,14 +167,43 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
     const { stepDefinition: step } = this.context;
     const useAi = step.executionType !== StepExecutionMode.Manual;
 
-    if (step.executionType === StepExecutionMode.FullyAutomated) {
-      return this.resolveAndLoadAutomatic();
-    }
+    try {
+      if (step.executionType === StepExecutionMode.FullyAutomated) {
+        return await this.resolveAndLoadAutomatic();
+      }
 
-    const target = await this.resolveTarget(useAi);
+      const target = await this.resolveTarget(useAi);
+      const sourceSchema = await this.getCollectionSchema(target.selectedRecordRef.collectionName);
+
+      return await this.saveAndAwaitInput(target, sourceSchema, useAi);
+    } catch (error) {
+      // AI failure/timeout in any AI mode (relation pick, field/record ranking, or a Full AI
+      // auto-load) degrades to Manual instead of erroring the run — never auto-skip.
+      if (useAi && error instanceof AiAssistUnavailableError) {
+        this.logAiDegrade(error.reason);
+
+        return this.degradeToManualAwaitInput();
+      }
+
+      throw error;
+    }
+  }
+
+  private logAiDegrade(reason: unknown): void {
+    this.context.logger(
+      'Warn',
+      'load-related-record: AI unavailable, degrading to manual selection',
+      { ...this.logCtx, error: extractErrorMessage(reason) },
+    );
+  }
+
+  // Degrades to the deterministic Manual path: first eligible relation, candidate list fetched
+  // without AI, user picks. The re-run is AI-free, so it cannot re-trigger the failure.
+  private async degradeToManualAwaitInput(): Promise<StepExecutionResult> {
+    const target = await this.resolveTarget(false);
     const sourceSchema = await this.getCollectionSchema(target.selectedRecordRef.collectionName);
 
-    return this.saveAndAwaitInput(target, sourceSchema, useAi);
+    return this.saveAndAwaitInput(target, sourceSchema, false);
   }
 
   // Picks the (record, relation) pair to follow. Unlike a separate record-then-relation choice,
@@ -179,7 +238,9 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
     // relation (getAvailableRecordRefs lists the base record first). The user switches the relation
     // at runtime; a non-base source needs deterministic "Related to" pinning (selectedRecordStepId).
     const chosen =
-      eligible.length > 1 && useAi ? await this.selectRelationToFollow(eligible) : eligible[0];
+      eligible.length > 1 && useAi
+        ? await this.withAiAssist(() => this.selectRelationToFollow(eligible))
+        : eligible[0];
 
     return this.targetFromCandidate(chosen);
   }
@@ -332,25 +393,29 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
   ): Promise<{
     availableRecordIds: LoadRelatedRecordCandidate[];
     suggestedRecord?: LoadRelatedRecordCandidate;
+    // Full AI only: the AI returned a best guess but could not confidently single one out among
+    // several viable candidates. Routes Full AI to a confirmation instead of an auto-load.
+    ambiguous: boolean;
   }> {
     if (target.relationType === 'BelongsTo' || target.relationType === 'HasOne') {
       const candidate = await this.fetchXToOneCandidate(target);
 
       return candidate
-        ? { availableRecordIds: [candidate], suggestedRecord: candidate }
-        : { availableRecordIds: [] };
+        ? { availableRecordIds: [candidate], suggestedRecord: candidate, ambiguous: false }
+        : { availableRecordIds: [], ambiguous: false };
     }
 
-    const { relatedData, bestIndex, relatedSchema } = await this.selectBestFromRelatedData(
-      target,
-      50,
-      // allowNone: the AI may judge no candidate relevant (→ "No X to load"); only meaningful when
-      // ranking (Manual passes rank=false and never calls the AI).
-      { rank: suggestViaAi, allowNone: true },
-    );
+    const { relatedData, bestIndex, confident, relatedSchema } =
+      await this.selectBestFromRelatedData(
+        target,
+        50,
+        // allowNone: the AI may judge no candidate relevant (→ "No X to load"); only meaningful when
+        // ranking (Manual passes rank=false and never calls the AI).
+        suggestViaAi ? { rank: true, allowNone: true } : { rank: false },
+      );
 
     if (relatedData.length === 0) {
-      return { availableRecordIds: [] };
+      return { availableRecordIds: [], ambiguous: false };
     }
 
     const referenceField = relatedSchema.referenceField ?? null;
@@ -364,6 +429,8 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
     return {
       availableRecordIds: relatedData.map(toCandidate),
       suggestedRecord: bestIndex >= 0 ? toCandidate(relatedData[bestIndex]) : undefined,
+      // -1 (none relevant) is not ambiguity; only a low-confidence positive pick is.
+      ambiguous: bestIndex >= 0 && !confident,
     };
   }
 
@@ -377,19 +444,26 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
   }
 
   private async resolveAndLoadAutomatic(): Promise<StepExecutionResult> {
-    // No source record throws (like Manual/AI-assisted) → the front offers "continue without" (PRD-550).
+    // No source record throws (like Manual/AI-assisted) → the front offers "continue without".
     const target = await this.resolveTarget(true);
-    const { availableRecordIds, suggestedRecord } = await this.collectCandidateIds(target, true);
+    const { availableRecordIds, suggestedRecord, ambiguous } = await this.collectCandidateIds(
+      target,
+      true,
+    );
 
-    // Full AI auto-loads a confident pick and advances. With nothing relevant to load it degrades to
-    // an AI-assisted confirmation (a human decides, "No X to load" pre-checked) instead of skipping —
-    // mirrors the Trigger Action automated→confirmation fallback.
-    if (!suggestedRecord) {
+    // Full AI auto-loads a confident pick and advances. It degrades to an AI-assisted confirmation
+    // (a human decides) instead of auto-loading when either:
+    //  - nothing is relevant to load → "No X to load" pre-checked (suggestNoRecord); or
+    //  - the AI cannot confidently single out one of several viable candidates → its best guess is
+    //    pre-selected for the human to confirm (no "No X to load").
+    // Mirrors the Trigger Action automated→confirmation fallback.
+    if (!suggestedRecord || ambiguous) {
       const sourceSchema = await this.getCollectionSchema(target.selectedRecordRef.collectionName);
 
       return this.persistAwaitInput(target, sourceSchema, {
         availableRecordIds,
-        suggestNoRecord: true,
+        suggestedRecord,
+        suggestNoRecord: !suggestedRecord,
       });
     }
 
@@ -479,10 +553,15 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
   private async selectBestFromRelatedData(
     target: Pick<RelationTarget, 'selectedRecordRef' | 'name' | 'relatedCollectionName'>,
     limit: number,
-    opts: { rank: boolean; allowNone: boolean } = { rank: true, allowNone: false },
+    // allowNone is meaningful only when ranking, so it's absent from the rank:false variant.
+    opts: { rank: true; allowNone: boolean } | { rank: false } = { rank: true, allowNone: false },
   ): Promise<{
     relatedData: RecordData[];
     bestIndex: number;
+    // Whether the suggested record is a confident pick. The deterministic/short-circuit paths
+    // (single candidate, no ranking) are confident by construction; the ranked path reflects the
+    // AI's own confidence flag.
+    confident: boolean;
     suggestedFields: string[];
     relatedSchema: CollectionSchema;
   }> {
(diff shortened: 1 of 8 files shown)


packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -168,8 +177,10 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
       );
     }
 
+    // Manual mode never invokes the AI relation chooser: with several eligible relations (legacy
+    // steps that pinned no relationName) it defaults to the first; the user can switch it at runtime.
     const chosen =
```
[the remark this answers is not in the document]
matthv: Clarified in ea0b5f833. `eligible[0]` isn't arbitrary: `getAvailableRecordRefs()` lists the workflow-start (base) record first, so Manual defaults to the base record's first relation — a predictable default, and the relation stays switchable at runtime. Picking a *non-base* source record is the job of the deterministic "Related to" pinning (`selectedRecordStepId`), which lands with the deterministic editor (PRD-546) and is out of scope here; a run-panel source picker would be a separate change.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -717,11 +782,12 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
       tool,
     );
 
-    // NOTE: The Zod schema's .min(0).max(maxIndex) shapes the tool prompt only — it is NOT
-    // validated against the AI response. This guard is the sole runtime enforcement.
-    if (!Number.isInteger(recordIndex) || recordIndex < 0 || recordIndex > maxIndex) {
+    // NOTE: The Zod schema's .min().max() shapes the tool prompt only — it is NOT validated against
+    // the AI response. This guard is the sole runtime enforcement. -1 (none relevant) is accepted
+    // only when allowNone.
+    if (!Number.isInteger(recordIndex) || recordIndex < minIndex || recordIndex > maxIndex) {
```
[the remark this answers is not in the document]
matthv: Good catch — fixed in 15793400c. `-1` ("none relevant") is now only offered/accepted when the list was **not** truncated (`noneAllowed = allowNone && !truncated`). On truncation the AI must pick from the candidates it actually saw, so a match in the unseen tail can no longer be silently skipped. (Went with disabling `allowNone` rather than falling back to the first candidate, so the AI's judgment over the shown subset is still used.)

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -338,14 +352,26 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
-  /** Branch B: fully automated. xToOne loads the linked record; HasMany ranks candidates via AI; BelongsToMany takes the first. */
-  private async resolveAndLoadAutomatic(target: RelationTarget): Promise<StepExecutionResult> {
+  private async resolveAndLoadAutomatic(useAi: boolean): Promise<StepExecutionResult> {
+    let target: RelationTarget;
+
+    try {
+      target = await this.resolveTarget(useAi);
+    } catch (error) {
+      // No source record to load from → Full AI auto-continues. (Manual/AI-assisted let this error
+      // surface so the front can offer "continue without" — PRD-550.)
```
hercemer42: **[Claude Opus 4.8 (1M context)] · Violates conventions**

Ticket ID `PRD-550` in a code comment. Project convention: no PRD/ticket IDs in comments (applies to code *and* tests). Drop `— PRD-550`; the sentence reads fine without it. Heads-up: the linter doesn't catch this — worth a lint rule if the team wants it enforced.

packages/workflow-executor/test/executors/load-related-record-step-executor.test.ts
```diff
@@ -3624,6 +3834,29 @@ describe('LoadRelatedRecordStepExecutor', () => {
...
+      expect(result.stepOutcome.status).toBe('success');
+      expect(runStore.saveStepExecution).toHaveBeenCalledWith(
+        'run-1',
+        expect.objectContaining({ executionResult: { skipped: true } }),
+      );
+    });
+
+    it('surfaces a "no source record" error in await modes when the source step loaded nothing', async () => {
+      // Only Full AI auto-skips; Manual / AI-assisted surface the error so the front can offer
+      // "continue without" (PRD-550).
```
hercemer42: **[Claude Opus 4.8 (1M context)] · Violates conventions**

`(PRD-550)` in a `//` comment — the no-ticket-IDs-in-comments rule applies to tests too. Drop the ID; the behavioral note is enough. (The `describe('… (PRD-148)')` titles are borderline and out of the comment rule, so left alone.)

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -681,26 +712,36 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
       this.context.logger('Warn', 'load-related-record: candidate list truncated for AI prompt', {
         ...this.logCtx,
         shown,
         total: candidates.length,
       });
     }
 
+    // "None relevant" (-1) is only trustworthy when the AI saw the whole list. If it was truncated,
+    // force a pick from what was shown — otherwise a match in the unseen tail would be silently skipped.
+    const noneAllowed = allowNone && !truncated;
```
hercemer42, line 783: **[Claude Opus 4.8 (1M context)] · Should fix**

The `noneAllowed = allowNone && !truncated` rule — the safety core of the feature (`-1` "none relevant" only trusted when the AI saw the whole list) — has no test. The truncation tests all run the await path (`allowNone:false`); the none-relevant test uses a 2-candidate, non-truncated list. So the Full-AI + truncated combination, where `noneAllowed` collapses to `false`, `minIndex`→0, and a `-1` response is *rejected* by the guard at line 764, is unverified — a regression dropping `!truncated` (silently skipping a record that matches in the unseen tail) would pass the whole suite. Add: Full-AI HasMany, candidates wide enough to truncate, AI returns `-1` → assert `InvalidAIResponseError`, not a skip.

packages/workflow-executor/src/types/validated/step-definition.ts
```diff
@@ -98,10 +98,10 @@ export type TriggerActionStepDefinition = z.infer<typeof TriggerActionStepDefini
 export const LoadRelatedRecordStepDefinitionSchema = z.object({
   ...sharedFields,
   type: z.literal(StepType.LoadRelatedRecord),
+  // No `.catch`: it would silently coerce `manual` to AutomatedWithConfirmation (AI prefill back on).
   executionType: z
-    .enum([AutomatedWithConfirmation, FullyAutomated])
-    .default(AutomatedWithConfirmation)
-    .catch(AutomatedWithConfirmation),
+    .enum([Manual, AutomatedWithConfirmation, FullyAutomated])
```
hercemer42, line 103: **[Claude Opus 4.8 (1M context)] · Preferential**

No test asserts this schema parses `executionType: 'manual'` and returns `'manual'`. The executor tests build definitions directly via `makeStep(...)`, bypassing the schema, and the only no-coercion tests live in an out-of-PR file (trigger-action / condition). A one-line `LoadRelatedRecordStepDefinitionSchema.parse({ …, executionType: 'manual' })` → `toBe('manual')` cheaply guards Manual mode against the enum being narrowed back later.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -448,6 +473,7 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
   private async selectBestFromRelatedData(
     target: Pick<RelationTarget, 'selectedRecordRef' | 'name' | 'relatedCollectionName'>,
     limit: number,
+    opts: { rank: boolean; allowNone: boolean } = { rank: true, allowNone: false },
```
hercemer42: **[Claude Opus 4.8 (1M context)] · Preferential**

`{ rank; allowNone }` permits the meaningless `{ rank:false, allowNone:true }` — when `rank:false` the method returns (line 492-494) before `allowNone` is ever read. A union `{ rank: false } | { rank: true; allowNone: boolean }` makes that illegal state unrepresentable. Minor (private, single-file); flag only for clarity should it grow more callers.

packages/workflow-executor/src/types/step-execution-data.ts
```diff
@@ -183,7 +183,8 @@ export interface LoadRelatedRecordStepExecutionData
     WithUserConfirmation<LoadRelatedRecordConfirmation> {
   type: 'load-related-record';
   pendingData?: LoadRelatedRecordPendingData;
-  selectedRecordRef: RecordRef;
+  // Absent only on the Full AI no-source auto-skip; always set on the await/load paths.
+  selectedRecordRef?: RecordRef;
```
hercemer42, line 190: **[Claude Opus 4.8 (1M context)] · Preferential**

Making `selectedRecordRef` optional widens the type to "sometimes absent", but the real invariant is narrower: absent ⟺ the Full-AI no-source skip (`executionResult: { skipped: true }`). That coupling is enforced only by this comment + scattered runtime guards, and `summary/step-execution-formatters.ts:99` derefs `selectedRecordRef.recordId` relying on it. Not a bug today (every producer upholds it), but a discriminated union (loaded vs skipped variant) would make the deref safe by construction and remove the `!selectedRecordRef` guards at lines 107/429.

packages/workflow-executor/src/types/validated/step-definition.ts
```diff
@@ -98,10 +98,10 @@ export type TriggerActionStepDefinition = z.infer<typeof TriggerActionStepDefini
...
+    .default(AutomatedWithConfirmation),
```
hercemer42, line 104: **[Claude Opus 4.8 (1M context)] · Preferential (discussion)**

Dropping `.catch` is the right call — it stops silently coercing values into AI-assisted. One downstream consequence to be aware of: an unknown/future `executionType` now throws a raw `ZodError`, and on the pending-run path `toStepDefinition` (run-to-available-step-mapper.ts:165) runs *outside* the try/catch that maps to `DomainValidationError` — so `getAvailableRuns` doesn't see it as a `WorkflowExecutorError` and logs-and-drops the run rather than reporting it malformed. Low reachability (server contract enum has 3 values), but this is exactly the executor version-skew case the conventions call out: an older executor receiving a newer 4th mode would silently drop the run instead of surfacing an "update needed" error. Fix is downstream/out of scope here. Also stale as a result: the comment at step-definition-mapper.ts:22 ("each schema's `.default().catch()` handles unsupported values") no longer holds for this schema or trigger-action.

hercemer42 (dismissed):

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
hercemer42: **[Claude Opus 4.8]** Fixed — removed the PRD-550 reference from the comment (kept the explanation).

packages/workflow-executor/test/executors/load-related-record-step-executor.test.ts
hercemer42: **[Claude Opus 4.8]** Fixed — removed the PRD-550 reference from the test comment.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
hercemer42: **[Claude Opus 4.8]** Confirmed intended (not a bug) — on a truncated list we withhold `-1` so the AI can't decline based on a partial view. Pinned with a test.

packages/workflow-executor/src/types/validated/step-definition.ts
hercemer42: **[Claude Opus 4.8]** Fixed — added a test that the schema parses `executionType: 'manual'` → 'manual'.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
hercemer42: **[Claude Opus 4.8]** Fixed — tightened the options to a discriminated union `{ rank: true; allowNone: boolean } | { rank: false }`.

packages/workflow-executor/src/types/step-execution-data.ts
hercemer42: **[Claude Opus 4.8]** Left optional, intentionally — `step-serializer.ts` guards its absence on the load-related wire path (unlike other step types), so narrowing risks the serializer/wire contract. Flagging the reasoning rather than changing it.

packages/workflow-executor/src/types/validated/step-definition.ts
hercemer42: **[Claude Opus 4.8]** Confirmed intended — dropping `.catch` stops silently coercing a bad value to AI-assisted; pinned with a "rejects invalid executionType" test.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -448,65 +544,56 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
   }> {
     const relatedSchema = await this.getCollectionSchema(target.relatedCollectionName);
     const relatedData = await this.fetchRelatedData(target, relatedSchema, limit);
 
-    // Empty (bestIndex unused — callers guard on length) or single → no ranking needed.
+    // length<=1 short-circuits before opts.allowNone, so Full AI with a single candidate always
+    // loads it — the AI is never asked to reject a sole option.
     if (relatedData.length <= 1) {
-      return { relatedData, bestIndex: 0, suggestedFields: [], relatedSchema };
+      return { relatedData, bestIndex: 0, confident: true, suggestedFields: [], relatedSchema };
```
Scra3: issue (blocking): For a HasMany relation resolving to exactly one candidate, this returns a confident pick without ever calling the AI, so Full AI auto-loads a possibly-irrelevant sole record instead of degrading with `-1` / "No X to load" (violates AC "Full AI: when the AI returns none ... it degrades"). xToOne never reaches this method, so the short-circuit only ever hits HasMany — gate it to `!opts.rank` (empty list → `bestIndex: -1`, single → `0`) and let ranking passes reach the AI even at one candidate.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -168,8 +230,12 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
       );
     }
 
+    // No-AI path picks eligible[0] — the base record's first relation (getAvailableRecordRefs lists
+    // it first). The user switches relation at runtime; a non-base source needs selectedRecordStepId pinning.
     const chosen =
-      eligible.length === 1 ? eligible[0] : await this.selectRelationToFollow(eligible);
+      eligible.length > 1 && useAi
+        ? await this.withAiAssist(() => this.selectRelationToFollow(eligible))
```
Scra3: issue (non-blocking): `withAiAssist` wraps `selectRelationToFollow`, whose `buildPreviousStepsMessages` reads the run store *before* the AI invoke, so a transient store failure is re-tagged `AiAssistUnavailableError` and degrades to Manual instead of surfacing — wrap only the `invokeWithTool` call.

Scra3 (changes requested):
Review of the 3-way execution mode against the PRD-148 acceptance criteria (plus a skeptic second pass on the findings). The edge-case handling is solid overall — Manual never calls the AI, no-source throws instead of auto-advancing, AI failure degrades to Manual, and `-1`/`allowNone` is correctly gated on a non-truncated list. Two points below: one blocking edge case in Full AI, one non-blocking error-classification smell.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -76,6 +82,18 @@ function isFollowableRelation(field: FieldSchema): boolean {
   return field.isRelationship && Boolean(field.relatedCollectionName || field.polymorphicTypeField);
 }
 
+// Marks any AI-call failure (timeout, outage, malformed/missing tool call, out-of-range pick) so
+// handlers degrade to the deterministic Manual path instead of erroring — never auto-skips. Internal only.
+class AiAssistUnavailableError extends Error {
```
Scra3: suggestion (non-blocking): `AiAssistUnavailableError` is the only error defined outside `src/errors.ts` and forces the file-level `max-classes-per-file` disable on line 1 — moving it to `errors.ts` unexported (like `AiInvokeTimeoutError`) matches house style and drops the whole-file suppression.

Scra3 (commented):
Follow-up architecture notes (non-blocking, complements the earlier review). Overall the executor conforms well to the `trigger-record-action` precedent — base machinery (`invokeWithTool`, `handleConfirmationFlow`, `patchAndReloadPendingData`) is reused rather than re-implemented, and the 3-mode schema mirrors TriggerAction. One inline placement suggestion, plus a nit: the `sharedFields` header comment (`src/types/validated/step-definition.ts:23-24`) still claims every schema normalizes via `.default().catch()`, now false for the LoadRelatedRecord and TriggerAction schemas that intentionally drop `.catch` — update it to avoid misleading future edits.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
hercemer42: [Claude Opus 4.8] Fixed in 04b2be81. Dropped the `length <= 1` confident short-circuit on the ranking path — a lone toMany candidate now goes through the AI (allowNone), so Full AI can decline it (-1 → "No X to load") instead of auto-loading a possibly-irrelevant sole record. xToOne is unaffected (still handled by fetchXToOneCandidate). Updated the 3 tests that pinned the old behavior and added one asserting the Full-AI-declines case.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
hercemer42: [Claude Opus 4.8] Fixed in 04b2be81. `withAiAssist` in selectRelationToFollow now wraps only the invokeWithTool call plus its response validation; buildPreviousStepsMessages runs outside it, so a transient run-store failure surfaces as a real error instead of being tagged AiAssistUnavailable and degraded to Manual.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
hercemer42: [Claude Opus 4.8] Done in 53670e43. Moved AiAssistUnavailableError into errors.ts as a WorkflowExecutorError subclass — exported there so the executor imports it, but kept out of the index barrel (like AiInvokeTimeoutError) so it stays internal and never HTTP-maps. The file-level max-classes-per-file disable is gone. Thanks for the pointer.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -448,65 +542,52 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
...
     const relatedSchema = await this.getCollectionSchema(target.relatedCollectionName);
     const relatedData = await this.fetchRelatedData(target, relatedSchema, limit);
 
-    // Empty (bestIndex unused — callers guard on length) or single → no ranking needed.
-    if (relatedData.length <= 1) {
-      return { relatedData, bestIndex: 0, suggestedFields: [], relatedSchema };
+    // Manual (user picks) and empty lists never rank. A lone candidate in an AI mode still goes
+    // through the AI below so Full AI can decline it (-1 → "No X to load") rather than auto-load.
+    if (!opts.rank || relatedData.length === 0) {
+      return { relatedData, bestIndex: -1, confident: true, suggestedFields: [], relatedSchema };
```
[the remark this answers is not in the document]
hercemer42: [Claude Opus 4.8] Good catch — fixed in 2f50ee32. Restored the sole-candidate prefill: the `!opts.rank` (Manual) branch now returns `bestIndex: 0` when there is exactly one candidate, and `-1` with several so multi-candidate Manual still lets the user pick. Scoped to `length === 1` rather than any non-empty list, so Manual with multiple candidates isn't wrongly pre-selected. Added a test for the sole-toMany-candidate case.

Scra3 (approved):
Approving — all findings from the earlier review are addressed and verified in the code, not just claimed:

- **Bug A (blocking)** — a lone HasMany candidate no longer auto-loads in Full AI: `selectBestFromRelatedData` now gates the short-circuit on `!opts.rank` and routes every ranking pass through the AI, so a sole irrelevant candidate degrades (`-1` → `suggestNoRecord`) instead of auto-loading. Covered by a dedicated test (Full AI + single HasMany candidate + AI returns -1 → `awaiting-input`, `suggestNoRecord: true`, no `executionResult`).
- **Finding B (non-blocking)** — `withAiAssist` moved inside `selectRelationToFollow` around the `invokeWithTool` call only; the run-store read (`buildPreviousStepsMessages`) is now outside the wrapper, so a store blip is no longer mislabeled as an AI failure.
- **Architecture** — `AiAssistUnavailableError` extracted to a shared error class; the whole-file `max-classes-per-file` suppression is gone.

Solid. 🚀

PR #1730, opened 2026-07-01
feat(agent-bff): BFF API key resolver wiring

nbouliol:
## Context

Mode 2 (developer path) for the BFF: a request presents an `fbff_<keyId>_<secret>` key in the `X-Forest-Bff-Key` header. The BFF resolves it server-to-server against the SaaS (`POST /liana/v1/bff-api-keys/resolve`, already shipped in forestadmin-server), caches the result briefly, and mints a fresh 5m agent JWT per request from the resolved identity. The minted JWT is never cached.

Slice 2 of the auth foundation. The actual agent call is Slice 3 — here the JWT is asserted as correctly minted and attached (`ctx.state.agentToken` / `ctx.state.apiKeyIdentity`), with the agent client out of scope.

fixes PRD-663

## What's in it

New `packages/agent-bff/src/api-key/`:
- `api-key.ts` — parse `fbff_<16hex>_<64hex>`, `hashApiKey` (cache key = sha256 of keyId+secret), `fingerprintApiKey` (logs)
- `api-key-client.ts` / `api-key-resolve-error.ts` — S2S resolve call, typed errors
- `agent-token.ts` — mint 5m JWT (camelCase + snake_case aliases, tags→Record, null→'')
- `resolve-cache.ts` — per-instance TTL cache (positive 60s / negative 10s), `now()` injected, bounded (10k)
- `api-key-error.ts` — nested `{ error: { type, status, message } }` contract
- `api-key-authenticator.ts` — orchestrates parse → cache → resolve → mint
- `api-key-middleware.ts` — reads the header, attaches to `ctx.state`, `Retry-After` on 503

Wired in `cli-core.ts` (mounted independently of OAuth) and exported from `index.ts`.

## Error mapping (SaaS → BFF)

- 401 → `invalid_api_key` (negative-cached ≤10s)
- 403 → `forest_identity_not_allowed` (negative-cached ≤10s)
- 400 → `invalid_request` (not cached)
- 429 / unreachable / 5xx → `503 key_resolution_unavailable` + `Retry-After` (not cached; SaaS `Retry-After` propagated on 429)
- malformed key shape → `401 invalid_api_key` locally, no SaaS call

## Notes / follow-ups

- The resolved identity has no `role`; the minted JWT omits it. To be revisited in Slice 3 when the agent call is wired.
- Mode 1/Mode 2 precedence, CORS, per-key `allowedOrigins` enforcement and cross-cutting contract tests are owned by T6 (PRD-666). This PR only exposes `allowedOrigins` on the attached identity for T6 to consume.

## Tests

49 focused tests (`test/api-key/`): parse, cache TTLs + bound, mint claims/expiry, resolve error mapping, authenticator cache behavior, middleware attach/error/passthrough, no raw secret in logs. Lint + tsc clean.

Includes a second commit addressing local review findings (downstream-error masking, `Retry-After` NaN, unguarded success JSON, `size()` staleness, test `fetch` restore).

> [!NOTE]
> ### Add Mode 2 BFF API key authentication middleware to `agent-bff`
> - Adds a full API key authentication pipeline: key parsing/hashing, a Forest server resolver client, a TTL-aware resolve cache (60s positive / 10s negative), an authenticator, and a Koa middleware gated on the `X-Forest-Bff-Key` header.
> - On successful authentication, mints a short-lived (5min) HS256 JWT with user identity claims and attaches it to Koa state alongside the resolved identity; sets `Cache-Control: no-store`.
> - On failure, returns structured `ApiKeyError` responses (400/401/403/503) with optional `Retry-After`; unexpected errors return a generic 500.
> - Middleware is conditionally registered in [`cli-core.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1730/files#diff-76c14e8251a9859dc672905a36febbeb3311acac1dd54a57a05f65282794f19c) when `forestServerUrl`, `forestEnvSecret`, and `forestAuthSecret` are all present in config; logs a warning and skips otherwise.
> - Risk: Requests authenticated via API key bypass existing OAuth middleware; cache-full behavior silently drops new entries rather than evicting old ones.
>
> 
> #### Changes since #1730 opened
>
> - Enforced strict validation in `ApiKeyClient.isResolvedIdentity` and added `ApiKeyClient.isIdentityUser` method to require complete identity structure [86caf50]
> - Implemented LRU eviction in `createResolveCache` store utility to remove oldest entry when cache reaches capacity [86caf50]
> - Restricted `ApiKeyClient.parseRetryAfter` to only accept positive integer values for Retry-After headers [86caf50]
> - Added `agentUrl` requirement to `resolveApiKeyConfig` configuration validation [86caf50]
> - Refactored `createApiKeyAuthenticator.authenticate` handler to mint `AuthenticatedApiKey` once before caching [86caf50]
> - Removed `agentUrl` requirement from `resolveApiKeyConfig` function in `agent-bff` package [8c99477]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bb63ce1.</sup>
> 
>
>

### DIFF
--- a/packages/agent-bff/src/api-key/agent-token.ts
+++ b/packages/agent-bff/src/api-key/agent-token.ts
@@ -1,0 +1,41 @@
+import type { ResolvedApiKeyIdentity } from './api-key-client';
+
+import jsonwebtoken from 'jsonwebtoken';
+
+export const AGENT_TOKEN_EXPIRES_IN = '5m';
+
+export interface IssueAgentTokenParams {
+  identity: ResolvedApiKeyIdentity;
+  authSecret: string;
+}
+
+function tagsToRecord(tags: { key: string; value: string }[]): Record<string, string> {
+  return tags.reduce((memo, { key, value }) => ({ ...memo, [key]: value }), {});
+}
+
+export function issueAgentToken({ identity, authSecret }: IssueAgentTokenParams): string {
+  const { user, renderingId } = identity;
+  const firstName = user.firstName ?? '';
+  const lastName = user.lastName ?? '';
+  const tags = tagsToRecord(user.tags);
+
+  // snake_case aliases: Ruby/Python agents splat JWT claims into Caller (snake_case kwargs).
+  return jsonwebtoken.sign(
+    {
+      id: user.id,
+      email: user.email,
+      firstName,
+      lastName,
+      team: user.team,
+      renderingId,
+      tags,
+      permissionLevel: user.permissionLevel,
+      first_name: firstName,
+      last_name: lastName,
+      rendering_id: renderingId,
+      permission_level: user.permissionLevel,
+    },
+    authSecret,
+    { algorithm: 'HS256', expiresIn: AGENT_TOKEN_EXPIRES_IN },
+  );
+}
--- a/packages/agent-bff/src/api-key/api-key-authenticator.ts
+++ b/packages/agent-bff/src/api-key/api-key-authenticator.ts
@@ -86,9 +86,10 @@ export default function createApiKeyAuthenticator({
         throw error;
       }
 
+      const authenticated = mint(identity);
       cache.setPositive(hash, identity);
 
-      return mint(identity);
+      return authenticated;
     },
   };
 }
--- a/packages/agent-bff/src/api-key/api-key-authenticator.ts
+++ b/packages/agent-bff/src/api-key/api-key-authenticator.ts
@@ -1,0 +1,94 @@
+import type ApiKeyClient from './api-key-client';
+import type { ResolvedApiKeyIdentity } from './api-key-client';
+import type { ApiKeyError } from './api-key-error';
+import type { ResolveCache } from './resolve-cache';
+
+import { issueAgentToken } from './agent-token';
+import { hashApiKey, parseApiKey } from './api-key';
+import { ApiKeyResolveError } from './api-key-client';
+import {
+  forestIdentityNotAllowed,
+  invalidApiKey,
+  invalidRequest,
+  keyResolutionUnavailable,
+} from './api-key-error';
+
+const UNAVAILABLE_RETRY_AFTER_SECONDS = 5;
+const NEGATIVE_CACHE_STATUSES = new Set([401, 403]);
+
+export interface ApiKeyAuthenticatorOptions {
+  client: ApiKeyClient;
+  cache: ResolveCache;
+  authSecret: string;
+}
+
+export interface AuthenticatedApiKey {
+  agentToken: string;
+  identity: ResolvedApiKeyIdentity;
+}
+
+export interface ApiKeyAuthenticator {
+  authenticate(rawKey: string): Promise<AuthenticatedApiKey>;
+}
+
+function mapResolveError(error: ApiKeyResolveError): ApiKeyError {
+  if (error.unreachable) return keyResolutionUnavailable(UNAVAILABLE_RETRY_AFTER_SECONDS);
+
+  switch (error.status) {
+    case 401:
+      return invalidApiKey();
+    case 403:
+      return forestIdentityNotAllowed();
+    case 400:
+      return invalidRequest();
+    case 429:
+      return keyResolutionUnavailable(error.retryAfter ?? UNAVAILABLE_RETRY_AFTER_SECONDS);
+    default:
+      return keyResolutionUnavailable(UNAVAILABLE_RETRY_AFTER_SECONDS);
+  }
+}
+
+export default function createApiKeyAuthenticator({
+  client,
+  cache,
+  authSecret,
+}: ApiKeyAuthenticatorOptions): ApiKeyAuthenticator {
+  function mint(identity: ResolvedApiKeyIdentity): AuthenticatedApiKey {
+    return { agentToken: issueAgentToken({ identity, authSecret }), identity };
+  }
+
+  return {
+    async authenticate(rawKey) {
+      const parsed = parseApiKey(rawKey);
+
+      if (!parsed) throw invalidApiKey();
+
+      const hash = hashApiKey(parsed.keyId, parsed.secret);
+
+      const cachedIdentity = cache.getPositive(hash);
+      if (cachedIdentity) return mint(cachedIdentity);
+
+      const cachedError = cache.getNegative(hash);
+      if (cachedError) throw cachedError;
+
+      let identity: ResolvedApiKeyIdentity;
+
+      try {
+        identity = await client.resolveApiKey(parsed);
+      } catch (error) {
+        if (error instanceof ApiKeyResolveError) {
+          const mapped = mapResolveError(error);
+          if (NEGATIVE_CACHE_STATUSES.has(mapped.status)) cache.setNegative(hash, mapped);
+
+          throw mapped;
+        }
+
+        throw error;
+      }
+
+      cache.setPositive(hash, identity);
+
+      return mint(identity);
+    },
+  };
+}
--- a/packages/agent-bff/src/api-key/api-key-client.ts
+++ b/packages/agent-bff/src/api-key/api-key-client.ts
@@ -61,17 +61,29 @@ export default class ApiKeyClient {
     if (!response.ok) {
       const body = (await response.json().catch(() => ({}))) as SaasErrorBody;
       const firstError = body.errors?.[0];
-      const retryAfterHeader = response.headers.get('retry-after');
 
       throw new ApiKeyResolveError({
         status: response.status,
         code: firstError?.meta?.code,
         name: firstError?.name,
-        retryAfter: retryAfterHeader ? Number(retryAfterHeader) : undefined,
+        retryAfter: ApiKeyClient.parseRetryAfter(response.headers.get('retry-after')),
       });
     }
 
-    return (await response.json()) as ResolvedApiKeyIdentity;
+    try {
+      return (await response.json()) as ResolvedApiKeyIdentity;
+    } catch {
+      throw new ApiKeyResolveError({ unreachable: true });
+    }
+  }
+
+  // Retry-After may be an HTTP-date rather than delay-seconds; only keep a finite number.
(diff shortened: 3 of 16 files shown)


packages/agent-bff/src/api-key/api-key-authenticator.ts
```diff
@@ -0,0 +1,94 @@
...
+          const mapped = mapResolveError(error);
+          if (NEGATIVE_CACHE_STATUSES.has(mapped.status)) cache.setNegative(hash, mapped);
+
+          throw mapped;
+        }
+
+        throw error;
+      }
+
+      cache.setPositive(hash, identity);
```
Tonours, line 90: We `setPositive` before minting, and `isResolvedIdentity` does not check `tags`. So a 200 without `tags` gets cached, then `tagsToRecord` throws a raw TypeError, served as a 500. Every subsequent request for that key hits the cached identity and 500s again for the full 60s TTL. Mint before caching, or assert `Array.isArray(user.tags)` in the guard.

packages/agent-bff/src/api-key/resolve-cache.ts
```diff
@@ -0,0 +1,99 @@
...
+
+      return undefined;
+    }
+
+    return entry;
+  }
+
+  function store(hash: string, entry: CacheEntry): void {
+    purgeExpired();
+    if (!entries.has(hash) && entries.size >= maxEntries) return;
```
Tonours: When the map is full, `store()` drops new keys instead of evicting. Negatives share the same map, so anyone spraying well-formed keys fills it with 401s and freezes it, and every valid key then re-hits the SaaS on each request. Evict the oldest, or bound negatives separately.

packages/agent-bff/src/api-key/api-key-client.ts
```diff
@@ -0,0 +1,114 @@
...
+    // A well-formed HTTP 200 with an incomplete body is an unusable resolution, not a caller
+    // error: surface it as unavailable rather than letting a later `user` deref throw a 500.
+    if (!ApiKeyClient.isResolvedIdentity(body)) {
+      throw new ApiKeyResolveError({ unreachable: true });
+    }
+
+    return body;
+  }
+
+  private static isResolvedIdentity(body: unknown): body is ResolvedApiKeyIdentity {
```
Tonours, line 90: This guard mints an auth token but only checks `user` and `renderingId`. Validate everything the mint reads (`id`, `email`, `team`, `permissionLevel`, `Array.isArray(tags)`) so a partial body fails here instead of minting a token with undefined claims.

packages/agent-bff/src/api-key/api-key-client.ts
```diff
@@ -0,0 +1,114 @@
...
+      candidate.user !== null &&
+      typeof candidate.renderingId === 'number'
+    );
+  }
+
+  // Retry-After may be an HTTP-date rather than delay-seconds; only keep a finite number.
+  private static parseRetryAfter(header: string | null): number | undefined {
+    if (header === null) return undefined;
+
+    const seconds = Number(header);
```
Tonours, line 123: `Number` treats `-5`, `3.7` and `0` (empty header) as finite, and `?? 5` does not catch `0` or `-5`. So we can emit `Retry-After: -5`. Require an integer `>= 0`.

packages/agent-bff/src/api-key/agent-token.ts
```diff
@@ -0,0 +1,41 @@
...
+  const { user, renderingId } = identity;
+  const firstName = user.firstName ?? '';
+  const lastName = user.lastName ?? '';
+  const tags = tagsToRecord(user.tags);
+
+  // snake_case aliases: Ruby/Python agents splat JWT claims into Caller (snake_case kwargs).
+  return jsonwebtoken.sign(
+    {
+      id: user.id,
+      email: user.email,
```
Tonours, line 26: This token shares `forestAuthSecret` with the OAuth `bff_access` token but has no `type`/`aud` claim, so a verifier cannot tell them apart; adding a `type` claim would fix it. It also emits `rendering_id` as a number while `bff-token.ts` and `mcp-server` both `String()` it, so it should align on the string form. Both are worth confirming with the agent team before Slice 3.

packages/agent-bff/src/cli-core.ts
```diff
@@ -75,12 +79,53 @@ async function buildOAuthMiddlewares(config: BFFConfig, logger: Logger): Promise
...
+interface ResolvedApiKeyConfig {
+  forestServerUrl: string;
+  forestEnvSecret: string;
+  forestAuthSecret: string;
+}
+
+function resolveApiKeyConfig(config: BFFConfig): ResolvedApiKeyConfig | undefined {
+  const { forestServerUrl, forestEnvSecret, forestAuthSecret } = config;
+
+  if (forestServerUrl && forestEnvSecret && forestAuthSecret) {
```
Tonours: Auth activates on 3 secrets, but `/health` degrades without `AGENT_URL`, so a node can resolve keys and mint tokens while reporting 503 with no proxy target to forward to. Adding `AGENT_URL` to the `resolveApiKeyConfig` check (or gating on `config.hasAllRequired`) keeps activation and health in sync.

packages/agent-bff/src/api-key/api-key-client.ts
```diff
@@ -0,0 +1,114 @@
...
+
+const DEFAULT_HEADERS = { 'Content-Type': 'application/json' } as const;
+const REQUEST_TIMEOUT_MS = 10_000;
+const RESOLVE_PATH = '/liana/v1/bff-api-keys/resolve';
+
+interface SaasErrorBody {
+  errors?: { name?: string; meta?: { code?: string } }[];
+}
+
+export default class ApiKeyClient {
```
Tonours, line 35: This re-implements `ForestServerClient` (same `url()`, headers, timeout), so `resolveApiKey` would be better as a method on it rather than a second client. And `issueAgentToken` duplicates `issueBffAccessToken` down to the copy-pasted snake_case comment, so pulling the claim building into a shared helper keeps the two token minters from drifting.

packages/agent-bff/src/api-key/api-key-authenticator.ts
nbouliol: Good catch. Fixed in 86caf50b0: the client guard now requires Array.isArray(user.tags) (and the other minted fields), and we mint before setPositive so an unmintable identity is never cached.

packages/agent-bff/src/api-key/resolve-cache.ts
nbouliol: Fixed in 86caf50b0: store() now evicts the oldest entry when full instead of dropping new ones, so a burst of negatives can no longer freeze the cache.

packages/agent-bff/src/api-key/api-key-client.ts
nbouliol: Done in 86caf50b0: isResolvedIdentity now validates id/email/team/permissionLevel/Array.isArray(tags) plus renderingId and allowedOrigins, so a partial body fails as unavailable before minting.

packages/agent-bff/src/api-key/api-key-client.ts
nbouliol: Fixed in 86caf50b0: parseRetryAfter now keeps only a positive integer; date/negative/zero fall back to the default backoff.

packages/agent-bff/src/cli-core.ts
nbouliol: Added AGENT_URL to the activation gate in 86caf50b0 (Mode 2 has no target without it). Making /health reflect the auth modes actually enabled is left to T6 (request-edge, PRD-666), which owns the edge and the health/CORS surface.

packages/agent-bff/src/api-key/agent-token.ts
nbouliol: Deferring both to Slice 3 as you suggest: that's where the real agent call lands and where the claim shape gets confirmed with the agent team. Will align rendering_id to the String() form and add a distinguishing type/aud claim then.

packages/agent-bff/src/api-key/api-key-client.ts
nbouliol: Deliberate for this slice: an isolated client keeps Mode 2 independently deployable/rollback-able (a ticket requirement) without touching the OAuth ForestServerClient or its public export. Promoting a shared Forest SaaS client + a shared claim-builder is planned as a follow-up cleanup, likely with T6.

packages/agent-bff/src/cli-core.ts
nbouliol: Following up: I reverted the AGENT_URL gate in 8c994772e. Gating on it silently disables Mode 2 auth when AGENT_URL is missing (the middleware doesn't use it, only server/env/auth secrets). Since AGENT_URL is already a REQUIRED_KEY, /health already reports degraded without it, so activation/health stay in sync without coupling auth to an unused field. Full health-reflects-modes stays with T6.

packages/agent-bff/src/cli-core.ts
```diff
@@ -75,12 +79,53 @@ async function buildOAuthMiddlewares(config: BFFConfig, logger: Logger): Promise
...
+
+function resolveApiKeyConfig(config: BFFConfig): ResolvedApiKeyConfig | undefined {
+  const { forestServerUrl, forestEnvSecret, forestAuthSecret, agentUrl } = config;
+
+  if (forestServerUrl && forestEnvSecret && forestAuthSecret && agentUrl) {
+    return { forestServerUrl, forestEnvSecret, forestAuthSecret };
+  }
+
+  return undefined;
+}
```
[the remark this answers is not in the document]
nbouliol: Agreed, fixed in 8c994772e: dropped agentUrl from resolveApiKeyConfig so the guard checks only the fields the middleware uses. AGENT_URL remains a REQUIRED_KEY so /health still degrades without it.

Tonours (approved):

PR #1731, opened 2026-07-01
feat(agent-bff): request-edge wiring (timezone + CORS + auth-mode precedence)

nbouliol:
Adds a hardened `/agent/*` request edge to `agent-bff` with a structured, type-first error contract (`{ error: { type, status, message, details? } }`) so consumers branch on `error.type` instead of message text.

## Scope

- **Auth-mode precedence** — `oauth` (Bearer) vs `api-key` (`X-Forest-Bff-Key`), mutually exclusive per request: both → `400 ambiguous_credentials`, neither → `401 unauthorized`. `bff_access` validated at the edge (expired → `session_expired`, bad → `unauthorized`). Fails closed (`401`) for api-key mode when the resolver is unconfigured.
- **CORS** — hand-rolled exact-origin allow-list (no wildcard, default ports normalized) from `BFF_ALLOWED_ORIGINS`, applied globally incl. `POST /oauth/token`; preflight short-circuits before auth. Per-key layer-2 intersection (`api-key`) → `403 origin_not_allowed`.
- **Timezone** — resolved header → body → `BFF_DEFAULT_TIMEZONE`, IANA-validated (mirrors the agent), injected into the (stubbed) agent query; missing → `400 missing_timezone`, invalid → `400 invalid_timezone`.
- **Error contract** — single shared error middleware; `api-key-middleware` now throws; malformed `/agent` bodies serialized into the structured shape.

The real agent proxy is Slice 3; the edge terminates in a `501` stub that exposes the built query for tests.

## Notes / decisions

- `/oauth/*` keep RFC-6749 error shape; `session_invalidated` stays a `/oauth/token` refresh-reuse error.
- A still-valid `bff_access` (≤15 min) remains usable at the edge after reuse-invalidation (standard short-lived-token trade-off) — documented in the README.

## Test plan

- `yarn workspace @forestadmin/agent-bff test` — 353 passing (auth precedence, CORS layers + preflight, timezone resolution + injection, full error-contract suite, config parsing).
- `yarn workspace @forestadmin/agent-bff lint` / `build` — clean.

Fixes PRD-666

> [!NOTE]
> ### Wire request edge for /agent/* with timezone, CORS, and auth-mode precedence in agent-bff
> - Adds a full request edge for `/agent/*` routes in [cli-core.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1731/files#diff-76c14e8251a9859dc672905a36febbeb3311acac1dd54a57a05f65282794f19c): CORS, structured error handling, auth-mode resolution, per-key origin enforcement, and timezone resolution are applied in order before reaching the agent stub.
> - Auth-mode middleware (`createAuthModeMiddleware`) resolves OAuth vs API-key from request headers, validates HS256-signed `bff_access` tokens, and throws structured errors (`400 ambiguous_credentials`, `401 unauthorized`, `401 session_expired`) on failure.
> - CORS is enforced via an allow-list from `BFF_ALLOWED_ORIGINS`; per-key origin restrictions add a second layer for API-key authenticated requests.
> - Timezone is resolved from `X-Forest-Timezone` header, request body, or `BFF_DEFAULT_TIMEZONE` fallback; invalid or missing values return structured 4xx errors.
> - Centralized error middleware serializes all `BffHttpError` instances to a consistent structured body; unknown errors map to `500 internal_error`.
> - Risk: the agent edge is disabled entirely (with a warning) when `FOREST_AUTH_SECRET` is absent, and API-key requests fail closed with `401` if no resolver is configured.
>
> 
> #### Changes since #1731 opened
>
> - Modified JWT token verification in `verifyBffAccess` utility to separate signature verification from expiration checking, validating token type ('bff_access') before manually inspecting the `exp` claim [75b8fb1]
> - Added specific error type mappings for HTTP status codes 413 and 415 in `createErrorMiddleware` factory [75b8fb1]
> - Updated `isValidTimezone` utility to cache canonical timezone names instead of raw input values [75b8fb1]
> - Added validation in `auth-mode-middleware.verifyBffAccess` to verify the `exp` claim is a number before checking token expiry, rejecting tokens with non-numeric or missing `exp` claims with unauthorized status [55db5ef]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 34c3be4.</sup>
> 
>
>

### DIFF
--- a/packages/agent-bff/src/auth/auth-mode-middleware.ts
+++ b/packages/agent-bff/src/auth/auth-mode-middleware.ts
@@ -38,7 +38,11 @@ function verifyBffAccess(token: string, authSecret: string): BffAccessTokenPaylo
 
   const { exp } = decoded as { exp?: unknown };
 
-  if (typeof exp === 'number' && exp * 1000 <= Date.now()) {
+  if (typeof exp !== 'number') {
+    throw unauthorized();
+  }
+
+  if (exp * 1000 <= Date.now()) {
     throw sessionExpired();
   }
 
--- a/packages/agent-bff/src/auth/auth-mode-middleware.ts
+++ b/packages/agent-bff/src/auth/auth-mode-middleware.ts
@@ -1,0 +1,55 @@
+import type { BffAccessTokenPayload } from '../oauth/bff-token';
+import type { Middleware } from 'koa';
+
+import jsonwebtoken from 'jsonwebtoken';
+
+import { extractBearerToken, resolveAuthMode } from './auth-mode';
+import { sessionExpired, unauthorized } from '../http/bff-http-error';
+import { BFF_ACCESS_TOKEN_TYPE } from '../oauth/bff-token';
+
+export const BFF_KEY_HEADER = 'X-Forest-Bff-Key';
+
+export interface AuthModeMiddlewareOptions {
+  authSecret: string;
+}
+
+function verifyBffAccess(token: string, authSecret: string): BffAccessTokenPayload {
+  let decoded: unknown;
+
+  try {
+    decoded = jsonwebtoken.verify(token, authSecret, { algorithms: ['HS256'] });
+  } catch (error) {
+    if (error instanceof jsonwebtoken.TokenExpiredError) throw sessionExpired();
+
+    throw unauthorized();
+  }
+
+  if (
+    typeof decoded !== 'object' ||
+    decoded === null ||
+    (decoded as { type?: unknown }).type !== BFF_ACCESS_TOKEN_TYPE
+  ) {
+    throw unauthorized();
+  }
+
+  return decoded as BffAccessTokenPayload;
+}
+
+export default function createAuthModeMiddleware({
+  authSecret,
+}: AuthModeMiddlewareOptions): Middleware {
+  return async function authModeMiddleware(ctx, next) {
+    const bearer = extractBearerToken(ctx.get('Authorization'));
+    const apiKey = ctx.get(BFF_KEY_HEADER);
+
+    const mode = resolveAuthMode({ hasBearer: bearer !== undefined, hasApiKey: apiKey !== '' });
+
+    ctx.state.authMode = mode;
+
+    if (mode === 'oauth') {
+      ctx.state.principal = verifyBffAccess(bearer as string, authSecret);
+    }
+
+    await next();
+  };
+}
--- a/packages/agent-bff/src/auth/auth-mode-middleware.ts
+++ b/packages/agent-bff/src/auth/auth-mode-middleware.ts
@@ -16,11 +16,15 @@ export interface AuthModeMiddlewareOptions {
 function verifyBffAccess(token: string, authSecret: string): BffAccessTokenPayload {
   let decoded: unknown;
 
+  // Verify the signature first but ignore expiration, so the token type can be
+  // checked before expiry: a wrong-typed token must be `unauthorized`, and only
+  // a genuine expired `bff_access` should map to `session_expired`.
   try {
-    decoded = jsonwebtoken.verify(token, authSecret, { algorithms: ['HS256'] });
-  } catch (error) {
-    if (error instanceof jsonwebtoken.TokenExpiredError) throw sessionExpired();
-
+    decoded = jsonwebtoken.verify(token, authSecret, {
+      algorithms: ['HS256'],
+      ignoreExpiration: true,
+    });
+  } catch {
     throw unauthorized();
   }
 
@@ -30,6 +34,12 @@ function verifyBffAccess(token: string, authSecret: string): BffAccessTokenPaylo
     (decoded as { type?: unknown }).type !== BFF_ACCESS_TOKEN_TYPE
   ) {
     throw unauthorized();
+  }
+
+  const { exp } = decoded as { exp?: unknown };
+
+  if (typeof exp === 'number' && exp * 1000 <= Date.now()) {
+    throw sessionExpired();
   }
 
   return decoded as BffAccessTokenPayload;
--- a/packages/agent-bff/src/cors/cors-middleware.ts
+++ b/packages/agent-bff/src/cors/cors-middleware.ts
@@ -1,0 +1,40 @@
+import type { Middleware } from 'koa';
+
+import { originAllowed } from './origin';
+
+export const ALLOWED_METHODS = 'GET, POST, PUT, PATCH, DELETE, OPTIONS';
+export const ALLOWED_HEADERS =
+  'Authorization, Content-Type, X-Forest-Timezone, X-Forest-Bff-Key, X-Request-Id';
+export const PREFLIGHT_MAX_AGE_SECONDS = 600;
+
+export interface CorsMiddlewareOptions {
+  allowedOrigins: string[];
+}
+
+export default function createCorsMiddleware({
+  allowedOrigins,
+}: CorsMiddlewareOptions): Middleware {
+  return async function corsMiddleware(ctx, next) {
+    const origin = ctx.get('Origin');
+
+    if (origin) ctx.vary('Origin');
+
+    const allowed = origin ? originAllowed(origin, allowedOrigins) : false;
+
+    if (allowed) ctx.set('Access-Control-Allow-Origin', origin);
+
+    if (ctx.method === 'OPTIONS') {
+      if (allowed) {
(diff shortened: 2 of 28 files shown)


packages/agent-bff/src/cors/cors-middleware.ts
```diff
@@ -0,0 +1,40 @@
...
+    const allowed = origin ? originAllowed(origin, allowedOrigins) : false;
+
+    if (allowed) ctx.set('Access-Control-Allow-Origin', origin);
+
+    if (ctx.method === 'OPTIONS') {
+      if (allowed) {
+        ctx.set('Access-Control-Allow-Methods', ALLOWED_METHODS);
+        ctx.set('Access-Control-Allow-Headers', ALLOWED_HEADERS);
+        ctx.set('Access-Control-Max-Age', String(PREFLIGHT_MAX_AGE_SECONDS));
+      }
```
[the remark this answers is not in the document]
nbouliol: Intentional, not applying this one.

We deliberately set `Access-Control-Allow-Origin` on **all** responses (including 4xx), not only on success. This edge exists to expose a type-first error contract (`{ error: { type, ... } }`) that browser consumers (e.g. the Zendesk iframe) read via `fetch`. If ACAO were only set after a successful `next()`, an allow-listed origin could no longer read `error.type` on `400 missing_timezone`, `401 unauthorized`, `invalid_api_key`, etc. — which defeats the contract.

On the specific `403 origin_not_allowed` case: the origin is already in the global `BFF_ALLOWED_ORIGINS` (a trusted BFF origin), the response body is a generic error with no sensitive data, and the protected resource is never reached (the per-key check throws before the agent stub). So exposing that error body to an already-trusted origin isn't a leak — it only tells it 'this key isn't usable from your origin'.

Keeping ACAO before `next()`.

packages/agent-bff/src/timezone/timezone.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (value === '') return false;
+  if (VALID_TIMEZONES.has(value)) return true;
+
+  try {
+    Intl.DateTimeFormat('en-US', { timeZone: value });
+  } catch {
+    return false;
+  }
+
+  VALID_TIMEZONES.add(value);
```
Tonours: `isValidTimezone` caches the raw header value, and Intl matches IANA names case-insensitively while we store the original casing. A client sending one valid zone in many casings (e.g. `America/Argentina/Buenos_Aires`, ~2^26 accepted variants) grows this Set without bound for the process lifetime, shared across all tenants: a memory-exhaustion vector on an unauthenticated-shaped input. What about caching the canonical form (`Intl.DateTimeFormat('en-US', { timeZone: value }).resolvedOptions().timeZone`) instead of `value`, or bounding the Set? Verified empirically: 50000 mixed-case variants of one zone left 49991 entries.

packages/agent-bff/src/auth/auth-mode-middleware.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export interface AuthModeMiddlewareOptions {
+  authSecret: string;
+}
+
+function verifyBffAccess(token: string, authSecret: string): BffAccessTokenPayload {
+  let decoded: unknown;
+
+  try {
+    decoded = jsonwebtoken.verify(token, authSecret, { algorithms: ['HS256'] });
```
Tonours: `jsonwebtoken.verify` throws `TokenExpiredError` before the `type` check on lines 27-33 runs, so a validly-signed but expired token of the wrong type returns `session_expired` instead of `unauthorized` (the README states wrong-typed must be `unauthorized`). Not a bypass since it is still 401, but it tells the client to refresh a session that was never a valid `bff_access`, risking a pointless refresh loop. What about verifying with `{ ignoreExpiration: true }`, checking `type` first, then rejecting expiry explicitly?

packages/agent-bff/src/http/error-middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+        return;
+      }
+
+      const clientStatus = clientErrorStatus(error);
+
+      if (clientStatus !== undefined) {
+        ctx.status = clientStatus;
+        ctx.body = toErrorBody({
+          status: clientStatus,
+          type: 'invalid_request',
```
Tonours: Any thrown error carrying a 4xx `status`/`statusCode` but no string `type` (e.g. a bodyParser `413`) is relabeled `invalid_request` here, so the response has a status/type mismatch (`413` body with `type: invalid_request`) and breaks the type-first contract consumers branch on. Not live in this slice since no bodyParser is wired to `/agent/*` yet, but the logic is merged. What about mapping known statuses to their type (413 to `payload_too_large`, 415 to `unsupported_media_type`) instead of a blanket `invalid_request`?

packages/agent-bff/src/http/bff-http-error.ts
```diff
@@ -0,0 +1,90 @@
...
+  | 'session_expired'
+  | 'origin_not_allowed'
+  | 'missing_timezone'
+  | 'invalid_timezone';
+
+export class BffHttpError extends Error {
+  readonly status: number;
+  readonly type: string;
+  readonly details?: unknown;
+  readonly retryAfter?: number;
```
Tonours: `retryAfter` is declared here but no `BffHttpError` factory ever sets it (only `ApiKeyError` does, and it is caught structurally), so the field is permanently dead on this class. Separately, `BffErrorType` is exported but `type` is typed as plain `string` on line 11, so the union constrains nothing: adding a member type-checks nothing. What about dropping `retryAfter` from `BffHttpError`, and either typing `type: BffErrorType` or deleting the union?

packages/agent-bff/src/timezone/timezone.ts
nbouliol: Good catch, fixed in 75b8fb1b8. `isValidTimezone` now caches the canonical form (`Intl.DateTimeFormat('en-US', { timeZone: value }).resolvedOptions().timeZone`) instead of the raw input, so mixed-casing variants of one zone all collapse to a single entry and the Set stays bounded to real IANA zones.

packages/agent-bff/src/auth/auth-mode-middleware.ts
nbouliol: Fixed in 75b8fb1b8. Now verifying with `{ ignoreExpiration: true }`, checking `type` first, then rejecting expiry explicitly — a wrong-typed token is `unauthorized` regardless of expiry, and only a genuine expired `bff_access` maps to `session_expired`.

packages/agent-bff/src/http/error-middleware.ts
nbouliol: Fixed in 75b8fb1b8. Known client statuses now map to their own type (413 → `payload_too_large`, 415 → `unsupported_media_type`), defaulting to `invalid_request` otherwise, so status and type stay consistent. (Minor: bodyParser is in fact wired to /agent/* — it's mounted globally and the agent-scoped error middleware precedes it — so this was live, not just merged.)

packages/agent-bff/src/http/bff-http-error.ts
nbouliol: Fixed in 75b8fb1b8. Dropped the dead `retryAfter` field from `BffHttpError` (only `ApiKeyError` sets it, and it's read structurally) and removed the unused `BffErrorType` union. Kept `type: string` on the class since the shared serializer also handles `ApiKeyError` and plain http-errors.

packages/agent-bff/src/auth/auth-mode-middleware.ts
```diff
@@ -0,0 +1,65 @@
...
+    (decoded as { type?: unknown }).type !== BFF_ACCESS_TOKEN_TYPE
+  ) {
+    throw unauthorized();
+  }
+
+  const { exp } = decoded as { exp?: unknown };
+
+  if (typeof exp === 'number' && exp * 1000 <= Date.now()) {
+    throw sessionExpired();
+  }
```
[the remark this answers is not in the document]
nbouliol: Fixed in 55db5ef1d. A missing or non-numeric `exp` is now rejected as `unauthorized` before the expiry comparison, so a `bff_access` without `exp` is no longer accepted indefinitely. (Legitimately-issued tokens always carry `exp` via `expiresIn`; this closes the ignoreExpiration path explicitly.)

Tonours (approved):

PR #1732, opened 2026-07-02
feat(agent-bff): fetch and cache agent schema, capabilities and allow-list

nbouliol:
Fixes PRD-668

## What

Adds the BFF read-model layer (Slice 2): an in-memory, per-instance cache of the agent schema and the metadata every later BFF slice reads instead of re-fetching.

- **Schema cache** (`schema-cache.ts`) — fetches via `SchemaService.getSchema()` (MCP precedent), 24h TTL, in-flight dedup. The TTL only *triggers* a refresh; the last good schema is served until a refresh succeeds. Cold cache → typed `SchemaUnavailableError`; warm refresh failure → serves last-good + `schema_cache_refresh_error`. Exposes `schema_cache_age_seconds`.
- **Read-model** (`read-model.ts`) — collection/relation/action allow-lists and relation targets, keyed `(collection, name)` (relation/action names are collection-scoped). Action allow-list = endpoint-map keys (endpoint-less actions are not exposed). Action-endpoint map typed as agent-client `ActionEndpointsByCollection`.
- **Action-endpoint resolver** (`action-endpoint-resolver.ts`) — never throws; emits `miss`/`error` counters tagged `rendering`/`collection`/`action`.
- **Capabilities cache** (`capabilities-cache.ts`) — per-collection, 24h, fetcher passed per call (bound to the request token), invalidated with the schema. A generation guard prevents an in-flight fetch from repopulating stale data after a refresh.
- **`ReadModelStore`** — single owner of the coupled schema + capabilities lifecycle: a successful refresh rebuilds the read-model and clears capabilities atomically.
- **`Metrics` port** + console adapter (no metrics backend exists yet).

## Out of scope (later slices)

No operator/field validation (PRD-640), no data endpoints (PRD-639), no nested-relation guard (PRD-669), no permissions/OpenAPI, no active cache invalidation. Nothing is wired into the HTTP runtime yet — Slice 3 (PRD-671) constructs and injects it.

## Tests

Build, lint, and the full suite pass (403 tests, incl. 42 new). Covers cache hit/miss, 24h boundary, cold/warm failure + re-attempt, schema-refresh invalidates capabilities, collection-scoped keying, and the miss/error counters.

> [!NOTE]
> ### Add agent schema and capabilities read-model with TTL caching to agent-bff
> - Introduces a `ReadModel` built from Forest schema that answers allow-list queries for collections, relations, and actions, and supplies immutable action-endpoint mappings.
> - Adds `SchemaCache` and `CapabilitiesCache` with 24h TTL, deduped concurrent fetches, and monotonic revision tracking; schema revision changes trigger capabilities invalidation via `ReadModelStore`.
> - Adds `ForestSchemaClient` (wrapping `SchemaService`) and `createAgentCapabilitiesFetcher` (wrapping `@forestadmin/agent-client`) as the two remote data sources.
> - Wires everything together in `createReadModel`, which returns a store and an `ActionEndpointResolver`; metrics default to a console logger sink if none are provided.
> - Risk: a cold schema fetch failure throws `SchemaUnavailableError` with no fallback; warm failures fall back to last known schema silently.
>
> 
> #### Changes since #1732 opened
>
> - Added defensive defaulting in `ReadModel.buildActionEndpoints` method for malformed action schemas [6486f7e]
> - Added test coverage for defensive defaulting of malformed action schemas [6486f7e]
> - Wrapped metrics emission calls in `ActionEndpointResolver.resolve` with a new `ActionEndpointResolver.safeIncrement` method that swallows exceptions thrown by the `Metrics` backend [2acfc83]
> - Replaced plain object literals with null-prototype objects in `ReadModel` for action endpoint storage [2acfc83]
> - Added test coverage for metrics backend exception handling in `ActionEndpointResolver` and prototype member collision scenarios in `ReadModel` [2acfc83]
> - Wrapped the `Metrics` instance with `safeMetrics` adapter in `createReadModel` factory so all components receive a metrics implementation that catches and swallows exceptions from `increment` and `gauge` calls [82f1155]
> - Refactored `ActionEndpointResolver.resolve` method to remove internal `safeIncrement` helper and call `this.metrics.increment` directly for `ACTION_ENDPOINT_ERROR` and `ACTION_ENDPOINT_MISS` events [82f1155]
> - Created `safeMetrics` adapter function that wraps a `Metrics` implementation and delegates `increment` and `gauge` calls within try-catch blocks [82f1155]
> - Added tests for `safeMetrics` adapter verifying delegation and exception swallowing behavior for `increment` and `gauge` methods [82f1155]
> - Removed tests asserting that `ActionEndpointResolver` never throws when metrics backend throws, and added test in `createReadModel` suite verifying the factory wraps throwing metrics backends [82f1155]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0d226e2.</sup>
> 
>
>

### DIFF
--- a/packages/agent-bff/src/read-model/action-endpoint-resolver.ts
+++ b/packages/agent-bff/src/read-model/action-endpoint-resolver.ts
@@ -37,7 +37,7 @@ export default class ActionEndpointResolver {
     try {
       readModel = await this.getReadModel();
     } catch {
-      this.safeIncrement(ACTION_ENDPOINT_ERROR, tags);
+      this.metrics.increment(ACTION_ENDPOINT_ERROR, tags);
 
       return undefined;
     }
@@ -45,20 +45,11 @@ export default class ActionEndpointResolver {
     const info = readModel.getActionEndpoints()[collection]?.[action];
 
     if (!info) {
-      this.safeIncrement(ACTION_ENDPOINT_MISS, tags);
+      this.metrics.increment(ACTION_ENDPOINT_MISS, tags);
 
       return undefined;
     }
 
     return info;
-  }
-
-  // Metrics is a user-supplied port; a throwing backend must not break the "never throws" contract.
-  private safeIncrement(name: string, tags: Record<string, string | number>): void {
-    try {
-      this.metrics.increment(name, tags);
-    } catch {
-      // Swallow: emitting a metric must never turn a missing action into a request failure.
-    }
   }
 }
--- a/packages/agent-bff/src/read-model/action-endpoint-resolver.ts
+++ b/packages/agent-bff/src/read-model/action-endpoint-resolver.ts
@@ -1,0 +1,55 @@
+import type ReadModel from './read-model';
+import type { Metrics } from '../ports/metrics-port';
+import type { ActionEndpointsByCollection } from '@forestadmin/agent-client';
+
+export const ACTION_ENDPOINT_MISS = 'action_endpoint_miss';
+export const ACTION_ENDPOINT_ERROR = 'action_endpoint_error';
+
+export type ActionEndpointInfo = ActionEndpointsByCollection[string][string];
+
+export type ReadModelProvider = () => Promise<ReadModel>;
+
+export interface ResolveActionContext {
+  rendering: string | number;
+}
+
+/**
+ * Resolves an action to its endpoint against the current read-model. Never throws: an absent
+ * mapping emits the miss counter, a failure to obtain the read-model emits the error counter.
+ * Both counters carry `rendering`, `collection`, and `action` tags. In normal flow the action
+ * allow-list already excludes endpoint-less actions, so the miss path is a defensive guard.
+ */
+export default class ActionEndpointResolver {
+  constructor(
+    private readonly getReadModel: ReadModelProvider,
+    private readonly metrics: Metrics,
+  ) {}
+
+  async resolve(
+    collection: string,
+    action: string,
+    { rendering }: ResolveActionContext,
+  ): Promise<ActionEndpointInfo | undefined> {
+    const tags = { rendering, collection, action };
+
+    let readModel: ReadModel;
+
+    try {
+      readModel = await this.getReadModel();
+    } catch {
+      this.metrics.increment(ACTION_ENDPOINT_ERROR, tags);
+
+      return undefined;
+    }
+
+    const info = readModel.getActionEndpoints()[collection]?.[action];
+
+    if (!info) {
+      this.metrics.increment(ACTION_ENDPOINT_MISS, tags);
+
+      return undefined;
+    }
+
+    return info;
+  }
+}
--- a/packages/agent-bff/src/read-model/action-endpoint-resolver.ts
+++ b/packages/agent-bff/src/read-model/action-endpoint-resolver.ts
@@ -37,7 +37,7 @@ export default class ActionEndpointResolver {
     try {
       readModel = await this.getReadModel();
     } catch {
-      this.metrics.increment(ACTION_ENDPOINT_ERROR, tags);
+      this.safeIncrement(ACTION_ENDPOINT_ERROR, tags);
 
       return undefined;
     }
@@ -45,11 +45,20 @@ export default class ActionEndpointResolver {
     const info = readModel.getActionEndpoints()[collection]?.[action];
 
     if (!info) {
-      this.metrics.increment(ACTION_ENDPOINT_MISS, tags);
+      this.safeIncrement(ACTION_ENDPOINT_MISS, tags);
 
       return undefined;
     }
 
     return info;
+  }
+
+  // Metrics is a user-supplied port; a throwing backend must not break the "never throws" contract.
+  private safeIncrement(name: string, tags: Record<string, string | number>): void {
+    try {
+      this.metrics.increment(name, tags);
+    } catch {
+      // Swallow: emitting a metric must never turn a missing action into a request failure.
+    }
   }
 }
--- a/packages/agent-bff/src/read-model/agent-capabilities-fetcher.ts
+++ b/packages/agent-bff/src/read-model/agent-capabilities-fetcher.ts
@@ -1,0 +1,21 @@
+import type { CapabilitiesFetcher } from './capabilities-cache';
+
+import { createRemoteAgentClient } from '@forestadmin/agent-client';
+
+export interface AgentCapabilitiesFetcherOptions {
+  agentUrl: string;
+  token: string;
+}
+
+/**
+ * Builds a capabilities fetcher bound to a request's agent token. The cache calls it only on a
+ * miss, so the token of whichever request first populates a collection is the one used.
+ */
+export default function createAgentCapabilitiesFetcher({
+  agentUrl,
+  token,
+}: AgentCapabilitiesFetcherOptions): CapabilitiesFetcher {
+  const client = createRemoteAgentClient({ url: agentUrl, token });
+
+  return collection => client.collection(collection).capabilities();
+}
--- a/packages/agent-bff/src/read-model/capabilities-cache.ts
+++ b/packages/agent-bff/src/read-model/capabilities-cache.ts
@@ -1,0 +1,68 @@
+import { ONE_DAY_MS } from './schema-cache';
+
+export interface CapabilitiesResult {
+  fields: { name: string; type: string; operators: string[] }[];
+}
+
+export type CapabilitiesFetcher = (collection: string) => Promise<CapabilitiesResult>;
+
+export interface CapabilitiesCacheOptions {
+  now?: () => number;
+  ttlMs?: number;
+}
+
+interface CacheEntry {
+  result: CapabilitiesResult;
+  fetchedAt: number;
+}
+
+/**
+ * Per-collection capabilities cache (24h). The fetcher is passed per call so it can be bound to
+ * the caller's request token. Invalidated together with the schema via `clear()`.
+ */
+export default class CapabilitiesCache {
+  private readonly now: () => number;
+  private readonly ttlMs: number;
+
+  private readonly entries = new Map<string, CacheEntry>();
+  private readonly inFlight = new Map<string, Promise<CapabilitiesResult>>();
+  private generation = 0;
+
+  constructor({ now = Date.now, ttlMs = ONE_DAY_MS }: CapabilitiesCacheOptions = {}) {
+    this.now = now;
+    this.ttlMs = ttlMs;
+  }
+
+  async get(collection: string, fetcher: CapabilitiesFetcher): Promise<CapabilitiesResult> {
+    const entry = this.entries.get(collection);
+    if (entry && this.now() - entry.fetchedAt < this.ttlMs) return entry.result;
+
+    const pending = this.inFlight.get(collection);
+    if (pending) return pending;
+
+    const { generation } = this;
+    const fetch = fetcher(collection)
+      .then(result => {
+        // Skip the write if a clear() (schema refresh) happened while this fetch was in flight —
+        // its result belongs to the old schema generation and must not repopulate the cache.
+        if (this.generation === generation) {
+          this.entries.set(collection, { result, fetchedAt: this.now() });
+        }
+
+        return result;
+      })
+      .finally(() => {
+        if (this.inFlight.get(collection) === fetch) this.inFlight.delete(collection);
+      });
+
+    this.inFlight.set(collection, fetch);
+
+    return fetch;
+  }
+
+  clear(): void {
+    this.generation += 1;
+    this.entries.clear();
+    this.inFlight.clear();
+  }
+}
--- a/packages/agent-bff/src/read-model/read-model-store.ts
+++ b/packages/agent-bff/src/read-model/read-model-store.ts
@@ -1,0 +1,50 @@
+import type CapabilitiesCache from './capabilities-cache';
+import type { CapabilitiesFetcher, CapabilitiesResult } from './capabilities-cache';
+import type SchemaCache from './schema-cache';
+import type { ForestSchemaCollection } from '@forestadmin/forestadmin-client';
+
+import ReadModel from './read-model';
+
+/**
+ * Single owner of the coupled schema + capabilities lifecycle. A successful schema refresh (a new
+ * collections reference from the cache) rebuilds the read-model and clears capabilities atomically,
+ * so the allow-list and capabilities never split-brain across schema generations.
+ */
+export default class ReadModelStore {
+  private readonly schemaCache: SchemaCache;
+  private readonly capabilitiesCache: CapabilitiesCache;
+
+  private lastCollections: ForestSchemaCollection[] | null = null;
+  private readModel: ReadModel | null = null;
+
+  constructor(schemaCache: SchemaCache, capabilitiesCache: CapabilitiesCache) {
+    this.schemaCache = schemaCache;
+    this.capabilitiesCache = capabilitiesCache;
+  }
+
+  async getReadModel(): Promise<ReadModel> {
+    const collections = await this.schemaCache.get();
+
+    if (collections !== this.lastCollections || !this.readModel) {
+      this.readModel = new ReadModel(collections);
+      this.lastCollections = collections;
+      this.capabilitiesCache.clear();
+    }
+
+    return this.readModel;
+  }
+
+  async getCapabilities(
+    collection: string,
+    fetcher: CapabilitiesFetcher,
+  ): Promise<CapabilitiesResult> {
+    // Ensure any pending schema refresh (and its capabilities invalidation) runs first.
+    await this.getReadModel();
+
+    return this.capabilitiesCache.get(collection, fetcher);
+  }
+
+  ageSeconds(): number | undefined {
+    return this.schemaCache.ageSeconds();
+  }
+}
--- a/packages/agent-bff/src/read-model/read-model-store.ts
+++ b/packages/agent-bff/src/read-model/read-model-store.ts
@@ -1,20 +1,19 @@
 import type CapabilitiesCache from './capabilities-cache';
 import type { CapabilitiesFetcher, CapabilitiesResult } from './capabilities-cache';
 import type SchemaCache from './schema-cache';
-import type { ForestSchemaCollection } from '@forestadmin/forestadmin-client';
 
 import ReadModel from './read-model';
 
 /**
- * Single owner of the coupled schema + capabilities lifecycle. A successful schema refresh (a new
- * collections reference from the cache) rebuilds the read-model and clears capabilities atomically,
- * so the allow-list and capabilities never split-brain across schema generations.
+ * Single owner of the coupled schema + capabilities lifecycle. A successful schema refresh (a bump
+ * of the cache `revision`) rebuilds the read-model and clears capabilities atomically, so the
+ * allow-list and capabilities never split-brain across schema generations.
  */
 export default class ReadModelStore {
   private readonly schemaCache: SchemaCache;
   private readonly capabilitiesCache: CapabilitiesCache;
 
-  private lastCollections: ForestSchemaCollection[] | null = null;
+  private builtRevision = -1;
   private readModel: ReadModel | null = null;
 
   constructor(schemaCache: SchemaCache, capabilitiesCache: CapabilitiesCache) {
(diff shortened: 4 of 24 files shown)


packages/agent-bff/src/read-model/read-model.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (!field.relationship) return null;
+
+  const type = field.relationship;
+
+  if (field.polymorphicReferencedModels && field.polymorphicReferencedModels.length > 0) {
+    return { type, polymorphic: true, targets: [...field.polymorphicReferencedModels] };
+  }
+
+  if (field.reference) {
+    return { type, polymorphic: false, target: field.reference.split('.')[0] };
```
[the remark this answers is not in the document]
nbouliol: Good catch — confirmed as a real bug, not a theoretical one. The mongoose datasource names nested collections with dots (`packages/datasource-mongoose/src/collection.ts:43`, e.g. `User.address`), so a reference like `User.address.id` was being truncated to `User`.

Since the schema builds `reference` as `${foreignCollection}.${targetKey}` and the target key (`foreignKeyTarget`/`originKeyTarget`) is always a single column name, I fixed it by dropping only the trailing key: `reference.split('.').slice(0, -1).join('.')`. Added a test for a dotted target collection.

Fixed in 8615aefee.

packages/agent-bff/src/read-model/schema-cache.ts
```diff
@@ -0,0 +1,99 @@
...
+  private inFlight: Promise<ForestSchemaCollection[]> | null = null;
+
+  constructor({ fetcher, metrics, now = Date.now, ttlMs = ONE_DAY_MS }: SchemaCacheOptions) {
+    this.fetcher = fetcher;
+    this.metrics = metrics;
+    this.now = now;
+    this.ttlMs = ttlMs;
+  }
+
+  async get(): Promise<ForestSchemaCollection[]> {
```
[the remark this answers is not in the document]
nbouliol: Intentional — this is synchronous refresh-on-read, chosen over stale-while-revalidate during design.

The "serve last-good" contract applies on a refresh **failure** (warm cache keeps serving stale + emits `schema_cache_refresh_error`), not while a refresh is in flight. When the TTL expires we deliberately block on the refresh: concurrent readers share a single in-flight fetch (dedup, so one network call), the wait is bounded by the client's 10s timeout, and it only happens at the 24h expiry boundary (~once/day). Serving a stale schema when a fresh one is one `await` away was a conscious no (it also matches the MCP server precedent).

Leaving as-is.

packages/agent-bff/src/read-model/read-model.ts
```diff
@@ -0,0 +1,100 @@
...
+      this.collections.add(collection.name);
+      this.buildRelations(collection);
+      this.buildActionEndpoints(collection);
+    }
+  }
+
+  private buildRelations(collection: ForestSchemaCollection): void {
+    const relations = new Map<string, RelationTarget>();
+
+    for (const field of collection.fields) {
```
Tonours: This throws on a malformed payload where `fields` is absent: guard with `collection.fields ?? []` like `buildActionEndpoints` does for `actions`. It runs outside SchemaCache try/catch, so a bad collection wedges the store for the whole 24h TTL with no SchemaUnavailableError and no metric.

packages/agent-bff/src/read-model/schema-cache.ts
```diff
@@ -0,0 +1,99 @@
...
+      });
+    }
+
+    return this.inFlight;
+  }
+
+  private async doRefresh(): Promise<ForestSchemaCollection[]> {
+    try {
+      const collections = await this.fetcher.fetchSchema();
+      this.entry = { collections, fetchedAt: this.now() };
```
Tonours, line 86: An empty `collections` from a transient 200 gets cached as last-good for 24h and silently denies every collection/relation/action. What about rejecting an empty result here so the cold-cache error path fires instead?

packages/agent-bff/src/read-model/capabilities-cache.ts
```diff
@@ -0,0 +1,68 @@
...
+  async get(collection: string, fetcher: CapabilitiesFetcher): Promise<CapabilitiesResult> {
+    const entry = this.entries.get(collection);
+    if (entry && this.now() - entry.fetchedAt < this.ttlMs) return entry.result;
+
+    const pending = this.inFlight.get(collection);
+    if (pending) return pending;
+
+    const { generation } = this;
+    const fetch = fetcher(collection)
+      .then(result => {
```
Tonours, line 45: Unlike `SchemaCache.doRefresh`, this has no `.catch`: a failing `capabilities()` propagates raw with no error counter and no last-good fallback. What about mirroring the schema cache here (emit a counter, serve stale)?

packages/agent-bff/src/read-model/capabilities-cache.ts
```diff
@@ -0,0 +1,68 @@
...
+  private readonly inFlight = new Map<string, Promise<CapabilitiesResult>>();
+  private generation = 0;
+
+  constructor({ now = Date.now, ttlMs = ONE_DAY_MS }: CapabilitiesCacheOptions = {}) {
+    this.now = now;
+    this.ttlMs = ttlMs;
+  }
+
+  async get(collection: string, fetcher: CapabilitiesFetcher): Promise<CapabilitiesResult> {
+    const entry = this.entries.get(collection);
```
Tonours, line 37: Cache is keyed by collection only while the token is bound in the fetcher, so the first caller result is served to every token for 24h. Can we confirm `/capabilities` is not token-scoped server-side? If it is, key by token.

packages/agent-bff/src/read-model/read-model-store.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  constructor(schemaCache: SchemaCache, capabilitiesCache: CapabilitiesCache) {
+    this.schemaCache = schemaCache;
+    this.capabilitiesCache = capabilitiesCache;
+  }
+
+  async getReadModel(): Promise<ReadModel> {
+    const collections = await this.schemaCache.get();
+
+    if (collections !== this.lastCollections || !this.readModel) {
```
Tonours: Refresh detection by array-reference identity is a contract only the doc-comment enforces, no test pins it: a future defensive copy in SchemaCache would rebuild and clear capabilities on every request. Returning a version counter from `get()` would make it testable.

packages/agent-bff/src/read-model/read-model.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    this.actionEndpoints[collection.name] = {};
+
+    for (const action of withEndpoint) {
+      this.actionEndpoints[collection.name][action.name] = {
+        id: action.id,
+        name: action.name,
+        endpoint: action.endpoint,
+        hooks: action.hooks,
+        fields: action.fields,
```
Tonours: `fields` and `hooks` are stored by reference from the shared cached schema, so a future mutating consumer would corrupt the cache for all requests. What about freezing or shallow-cloning them here?

packages/agent-bff/test/read-model/create-read-model.test.ts
```diff
@@ -0,0 +1,68 @@
...
+      rendering: 3,
+      collection: 'users',
+      action: 'missing',
+    });
+  });
+
+  it('should default to console metrics when none is provided', () => {
+    const bundle = createReadModel({ forestServerUrl: 'x', envSecret: 'y' });
+
+    expect(bundle.store).toBeDefined();
```
Tonours: These assert only `toBeDefined`, not that console metrics were selected: the `metrics ?? createConsoleMetrics` fallback could be deleted and the test still passes. Assert an `increment` call routes through the console logger.

packages/agent-bff/src/read-model/read-model.ts
nbouliol: Fixed in 33c7bc9 — `buildRelations` now guards `collection.fields ?? []`, mirroring `buildActionEndpoints`, so a malformed collection can't wedge the store for the 24h TTL.

packages/agent-bff/src/read-model/schema-cache.ts
nbouliol: Fixed in 33c7bc9 — an empty schema is now treated as a failed fetch (cold → `SchemaUnavailableError` + `schema_cache_refresh_error`, warm → keep serving last good), instead of being cached as an all-denying schema for 24h.

packages/agent-bff/src/read-model/capabilities-cache.ts
nbouliol: Kept as-is, on purpose:

1. The ticket's counter contract only names the schema + action-endpoint counters — a capabilities counter would be new contract.
2. "Serve stale" is unsafe for capabilities specifically: stale operators would make Slice 4 validate filters against the wrong operator set. And there's no last-good on the first fetch anyway.
3. The current behavior already avoids poisoning — the rejection propagates, the in-flight entry is cleared, and the next read re-attempts. The failure belongs to the request-scoped consumer (Slice 3/4), which has the context to map it to HTTP.

Happy to add a capabilities failure counter in Slice 4 if ops wants the signal.

packages/agent-bff/src/read-model/capabilities-cache.ts
nbouliol: Confirmed not token-scoped. `/forest/_internal/capabilities` (`packages/agent/src/routes/capabilities.ts`, `fetchCapabilities`) derives operators from each field's `filterOperators` on the datasource schema — it does not read the caller / rendering / permissions. It's a `PrivateRoute` (needs a valid JWT) but the response is identical for every caller, so process-wide keying by collection is correct. Kept as-is.

packages/agent-bff/src/read-model/read-model-store.ts
nbouliol: Fixed in 33c7bc9 — `SchemaCache` now exposes a monotonic `revision` (bumped only on a successful refresh), and the store rebuilds on a `revision` change instead of array-reference identity. Pinned by tests: revision increments on refresh but not on a cache hit or a warm failure, and the store doesn't rebuild/clear capabilities on a failed refresh.

packages/agent-bff/src/read-model/read-model.ts
nbouliol: Fixed in 33c7bc9 — action `fields`/`hooks` are now shallow-copied when building the endpoint map, so a mutating consumer can't corrupt the shared cached schema (consistent with the polymorphic-targets copy already done). Asserted by a test.

packages/agent-bff/test/read-model/create-read-model.test.ts
nbouliol: Fixed in 33c7bc9 — the test now spies the console logger and asserts an `increment` is routed through it (deleting the `?? createConsoleMetrics` fallback now fails the test), instead of only asserting `toBeDefined`.

packages/agent-bff/src/read-model/read-model.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  getRelationTarget(collection: string, relation: string): RelationTarget | undefined {
+    return this.relations.get(collection)?.get(relation);
+  }
+
+  isActionAllowed(collection: string, action: string): boolean {
+    return this.actionEndpoints[collection]?.[action] !== undefined;
+  }
+
+  getActionEndpoints(): ActionEndpointsByCollection {
```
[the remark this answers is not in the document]
nbouliol: Fixed. The built structures are now deep-frozen once at construction (after the field/hook clone that detaches them from the source schema), so `getActionEndpoints()` can safely return the live reference and any consumer mutation throws.

Chose freeze-at-build over deep-copy-on-read: this getter is called on every agent-client build, so a per-request copy would be wasteful, and the only consumer (agent-client) reads `actionEndpoints` without mutating it. Relation targets are frozen too. Landing in the next push.

packages/agent-bff/src/read-model/read-model-store.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  async getCapabilities(
+    collection: string,
+    fetcher: CapabilitiesFetcher,
+  ): Promise<CapabilitiesResult> {
+    // Ensure any pending schema refresh (and its capabilities invalidation) runs first.
+    await this.getReadModel();
+
+    return this.capabilitiesCache.get(collection, fetcher);
```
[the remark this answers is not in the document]
nbouliol: Valid TOCTOU. Deferring the actual fix to the data-endpoint wiring rather than adding it here, for three reasons:

1. This method is not called from request handling yet (that's the data-endpoints slice).
2. Cache poisoning is already prevented: a generation guard means a fetch that started before a `clear()` never writes to the cache, so no stale entry is *served from cache* after a refresh (the ticket's coupling AC).
3. The residual window — a single caller receiving previous-generation capabilities while a refresh happens mid-flight — is near-nil impact: it needs a concurrent refresh exactly at the 24h boundary during an in-flight capabilities fetch AND the operators to actually differ between the two schema generations.

I've added a code note at the call site to re-check `schemaCache.revision` after the fetch and retry on a mismatch when the endpoints are wired, so the fix lands where the method is actually consumed.

packages/agent-bff/src/read-model/read-model.ts
```diff
@@ -0,0 +1,114 @@
...
+    this.actionEndpoints[collection.name] = {};
+
+    for (const action of withEndpoint) {
+      // Copy fields/hooks so a mutating consumer can't corrupt the shared cached schema.
+      this.actionEndpoints[collection.name][action.name] = {
+        id: action.id,
+        name: action.name,
+        endpoint: action.endpoint,
+        hooks: { ...action.hooks, change: [...action.hooks.change] },
+        fields: action.fields.map(field => ({ ...field })),
```
PMerlet: `ForestSchemaAction` declares `hooks`/`fields` as required, but that's only the declared type: `ForestHttpApi.getSchema()` deserializes JSON:API attributes typed `Record<string, unknown>` and casts the result, so nothing guarantees the runtime shape (the Forest server serves whatever apimap the customer's agent posted, across agent versions/lianas).

If a single action comes back without `hooks` or `fields`, the failure mode is nasty: the schema fetch **succeeds** → `SchemaCache` stores the entry → `new ReadModel()` throws a `TypeError` in `ReadModelStore.getReadModel()`. Since the cache is warm, no refresh is re-attempted: every read-model consumer throws for up to 24h (until restart or schema change), `schema_cache_refresh_error` never fires (it only covers the fetch), and one malformed action bricks the allow-list of *all* collections.

Suggestion — make the copy defensive:

```suggestion
        hooks: { load: action.hooks?.load ?? false, change: [...(action.hooks?.change ?? [])] },
        fields: (action.fields ?? []).map(field => ({ ...field })),
```

Worth a companion test ("action with missing hooks/fields") next to the existing "collection with no actions/fields key" cases.

nbouliol: Good catch — fixed in 6486f7e.

Same failure class as the collection-level `fields`/`actions` guards, and you're right that `ForestHttpApi.getSchema()` casts the JSON:API attributes so the declared required-ness of `hooks`/`fields` isn't guaranteed at runtime. Applied your suggestion:

```ts
hooks: { load: action.hooks?.load ?? false, change: [...(action.hooks?.change ?? [])] },
fields: (action.fields ?? []).map(field => ({ ...field })),
```

and added a companion test ("should default hooks and fields when a malformed action omits them") next to the existing malformed-collection cases.

packages/agent-bff/src/read-model/read-model.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+export default class ReadModel {
+  private readonly collections: Set<string>;
+  private readonly relations: Map<string, Map<string, RelationTarget>>;
+  private readonly actionEndpoints: ActionEndpointsByCollection;
+
+  constructor(collections: ForestSchemaCollection[]) {
+    this.collections = new Set();
+    this.relations = new Map();
+    this.actionEndpoints = {};
```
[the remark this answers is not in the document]
nbouliol: Fixed in 2acfc83. Good catch — the allow-list is a security boundary, so a prototype member falsely passing it is a real bug. The action-endpoint map (outer + per-collection objects) is now built with `Object.create(null)`, so `isActionAllowed('users', 'constructor' | 'toString' | '__proto__' | …)` returns `false`. `collections` (Set) and `relations` (Map) were already immune. Added a test asserting prototype-member action names aren't allowed.

packages/agent-bff/src/read-model/action-endpoint-resolver.ts
```diff
@@ -0,0 +1,55 @@
...
+  async resolve(
+    collection: string,
+    action: string,
+    { rendering }: ResolveActionContext,
+  ): Promise<ActionEndpointInfo | undefined> {
+    const tags = { rendering, collection, action };
+
+    let readModel: ReadModel;
+
+    try {
```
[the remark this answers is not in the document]
nbouliol: Fixed in 2acfc83. Right — `Metrics` is a user-supplied port, so a throwing `increment()` would have broken the documented "never throws" contract on both the miss and error paths. The resolver now routes its emissions through a `safeIncrement` that swallows metrics-backend failures. Added two tests (miss path and error path) with a metrics backend that throws, asserting `resolve()` still resolves to `undefined`.

packages/agent-bff/src/read-model/agent-capabilities-fetcher.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface AgentCapabilitiesFetcherOptions {
+  agentUrl: string;
+  token: string;
+}
+
+/**
+ * Builds a capabilities fetcher bound to a request's agent token. The cache calls it only on a
+ * miss, so the token of whichever request first populates a collection is the one used.
+ */
+export default function createAgentCapabilitiesFetcher({
```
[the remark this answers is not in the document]
nbouliol: Kept as-is — the premise doesn't hold: `/forest/_internal/capabilities` is not permission- or token-scoped. `fetchCapabilities` (`packages/agent/src/routes/capabilities.ts`) derives each field's operators from `filterOperators` on the datasource schema; it doesn't read the caller, rendering, or permissions. It's a `PrivateRoute` (needs a valid JWT) but the response is identical for every valid token, so there's no "capabilities computed under the wrong permissions".

CRUD/action permissions are a separate concern (PermissionService → `/liana/v1/permissions`, a later slice), not capabilities. So process-wide keying by collection is correct and the cross-token reuse is intended. (Same as the earlier thread on this file.)

packages/agent-bff/src/read-model/schema-cache.ts
```diff
@@ -0,0 +1,111 @@
...
+  constructor({ fetcher, metrics, now = Date.now, ttlMs = ONE_DAY_MS }: SchemaCacheOptions) {
+    this.fetcher = fetcher;
+    this.metrics = metrics;
+    this.now = now;
+    this.ttlMs = ttlMs;
+  }
+
+  async get(): Promise<ForestSchemaCollection[]> {
+    if (this.entry && this.now() - this.entry.fetchedAt < this.ttlMs) {
+      this.emitAge();
```
PMerlet, line 49: 2acfc837 hardened the resolver on the grounds that *"Metrics is a user-supplied port; a throwing backend must not break requests"* — but that invariant isn't applied here, and this call site is the worse exposure:

- `emitAge()` sits on the **cache-hit** path: a throwing `metrics.gauge` makes every `get()` throw even with a warm cache, so the whole read-model (store, resolver via its error path, capabilities) degrades continuously — exactly the failure class `safeIncrement` was added to prevent.
- Same on the refresh success path (`doRefresh`): the entry is stored, then `emitAge()` can reject a refresh that actually succeeded.

Rather than spreading try/catch per call site, consider wrapping the port once in `create-read-model.ts` (a `safeMetrics(resolvedMetrics)` decorator that guards `increment`/`gauge`) and dropping the resolver-local `safeIncrement` — one place, whole bundle covered, consistent with the stated contract.

nbouliol: Done in 82f1155c — you're right, and `emitAge()` on the cache-hit / refresh-success path was the worse exposure.

Wrapped the port once with a `safeMetrics(metrics)` decorator (guards both `increment` and `gauge`), applied in `create-read-model.ts` right after resolving the default console metrics, so the whole bundle (`SchemaCache` + resolver) is covered. Dropped the resolver-local `safeIncrement`.

Tests: `safe-metrics` unit tests (delegates args; swallows a throwing `increment`/`gauge`) plus a `createReadModel` integration test that a throwing backend doesn't break `getReadModel()` (cold fetch + cache-hit gauge) or `resolve()`.

PMerlet (approved):
LGTM

PR #1736, opened 2026-07-03
feat(agent-bff): reject nested relation field paths in top-level list and count

nbouliol:
## What

Adds a pure, syntactic guard in `agent-bff` that rejects any **top-level** list/count field path carrying the relation separator `:` (e.g. `company:name`) with `422 relation_field_not_supported`.

## Why

The agent has an authority gap: a top-level `list` checks browse/scope on the top collection only (`packages/agent/src/routes/access/list.ts`), whereas `listRelated` browse-checks the foreign collection (`list-related.ts`). A relation-target field can thus be projected without the target collection browse/scope check. The BFF must not propagate that gap, so v1 rejects nested relation paths on top-level list/count outright.

## Scope

- `relationFieldNotSupported(fields)` factory on the existing `BffHttpError` contract — envelope `{ error: { type, status, message, details: { fields } } }`, `details.fields` always an array.
- `assertNoRelationFieldPaths(paths)` — throwing guard; offending paths deduped, first-seen order; direct-only and empty input pass through. Syntactic only (no schema/allow-list consulted).
- Guard is internal (imported by path); not exported from the public `index.ts`.

## Non-goals

- Wiring into live list/count handlers (data-endpoints slice / PRD-639).
- `listRelated`/`countRelated`, allow-list validation, capabilities validation.

## Tests

10 focused unit tests: nested path in projection/filter/sort (list) and filter (count); multiple offending; single-as-array; dedupe + order; direct-field & empty pass-through; one `toErrorBody` wire-envelope assertion. Full `agent-bff` suite green (374/374), lint clean.

Fixes PRD-669

> [!NOTE]
> ### Reject nested relation field paths in top-level list and count requests
> - Adds [`assertNoRelationFieldPaths`](https://github.com/ForestAdmin/agent-nodejs/pull/1736/files#diff-d5dff585cbbbfecf5fc60610db412c3453cf4ea08e669cd400c6501426622570) that scans field paths for `:` (relation separator) and throws a 422 error listing the offending paths if any are found.
> - Adds a [`relationFieldNotSupported`](https://github.com/ForestAdmin/agent-nodejs/pull/1736/files#diff-74600553bc686d78db3d2c7a16074e0916e788e92797d5ee9a9e3de9130a8b72) error factory producing a `BffHttpError` with status 422, type `relation_field_not_supported`, and the offending fields in `details.fields`.
> - Behavioral Change: top-level list and count requests that include nested relation field paths now return 422 instead of being processed.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 843b80c.</sup>
> 
>

### DIFF
--- a/packages/agent-bff/src/http/bff-http-error.ts
+++ b/packages/agent-bff/src/http/bff-http-error.ts
@@ -79,3 +79,12 @@ export function missingTimezone(
 export function invalidTimezone(value: string): BffHttpError {
   return new BffHttpError(400, 'invalid_timezone', `Invalid timezone: "${value}"`);
 }
+
+export function relationFieldNotSupported(fields: string[]): BffHttpError {
+  return new BffHttpError(
+    422,
+    'relation_field_not_supported',
+    'Nested relation field paths are not supported on top-level list and count',
+    { fields },
+  );
+}
--- a/packages/agent-bff/src/validation/relation-field-guard.ts
+++ b/packages/agent-bff/src/validation/relation-field-guard.ts
@@ -1,0 +1,15 @@
+import { relationFieldNotSupported } from '../http/bff-http-error';
+
+const RELATION_SEPARATOR = ':';
+
+export default function assertNoRelationFieldPaths(paths: string[]): void {
+  const offending: string[] = [];
+
+  for (const path of paths) {
+    if (path.includes(RELATION_SEPARATOR) && !offending.includes(path)) {
+      offending.push(path);
+    }
+  }
+
+  if (offending.length > 0) throw relationFieldNotSupported(offending);
+}
--- a/packages/agent-bff/test/validation/relation-field-guard.test.ts
+++ b/packages/agent-bff/test/validation/relation-field-guard.test.ts
@@ -1,0 +1,101 @@
+import { type BffHttpError, toErrorBody } from '../../src/http/bff-http-error';
+import assertNoRelationFieldPaths from '../../src/validation/relation-field-guard';
+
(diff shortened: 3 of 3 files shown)


nbouliol:
## Code review — commit 843b80c

Reviewed on two axes (repo standards + PRD-669 spec). Verdict: clean, mergeable.

### Standards
No hard violations.
- Inline `422` / `relation_field_not_supported` in `bff-http-error.ts` are magic values, but consistent with the existing factory pattern in that file (`400, 'invalid_timezone'`, etc.) — fine as-is.
- Low-priority judgement calls: O(n²) dedup via `!offending.includes(path)` in the loop (a `Set` would be linear, though input cardinality is tiny); `captureError` helper used by only 3 of the tests while the rest use `toThrow` (mild dual-style, but the `toEqual`-on-`details` cases legitimately need it).
- Tests are strong: name = assertion, one behavior each, edge cases covered (empty/single/multiple/dedup/passthrough), exact wire-envelope assertion.

### Spec (PRD-669)
- All 5 acceptance criteria implemented and tested. Single path-list primitive is explicitly permitted by the spec.
- `details.fields` always an array incl. single-path case ✓. Syntactic-only (no schema/allow-list) ✓.
- Dedup + first-seen ordering is not named in the spec but is a defensible reading of AC4 and is tested.
- Guard being uncalled by any route is expected — wiring is a non-goal (Slice 3 / PRD-639).

No blocking issues.

PMerlet (approved):
LGTM

PR #1737, opened 2026-07-06
feat(mcp-server): configurable basePath to avoid host OAuth route collisions

Scra3:
## Context

When the MCP server is embedded via `agent.mountAiMcpServer()`, it registers its OAuth (`/oauth/authorize`, `/oauth/token`), `.well-known/*` and `/mcp` routes **at the host root**. If the host app already does OAuth, the MCP server **overrides its routes** and breaks the app's authentication (reported by a customer). The only workaround today is running the MCP server standalone.

This PR adds an **opt-in `basePath`** option that scopes every MCP route under a prefix, so the MCP server coexists with the host's OAuth.

```ts
agent.mountAiMcpServer({ basePath: '/ai' }); // routes under /ai/*
agent.mountAiMcpServer();                     // default: root, unchanged
```

## How it works

- The prefix is applied to the active routes (`/ai/oauth/authorize`, `/ai/oauth/token`, `/ai/mcp`) and, per RFC 8414/9728, as a **suffix** to the root-anchored well-known paths (`/.well-known/oauth-authorization-server/ai`). The MCP client discovers all endpoints dynamically from the advertised metadata, so no path is hardcoded on the client side.
- The authorization-server metadata is mounted manually at the suffixed path because the SDK's `mcpAuthMetadataRouter` only serves it at the root.
- **Default (no `basePath`) is byte-identical** to the previous behavior — zero breaking change for deployed MCP servers.

⚠️ `basePath: '/mcp'` yields the MCP endpoint at `/mcp/mcp` (the prefix applies to all routes). A distinct prefix like `/ai` avoids the doubling — to be documented.

## Validation

`normalizeMountPath` rejects any input that `new URL()` would rewrite differently from the raw-string route mounts (backslash, `..`, `%`-escapes, path-to-regexp metacharacters), preventing a silent advertised-vs-mounted route desync at this user-facing trust boundary.

## Tests

- `mcp-paths.test.ts` (new): normalization, validation rejects, prefixed/nested path building, idempotence
- `server.test.ts`: prefix-mode metadata + endpoints + `WWW-Authenticate`, `getHttpCallback` route filtering (host routes pass through, prefixed routes delegated), nested `/api/mcp`, base-path preservation
- `mcp-middleware.test.ts` (new): Koa matcher gating
- `agent.test.ts`: `basePath` threading

mcp-server: 593 tests pass. Reviewed with the pr-review-toolkit agents + skeptic pass; findings addressed.

### General

- [x] Explicit title following Conventional Commits
- [x] Validate the code quality

### Security

- [x] Consider the security impact — input validation added at the `basePath` trust boundary; default behavior unchanged.

> [!NOTE]
> ### Add configurable `basePath` to `ForestMCPServer` to avoid OAuth route collisions with the host app
> - Adds an optional `mountPath` to `ForestMCPServerOptions` that scopes all MCP routes (`/oauth`, `/.well-known`, `/mcp`) under a URL prefix, avoiding collisions with host app routes.
> - Introduces `normalizeMountPath`, `buildMcpPaths`, and `makeIsMcpRoute` in [mcp-paths.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1737/files#diff-53973bdee82330bf78536c987252e943f34d14dd93b88066697528d8483aa0bf) to validate prefixes, construct prefixed route lists, and match routes at segment boundaries.
> - `ForestMCPServer.buildExpressApp` mounts OAuth and MCP endpoints under the prefix and serves RFC 8414 authorization-server metadata at the prefixed `.well-known` path instead of the host root.
> - The `McpMiddleware` and `FrameworkMounter` are updated to accept a custom route matcher so that only prefixed MCP routes are intercepted by the middleware.
> - Behavioral Change: the default (no prefix) `isMcpRoute` now uses segment-boundary semantics, so paths like `/mcp-dashboard` no longer match as MCP routes.
>
> 
> #### Changes since #1737 opened
>
> - Removed re-exports of `buildMcpPaths`, `normalizeMountPath`, and `McpRouteMatcher` type from `mcp-server` package [8f18b83]
> - Added validation in `ForestMCPServer.buildExpressApp` that throws an error when a `basePath` is configured but the effective base URL contains a non-root pathname [c293175]
> - Restricted accepted characters in mount path validation within `normalizeMountPath` function to only allow letters, digits, hyphens, and underscores per segment using regex pattern `^(\/[A-Za-z0-9_-]+)+$` [c293175]
> - Modified `makeIsMcpRoute` function to extract pathname by splitting on query string and fragment delimiters before performing route matching [c293175]
> - Renamed `mountPath` option to `basePath` throughout `mcp-server` package in `ForestMCPServerOptions` interface, `ForestMCPServer` class field, and all call sites in `agent` package [c293175]
> - Updated documentation in `ForestMCPServerOptions` and `Agent.mountAiMcpServer` JSDoc to clarify that well-known OAuth discovery endpoints remain at origin root while OAuth and MCP protocol routes are mounted under the `basePath` [c293175]
> - Modified OAuth discovery metadata serving in `ForestMCPServer.buildExpressApp` to conditionally mount endpoints based on prefix presence [77cd41e]
> - Added test verifying authorization-server metadata is not served at bare origin root when prefix is configured [77cd41e]
> - Renamed Jest describe block in `mcp-server` test suite from 'mountPath prefix' to 'basePath prefix' [5d1e2c8]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized d872b93.</sup>
> 
>
>

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -4,6 +4,7 @@ import type {
   AgentOptions,
   AgentOptionsWithDefaults,
   HttpCallback,
+  McpRouteMatcher,
   WorkflowExecutorEmbedOptions,
 } from './types';
 import type { AiProviderDefinition } from '@forestadmin/agent-toolkit';
@@ -55,6 +56,7 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
   /** Whether MCP server should be mounted */
   private mcpEnabled = false;
   private mcpEnabledTools?: ToolName[];
+  private mcpBasePath?: string;
 
   /** In-process workflow executor, created only when addWorkflowExecutor() is called. */
   private embeddedExecutor: EmbeddedWorkflowExecutor | null = null;
@@ -94,12 +96,12 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
    */
   async start(): Promise<void> {
     try {
-      const { router, mcpHttpCallback } = await this.buildRouterAndSendSchema();
+      const { router, mcpHttpCallback, mcpIsMcpRoute } = await this.buildRouterAndSendSchema();
 
       await this.options.forestAdminClient.subscribeToServerEvents();
       this.options.forestAdminClient.onRefreshCustomizations(this.restart.bind(this));
 
-      this.setMcpCallback(mcpHttpCallback ?? null);
+      this.setMcpCallback(mcpHttpCallback ?? null, mcpIsMcpRoute);
       await this.mount(router);
 
       // Boot after mount(): the embedded executor reaches the agent over HTTP, and the
@@ -140,9 +142,9 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
 
     try {
       // We force sending schema when restarting
-      const { router, mcpHttpCallback } = await this.buildRouterAndSendSchema();
+      const { router, mcpHttpCallback, mcpIsMcpRoute } = await this.buildRouterAndSendSchema();
 
-      this.setMcpCallback(mcpHttpCallback ?? null);
+      this.setMcpCallback(mcpHttpCallback ?? null, mcpIsMcpRoute);
       await this.remount(router);
     } finally {
       this.isRestarting = false;
@@ -247,10 +249,13 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
    * agent.mountAiMcpServer();
    * // Example: read-only mode (only browse data, no create/update/delete/actions)
    * agent.mountAiMcpServer({ enabledTools: ['describeCollection', 'list', 'listRelated'] });
+   * // Example: scope MCP routes under a prefix to avoid colliding with your app's own OAuth routes
+   * agent.mountAiMcpServer({ basePath: '/ai' });
    */
-  mountAiMcpServer(options?: { enabledTools?: ToolName[] }): this {
+  mountAiMcpServer(options?: { enabledTools?: ToolName[]; basePath?: string }): this {
     this.mcpEnabled = true;
     this.mcpEnabledTools = options?.enabledTools;
+    this.mcpBasePath = options?.basePath;
 
     return this;
   }
@@ -350,9 +355,11 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
    * Create an http handler which can respond to all queries which are expected from an agent.
    * Returns the main router and optional MCP HTTP callback.
    */
-  private async getRouter(
-    dataSource: DataSource,
-  ): Promise<{ router: Router; mcpHttpCallback?: HttpCallback }> {
+  private async getRouter(dataSource: DataSource): Promise<{
+    router: Router;
+    mcpHttpCallback?: HttpCallback;
+    mcpIsMcpRoute?: McpRouteMatcher;
+  }> {
     // Bootstrap app
     const services = makeServices(this.options);
     const routes = this.getRoutes(dataSource, services);
@@ -361,9 +368,11 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
 
     // Initialize MCP server if enabled via mountAiMcpServer()
     let mcpHttpCallback: HttpCallback | undefined;
+    let mcpIsMcpRoute: McpRouteMatcher | undefined;
 
     if (this.mcpEnabled) {
-      mcpHttpCallback = await this.initializeMcpServer();
+      ({ httpCallback: mcpHttpCallback, isMcpRoute: mcpIsMcpRoute } =
+        await this.initializeMcpServer());
     }
 
     // Build main router
@@ -379,7 +388,7 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
     );
     routes.forEach(route => route.setupRoutes(router));
 
-    return { router, mcpHttpCallback };
+    return { router, mcpHttpCallback, mcpIsMcpRoute };
   }
 
   /**
@@ -387,12 +396,17 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
    * Uses dynamic import to defer loading until mountAiMcpServer() is actually used.
    * This avoids loading the mcp-server dependency at startup for users who don't use MCP.
    */
-  private async initializeMcpServer(): Promise<HttpCallback> {
+  private async initializeMcpServer(): Promise<{
+    httpCallback: HttpCallback;
+    isMcpRoute: McpRouteMatcher;
+  }> {
     const mcpLogger = (level, message) => this.options.logger(level, `[MCP] ${message}`);
 
     try {
       // Dynamic import to defer loading until actually needed
-      const { ForestMCPServer, ForestServerClientImpl } = await import('@forestadmin/mcp-server');
+      const { ForestMCPServer, ForestServerClientImpl, makeIsMcpRoute } = await import(
+        '@forestadmin/mcp-server'
+      );
 
       const forestServerClient = new ForestServerClientImpl(
         this.options.forestAdminClient.schemaService,
@@ -408,13 +422,15 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
         logger: mcpLogger,
         forestServerClient,
         enabledTools: this.mcpEnabledTools,
+        mountPath: this.mcpBasePath,
       });
 
       const httpCallback = await mcpServer.getHttpCallback();
+      const isMcpRoute = makeIsMcpRoute(this.mcpBasePath);
 
       mcpLogger('Info', 'Server initialized successfully');
 
-      return httpCallback;
+      return { httpCallback, isMcpRoute };
     } catch (error) {
       const { message } = error as Error;
       mcpLogger('Error', `Failed to initialize MCP server: ${message}`);
@@ -443,6 +459,7 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
   private async buildRouterAndSendSchema(): Promise<{
     router: Router;
     mcpHttpCallback?: HttpCallback;
+    mcpIsMcpRoute?: McpRouteMatcher;
   }> {
     const { isProduction, logger, typingsPath, typingsMaxDepth } = this.options;
 
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -249,7 +249,9 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
    * agent.mountAiMcpServer();
    * // Example: read-only mode (only browse data, no create/update/delete/actions)
    * agent.mountAiMcpServer({ enabledTools: ['describeCollection', 'list', 'listRelated'] });
-   * // Example: scope MCP routes under a prefix to avoid colliding with your app's own OAuth routes
+   * // Example: scope MCP routes under a prefix to avoid colliding with your app's own OAuth routes.
+   * // OAuth discovery metadata stays at the origin root (prefix-suffixed), so root `.well-known`
+   * // traffic must still reach the agent.
    * agent.mountAiMcpServer({ basePath: '/ai' });
    */
   mountAiMcpServer(options?: { enabledTools?: ToolName[]; basePath?: string }): this {
@@ -422,7 +424,7 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
         logger: mcpLogger,
         forestServerClient,
         enabledTools: this.mcpEnabledTools,
-        mountPath: this.mcpBasePath,
+        basePath: this.mcpBasePath,
       });
 
       const httpCallback = await mcpServer.getHttpCallback();
--- a/packages/agent/src/mcp-middleware.ts
+++ b/packages/agent/src/mcp-middleware.ts
@@ -1,4 +1,4 @@
-import type { HttpCallback } from './types';
+import type { HttpCallback, McpRouteMatcher } from './types';
 import type Koa from 'koa';
 
 import expressToKoa from './utils/express-to-koa';
@@ -22,12 +22,11 @@ function isMcpRoute(url: string): boolean {
  */
 export default class McpMiddleware {
   private mcpHttpCallback: HttpCallback | null = null;
+  private routeMatcher: McpRouteMatcher = isMcpRoute;
 
-  /**
-   * Set the MCP HTTP callback. Call this before using the middleware.
-   */
-  setCallback(callback: HttpCallback | null): void {
+  setCallback(callback: HttpCallback | null, routeMatcher?: McpRouteMatcher): void {
     this.mcpHttpCallback = callback;
+    this.routeMatcher = routeMatcher ?? isMcpRoute;
   }
 
   /**
@@ -64,8 +63,7 @@ export default class McpMiddleware {
           next();
         }
       },
-      // MCP routes are at root: /.well-known/*, /oauth/*, /mcp
-      isMcpRoute,
+      url => this.routeMatcher(url),
     );
   }
 }
--- a/packages/agent/test/agent.test.ts
+++ b/packages/agent/test/agent.test.ts
@@ -581,6 +581,16 @@ describe('Agent', () => {
       );
     });
 
+    test('should pass basePath to ForestMCPServer as mountPath', async () => {
+      const options = factories.forestAdminHttpDriverOptions.build();
+      const agent = new Agent(options);
+
+      agent.mountAiMcpServer({ basePath: '/mcp' });
+      await agent.start();
+
+      expect(mcpServerSpy).toHaveBeenCalledWith(expect.objectContaining({ mountPath: '/mcp' }));
+    });
+
     test('should log error when MCP initialization fails', async () => {
       const mockLogger = jest.fn();
       const options = factories.forestAdminHttpDriverOptions.build({ logger: mockLogger });
--- a/packages/agent/test/agent.test.ts
+++ b/packages/agent/test/agent.test.ts
@@ -581,14 +581,31 @@ describe('Agent', () => {
       );
     });
 
-    test('should pass basePath to ForestMCPServer as mountPath', async () => {
+    test('should pass basePath to ForestMCPServer', async () => {
       const options = factories.forestAdminHttpDriverOptions.build();
       const agent = new Agent(options);
 
       agent.mountAiMcpServer({ basePath: '/mcp' });
       await agent.start();
 
-      expect(mcpServerSpy).toHaveBeenCalledWith(expect.objectContaining({ mountPath: '/mcp' }));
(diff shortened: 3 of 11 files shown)


packages/mcp-server/src/mcp-paths.ts
```diff
@@ -1,7 +1,58 @@
...
+  return [...wellKnown, `${normalized}/oauth/`, `${normalized}/mcp`];
 }
+
+export function makeIsMcpRoute(prefix = ''): McpRouteMatcher {
+  const paths = buildMcpPaths(prefix);
+
+  // Match on a segment boundary so a claimed path like '/ai/mcp' does not shadow an
+  // unrelated host route like '/ai/mcp-dashboard'.
+  return (url: string) =>
+    paths.some(p => url === p || url.startsWith(p.endsWith('/') ? p : `${p}/`));
```
hercemer42: **Claude Fable 5** — [Should fix] The matcher receives the raw `req.url`/`ctx.url` *including the query string*, so exact-match paths no longer match when one is present: `isMcpRoute('/mcp?foo=1')` → `false` (was `true` on main via `startsWith('/mcp')`); same for `/mcp/mcp?x` and the suffixed well-known paths under a prefix. That's a second default-mode behavior change beyond the documented `/mcp-dashboard` one — a query-string request to the MCP endpoint now silently falls through to the host. Match on the pathname (split at `?`/`#`, or accept `${p}?` as a boundary too) and add a `'/mcp?foo=1'` case to the matcher tests.

packages/mcp-server/src/mcp-paths.ts
```diff
@@ -1,7 +1,58 @@
...
+
+  // The prefix is interpolated raw into Express/Koa route patterns AND resolved through
+  // new URL() to advertise the OAuth metadata. Reject anything that makes those diverge:
+  // path-to-regexp metacharacters, or segments new URL() rewrites (backslash, '.', '..', %-escapes).
+  const urlPath = new URL(collapsed.slice(1), 'http://forestadmin.invalid/').pathname.replace(
+    /\/+$/,
+    '',
+  );
+
+  if (/[?#\s\\:*()+%[\]]/.test(collapsed) || urlPath !== collapsed) {
```
hercemer42: **Claude Fable 5** — [Should fix] The denylist misses `!`, which path-to-regexp v8 (Express 5) treats as reserved: `normalizeMountPath('/a!b')` passes, then `app.use('/a!b/oauth/token', …)` throws `Unexpected ! at index 2 …` at startup — the cryptic failure this validator exists to prevent (verified against express 5.2.1). A per-segment allowlist would be more robust across path-to-regexp versions than a denylist, and also closes `|`/`^` (WHATWG-legal, so they survive the `new URL()` round-trip, but RFC 3986-illegal — a conforming client may percent-encode them and miss the byte-exact matcher). Minimum fix: add `!` to the character class.

packages/mcp-server/src/server.ts
```diff
@@ -477,22 +486,31 @@ export default class ForestMCPServer {
       );
     }
 
+    const { mountPath: prefix } = this;
+    // Resolve the prefix relatively against a trailing-slash base so a path on the base URL
+    // (e.g. https://host/agent) is preserved rather than dropped by URL resolution.
+    const baseWithSlash = effectiveBaseUrl.href.endsWith('/')
+      ? effectiveBaseUrl
+      : new URL(`${effectiveBaseUrl.href}/`);
+    const mountBase = prefix ? new URL(`${prefix.slice(1)}/`, baseWithSlash) : baseWithSlash;
```
hercemer42: **Claude Fable 5** — [Should fix] When `effectiveBaseUrl` carries a path (e.g. `https://host/agent`) and a prefix is set, only the *advertised* URLs include the base path: the issuer becomes `https://host/agent/mcp`, so an RFC 8414 client fetches `/.well-known/oauth-authorization-server/agent/mcp` — but that route is mounted (L564) and claimed by `makeIsMcpRoute` at `/.well-known/oauth-authorization-server/mcp` only. Likewise the advertised token endpoint path is `/agent/mcp/oauth/token` while Express mounts `/mcp/oauth/token`. This works only if a proxy strips the base path — which can't happen for the root-anchored well-known path. The test "preserves a base path on the origin" (server.test.ts:2630) pins the metadata half of this desync as expected behavior. Either derive the well-known mount + matcher from base path + prefix, or reject/warn loudly on a pathful base URL when `mountPath` is set.

packages/mcp-server/src/server.ts
```diff
@@ -117,6 +118,12 @@ export interface ForestMCPServerOptions {
   forestServerClient?: ForestServerClient;
   /** List of tool names to enable (allowlist). Only these tools will be exposed. New tools in future releases will NOT be auto-enabled. */
   enabledTools?: ToolName[];
+  /**
+   * Opt-in path prefix under which all MCP routes (OAuth, .well-known, /mcp) are mounted,
+   * e.g. '/mcp'. Defaults to the origin root. Use it to avoid colliding with a host app's
+   * own OAuth routes when the server is embedded in an agent.
+   */
```
hercemer42, line 127: **Claude Fable 5** — [Should fix] This says all MCP routes including `.well-known` are mounted *under* the prefix, but `buildMcpPaths` and L564 anchor the well-known routes at the origin root with the prefix as a suffix (`/.well-known/oauth-authorization-server/mcp`), per RFC 8414/9728. A user who reads this and proxies only `/mcp/*` to the agent breaks OAuth discovery. The same caveat is missing from the `basePath` example on `mountAiMcpServer` (agent.ts:249-256). One line on each: discovery metadata stays at the origin root, so root `.well-known` traffic must still reach the agent.

packages/mcp-server/src/server.ts
```diff
@@ -117,6 +118,12 @@ export interface ForestMCPServerOptions {
...
+  mountPath?: string;
```
hercemer42: **Claude Fable 5** — [Should fix] One concept, two public names: the agent option is `basePath` (agent.ts:255, and the PR title/body) while the exported `ForestMCPServerOptions` calls it `mountPath`. Both ship as public API in this PR and renaming later is a breaking change on mcp-server — align on one name now (or cross-reference each in the JSDoc if the split is deliberate).

packages/agent/test/agent.test.ts
```diff
@@ -581,6 +581,16 @@ describe('Agent', () => {
       );
     });
 
+    test('should pass basePath to ForestMCPServer as mountPath', async () => {
```
hercemer42: **Claude Fable 5** — [Should fix] This asserts the constructor arg only; no test pins that `basePath` reaches the Koa route gate (`makeIsMcpRoute(this.mcpBasePath)` → `setMcpCallback(cb, matcher)` → `McpMiddleware.routeMatcher`). Dropping the matcher argument at the `start()` call site keeps every test in this PR green, while Koa-mounted agents with a `basePath` would never route `/<prefix>/mcp` to the MCP server. Add an assertion that the matcher passed to `setCallback` claims `/<prefix>/mcp` and rejects root `/oauth/token`.

packages/mcp-server/test/server.test.ts
```diff
@@ -2528,6 +2528,169 @@ describe('getHttpCallback', () => {
...
+      expect(response.status).toBe(200);
+      expect(response.body.issuer).toBe('http://localhost:3000/mcp');
+      expect(response.body.authorization_endpoint).toBe(
+        'http://localhost:3000/mcp/oauth/authorize',
+      );
+      expect(response.body.token_endpoint).toBe('http://localhost:3000/mcp/oauth/token');
+    });
+
+    it('serves protected-resource metadata under the prefixed resource path', async () => {
+      const response = await request(app).get('/.well-known/oauth-protected-resource/mcp/mcp');
```
hercemer42, line 2589: **Claude Fable 5** — [Should fix] The prefix-mode metadata tests assert `resource` but never `authorization_servers` (the field the client actually follows to reach the issuer) nor `registration_endpoint` (which must remain the unprefixed Forest server URL, server.ts:515). If `mcpAuthMetadataRouter` were fed the wrong issuer, discovery would break with all tests green. Two one-line assertions close this.

packages/agent/src/mcp-middleware.ts
```diff
@@ -22,12 +22,11 @@ function isMcpRoute(url: string): boolean {
...
   private mcpHttpCallback: HttpCallback | null = null;
+  private routeMatcher: McpRouteMatcher = isMcpRoute;
 
-  /**
-   * Set the MCP HTTP callback. Call this before using the middleware.
-   */
-  setCallback(callback: HttpCallback | null): void {
+  setCallback(callback: HttpCallback | null, routeMatcher?: McpRouteMatcher): void {
     this.mcpHttpCallback = callback;
+    this.routeMatcher = routeMatcher ?? isMcpRoute;
```
hercemer42, line 29: **Claude Fable 5** — [Preferential] The local fallback `isMcpRoute` (L9-16) keeps the old bare-`startsWith` semantics — it still claims `/mcp-dashboard`, which `makeIsMcpRoute` deliberately stopped matching — and the class JSDoc above still presents the root paths as unconditional. The fallback is nearly dead (null callback ⇒ pass-through), but nothing marks it as fallback-only, so a future reader may trust the stale semantics. Align it with the segment-boundary matcher or mark it explicitly as the null-callback fallback.

packages/agent/src/agent.ts
```diff
@@ -408,13 +422,15 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
         logger: mcpLogger,
         forestServerClient,
         enabledTools: this.mcpEnabledTools,
+        mountPath: this.mcpBasePath,
       });
 
       const httpCallback = await mcpServer.getHttpCallback();
+      const isMcpRoute = makeIsMcpRoute(this.mcpBasePath);
```
hercemer42, line 431: **Claude Fable 5** — [Preferential] The matcher is rebuilt here from the raw string while the callback's internal filter builds its own from the normalized field (server.ts:646) — they agree only because both call the same function, and `{ mcpHttpCallback?; mcpIsMcpRoute? }` makes callback-without-matcher representable (silently falling back to the root-path matcher in `McpMiddleware`). Having `getHttpCallback()` return its matcher — or passing the pair as one object to `setMcpCallback` — makes the mismatch unrepresentable, and would let agent/types.ts reuse `McpRouteMatcher` via `import type` instead of duplicating it.

packages/mcp-server/src/server.ts
```diff
@@ -617,6 +643,7 @@ export default class ForestMCPServer {
    */
   async getHttpCallback(baseUrl?: URL): Promise<HttpCallback> {
     const app = await this.buildExpressApp(baseUrl);
+    const isMcpRoute = makeIsMcpRoute(this.mountPath);
```
hercemer42: **Claude Fable 5** — [Code style] The `getHttpCallback` docblock above still says the callback handles `/.well-known/*`, `/oauth/*`, `/mcp` — with `mountPath` set it handles the prefixed/suffixed set only, and root `/oauth/*` passes through (as the new pass-through test asserts). Worth updating while touching this.

hercemer42 (changes requested):
I burn my tokens for you ! 😆

packages/mcp-server/src/mcp-paths.ts
Scra3: Fixed in c293175 — `makeIsMcpRoute` now matches on the pathname (`url.split(/[?#]/, 1)`), so `/mcp?foo=1` and the suffixed well-known paths match again. Added a `/mcp?foo=1` case to the matcher tests.

packages/mcp-server/src/mcp-paths.ts
Scra3: Fixed in c293175 — switched `normalizeMountPath` to a per-segment allowlist (`/^(\/[A-Za-z0-9_-]+)+$/`), which rejects `!`, `|`, `^` and every other path-to-regexp/URL-reserved char in one rule (verified `/a!b` now throws). This also let me drop the `new URL()` round-trip check.

packages/mcp-server/src/server.ts
Scra3: Fixed in c293175 — rather than half-supporting a path-based base URL, `buildExpressApp` now throws when `basePath` is set and the base URL has a non-root path (OAuth discovery must live at the origin root). Removed the `baseWithSlash` preservation and the test that pinned the desync; added a rejection test.

packages/mcp-server/src/server.ts
Scra3: Fixed in c293175 — JSDoc on the option now states discovery metadata stays at the origin root (prefix-suffixed), and the `mountAiMcpServer` example carries the same note. (Docs PR ForestAdmin/docs#10 updated too.)

packages/mcp-server/src/server.ts
Scra3: Fixed in c293175 — renamed the mcp-server option `mountPath` → `basePath` (and the internal field) so it matches the agent's public `mountAiMcpServer({ basePath })`. One name across both packages now.

packages/agent/test/agent.test.ts
Scra3: Fixed in c293175 — added a test that spies `setMcpCallback` and asserts the threaded matcher claims `/ai/mcp` and rejects root `/oauth/token`.

packages/mcp-server/test/server.test.ts
Scra3: Fixed in c293175 — the prefix metadata tests now assert `authorization_servers` (the issuer the client follows) and `registration_endpoint` (unprefixed Forest server URL).

packages/mcp-server/src/server.ts
Scra3: Fixed in c293175 — updated the `getHttpCallback` docblock to say it handles the OAuth/.well-known/protocol routes scoped under `basePath`, and passes through everything else.

packages/agent/src/mcp-middleware.ts
Scra3: Kept as-is (preferential). The local `isMcpRoute` is only the default value before a matcher is set / when the callback is null — a pass-through case. When MCP is enabled the agent always passes the prefix-aware matcher via `setCallback(cb, matcher)`, so the bare-`startsWith` semantics are never exercised in a live MCP mount. Importing the real matcher from `@forestadmin/mcp-server` here would defeat the dynamic-import laziness (non-MCP users must not load the package), which is why the local copy exists.

packages/agent/src/agent.ts
Scra3: Acknowledged (preferential). The two matchers agree by construction — both are `makeIsMcpRoute` fed the same `basePath` (normalized identically), so they can't diverge without editing the function itself. Returning the matcher from `getHttpCallback` would change its return contract and ripple into the standalone `run()` path and its tests; I kept the threading (which mirrors the existing `httpCallback` channel) to keep the diff contained. Happy to do the object-return refactor as a follow-up if you prefer it.

Scra3:
Thanks for the thorough review 🙏 — addressed in c293175. Summary:

**Should-fix (all done):**
- Query string no longer breaks the matcher — matches on the pathname now (`/mcp?x=1` handled). +test
- `normalizeMountPath` uses an unreserved-char allowlist → rejects `!`, `|`, `^` and all path-to-regexp metacharacters up front (no more boot crash). +tests
- Path-based base URL + `basePath` now **throws** instead of advertising discovery URLs the mounted routes don't serve. Removed the desync-pinning test, added a rejection test.
- Renamed the mcp-server option `mountPath` → `basePath` to match the agent's public API.
- JSDoc (option + `mountAiMcpServer` example + `getHttpCallback`) now states discovery metadata stays at the origin root.
- Added the matcher-threading test (spies `setMcpCallback`, asserts `/ai/mcp` claimed, root `/oauth/token` rejected) and the `authorization_servers` / `registration_endpoint` metadata assertions.

**Preferential (kept, rationale in-thread):** the agent's local fallback matcher (dynamic-import laziness) and the twin-matcher threading vs returning it from `getHttpCallback`.

mcp-server: 602 tests green, lint clean. Docs updated in ForestAdmin/docs#10.

packages/mcp-server/src/server.ts
```diff
@@ -521,17 +547,16 @@ export default class ForestMCPServer {
...
-    app.use('/oauth/token', tokenHandler({ provider: oauthProvider }));
+    app.use(`${prefix}/oauth/token`, tokenHandler({ provider: oauthProvider }));
 
     // Mount metadata router with custom metadata
-    // The resourceServerUrl should include the /mcp path to match RFC 9728 requirements.
-    // This creates the .well-known/oauth-protected-resource/mcp endpoint.
-    const mcpResourceUrl = new URL('mcp', effectiveBaseUrl);
+    // The resourceServerUrl includes the /mcp path to match RFC 9728 requirements.
+    const mcpResourceUrl = new URL('mcp', mountBase);
     app.use(
```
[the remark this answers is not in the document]
Scra3: Fixed in 77cd41e — in prefix mode `buildExpressApp` no longer calls `mcpAuthMetadataRouter` (which mounts the authorization-server metadata at the bare origin root). It now mounts both discovery documents itself at the RFC 8414/9728 suffixed paths (`/.well-known/oauth-authorization-server${prefix}` and `/.well-known/oauth-protected-resource${resourcePath}`), so nothing is registered at the root to shadow a host app's own discovery. Non-prefix mode still uses the router unchanged. Added a test asserting the bare root no longer serves the AS metadata.

packages/mcp-server/test/server.test.ts
```diff
@@ -2528,6 +2528,174 @@ describe('getHttpCallback', () => {
   });
 });
 
+describe('mountPath prefix', () => {
```
hercemer42: **Claude Fable 5** — [Code style] `describe('mountPath prefix', …)` — stale name after the `mountPath` → `basePath` rename in c2931750.

hercemer42 (approved):

packages/mcp-server/test/server.test.ts
Scra3: Fixed in 5d1e2c8 — renamed the describe to `'basePath prefix'`.

Scra3:
All review comments addressed 🙏

**@hercemer42** — thanks for the deep pass, approved ✅. Every thread answered inline; the substantive ones landed in `c293175` (query-string matching, allowlist validation incl. `!`, non-root base rejection, `mountPath`→`basePath` rename, JSDoc accuracy, matcher-threading + metadata tests) and `5d1e2c8` (the stale `describe` name).

**Macroscope** — all findings auto-resolved:
- root authorization-server metadata no longer mounted under a `basePath` (`77cd41e`)
- query strings no longer break the route matcher (`c293175`)
- path-based base URL now rejected instead of desyncing (`c293175`)

**Docs** (ForestAdmin/docs#10) updated in lockstep — the four doc comments (blast-radius wording, `GET, POST` on authorize, proxy caveat, param accuracy) are fixed there.

CI is re-running on the latest commits. Ready to merge once green — noting the docs PR should merge only after this ships in a release.

PR #1739, opened 2026-07-07
fix(mcp-server): include agent error detail and status in tool errors

hercemer42:
## What

MCP tool error results now surface the agent's real error — the JSON:API `errors[0].detail` plus the HTTP status (e.g. `The value "reason" is required (HTTP 422)`) — instead of the generic `Agent responded with HTTP {status}`.

## Why

`registerToolWithLogging`'s catch built the `{ isError: true }` text from `error.message`. For a raw `AgentHttpError` (e.g. `getActionForm`) that was the generic string, leaving the real cause buried in `error.body`.

## Changes

- `error-parser.ts` — `parseAgentError` appends `(HTTP <status>)` to the extracted JSON:API detail; falls back to the raw message (no doubled status) when no detail can be extracted.
- `tool-with-logging.ts` — the catch builds the message via `parseAgentError(error)`, preserving the existing non-`Error` fallback (`JSON.stringify ?? String`).
- Tests updated/extended: `error-parser`, `tool-with-logging`, `with-activity-log`.

## Scope / notes

- Both error paths funnel through `parseAgentError`, so the suffix reaches all `withActivityLog` data tools (`list`/`create`/`update`/`delete`/`associate`/`dissociate`/`listRelated`) and `getActionForm`.
- `executeAction` is unchanged: `agent-client` converts its `AgentHttpError` to semantic `ActionFormValidationError`/`ActionRequiresApprovalError` upstream, dropping the status — out of scope.
- Response envelope unchanged (already MCP-spec-compliant).

Tests: `test/utils` (110) and `test/tools` (244) green; eslint / prettier / tsc clean.

fixes PRD-729

> [!NOTE]
> ### Include HTTP status and JSON:API detail in MCP server tool errors
> - Updates `parseAgentError` in [error-parser.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1739/files#diff-bdf07c04d227d6a9b8d3effff0b56f64207f15a160d4ff0cb39d8de3a3c7eb42) to append the HTTP status code to JSON:API error details, returning `'<detail> (HTTP <status>)'` instead of just `'<detail>'`.
> - Updates `registerToolWithLogging` in [tool-with-logging.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1739/files#diff-1989bb6374348e66de98a0d536b15201ebde9a1430c7f5ded16df0addd68fba6) to use `parseAgentError` in its catch block, surfacing structured error details for `AgentHttpError` instead of a raw stringified error.
> - Behavioral Change: tool error messages now include HTTP status codes when a JSON:API detail is present; consumers parsing error strings may need to account for the new format.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9caee51.</sup>
> 
>

### DIFF
--- a/packages/mcp-server/src/utils/tool-with-logging.ts
+++ b/packages/mcp-server/src/utils/tool-with-logging.ts
@@ -10,6 +10,8 @@ import type {
 
 import { z } from 'zod';
 
+import parseAgentError from './error-parser';
+
 // -----------------------------------------------------------------------------
 // Types
 // -----------------------------------------------------------------------------
@@ -114,7 +116,7 @@ export default function registerToolWithLogging<
         let message: string;
 
         try {
-          message = error instanceof Error ? error.message : JSON.stringify(error) ?? String(error);
+          message = parseAgentError(error) ?? JSON.stringify(error) ?? String(error);
         } catch {
           message = String(error);
         }
--- a/packages/mcp-server/src/utils/error-parser.ts
+++ b/packages/mcp-server/src/utils/error-parser.ts
@@ -17,10 +17,13 @@ function jsonApiDetail(body: unknown, text?: string): string | null {
   return null;
 }
 
-// Turn an agent RPC error into a human-readable message (AgentHttpError detail, else raw message).
+// Turn an agent RPC error into a human-readable message: the JSON:API detail with its HTTP status
+// when one can be extracted, else the raw message (which already carries the status).
 export default function parseAgentError(error: unknown): string | null {
   if (error instanceof AgentHttpError) {
-    return jsonApiDetail(error.body, error.responseText) ?? error.message;
+    const detail = jsonApiDetail(error.body, error.responseText);
+
+    return detail ? `${detail} (HTTP ${error.status})` : error.message;
   }
 
   if (error && typeof error === 'object' && 'message' in error) {
--- a/packages/mcp-server/test/utils/error-parser.test.ts
+++ b/packages/mcp-server/test/utils/error-parser.test.ts
@@ -3,22 +3,22 @@ import { AgentHttpError } from '@forestadmin/agent-client';
 import parseAgentError from '../../src/utils/error-parser';
 
 describe('parseAgentError', () => {
-  it('extracts the JSON:API detail from the parsed body', () => {
+  it('appends the HTTP status to the JSON:API detail from the parsed body', () => {
     const error = new AgentHttpError(400, {
       errors: [{ name: 'ValidationError', detail: 'Invalid filters provided' }],
     });
 
-    expect(parseAgentError(error)).toBe('Invalid filters provided');
+    expect(parseAgentError(error)).toBe('Invalid filters provided (HTTP 400)');
   });
 
-  it('falls back to parsing responseText when the body is not a parsed object', () => {
+  it('appends the HTTP status to the JSON:API detail parsed from responseText', () => {
     const error = new AgentHttpError(
       400,
       undefined,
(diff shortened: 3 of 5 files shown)


packages/mcp-server/src/utils/tool-with-logging.ts
```diff
@@ -114,7 +116,7 @@ export default function registerToolWithLogging<
         let message: string;
 
         try {
-          message = error instanceof Error ? error.message : JSON.stringify(error) ?? String(error);
+          message = parseAgentError(error) ?? JSON.stringify(error) ?? String(error);
```
hercemer42, line 119: **Claude Opus 4.8 (1M) — Preferential**

Routing the catch through `parseAgentError` makes a pre-existing type leak newly reachable here. `parseAgentError` returns `(error as { message: string }).message` unchecked, so a thrown non-`Error` object with a non-string `message` (e.g. `{ message: 42 }`) puts a non-string into `content[0].text` (typed `string`). The old code JSON-stringified non-`Error` throws, so this only surfaces now. Very rare in practice (handlers throw `Error`/`AgentHttpError`) — flagging, not blocking.

hercemer42, line 119: **Claude Opus 4.8 (1M) — Preferential**

Minor behavior change for an empty-message `Error`: `new Error('')` → `parseAgentError` returns `null` (`'' || null`) → falls through to `JSON.stringify(error)` → `'{}'`. The previous `error instanceof Error ? error.message` branch yielded `''`. Both are unhelpful, but `'{}'` misleadingly implies an empty object was thrown. Not worth fixing — noting for awareness.

hercemer42:
**Claude Opus 4.8 (1M) — code review of #1739**

No must-fix. The core fix is correct: the `(HTTP <status>)` suffix is appended only when a JSON:API detail is extracted (no double status on non-JSON:API bodies), the two-layer `withActivityLog` → `tool-with-logging` path does not double-append (the re-thrown `Error` hits the object-with-`message` branch, not the `AgentHttpError` branch), and `undefined`/`null`/circular throws are handled. Both changed source lines are covered by exact `toEqual`/`toBe` assertions.

Two preferential edge-case notes inline (both newly reachable via the `parseAgentError` switch). No convention violations. The `executeAction` path surfacing the detail without an HTTP suffix is the documented, intended asymmetry (`agent-client` converts its `AgentHttpError` to semantic errors upstream before mcp-server sees it).

packages/mcp-server/src/utils/tool-with-logging.ts
hercemer42: **Claude Opus 4.8 (1M):** Acknowledged — not fixing in this PR. The unchecked `.message` cast lives in `parseAgentError` and pre-dates this change; the trigger (`throw { message: <non-string> }`) is unreachable from the current handlers, which throw `Error`/`AgentHttpError` or agent-client's semantic errors — never a bare object with a numeric `message`. Tightening the cast here would be opportunistic scope-creep into an unrelated util, so it's better as a standalone cleanup if we want it.

hercemer42: **Claude Opus 4.8 (1M):** Not fixing — agreed it isn't worth it. `new Error('')` isn't thrown anywhere in the tool paths, and the prior output (`''`) was equally unhelpful, so there's no meaningful regression. A dedicated empty-message branch would be added complexity for a ~0-probability case.

christophebrun-forest (approved):
Manual test with script : OK
LGTM

PR #1747, opened 2026-07-09
chore(eslint): enforce kebab-case filenames, purge convention prose

hercemer42:
## What

Encodes the one enforceable agent-nodejs convention as an ESLint rule (from the PRD-711 conventions analysis) and drops the unmeasurable clean-code prose from `CLAUDE.md`.

- `unicorn/filename-case` — **error**, kebab-case. Prevents a real bug class: macOS (case-insensitive) vs Linux CI (case-sensitive) import-path mismatches. Relaxed via ignore-globs for the generated ANTLR parser (can't rename), snake_case `datasource-demo-fintech` fixtures, and `_`-grouped test files — every existing violation is covered, so CI stays green while new violations are blocked.
- One new devDep, pinned for ESLint 8: `eslint-plugin-unicorn@^48.0.1` (later majors require ESLint ≥8.56).
- `CLAUDE.md`: purged DRY/KISS/YAGNI, single-responsibility, AAA, pure-functions, comments-why-not-what, descriptive-names — generic prose that carried no repo-specific information.

**Why cutting the prose costs the models nothing:** `CLAUDE.md` is injected into every agent session, so each line spends context budget. DRY/KISS/YAGNI/AAA-style advice is the most over-represented guidance in any model's training data — models (and experienced devs) already default to it unprompted, so restating it changes zero behavior. The repo's own measurements confirm the habits hold without the prose (median 8 lines/function, 99.9% of ~46k functions ≤4 params, kebab-case at 846/849). Meanwhile every generic line dilutes the doc's real job: carrying the Forest-specific signal a model *can't* guess — architecture, workflows, and the conventions the linter actually enforces.

## What was considered and dropped

Three advisory rules (`max-params: 4`, `max-lines-per-function: 50`, `promise/prefer-await-to-then`) were measured (46 / 131 / 16 existing src violations) and initially added as `warn` — then removed: CI has no `--max-warnings` gate, so a warning nobody fails on is dormant documentation, not enforcement. They can return as `error` + bulk suppressions after the ESLint 9 upgrade (blocked today by airbnb-config deprecation, separate ticket). `max-lines-per-function` is the least likely to return as-is: the length distribution is bimodal (median 8, but a legitimate tail of long-declarative route handlers and decorator wiring), so a blunt line count can't separate justified outliers from real smells — that's per-function judgment, which belongs in the conventions corpus, not lint.

## Notes for reviewer

- Config + docs only — no production code touched.
- Local lint on an unbuilt worktree shows 4 pre-existing `@typescript-eslint/no-throw-literal` errors in `agent/src/routes/.../action-authorization.ts`; this is a type-resolution artifact of the unbuilt tree (CI builds before linting and is clean). Unrelated to this change.

Refs PRD-759

> [!NOTE]
> ### Enforce kebab-case filenames via ESLint unicorn plugin
> - Adds `eslint-plugin-unicorn` and enables the `unicorn/filename-case` rule (error level) in [.eslintrc.js](https://github.com/ForestAdmin/agent-nodejs/pull/1747/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5), requiring kebab-case filenames across the repo.
> - Overrides disable the rule for generated parser files, demo fintech customizations, and test files.
> - Removes prescriptive coding convention prose from [CLAUDE.md](https://github.com/ForestAdmin/agent-nodejs/pull/1747/files#diff-6ebdb617a8104a7756d0cf36578ab01103dc9f07e4dc6feb751296b9c402faf7), replacing it with a note that mechanical rules are now owned by ESLint.
>
> 
> #### Changes since #1747 opened
>
> - Updated `eslint` dev dependency version constraint from `^8.31.0` to `^8.44.0` [ff0c2e0]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0380cb4.</sup>
> 
>

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,11 +17,24 @@ module.exports = {
     sourceType: 'module',
     project: ['./tsconfig.eslint.json'],
   },
-  plugins: ['@typescript-eslint', 'prettier', 'jest', 'jest-formatting'],
+  plugins: ['@typescript-eslint', 'prettier', 'jest', 'jest-formatting', 'unicorn', 'promise'],
   rules: {
     /**********/
     /** Style */
     /**********/
+
+    // Kebab-case filenames avoid case-sensitivity mismatches between macOS and Linux CI.
+    // Generated parser, snake-case demo fixtures, and _-grouped tests are relaxed below.
+    'unicorn/filename-case': ['error', { case: 'kebabCase' }],
+
+    // Prefer async/await over .then chains; a smell, not a build blocker.
+    'promise/prefer-await-to-then': 'warn',
+
+    // Options-object pattern is the norm; flag drift without forcing signature refactors.
+    'max-params': ['warn', 4],
+
+    // Long functions are a smell; scoped to src (excluded for tests below).
+    'max-lines-per-function': ['warn', 50],
 
     // Code spacing
     'prettier/prettier': 'error',
@@ -142,6 +155,24 @@ module.exports = {
             ],
           },
         ],
+      },
+    },
+    {
+      // Generated ANTLR parser: cannot rename without breaking regeneration.
+      files: ['**/generated-parser/**/*'],
+      rules: { 'unicorn/filename-case': 'off' },
+    },
+    {
+      // Demo fixtures mirror snake_case database table names.
+      files: ['packages/datasource-demo-fintech/**/*'],
+      rules: { 'unicorn/filename-case': 'off' },
+    },
+    {
+      // Tests use _-grouped filenames and long describe/it callbacks.
+      files: ['**/test/**/*'],
+      rules: {
+        'unicorn/filename-case': 'off',
+        'max-lines-per-function': 'off',
       },
     },
   ],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,8 +33,8 @@ module.exports = {
     // Options-object pattern is the norm; flag drift without forcing signature refactors.
     'max-params': ['warn', 4],
 
-    // Long functions are a smell; scoped to src (excluded for tests below).
-    'max-lines-per-function': ['warn', 50],
+    // Long functions are a smell; count code lines only, scoped to src (excluded for tests below).
+    'max-lines-per-function': ['warn', { max: 50, skipBlankLines: true, skipComments: true }],
 
     // Code spacing
     'prettier/prettier': 'error',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,7 +17,7 @@ module.exports = {
     sourceType: 'module',
     project: ['./tsconfig.eslint.json'],
   },
-  plugins: ['@typescript-eslint', 'prettier', 'jest', 'jest-formatting', 'unicorn', 'promise'],
+  plugins: ['@typescript-eslint', 'prettier', 'jest', 'jest-formatting', 'unicorn'],
   rules: {
     /**********/
     /** Style */
@@ -26,15 +26,6 @@ module.exports = {
     // Kebab-case filenames avoid case-sensitivity mismatches between macOS and Linux CI.
     // Generated parser, snake-case demo fixtures, and _-grouped tests are relaxed below.
     'unicorn/filename-case': ['error', { case: 'kebabCase' }],
-
-    // Prefer async/await over .then chains; a smell, not a build blocker.
-    'promise/prefer-await-to-then': 'warn',
-
-    // Options-object pattern is the norm; flag drift without forcing signature refactors.
-    'max-params': ['warn', 4],
-
-    // Long functions are a smell; count code lines only, scoped to src (excluded for tests below).
-    'max-lines-per-function': ['warn', { max: 50, skipBlankLines: true, skipComments: true }],
 
     // Code spacing
     'prettier/prettier': 'error',
@@ -168,12 +159,9 @@ module.exports = {
       rules: { 'unicorn/filename-case': 'off' },
     },
     {
-      // Tests use _-grouped filenames and long describe/it callbacks.
+      // Tests use _-grouped filenames.
       files: ['**/test/**/*'],
-      rules: {
-        'unicorn/filename-case': 'off',
-        'max-lines-per-function': 'off',
-      },
+      rules: { 'unicorn/filename-case': 'off' },
     },
   ],
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -155,11 +155,12 @@ module.exports = {
     },
     {
       // Demo fixtures mirror snake_case database table names.
-      files: ['packages/datasource-demo-fintech/**/*'],
+      files: ['packages/datasource-demo-fintech/src/customizations/**/*'],
       rules: { 'unicorn/filename-case': 'off' },
     },
     {
-      // Tests use _-grouped filenames.
+      // Test filenames separate subject from case with an underscore; nothing imports
+      // them, so their casing cannot break a build.
       files: ['**/test/**/*'],
       rules: { 'unicorn/filename-case': 'off' },
     },
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -73,8 +73,7 @@ yarn workspace @forestadmin/agent test
 
 ## Code Review Principles
 
-Mechanical conventions (function size, parameter count, filename case, async/await
-over `.then`) are enforced by ESLint — see `.eslintrc.js`, not this file.
+Mechanical conventions are enforced by ESLint — see `.eslintrc.js`, not this file.
 
 ### Function Design
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -67,42 +67,28 @@ yarn workspace @forestadmin/agent test
   - ✅ `expect(mock).toHaveBeenCalledWith({ host: 'localhost', port: 3000 })` - verifies exact arguments
   - ✅ `expect(mock).toHaveBeenCalledWith(expect.objectContaining({ host: 'localhost' }))` - partial match when some values are dynamic
 - **One behavior per test** - Each test should verify one specific behavior
-- **Use AAA pattern** - Arrange (setup), Act (execute), Assert (verify) - keep these sections clear
 - **Test edge cases** - Not just happy paths; test errors, null values, empty arrays, boundaries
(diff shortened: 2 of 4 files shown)


hercemer42:
**Claude Fable 5 (claude-fable-5)** — review + scope history

1. A `/validator` pending review found one Preferential issue (`max-lines-per-function` counted comment/blank lines); fixed in `29e9691d`.
2. Scope was then cut in `84fa5312`: all three advisory `warn` rules (`max-params`, `max-lines-per-function`, `promise/prefer-await-to-then`) removed — CI has no `--max-warnings` gate, so ungated warnings are dormant documentation, not enforcement. The pending review was discarded as moot (its finding anchored to a removed rule).

Final scope: `unicorn/filename-case` (**error**, kebab-case, ignore-globs for generated parser / demo fixtures / tests) + the `CLAUDE.md` prose purge. See the PR description.

CI note: the `workflow-executor` test failure (`AI payload bounding (HasMany)`) is pre-existing on `main` — the run for `b4e7fcd2` (this branch's fork point) fails the same job. The three "failed" datasource jobs are fail-fast cancellations; every lint step that ran on this PR passed.

CLAUDE.md
```diff
@@ -67,42 +67,27 @@ yarn workspace @forestadmin/agent test
...
 - **Coverage ≠ quality** - 100% coverage with weak assertions is worse than 80% with strong assertions
 
 ## Code Review Principles
 
-### Core Principles
-
-- **DRY** (Don't Repeat Yourself) - Extract common logic into reusable functions/services
-- **KISS** (Keep It Simple, Stupid) - Prefer simple, readable solutions over clever ones
-- **YAGNI** (You Aren't Gonna Need It) - Don't implement features "just in case"
+Mechanical conventions are enforced by ESLint — see `.eslintrc.js`, not this file.
```
hercemer42: Are these mechanical though ? They seem very general. And if they are, I don't see how we added those as enforcement to ESLint.

CLAUDE.md
```diff
@@ -67,42 +67,27 @@ yarn workspace @forestadmin/agent test
...
 
 ### Function Design
 
-- **Single Responsibility** - One function = one task
-- **Size** - Keep functions short (20-30 lines max); split if longer
-- **Naming** - Descriptive names that explain *what*, not *how* (`getUserPermissions` not `getData`)
-- **Parameters** - Max 3-4 parameters; use an options object for more
```
hercemer42, line 88: Do we enforce these in lint ?

CLAUDE.md
```diff
@@ -67,42 +67,27 @@ yarn workspace @forestadmin/agent test
...
 - **Return early** - Guard clauses at the top, avoid deep nesting
 
 ### Code Quality
 
 - **No magic numbers/strings** - Use named constants
-- **Avoid side effects** - Pure functions when possible
 - **Handle errors explicitly** - No silent failures
-- **Comments** - Explain *why*, not *what* (code should be self-documenting)
```
hercemer42, line 96: Likewise - on side effects and comments, enforced in lint ?

CLAUDE.md
```diff
@@ -67,42 +67,27 @@ yarn workspace @forestadmin/agent test
...
 
 ### Review Checklist
 
-1. Does this code do one thing well?
-2. Can I understand it in 30 seconds?
-3. Is there duplication that should be extracted?
-4. Are there unused variables/imports?
-5. Are edge cases handled?
-6. Is the naming clear and consistent?
```
hercemer42, line 105: Enforced in lint ?

CLAUDE.md
hercemer42: Claude Fable 5 (claude-fable-5): Agreed — as written the sentence implied DRY/KISS/YAGNI moved to ESLint. They didn't, and they're not mechanical: they were deleted per the ticket's purge list (generic clean-code prose carrying no repo-specific signal), independently of any lint rule. After the scope cut, the only convention this PR actually encodes in ESLint is kebab-case file naming. Reworded in 99443ce4 to say precisely what the linter owns (file naming, formatting, unused variables — the latter two pre-existing) and that the remaining bullets are judgment calls.

CLAUDE.md
hercemer42: Claude Fable 5 (claude-fable-5): No — none of these four are enforced in lint. Two different reasons: **Single Responsibility** and **Naming** are unmeasurable judgment prose, purged per the ticket. **Size** and **Parameters** were lint-encoded in this PR's first commit (max-lines-per-function, max-params as warn) and then dropped in 84fa5312 when we cut the ungated warns — so their "moved to lint" premise did lapse. I'd still keep them deleted: the corpus measurements show they're self-sustaining habits (median 8 lines/function, 99.9% of ~46k functions ≤4 params) — prose neither created nor maintains that. If you want them recorded somewhere, the right vehicle is the ESLint 9 revisit (error + bulk suppressions), not CLAUDE.md.

CLAUDE.md
hercemer42: Claude Fable 5 (claude-fable-5): No — neither is lintable, and neither was meant to be. "Avoid side effects / pure functions" and "comments explain why not what" are both on the ticket's explicit purge list: removed as generic prose, not migrated to ESLint. (For comment style specifically, nothing mechanical exists to enforce; it stays a review judgment.)

CLAUDE.md
hercemer42: Claude Fable 5 (claude-fable-5): Mixed — and you caught an inconsistency. The four deleted items (one-thing-well, 30-second readability, duplication, naming clarity) are judgment, purged per the ticket. But of the two items I originally kept, "unused variables/imports" IS lint-enforced (@typescript-eslint/no-unused-vars is error, pre-existing) — so keeping it in prose was exactly the redundancy this PR removes. Dropped it in 99443ce4; the checklist is now just the edge-cases prompt, which no linter can check.

Scra3 (commented):
Automated review — safe to merge; globs verified covering 100% of existing violations and the `^48` pin is sound (walled below unicorn 49/50 which need newer eslint). Two non-blocking scope suggestions plus a peer-range nit.

.eslintrc.js
```diff
@@ -144,5 +148,20 @@ module.exports = {
...
       },
     },
+    {
+      // Generated ANTLR parser: cannot rename without breaking regeneration.
+      files: ['**/generated-parser/**/*'],
+      rules: { 'unicorn/filename-case': 'off' },
+    },
+    {
+      // Demo fixtures mirror snake_case database table names.
+      files: ['packages/datasource-demo-fintech/**/*'],
```
Scra3, line 158: suggestion: this exempts the entire `datasource-demo-fintech` package, but only the 5 snake_case files under `src/customizations/**` actually violate the rule — scope to `packages/datasource-demo-fintech/src/customizations/**/*` so future files stay checked.

.eslintrc.js
```diff
@@ -144,5 +148,20 @@ module.exports = {
...
+      rules: { 'unicorn/filename-case': 'off' },
+    },
+    {
+      // Tests use _-grouped filenames.
+      files: ['**/test/**/*'],
```
Scra3, line 163: suggestion: disabling kebab-case for every `**/test/**` file also exempts all future tests that would otherwise comply — reasonable trade-off, flagging as the widest of the three globs.

package.json
```diff
@@ -20,6 +20,7 @@
     "eslint-plugin-jest": "^27.2.1",
     "eslint-plugin-jest-formatting": "^3.1.0",
     "eslint-plugin-prettier": "^4.2.1",
+    "eslint-plugin-unicorn": "^48.0.1",
```
Scra3, line 23: nit: unicorn@48 requires eslint >=8.44 while the repo declares `eslint: ^8.31.0` — fine in practice (lock resolves 8.53), but a pin to 8.31–8.43 would trip the peer.

.eslintrc.js
hercemer42: **Claude Opus 5 (claude-opus-5[1m])**: Accepted — narrowed to `packages/datasource-demo-fintech/src/customizations/**/*` in `0380cb4c`.

Verified against the real `.eslintrc.js`, not a scratch config:

- all five snake_case files in the package (`kyc_cases`, `kyc_documents`, `aml_alerts`, `sar_reports`, `refund_requests`) sit under `src/customizations`, and `eslint src test` on the package is still clean;
- the narrowing isn't cosmetic — a probe `src/TempProbe.ts` now errors (`Filename is not in kebab case. Rename it to temp-probe.ts`) where the package-wide glob let it through.

.eslintrc.js
hercemer42: **Claude Opus 5 (claude-opus-5[1m])**: Keeping the glob — but the comment above it was wrong, and that is fixed in `0380cb4c`.

Leading underscores were never the reason. unicorn strips them before checking, so `test/_helpers.ts` and the four `_build-models.ts` never violated the rule at all. The real violators are `snake_case` (20 files) and mixed separators — `generator-fields_column.test.ts`, `collection_mixed-relations.test.ts` (13 files) — where the `_` separates subject from case.

That also rules out the narrower alternative: `cases: { kebabCase: true, snakeCase: true }` still rejects those 13, so the only narrowing left is enumerating ~6 directories, which rots the first time a test file moves.

And the breadth costs nothing in bug terms. The rule is here for macOS↔Linux import-casing mismatches; jest discovers test files by glob and nothing imports them, so a miscased test filename cannot produce the failure this rule guards. The enforcement that pays is on `src`, which is fully covered.

package.json
hercemer42: **Claude Opus 5 (claude-opus-5[1m])**: Fixed in `ff0c2e04` — root `eslint` declared range `^8.31.0` → `^8.44.0`, with the matching `yarn.lock` spec-key update. Resolution is unchanged (8.53.0 satisfies both specs, and root is the only `package.json` declaring eslint); `yarn install --frozen-lockfile` passes on the result.

For the record it was unreachable rather than merely unlikely: the lock pins 8.53.0, yarn 1 only warns on a peer mismatch, and CI installs from the committed lock (`yarn && yarn bootstrap --ci`) — you'd have had to deliberately downgrade eslint into 8.31–8.43. Bumped anyway, because the declared floor should state the real one.

Tonours (approved):

PR #1752, opened 2026-07-10
feat(agent-bff): expose the action execute endpoint with result normalization

nbouliol:
## What

Exposes `POST /v1/:collection/actions/:action/execute` on the BFF, stacked on the action-form PR (#1748). It rejects unknown submitted fields, runs the Smart Action, and normalizes the agent result into one flat wrapper — success / error / webhook / redirect — returning a structured 501 for File results, which are unimplemented in v1.

Fixes PRD-674

> Base branch is `feature/prd-673-expose-the-action-form-endpoint` (stacked). Review after #1748 merges; rebase onto `main` at that point.

## How

**`agent-client`** (2 additive, backward-compatible changes — a version bump ships with this, so rollback is no longer "redeploy agent-bff alone"):
- `ActionFormValidationError` now carries `html`; `toActionError` forwards it from the native action `Error` payload (`{ error, html }`). Without this the html is dropped before reaching the BFF.
- `setFields` throws a typed `UnknownActionFieldError` (same message as before) so callers route an unknown field to a client error instead of a generic 500. Existing consumers (mcp-server, workflow-executor) catch generically and are unaffected.

**`agent-bff`**:
- `action-execute-mapper.ts` — pure mapper: the agent 200 payload is discriminated positively (webhook / redirect / success); any unrecognized body (a streamed File) falls through to a `501 unsupported_action_result`. `HttpRequester` does not expose response headers, so `Content-Disposition` cannot be read — the fall-through is the only viable File detection.
- `agent-action-client.ts` — a single `loadAction` method (form and execute load the same agent-client `Action` object); `Action extends ActionForm`.
- `action-routes-middleware.ts` — one middleware matching both `/form` and `/execute` with shared body parsing and action resolution. `execute()` gets its own try/catch (it cannot use the generic `callAgent`, which would mislabel the action-Error 400 as a transport 502): `UnknownActionFieldError` → 400 `invalid_request`, `ActionFormValidationError` → 400 `{ type: "error", ..., html }`, `ActionRequiresApprovalError` → 403 `action_requires_approval`, transport 5xx → `agent_unavailable`.
- `bff-local-errors.ts` — new `actionRequiresApproval` (403).

## Result shapes

| Agent result | HTTP | BFF body |
|---|---|---|
| Success | 200 | `{ type: "success", message, invalidated: [...], html }` |
| Error (native) | 400 | `{ type: "error", status: 400, message, html }` |
| Webhook | 200 | `{ type: "webhook", url, method, headers, body }` |
| Redirect | 200 | `{ type: "redirect", path }` |
| File | 501 | `{ error: { type: "unsupported_action_result", status: 501 } }` |

## Out of scope / follow-ups

- **Approval-gated actions** are not wired (no `createApprovalRequest`); they surface as `403 action_requires_approval` (with `roleIdsAllowedToApprove`). Full wiring depends on the auth foundation (PRD-637) providing a per-mode Forest-server token + rendering context, and would add an `approvalRequested` wrapper shape. To confirm with the spec author.
- `action_not_allowed` (403) still has no local trigger — every non-exposed action maps to 404, mirroring the collection/relation routes (existing TODO).

## Tests

- `agent-client`: html carry on the error path, typed `UnknownActionFieldError` from `setFields`.
- `agent-bff`: pure execute mapper (all shapes + 501), the `/execute` route (missing recordIds, unknown field, action Error + html, approval, File, unknown action, transport 5xx vs action-Error 400, success/webhook/redirect), and the shared `loadAction`.
- No regression in `setFields` consumers: mcp-server (24) and workflow-executor adapter (63) green.

> [!NOTE]
> ### Expose action execute endpoint in agent-bff with result normalization and capability validation
> - Adds a new `/execute` route to [`action-routes-middleware.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1752/files#diff-1507f08ca51ccc3a715be158e30f8c80ccfb52881e3853b72775f1e677981d4d) that loads the action, applies field validation, executes it, and returns a normalized response.
> - Introduces [`action-execute-mapper.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1752/files#diff-adf19282557435797abff69c50303936dd52dc28025a4ebaded3bf3159c81f77) to normalize raw agent execute results into structured success/webhook/redirect bodies, returning 501 for unrecognized payloads.
> - Renames `loadActionForm` to `loadAction` on `AgentActionClient` and extends the returned `Action` interface with `setFields` and `execute` methods.
> - Adds typed `UnknownActionFieldError` to `agent-client` so the execute handler can return a 400 when a field is missing, and extends `ActionFormValidationError` with an optional `html` payload.
> - Adds capabilities validation to list and count routes in [`data-routes-middleware.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1752/files#diff-f72d9e0085f04fbc09b3ee055760d1ab06989a7f7609ff77350b0c92ec43be20), rejecting requests with invalid fields or operators before querying the agent.
> - `ReadModelStore.getCapabilities` now returns capabilities paired with their read-model and retries on schema-generation races.
> - Behavioral Change: `AgentActionClient.loadActionForm` is replaced by `loadAction`; any callers relying on the old method name will break.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1d9392a.</sup>
> 
>

### DIFF
--- a/packages/agent-bff/package.json
+++ b/packages/agent-bff/package.json
@@ -32,6 +32,7 @@
   },
   "dependencies": {
     "@forestadmin/agent-client": "1.12.2",
+    "@forestadmin/datasource-toolkit": "1.54.0",
     "@forestadmin/forestadmin-client": "1.41.3",
     "@koa/bodyparser": "^6.1.0",
     "jsonwebtoken": "^9.0.3",
--- a/packages/agent-bff/src/action/action-execute-mapper.ts
+++ b/packages/agent-bff/src/action/action-execute-mapper.ts
@@ -33,10 +33,10 @@ export interface ActionExecuteMapped {
   body: ActionExecuteMappedBody;
 }
 
-// Normalizes the agent's 200 execute payload into the flat BFF wrapper. The native ActionResult
-// union never reaches the BFF (agent-client `execute()` is typed `{ success, html? }`), so we
-// discriminate on the agent HTTP payload shape. A File result streams a binary with no JSON marker,
-// so any unrecognized 200 body falls through to a structured 501 rather than being mislabelled.
+// Normalizes the agent's 200 execute payload into the flat BFF wrapper. The execute result is
+// untyped at the BFF boundary (`Action.execute(): Promise<unknown>`), so we discriminate on the
+// agent HTTP payload shape. A File result streams a binary with no JSON marker, so any unrecognized
+// 200 body falls through to a structured 501 rather than being mislabelled.
 export function mapActionExecuteResult(raw: unknown): ActionExecuteMapped {
   const body = (typeof raw === 'object' && raw !== null ? raw : {}) as Record<string, unknown>;
 
--- a/packages/agent-bff/src/action/action-execute-mapper.ts
+++ b/packages/agent-bff/src/action/action-execute-mapper.ts
@@ -40,10 +40,11 @@ export interface ActionExecuteMapped {
 export function mapActionExecuteResult(raw: unknown): ActionExecuteMapped {
   const body = (typeof raw === 'object' && raw !== null ? raw : {}) as Record<string, unknown>;
 
-  if ('webhook' in body) {
-    const hook = (
-      typeof body.webhook === 'object' && body.webhook !== null ? body.webhook : {}
-    ) as Record<string, unknown>;
+  // Each branch validates the value shape, not just key presence: a malformed payload
+  // (`{ webhook: null }`, `{ redirectTo: {} }`, `{ success: {} }`) must fall through to the 501
+  // path rather than be surfaced as a 200 with null fields the client cannot tell from a real one.
+  if (typeof body.webhook === 'object' && body.webhook !== null) {
+    const hook = body.webhook as Record<string, unknown>;
 
     return {
       status: 200,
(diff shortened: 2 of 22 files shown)


ShohanRahman:
## Review — key points

Multi-perspective review (correctness, tests, error-handling, types, comments, simplification). **No Critical/blocking bugs.** Verified the mapper's wire-shape discrimination (`success`/`redirectTo`/`webhook`/`refresh`) matches the agent serialization in `agent/src/routes/modification/action/action.ts:122-140`, and the `handleExecute` error routing is sound.

### Missing test coverage (worth adding)

1. **`setFields` catch → `mapAgentError` fallback is untested** — `action-routes-middleware.ts:99-104`. Tests only reject `setFields` with `UnknownActionFieldError` (→400); no test drives a non-unknown-field rejection (e.g. an `AgentHttpError` from the change-hook) through the `mapAgentError` branch. The path is correct (`setFieldValue` throws raw `AgentHttpError`, never `ActionFormValidationError`), but the branch is uncovered.
2. **`ActionRequiresApprovalError` without `roleIdsAllowedToApprove`** — `action-routes-middleware.ts:114-121`. The `details`-omitted branch (403 body with no `roleIdsAllowedToApprove`) has zero coverage; tests only pass `[7,9]`/`[]`.
3. **Non-array `refresh.relationships` fallback** — `action-execute-mapper.ts:70,77`. The `: []` guard that malformed payloads exist to exercise is never tested (`refresh: {}` or `relationships: 'x'`).
4. **`html`-absent on the error path** — `action-routes-middleware.ts:125`. `error.html ?? null` is only tested with `html` present; the null-default on the 400 error body is uncovered (the success-path default is covered).

### Advisory — worth a conscious decision before merge

- **Inconsistent error-body shape for action-validation failures** — `action-routes-middleware.ts:123-128`. This branch hand-builds a flat `{ type: 'error', status, message, html }`, bypassing the `{ error: { type, status, message, details? } }` convention used by every other BFF error (including `actionRequiresApproval` a few lines above). A client parsing `body.error.type` uniformly breaks only for this case. It's deliberate (asserted at test `:499`) — flagging for error-contract consistency.
- **No logging on the 501 fall-through** — `action-execute-mapper.ts:83`. Any unrecognized payload becomes a generic 501 with no record of the raw body; a `Warn` log would help diagnose a File result or a new agent result type reaching this path.
- **Comment at `action-execute-mapper.ts:37`** slightly misstates the reason shape-sniffing is needed: the BFF's own `execute()` is typed `Promise<unknown>` (`agent-action-client.ts:26`), which is the real driver — not agent-client's typing.

nbouliol:
Addressed in 308d8619a.

**Test coverage (1–4):** all four branches now covered — `setFields` `AgentHttpError` → `mapAgentError` (422), approval without `roleIdsAllowedToApprove` (details omitted), non-array/absent `refresh.relationships` → `[]`, and `html`-absent → `null` on the error body.

**501 fall-through:** added a `Warn` logging the payload keys (in the middleware, keeping the mapper pure).

**Mapper comment:** corrected — the real driver is `Action.execute(): Promise<unknown>` at the BFF boundary, not agent-client's typing.

**Kept — flat error body shape (`{ type: 'error', status, message, html }`):** intentional, not an oversight. This branch is *not* a BFF error envelope — it's the normalized action-result wrapper mandated by the PRD (API Contract: success/error/webhook/redirect share one flat `{ type, ... }` shape; File → 501). A client reads `body.type` to discriminate the action outcome, distinct from transport/BFF errors which keep the `{ error: { type, status, ... } }` envelope. Aligning it to the envelope would break the documented contract, so it's left as-is and asserted in the tests.

Tonours (approved):

PR #1760, opened 2026-07-15
feat(workflow-executor): deterministic source record for Get Data steps (PRD-775)

EnkiP:
Part of the deterministic Get Data step work.

fixes PRD-775

## What

The `read-record` (Get Data) step can now resolve its source record deterministically from a build-time-configured **stable BPMN step id** (`selectedRecordStepId`), mirroring how `trigger-action` and `load-related-record` already work — instead of always asking the AI to pick among the available records.

## Changes

- `ReadRecordStepDefinitionSchema.preRecordedArgs` gains `selectedRecordStepId` (legacy `selectedRecordStepIndex` kept for back-compat).
- `read-record-step-executor`: when `selectedRecordStepId` is set, resolve via `resolveSourceRecordRef` (revise-safe; `WORKFLOW_START_STEP_ID` → base record). Falls back to legacy index / AI selection when unset.
- Server contract (`ServerWorkflowTaskGetData`) + mapper forward `preRecordedArgs` for `get-data`.

The `fieldNames` pre-recorded path (build-time field selection) already existed and is unchanged.

## Tests

- Deterministic resolve skips the select-record AI round; `WORKFLOW_START_STEP_ID` → base record; unresolvable id → error.
- Mapper forwards `preRecordedArgs.selectedRecordStepId`.

Companion PRs: forestadmin-server (orchestrator), forestadmin (editor UI).

> [!NOTE]
> ### Add deterministic source record resolution by step ID to `ReadRecordStepExecutor`
> - Adds an optional `selectedRecordStepId` field to `preRecordedArgs` in `ReadRecordStepDefinition` and the `ServerWorkflowTaskGetData` interface, alongside the legacy `selectedRecordStepIndex` for back-compat.
> - When `preRecordedArgs.selectedRecordStepId` is set, `ReadRecordStepExecutor.doExecute` calls `resolveSourceRecordRef(selectedRecordStepId)` and skips the interactive record selection flow entirely.
> - The `mapTask` mapper in [step-definition-mapper.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1760/files#diff-b7de2aee7747c46286af27c1dac9169be8f2469b8a2bebe327065e1cc51efe02) now forwards `preRecordedArgs` from `GetData` server tasks to the resulting `ReadRecord` step definition.
> - Behavioral Change: executions with `selectedRecordStepId` set will bypass record selection; an unmatched step ID results in an error rather than falling back to selection.
>
> 
> #### Changes since #1760 opened
>
> - Modified field name resolution logic in `ReadRecordStepExecutor.doExecute` method to check both existence and non-zero length of `preRecordedArgs.fieldNames` before using pre-recorded values, falling back to `selectFields` call when `fieldNames` is undefined, null, or an empty array [1b73908]
> - Added test coverage for error handling when pinned source step returns no record and validation of deterministic execution path when both `selectedRecordStepId` and `fieldNames` are pre-recorded [1b73908]
> 
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized 1b73908. 2 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>
>

### DIFF
--- a/packages/workflow-executor/src/adapters/server-types.ts
+++ b/packages/workflow-executor/src/adapters/server-types.ts
@@ -66,6 +66,7 @@ export interface ServerWorkflowTaskGuideline extends ServerWorkflowTaskBase {
 interface ServerWorkflowTaskGetData extends ServerWorkflowTaskBase {
   taskType: ServerTaskTypeEnum.GetData;
   executionType: ServerStepExecutionTypeEnum.FullyAutomated;
+  preRecordedArgs?: { selectedRecordStepId?: string };
 }
 
 interface ServerWorkflowTaskUpdateData extends ServerWorkflowTaskBase {
--- a/packages/workflow-executor/src/adapters/server-types.ts
+++ b/packages/workflow-executor/src/adapters/server-types.ts
@@ -66,7 +66,7 @@ export interface ServerWorkflowTaskGuideline extends ServerWorkflowTaskBase {
 interface ServerWorkflowTaskGetData extends ServerWorkflowTaskBase {
   taskType: ServerTaskTypeEnum.GetData;
   executionType: ServerStepExecutionTypeEnum.FullyAutomated;
-  preRecordedArgs?: { selectedRecordStepId?: string };
+  preRecordedArgs?: { selectedRecordStepId?: string; fieldNames?: string[] };
 }
 
 interface ServerWorkflowTaskUpdateData extends ServerWorkflowTaskBase {
--- a/packages/workflow-executor/src/executors/read-record-step-executor.ts
+++ b/packages/workflow-executor/src/executors/read-record-step-executor.ts
@@ -30,8 +30,9 @@ export default class ReadRecordStepExecutor extends RecordStepExecutor<ReadRecor
           preRecordedArgs?.selectedRecordStepIndex,
         );
     const schema = await this.getCollectionSchema(selectedRecordRef.collectionName);
-    const fieldNames =
-      preRecordedArgs?.fieldNames ?? (await this.selectFields(schema, step.prompt));
+    const fieldNames = preRecordedArgs?.fieldNames?.length
+      ? preRecordedArgs.fieldNames
+      : await this.selectFields(schema, step.prompt);
     const selectedFields = fieldNames.map(requested => ({
       requested,
       field: this.findFieldByTechnicalName(schema, requested),
--- a/packages/workflow-executor/src/executors/read-record-step-executor.ts
+++ b/packages/workflow-executor/src/executors/read-record-step-executor.ts
@@ -21,13 +21,14 @@ export default class ReadRecordStepExecutor extends RecordStepExecutor<ReadRecor
   protected async doExecute(): Promise<StepExecutionResult> {
     const { stepDefinition: step } = this.context;
     const { preRecordedArgs } = step;
-    const records = await this.getAvailableRecordRefs();
 
-    const selectedRecordRef = await this.resolveRecordRef(
-      records,
-      step.prompt,
-      preRecordedArgs?.selectedRecordStepIndex,
-    );
+    const selectedRecordRef = preRecordedArgs?.selectedRecordStepId
+      ? await this.resolveSourceRecordRef(preRecordedArgs.selectedRecordStepId)
+      : await this.resolveRecordRef(
+          await this.getAvailableRecordRefs(),
+          step.prompt,
+          preRecordedArgs?.selectedRecordStepIndex,
+        );
     const schema = await this.getCollectionSchema(selectedRecordRef.collectionName);
     const fieldNames =
       preRecordedArgs?.fieldNames ?? (await this.selectFields(schema, step.prompt));
--- a/packages/workflow-executor/test/executors/read-record-step-executor.test.ts
+++ b/packages/workflow-executor/test/executors/read-record-step-executor.test.ts
@@ -13,7 +13,11 @@ import AgentWithLog from '../../src/executors/agent-with-log';
 import ReadRecordStepExecutor from '../../src/executors/read-record-step-executor';
 import SchemaCache from '../../src/schema-cache';
 import SchemaResolver from '../../src/schema-resolver';
-import { StepExecutionMode, StepType } from '../../src/types/validated/step-definition';
+import {
+  StepExecutionMode,
+  StepType,
+  WORKFLOW_START_STEP_ID,
+} from '../../src/types/validated/step-definition';
 
 function makeStep(overrides: Partial<ReadRecordStepDefinition> = {}): ReadRecordStepDefinition {
   return {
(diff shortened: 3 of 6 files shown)


packages/workflow-executor/src/executors/read-record-step-executor.ts
```diff
@@ -21,13 +21,14 @@ export default class ReadRecordStepExecutor extends RecordStepExecutor<ReadRecor
...
     const { stepDefinition: step } = this.context;
     const { preRecordedArgs } = step;
-    const records = await this.getAvailableRecordRefs();
 
-    const selectedRecordRef = await this.resolveRecordRef(
-      records,
-      step.prompt,
-      preRecordedArgs?.selectedRecordStepIndex,
-    );
+    const selectedRecordRef = preRecordedArgs?.selectedRecordStepId
```
matthv, line 25: 🟠 **Truthiness gate: an empty-string `selectedRecordStepId` silently downgrades to AI/legacy resolution**

The branch is `preRecordedArgs?.selectedRecordStepId ? …` (truthiness), but the zod schema types it `z.string().optional()`, which accepts `''`. If a config ever persists `selectedRecordStepId: ''` (partial revision, serialization glitch, a cleared field), `'' ? A : B` takes the legacy/AI branch — a step the operator configured as deterministic silently reads an AI-chosen record.

Suggest gating on presence rather than truthiness (`!= null`) or rejecting empty in the schema (`z.string().min(1)`), so a malformed id fails loudly instead of degrading.

🟡 Related (observability): nothing here logs *which* path ran. When `selectedRecordStepId`/`fieldNames` are absent (e.g. after a revision that dropped them), the step runs fully AI-driven with `status: success` and no trace distinguishing "resolved deterministically" from "fell back". On a feature branch literally named *deterministic-get-data*, a `Warn` log on the fallback would save a lot of debugging.

packages/workflow-executor/src/executors/read-record-step-executor.ts
```diff
@@ -21,13 +21,14 @@ export default class ReadRecordStepExecutor extends RecordStepExecutor<ReadRecor
...
+      ? await this.resolveSourceRecordRef(preRecordedArgs.selectedRecordStepId)
+      : await this.resolveRecordRef(
+          await this.getAvailableRecordRefs(),
+          step.prompt,
+          preRecordedArgs?.selectedRecordStepIndex,
+        );
     const schema = await this.getCollectionSchema(selectedRecordRef.collectionName);
     const fieldNames =
       preRecordedArgs?.fieldNames ?? (await this.selectFields(schema, step.prompt));
```
matthv: 🟠 **An explicitly-configured empty `fieldNames: []` throws an error that blames the AI**

`preRecordedArgs?.fieldNames ?? (await this.selectFields(...))` uses `??`, so an explicit empty array `[]` is kept (not replaced by the AI). Downstream that yields `resolvedFieldNames.length === 0` → `NoResolvedFieldsError`, whose `userMessage` is *"The AI selected fields that don't exist on this record. Try rephrasing the step's prompt."* — but in this case the AI was never invoked; the cause is an empty/misconfigured deterministic `fieldNames`. The operator will chase a non-existent AI/prompt issue.

The message (and ideally the error type) should branch on whether the fields came from config vs the AI. Note this path is newly reachable: before this PR the mapper didn't forward `preRecordedArgs` to get-data, so `fieldNames` was always `undefined` here.

packages/workflow-executor/src/adapters/server-types.ts
```diff
@@ -66,6 +66,7 @@ export interface ServerWorkflowTaskGuideline extends ServerWorkflowTaskBase {
 interface ServerWorkflowTaskGetData extends ServerWorkflowTaskBase {
   taskType: ServerTaskTypeEnum.GetData;
   executionType: ServerStepExecutionTypeEnum.FullyAutomated;
+  preRecordedArgs?: { selectedRecordStepId?: string };
```
matthv: 🟡 **This "local mirror of the orchestrator contract" is now inaccurate for get-data**

The added type declares only `{ selectedRecordStepId?: string }`, but `doExecute` actively consumes two more keys the orchestrator sends in the same object: `fieldNames` (`read-record-step-executor.ts`) and `selectedRecordStepIndex` (legacy). Runtime is fine (the mapper forwards `task.preRecordedArgs` wholesale and the zod schema keeps both), so this is **type-only** — but since this file is documented as the orchestrator contract mirror, it should reflect what read-record actually reads:

```ts
preRecordedArgs?: {
  selectedRecordStepId?: string;
  selectedRecordStepIndex?: number; // legacy
  fieldNames?: string[];
};
```

packages/workflow-executor/test/executors/read-record-step-executor.test.ts
```diff
@@ -1141,6 +1145,82 @@ describe('ReadRecordStepExecutor', () => {
...
+
+    it('returns error when selectedRecordStepId matches no source step', async () => {
+      const context = makeContext({
+        stepDefinition: makeStep({ preRecordedArgs: { selectedRecordStepId: 'nope' } }),
+      });
+      const executor = new ReadRecordStepExecutor(context);
+
+      const result = await executor.execute();
+
+      expect(result.stepOutcome.status).toBe('error');
```
matthv, line 1222: 🟡 **This negative test is too weak to tell the two error paths apart**

It asserts only `status).toBe('error')`. Per this repo's CLAUDE.md ("assertions must verify behavior, not just that something happened") — and the file's own pattern elsewhere (`expect(result.stepOutcome.error).toBe(...)`) — it should assert the error type/message (`InvalidPreRecordedArgsError`, "No source record found for step…"). As-is, this test still passes if `InvalidPreRecordedArgsError` and `SourceRecordMissingError` were swapped, since both produce `status: 'error'`.

Two coverage gaps in the same area worth adding:
- **`SourceRecordMissingError`** (source step pinned but it loaded nothing) — the one branch of the new resolution chain with zero coverage.
- **Fully-deterministic → zero AI**: a test with both `selectedRecordStepId` *and* `fieldNames` asserting `bindTools` is called **0** times. That's the headline promise of PRD-775; the current deterministic test still runs `selectFields` (1 AI round).

packages/workflow-executor/src/executors/read-record-step-executor.ts
EnkiP: irrelevant: empty selectedRecordStepId should indeed fallback to the AI branch

packages/workflow-executor/src/executors/read-record-step-executor.ts
EnkiP: yep, the condition should silently fallback to the AI branch when fieldNames.length = 0

packages/workflow-executor/src/adapters/server-types.ts
EnkiP: selectedRecordStepIndex should be dropped, it's normal.
but yeah, fieldNames should be added

packages/workflow-executor/src/executors/read-record-step-executor.ts
```diff
@@ -21,16 +21,18 @@ export default class ReadRecordStepExecutor extends RecordStepExecutor<ReadRecor
...
-      step.prompt,
-      preRecordedArgs?.selectedRecordStepIndex,
-    );
+    const selectedRecordRef = preRecordedArgs?.selectedRecordStepId
+      ? await this.resolveSourceRecordRef(preRecordedArgs.selectedRecordStepId)
+      : await this.resolveRecordRef(
+          await this.getAvailableRecordRefs(),
+          step.prompt,
+          preRecordedArgs?.selectedRecordStepIndex,
+        );
```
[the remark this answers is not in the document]
EnkiP: expected behaviour

matthv (approved):
LGTM

PR #1775, opened 2026-07-23
fix(agent-client): build static forms from the schema instead of probing /hooks/load [PRD-802]

matthv:
## Context

`@forestadmin/agent-client` (used by the workflow executor) POSTs `<action>/hooks/load` unconditionally when opening an action form, even when the schema says `hooks.load: false` (static form). Agents that don't expose the route answer 404:

- forest_liana < 9.20.1, for every action without a load hook — fixed agent-side in 9.20.1 ([PRD-801](https://linear.app/forestadmin/issue/PRD-801)), but only for endpoints under the `/forest` engine mount;
- actions whose custom endpoints are **proxied outside the agent** (reported by Qonto): no agent-side fix can cover those, the 404s keep polluting the customer's HTTP metrics.

The probe is unnecessary for static forms: the Node agent sets `hooks.load = !schema.staticForm`, and for static forms the schema already carries the complete form — fields with materialized defaults (`setFieldsDefaultValue`) and layout when any (`generator-actions.ts`).

## Changes

- **agent-client** — `FieldFormStates.loadInitialState`: when `hooks.load` is false and the schema provided fields, build the form from `fields`/`layout` without any HTTP call. When the schema gave no fields, keep the existing behavior (probe + swallow 404 as "no dynamic fields"). Fallback logic extracted into `applyFallbackForm`, which now also restores the schema layout.
- **forestadmin-client** — `ForestSchemaAction` gains the optional `layout` that the Node agent already emits for static forms.
- **workflow-executor** — the schema envelope validates and threads `layout` through `buildActionEndpoints`.

## Tests

- new: `hooks.load: false` + schema fields → form built, **no request at all**;
- new: schema layout restored in that path;
- new: empty static form (`fields: []`) → no request, empty form;
- kept: no schema fields → probe still happens, 404 swallowed, 500 rethrown, successful response used (Node compatibility for old schemas).

347 agent-client + 1466 workflow-executor + 283 forestadmin-client tests green.

## Known limitation

The orchestrator's `getCollectionSchema` response doesn't expose action `layout` yet, so static Node forms consumed through the executor lose their layout elements until forestadmin-server passes it through (fields and values are unaffected — same degradation the 404 fallback already had). Tracked as a follow-up in PRD-802.

Linear: PRD-802

> [!NOTE]
> ### Build static action forms from schema layout instead of probing `/hooks/load`
> - When `hooks.load` is `false` and fallback fields are present in the schema, `FieldFormStates.loadInitialState` now skips the `/hooks/load` request and builds the form directly from schema fields and layout.
> - A new private `applyFallbackForm` method in [`field-form-states.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1775/files#diff-7ab5307c436664b3183f218254bb7e09882031d99c6b49a865e1280f22b2531d) consolidates fallback form construction, including clearing state, applying layout, and setting field defaults.
> - [`ActionEndpointsByCollection`](https://github.com/ForestAdmin/agent-nodejs/pull/1775/files#diff-c434b5b3c0debe1a1c3629c2828238cc6ab1609dce58c3b2f6dae384ef677b81) now treats `hooks` and `fields` as optional and includes `layout`; omitting them preserves legacy probing behavior.
> - [`AgentClientAgentPort.buildActionEndpoints`](https://github.com/ForestAdmin/agent-nodejs/pull/1775/files#diff-1e123e0cbe27612cfd4c2f0c74aa6529888468059a8106cdc6bab8375b9b141a) forwards `hooks`, `fields`, and `layout` as-is from the schema without defaulting, so absence is propagated correctly.
> - Behavioral Change: 404 responses from `/hooks/load` are now rethrown when `hooks.load` is `true`; they are only swallowed when hooks are absent or `load` is `false`.
>
> 
> #### Changes since #1775 opened
>
> - Modified `FieldFormStates.applyFallbackForm` method to preserve all schema-provided field properties when applying fallback forms for static actions [1178315]
> - Extended `ForestSchemaAction` interface in `forestadmin-client` to include optional field properties [1178315]
> - Renamed `ActionFieldsSchema` constant to `LooseObjectArraySchema` in `workflow-executor` validated collection module [1178315]
> - Added `widgetEdit` property with `parameters.static.options`, `enableOpacity`, and `quickPalette` to the `ForestSchemaAction` interface fields definition in `forestadmin-client` to support static form rendering without probing `/hooks/load` [3e1646f]
> - Extended `FieldFormStates.loadInitialState` test to verify that `getMultipleChoiceField` returns options from schema-provided `widgetEdit.parameters.static.options` when load hook is disabled [3e1646f]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2bc4223.</sup>
> 
>
> 

## Behavior changes (for the changelog)

- `hooks.load: false` + `fields` provided now means **"trust the schema, skip the `/hooks/load` call"** (previously: probe and swallow an eventual 404). External consumers of `@forestadmin/agent-client` that coerced absent hooks/fields to `{load: false}` / `[]` (as the executor did) must forward them as-is, or every legacy schema would render as an empty static form.
- Legacy schemas (no `hooks` at all) whose agent 404s on `/hooks/load`: the 404 was previously **rethrown** (visible error at form load) and is now **swallowed** (empty form; a real problem surfaces at action execution instead). Schemas that explicitly declare `hooks.load: true` still rethrow the 404.

### DIFF
--- a/packages/agent-client/src/action-fields/field-form-states.ts
+++ b/packages/agent-client/src/action-fields/field-form-states.ts
@@ -18,6 +18,7 @@ export default class FieldFormStates {
   private readonly layout: ForestServerActionFormLayoutElement[];
   private readonly hooks?: ForestSchemaAction['hooks'];
   private readonly fallbackFields?: ForestSchemaAction['fields'];
+  private readonly fallbackLayout?: ForestSchemaAction['layout'];
 
   constructor(
     actionName: string,
@@ -27,6 +28,7 @@ export default class FieldFormStates {
     ids: string[],
     hooks?: ForestSchemaAction['hooks'],
     fallbackFields?: ForestSchemaAction['fields'],
+    fallbackLayout?: ForestSchemaAction['layout'],
   ) {
     this.fields = [];
     this.actionName = actionName;
@@ -37,6 +39,7 @@ export default class FieldFormStates {
     this.layout = [];
     this.hooks = hooks;
     this.fallbackFields = fallbackFields;
+    this.fallbackLayout = fallbackLayout;
   }
 
   getFieldValues(): Record<string, unknown> {
@@ -78,6 +81,16 @@ export default class FieldFormStates {
   }
 
   async loadInitialState(): Promise<void> {
+    // hooks.load: false means the form is static: the schema already carries the complete form
+    // (fields with materialized defaults, and layout when any), so the request could only return
+    // the same data — and forest_liana agents < 9.20.1 don't even expose the route (404 noise in
+    // customers' metrics). Only probe when the schema gave us no fields to fall back on.
+    if (this.hooks && !this.hooks.load && this.fallbackFields) {
+      this.applyFallbackForm();
+
+      return;
+    }
+
     const requestBody = {
       data: {
         attributes: {
@@ -100,40 +113,34 @@ export default class FieldFormStates {
       this.layout.push(...(queryResults.layout ?? []));
       this.addFields(queryResults.fields);
     } catch (error) {
-      // When hooks.load is false, the behavior differs between backends:
-      //
-      // - Node agent (@forestadmin/agent): always responds to POST /hooks/load
-      //   with the form fields, even when hooks.load is false in the schema.
-      //   In this case the call above succeeds and fields are loaded normally.
-      //
-      // - Ruby agent (forest_liana): does NOT register a route for /hooks/load
-      //   when hooks.load is false. The POST returns a 404.
-      //   In this case we catch the 404 and continue with an empty form,
-      //   which matches the expected behavior (no dynamic fields to load).
-      //
-      // We always attempt the call so Node users get their fields,
-      // and only swallow 404 errors for Ruby users. Other errors (401, 500,
-      // network failures) are rethrown so they surface properly.
+      // Reached only when the schema provided no fallback fields for a static form: probe the
+      // agent anyway (Node agents answer with the form), and swallow the forest_liana < 9.20.1
+      // 404 as "no dynamic fields to load". Other errors (401, 500, network) surface properly.
       if (this.hooks && !this.hooks.load && HttpRequester.is404Error(error)) {
-        this.clearFieldsAndLayout();
-
-        if (this.fallbackFields?.length) {
-          this.addFields(
-            this.fallbackFields.map(f => ({
-              field: f.field,
-              type: f.type,
-              isRequired: f.isRequired ?? false,
-              isReadOnly: false,
-              value: f.defaultValue,
-              hook: f.hook,
-            })),
-          );
-        }
+        this.applyFallbackForm();
 
         return;
       }
 
       throw error;
+    }
+  }
+
+  private applyFallbackForm(): void {
+    this.clearFieldsAndLayout();
+    this.layout.push(...(this.fallbackLayout ?? []));
+
+    if (this.fallbackFields?.length) {
+      this.addFields(
+        this.fallbackFields.map(f => ({
+          field: f.field,
+          type: f.type,
+          isRequired: f.isRequired ?? false,
+          isReadOnly: false,
+          value: f.defaultValue,
+          hook: f.hook,
+        })),
+      );
     }
   }
 
--- a/packages/agent-client/src/action-fields/field-form-states.ts
+++ b/packages/agent-client/src/action-fields/field-form-states.ts
@@ -132,14 +132,15 @@ export default class FieldFormStates {
     this.layout.push(...(this.fallbackLayout ?? []));
 
     if (this.fallbackFields?.length) {
+      // Spread first: the schema is now the ONLY source for static forms, so every field prop
+      // (enums, description, widgetEdit, ...) must survive — dropping one starves consumers
+      // like ActionFieldEnum.getOptions() that the probe used to feed.
       this.addFields(
         this.fallbackFields.map(f => ({
-          field: f.field,
-          type: f.type,
+          ...f,
           isRequired: f.isRequired ?? false,
-          isReadOnly: false,
+          isReadOnly: f.isReadOnly ?? false,
           value: f.defaultValue,
-          hook: f.hook,
         })),
       );
     }
--- a/packages/agent-client/src/action-fields/field-form-states.ts
+++ b/packages/agent-client/src/action-fields/field-form-states.ts
@@ -113,10 +113,11 @@ export default class FieldFormStates {
       this.layout.push(...(queryResults.layout ?? []));
       this.addFields(queryResults.fields);
     } catch (error) {
-      // Reached only when the schema provided no fallback fields for a static form: probe the
-      // agent anyway (Node agents answer with the form), and swallow the forest_liana < 9.20.1
-      // 404 as "no dynamic fields to load". Other errors (401, 500, network) surface properly.
-      if (this.hooks && !this.hooks.load && HttpRequester.is404Error(error)) {
+      // Reached when the schema provided no fallback fields, or no hooks at all (legacy schemas):
+      // probe the agent anyway (Node agents answer with the form), and swallow the 404 of agents
+      // that don't expose the route as "no dynamic fields to load" — unless the schema explicitly
+      // declares a load hook, in which case a 404 is a real error. 401/500/network still surface.
+      if (!this.hooks?.load && HttpRequester.is404Error(error)) {
         this.applyFallbackForm();
 
         return;
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -406,18 +406,19 @@ export default class AgentClientAgentPort implements AgentPort {
       endpoints[collectionName] = {};
 
       for (const action of schema.actions) {
-        // agent-client POSTs /hooks/load unconditionally; `hooks.load` tells it whether a 404
-        // there is expected (Ruby agent, swallowed → fallback to the static `fields` below) or
-        // a real error. Both `hooks` and `fields` must mirror the agent's real schema for form
-        // detection to work on Ruby agents.
+        // agent-client only POSTs /hooks/load when `hooks.load` is true (dynamic form); for static
+        // forms it builds the form from `fields`/`layout` directly, so no 404 probe reaches agents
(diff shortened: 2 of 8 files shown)


packages/agent-client/src/action-fields/field-form-states.ts
```diff
@@ -78,6 +81,16 @@ export default class FieldFormStates {
   }
 
   async loadInitialState(): Promise<void> {
+    // hooks.load: false means the form is static: the schema already carries the complete form
+    // (fields with materialized defaults, and layout when any), so the request could only return
+    // the same data — and forest_liana agents < 9.20.1 don't even expose the route (404 noise in
+    // customers' metrics). Only probe when the schema gave us no fields to fall back on.
+    if (this.hooks && !this.hooks.load && this.fallbackFields) {
```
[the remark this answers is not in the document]
matthv: Fixed in 2bc422317, at the source rather than with a non-empty check: the executor no longer coerces omitted `hooks`/`fields` (`?? {load: false}` / `?? []`), so an absent field list stays distinguishable from an explicitly empty static form — legacy schemas keep probing `/hooks/load`, and `fields: []` remains a legitimate empty-form skip. The same coercion on `hooks` had the worse twin of this bug: a load-hook action on a pre-9.17 gem (no hooks in schema) would have been rendered as a static form; also covered. The 404 swallow now lives in agent-client and is rethrown when the schema explicitly declares a load hook.

packages/agent-client/src/action-fields/field-form-states.ts
```diff
@@ -137,6 +127,24 @@ export default class FieldFormStates {
     }
   }
 
+  private applyFallbackForm(): void {
+    this.clearFieldsAndLayout();
+    this.layout.push(...(this.fallbackLayout ?? []));
+
+    if (this.fallbackFields?.length) {
+      this.addFields(
+        this.fallbackFields.map(f => ({
```
Scra3, line 139: issue (blocking): This mapping drops `enums` (and `description`/`widgetEdit`/`label`) from the schema fields — it was tolerable as the legacy-Ruby 404 fallback, but it is now the main path for every static form on every agent: `ActionFieldEnum.getOptions()` reads `plainField.enums`, so the executor's AI fill loses the allowed values and required Enum fields become unfillable (FullyAutomated falls back to pause). Skipping the probe must not return less data than the probe did (forest_liana 9.20.1 now answers it with the full form) — propagate the remaining field props and add a test asserting enum options survive the skip path.

packages/agent-client/src/action-fields/field-form-states.ts
```diff
@@ -100,35 +113,12 @@ export default class FieldFormStates {
...
-              hook: f.hook,
-            })),
-          );
-        }
+      // Reached when the schema provided no fallback fields, or no hooks at all (legacy schemas):
+      // probe the agent anyway (Node agents answer with the form), and swallow the 404 of agents
+      // that don't expose the route as "no dynamic fields to load" — unless the schema explicitly
+      // declares a load hook, in which case a 404 is a real error. 401/500/network still surface.
+      if (!this.hooks?.load && HttpRequester.is404Error(error)) {
+        this.applyFallbackForm();
```
Scra3, line 121: issue (non-blocking): Behavior change worth a line in the PR description — a legacy schema (no `hooks`) whose agent 404s on /hooks/load previously rethrew (visible error at form load) and now yields an empty form, moving the failure to action execution time.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
```diff
@@ -406,18 +406,20 @@ export default class AgentClientAgentPort implements AgentPort {
...
+        // silently render dynamic or unknown forms as empty static ones.
         endpoints[collectionName][action.name] = {
           id: action.name,
           name: action.name,
           endpoint: action.endpoint,
-          hooks: action.hooks ?? { load: false, change: [] },
-          // Zod envelope-validates `fields` as an array of opaque objects. Inner widget/parameters
-          // shape is owned by @forestadmin/forestadmin-client and consumed by agent-client below.
-          fields: (action.fields ?? []) as ActionEndpointsByCollection[string][string]['fields'],
+          hooks: action.hooks,
```
Scra3, line 418: praise: Removing the `?? {load: false, change: []}` / `?? []` coercion is the critical piece — with the new skip semantics that coercion would have silently rendered every legacy schema as an empty static form, and the regression test locks it in.

packages/workflow-executor/src/types/validated/collection.ts
```diff
@@ -73,8 +73,10 @@ export const ActionSchemaSchema = z.object({
   displayName: z.string().min(1),
   endpoint: z.string().min(1),
   type: z.enum(['single', 'bulk', 'global']).optional(),
-  /** Static form fields. Used as fallback when the agent's /hooks/load route 404s (old Ruby agents). */
+  /** Static form fields. Lets agent-client skip the /hooks/load probe on static forms. */
   fields: ActionFieldsSchema,
+  /** Static form layout, when the schema carries one. Envelope-validated like `fields`. */
+  layout: ActionFieldsSchema,
```
Scra3: nit: `ActionFieldsSchema` now validates `layout` too, so the name lies — rename to something shape-based like `LooseObjectArraySchema` (or alias it).

Scra3 (changes requested):
Solid design overall: the probe-skip decision is fully owned by agent-client, the executor adapter stops falsifying absent hooks/fields (and mcp-server already forwards faithfully, so no other caller is trapped), and the test matrix covers the skip/legacy/404/500 paths well. One blocking regression on the static path (enum options), plus a server-side follow-up: the orchestrator's getCollectionSchema strips `enums` from action fields (workflow-orchestrator-service.ts:1125), so the executor path needs a companion forestadmin-server PR to benefit — I'll open the ticket. Last point for the changelog: this PR changes the *meaning* of existing inputs (`hooks.load: false` + `fields` no longer means "probe and maybe swallow 404" but "trust the schema, skip the call") — agent-client is a published package, so any external consumer that coerced absent hooks/fields like the executor did will silently get empty static forms; this deserves an explicit changelog entry.

packages/agent-client/src/action-fields/field-form-states.ts
```diff
@@ -137,6 +127,25 @@ export default class FieldFormStates {
     }
   }
 
+  private applyFallbackForm(): void {
+    this.clearFieldsAndLayout();
+    this.layout.push(...(this.fallbackLayout ?? []));
+
+    if (this.fallbackFields?.length) {
+      // Spread first: the schema is now the ONLY source for static forms, so every field prop
+      // (enums, description, widgetEdit, ...) must survive — dropping one starves consumers
```
Scra3, line 136: issue (blocking, widening my enums comment): this mapping cherry-picks `{field, type, isRequired, isReadOnly, value, hook}` and drops `enums`, `widgetEdit`, and `description` — all read downstream (`ActionFieldEnum.getOptions()` → `enums`; dropdown/radio/checkbox → `widgetEdit.parameters.static.options`; color → `widgetEdit.parameters.static.{enableOpacity,quickPalette}`; AI fill uses `description` as the per-field hint). Forward the full `PlainField` shape rather than a subset, and add a test asserting a dropdown's options and an enum's values both survive the skip path.

Scra3 (commented):
Follow-up to my earlier enums comment — broadening it. `applyFallbackForm` is now the primary path for every static form, so it must carry the *full* field shape the form layer reads (`FieldGetter`/`PlainField`), not just enums: `widgetEdit`, `description`, `isReadOnly` in addition to `hook`/`value`/`isRequired`. `widgetEdit` is the same functional class as enums — dropdown/radio/checkbox/color fields carry their allowed options under `widgetEdit.parameters.static.options`, so dropping it makes those required fields unfillable by the AI too. Companion orchestrator PR (forestadmin-server#8393) now forwards the full set (enums + widgetEdit + description + isReadOnly + hook), verified shape-compatible; agent-client's fallback mapping needs to stop cherry-picking so the data survives this second chokepoint.

packages/agent-client/src/action-fields/field-form-states.ts
matthv: The cherry-picking was already replaced by a full spread in 117831529 — our pushes crossed. 3e1646f8c completes your point: `widgetEdit` is now declared on `ForestSchemaAction.fields` (mirroring the shape `PlainField` reads), and the skip-path test asserts both an enum's values (`ActionFieldEnum.getOptions()`) and a dropdown's options (`getMultipleChoiceField().getOptions()` → `widgetEdit.parameters.static.options`) survive.

Scra3 (approved):
Verified 117831529 + 3e1646f8c: the full spread with post-spread overrides (`value: f.defaultValue`, `isRequired/isReadOnly ?? false`) resolves my blocking comment — and respecting the schema's `isReadOnly` (previously hardcoded `false`) is a bonus fix. The skip-path test pins exactly what I asked (enum values via `getOptions()` AND dropdown options via `getMultipleChoiceField()`), and typing `widgetEdit` on `ForestSchemaAction.fields` locks the contract. CI green. One reminder for the merge, non-blocking: a changelog/PR-description line about the semantic change of `hooks.load: false` + `fields` for external agent-client consumers. Approving.

packages/agent-client/src/action-fields/field-form-states.ts
matthv: Fixed in 117831529: the mapping now spreads the whole schema field (only `isRequired`/`isReadOnly` get defaults, plus the `defaultValue` → `value` projection), so `enums`/`description`/`widgetEdit`/`label` all survive, with a test pinning enum values on the skip path. 3e1646f8c extends it per your follow-up (widgetEdit typing + dropdown options assertion).

packages/agent-client/src/action-fields/field-form-states.ts
matthv: Done — added a "Behavior changes (for the changelog)" section to the PR description covering both this legacy-404 rethrow→swallow change and the new meaning of `hooks.load: false` + `fields` for external consumers.

packages/workflow-executor/src/types/validated/collection.ts
matthv: Renamed to `LooseObjectArraySchema` in 117831529.

PR #1787, opened 2026-07-28
fix(workflow-executor): resolve xToOne relations with PascalCase names

Scra3:
A client (swaive) reported a `load-related-record` step returning **no record** on a 1-1 relation, even though the related record exists.

## Root cause

xToOne relations have no `/relationships/<relation>` route on the agent, so `getSingleRelatedData` reads the parent record with a `<relation>@@@<field>` projection and unpacks the linkage agent-client returns:

```ts
const linkage = parent.values[toCamelCase(relation)];
```

agent-client deserializes every response key with jsonapi-serializer's `keyForAttribute: 'camelCase'` (`packages/agent-client/src/http-requester.ts:31`), which **lowercases the initial letter**. The local `toCamelCase` only handled snake_case (`_x` → `X`) and left PascalCase untouched:

```
agent serializes the relation as   AddressProofMetadatum   (serializer.ts:111, raw fieldName)
agent-client returns the key as    addressProofMetadatum   (initial letter lowercased)
toCamelCase produced              AddressProofMetadatum   (unchanged) → lookup misses → null
```

So the step reported "no record" for any xToOne relation whose technical name starts with an uppercase letter — common with Sequelize/Prisma models (`LegalEntity`, `UserDetail`, `RibMetadatum`, `ReferenceAccount`…). Relations named in lowercase (`entity`, `person`) worked, which is why this looked intermittent.

`restoreFieldNames` had the same gap for PascalCase columns (e.g. `currentAUM`, whose real key is `currentAum`).

The same class of bug was fixed once before for snake_case only (regression test `billing_address` → `billingAddress`); PascalCase slipped through.

## Fix

`toCamelCase` now mirrors the inflection agent-client actually applies — `inflected.underscore` then `camelize` without an initial capital. Not a plain first-letter lowercase, which would break acronyms (`KYCEvents` must yield `kycEvents`, not `kYCEvents`).

Verified **equivalent to jsonapi-serializer's inflector** on 300k fuzzed names (0 divergence) plus the real client schema names.

## Why this cannot regress

Since the new function is exactly equivalent to the inflector that *produces* the keys, it matches every key correctly — so it can never lose a match the old one had. Checked against 42 real field/relation names from the client schema: 18 change behavior, **18 are fixes, 0 regressions**. `toCamelCase` is module-private with 2 call sites, so the blast radius is this single file.

## Tests

Three regression tests added alongside the existing snake_case one: PascalCase relation name, acronym-prefixed name, and PascalCase field restoration. All three **fail before the fix, pass after**.

- `yarn workspace @forestadmin/workflow-executor test` → 1474 passed
- `yarn workspace @forestadmin/workflow-executor lint` → 0 errors

> [!NOTE]
> ### Fix `toCamelCase` util to resolve xToOne relations with PascalCase names
> The `toCamelCase` utility in [agent-client-agent-port.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1787/files#diff-1e123e0cbe27612cfd4c2f0c74aa6529888468059a8106cdc6bab8375b9b141a) previously only handled snake_case input, causing silent lookup misses for relation names in PascalCase, acronym-leading, or kebab-case formats (e.g., `LegalEntity`, `KYCEvent`, `billing-address`).
>
> - Rewrites `toCamelCase` as a two-step transform: normalizes arbitrary input to snake_case first (handling PascalCase, camelCase, acronyms, digits, kebab-case), then camelizes to match `inflected`'s behavior.
> - Adds tests covering PascalCase fields, acronym-prefixed relations, kebab-case relations, and composite PKs in [agent-client-agent-port.test.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1787/files#diff-22a2115dddda11057fe52890dae8030174d6b71d092e9d794b34237c9c30e420).
>
> 
> #### Changes since #1787 opened
>
> - Added assertions to verify field projection behavior in `AgentClientAgentPort` xToOne relation tests [f2b8d33]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ca54c54.</sup>
> 
>
>

### DIFF
--- a/packages/workflow-executor/test/adapters/agent-client-agent-port.test.ts
+++ b/packages/workflow-executor/test/adapters/agent-client-agent-port.test.ts
@@ -668,7 +668,7 @@ describe('AgentClientAgentPort', () => {
     // Regression: jsonapi-serializer emits the nested linkage with camelCased attribute
     // keys (full_name → fullName). The adapter must restore those keys before returning
     // values, otherwise snake_case referenceFields silently resolve to undefined upstream.
-    it('restores camelCased keys on the linkage when fields are snake_case', async () => {
+    it('projects the raw snake_case field name and restores it on the linkage', async () => {
       const snakeSchema = {
         ...ordersSchema,
         fields: [
@@ -694,6 +694,7 @@ describe('AgentClientAgentPort', () => {
         user,
       );
 
+      expect(mockCollection.getOne).toHaveBeenCalledWith([42], { fields: ['order@@@full_name'] });
       expect(result?.values).toEqual({ id: '99', full_name: 'John Doe' });
     });
 
@@ -784,7 +785,7 @@ describe('AgentClientAgentPort', () => {
       expect(result?.recordId).toEqual(['ba-1']);
     });
 
-    it('restores PascalCase field names on the linkage', async () => {
+    it('projects the raw PascalCase field name and restores it on the linkage', async () => {
       mockCollection.getOne.mockResolvedValue({ order: { id: '99', fullName: 'John Doe' } });
 
       const result = await port.getSingleRelatedData(
@@ -798,6 +799,7 @@ describe('AgentClientAgentPort', () => {
         user,
       );
 
+      expect(mockCollection.getOne).toHaveBeenCalledWith([42], { fields: ['order@@@FullName'] });
       expect(result?.values).toEqual({ id: '99', FullName: 'John Doe' });
     });
 
--- a/packages/workflow-executor/test/adapters/agent-client-agent-port.test.ts
+++ b/packages/workflow-executor/test/adapters/agent-client-agent-port.test.ts
@@ -685,9 +685,8 @@ describe('AgentClientAgentPort', () => {
       });
     });
 
-    // Regression: the relation name can be PascalCase (Sequelize/Prisma models often are).
-    // jsonapi-serializer lowercases the first letter, so the raw name found no linkage and the
-    // step reported "no record" even though the related record existed.
+    // Regression: jsonapi-serializer lowercases the initial letter, so `LegalEntity` arrives as
+    // `legalEntity` — looking it up by the raw name found nothing and the step reported "no record".
     it('finds the linkage when the relation name is PascalCase', async () => {
       mockCollection.getOne.mockResolvedValue({ legalEntity: { id: 'le-1' } });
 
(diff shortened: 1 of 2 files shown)


packages/workflow-executor/test/adapters/agent-client-agent-port.test.ts
```diff
@@ -685,6 +723,84 @@ describe('AgentClientAgentPort', () => {
...
+          collection: 'users',
+          id: [42],
+          relation: 'order',
+          relatedSchema: ordersSchema,
+          fields: ['FullName'],
+        },
+        user,
+      );
+
+      expect(result?.values).toEqual({ id: '99', FullName: 'John Doe' });
```
matthv, line 803: The "raw name for the projection, camelCased name for the lookup" invariant is pinned for **relation** names (the three tests above all assert `getOne` was called with `LegalEntity@@@id` / `KYCEvent@@@id` / `billing-address@@@id`), but nowhere for **field** names — this test only asserts `result?.values`.

That leaves a whole class of wrong fix green: camelCasing the projected field too, i.e.

```ts
fields: [`${relation}@@@${toCamelCase(projectedField)}`]   // wrong, but passes the suite
```

The mock returns `fullName` regardless of what was asked, so nothing catches it here. Against a real agent it would send `order@@@fullName` for a column named `FullName` — the ValidationError case documented at `agent-client-agent-port.ts:218-222`.

One line closes it:

```ts
expect(mockCollection.getOne).toHaveBeenCalledWith([42], { fields: ['order@@@FullName'] });
```

Scra3: Fixed in f2b8d33d6. Added the projection assertion on both linkage-restore tests, not just the PascalCase one — `restores camelCased keys on the linkage when fields are snake_case` had the exact same hole (`order@@@full_name` was never pinned either):

```js
expect(mockCollection.getOne).toHaveBeenCalledWith([42], { fields: [order@@@full_name] });
expect(mockCollection.getOne).toHaveBeenCalledWith([42], { fields: [order@@@FullName] });
```

Renamed both to `projects the raw <case> field name and restores it on the linkage` so the names match what they now assert.

Verified by mutation: applying the wrong fix (`toCamelCase(projectedField)` in the projection) fails exactly these two tests and nothing else — so the class of green-but-wrong fix you described is now closed.

matthv (approved):

PR #1797, opened 2026-08-03
feat(mcp-server): restrict oauth clients to an allowlist of domains

hercemer42:
## Description

Adds an `allowedOAuthClients` option to the MCP server so administrators can restrict which OAuth client applications may connect. DCR means any client can register and — once a legitimate user signs in — obtain tokens; zero-trust customers asked to cap that to approved clients only (e.g. Dust).

- `agent.mountAiMcpServer({ allowedOAuthClients: ['dust.tt'] })` and `FOREST_MCP_ALLOWED_OAUTH_CLIENTS` (standalone CLI) — both entry points, mirroring `tokenTtl`/`enabledTools`.
- Enforcement lives in `ForestOAuthProvider.clientsStore.getClient()`, the single choke point the MCP SDK's authorize and token handlers resolve clients through — covering authorization, code exchange, and refresh with one check and no extra HTTP round trip.
- A client is allowed only when it has at least one registered redirect URI and every one is on an allowed domain or subdomain (case-insensitive, lookalike-proof). Redirect URIs are matched because they are the one piece of registration metadata an impostor cannot benefit from; self-declared `client_name`/`client_uri` are ignored.
- Disallowed clients get a standard `invalid_client` (4xx, never a 500, never a redirect to the unvalidated URI) with a targeted `error_description` that names no allowed domain; the rejection is logged with client id + URIs for support.
- Omitted option = behavior unchanged. Empty array = startup error. Registration itself is not blocked (it happens on the Forest server); enforcement is use-time only.
- Native/loopback clients are inherently rejected when the allowlist is set — deliberate, documented as a non-goal in the README.

## Testing

- TDD: 28 new tests across provider (unit), server (supertest through the real express app: exchange+refresh success, authorize 302, 400 rejections on all three grant paths, empty-list construction throw), env parser, and agent pass-through.
- Scoped runs green: provider+parser 71, server 109, agent 42. eslint/tsc/prettier clean on changed files.

fixes PRD-861

> [!NOTE]
> ### Restrict MCP server OAuth clients to an allowlist of domains
> - Adds an `allowedOAuthClients` option to `ForestMCPServer` and `ForestOAuthProvider` that accepts a list of domains; when set, OAuth clients whose redirect URIs do not match an allowed domain (or subdomain) are rejected with a 400 `invalid_client` error.
> - The allowlist is parsed from the `FOREST_MCP_ALLOWED_OAUTH_CLIENTS` environment variable (comma-separated) in both the standalone CLI ([cli.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1797/files#diff-fde9d45c049a9795bb937299c77392960550d4d4d5c3dd6aa2b2eaa9bd6a0c25)) and the example agent ([agent.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1797/files#diff-b48e506f18c2fc5ca1454df7ef2ed5ef1f1773c1edd5a604ffae6e8c14b50bdd)).
> - Domain matching is case-insensitive and covers subdomains; non-http(s) redirect URI schemes are always rejected when the allowlist is active.
> - Risk: constructing `ForestMCPServer` with an explicitly empty `allowedOAuthClients` array now throws, preventing silent rejection of all clients.
>
> 
> #### Changes since #1797 opened
>
> - Added domain list validation and normalization utility [23eac8b]
> - Integrated domain validation into MCP server initialization [23eac8b]
> - Fixed environment variable parsing in agent initialization [23eac8b]
> - Added server tests for OAuth domain allowlist validation [23eac8b]
> - Documented OAuth domain allowlist validation rules [23eac8b]
> - Extended `normalizeDomainList` function to reject domain entries containing '@', '?', or '#' characters in addition to the existing rejections for '/', ':', and whitespace, preventing incorrect hostname parsing when these URL delimiter characters are present [ad46ed3]
> - Updated documentation and comments to clarify OAuth client allowlist parsing behavior and loopback redirect URI handling [ad46ed3]
> - Added backslash character to the invalid character validation in `normalizeDomainList` function within the `mcp-server` package [6344645]
> - Modified `normalizeDomainList` function to strip trailing root dots from FQDN-style domain entries and added test coverage [691b6d7]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6ff0797.</sup>
> 
>
>


packages/mcp-server/src/forest-oauth-provider.ts
```diff
@@ -165,12 +169,58 @@ export default class ForestOAuthProvider implements OAuthServerProvider {
...
+    );
+
+    // The description names no allowed domain on purpose; it is served to arbitrary clients.
+    throw new InvalidClientError(
+      'This MCP server only accepts approved client applications. ' +
+        'Contact your Forest Admin administrator.',
+    );
+  }
+
+  private static isUriOnAllowedDomain(redirectUri: string, allowedDomains: string[]): boolean {
```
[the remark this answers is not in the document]
hercemer42: Claude Fable 5 (claude-fable-5): Confirmed and fixed in b0e20ef9. WHATWG URL parsing does yield `hostname === 'dust.tt'` for non-special schemes, so `attacker-app://dust.tt/callback` passed the domain check while the OS would hand the callback to whatever app registered the scheme — the hostname is meaningless for delivery there. `isUriOnAllowedDomain` now rejects any scheme other than `http:`/`https:` before domain matching; covered by a new unit test (`rejects a custom-scheme redirect URI even when its hostname is on an allowed domain`), and the README + PRD-861 business rule were updated to state the http(s) requirement. Note the practical exploit still requires the attacker's app to be installed on the victim's device and the victim to complete the login — but the fix is a two-line fail-closed check, so severity arguing wasn't worth it.

hercemer42:
Claude Fable 5 (claude-fable-5): On the qlty **High total complexity (count = 56)** structure issue — pushing back on addressing it in this PR. The flagged total is file-level complexity on `forest-oauth-provider.ts`, which was already the package's largest source file before this change; the PR adds one focused enforcement method (`assertClientIsAllowed`) and one pure static matcher (`isUriOnAllowedDomain`), both small and single-responsibility. Reducing the file's total complexity means splitting the provider (token exchange / refresh / verification / client store), which is exactly the opportunistic refactoring our PR conventions exclude from feature PRs. If the team wants that split, it should be its own `refactor(mcp-server)` PR — happy to file a ticket.

packages/mcp-server/src/server.ts
```diff
@@ -148,6 +148,13 @@ export interface ForestMCPServerOptions {
    * logins. Minimum 60s for either.
    */
   tokenTtl?: TokenTtlOptions;
+  /**
+   * Domains of the OAuth client applications allowed to use this server, e.g. ['dust.tt'].
+   * A client is allowed only when every redirect URI it registered is on one of these domains
+   * or a subdomain of one. Omit to accept any dynamically registered client. Native desktop
+   * clients register loopback redirect URIs, so they are always rejected when this is set.
```
hercemer42: Is the last phrase important to mention ?  I'm sure we could shorten this comment to two lines.

packages/mcp-server/test/utils/parse-domain-list.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  it('should filter out empty entries from repeated or trailing commas', () => {
+    expect(parseDomainList('dust.tt,,claude.ai,')).toEqual(['dust.tt', 'claude.ai']);
+  });
+
+  it('should handle a single domain', () => {
+    expect(parseDomainList('dust.tt')).toEqual(['dust.tt']);
+  });
+});
```
hercemer42, line 38: What about sub domains ? like `sub.dust.tt` ?

packages/mcp-server/test/forest-oauth-provider.test.ts
```diff
@@ -1316,4 +1317,172 @@ describe('ForestOAuthProvider', () => {
...
+      const clientData = registerClient(['https://dust.tt/oauth/mcp_static/finalize']);
+      const provider = createRestrictedProvider(['dust.tt']);
+
+      const client = await provider.clientsStore.getClient(CLIENT_ID);
+
+      expect(client).toEqual(clientData);
+      expect(mockServer.fetch).toHaveBeenCalledTimes(1);
+    });
+
+    it('returns the client when redirect URIs are on subdomains of an allowed domain', async () => {
```
hercemer42, line 1371: Do we handle 4th (and greater) level domains ?

packages/mcp-server/README.md
```diff
@@ -8,18 +8,18 @@ This MCP server provides HTTP REST API access to Forest Admin operations, enabli
 
 ### Available Tools
 
-| Tool | Description |
-|------|-------------|
+| Tool                 | Description                                               |
+| -------------------- | --------------------------------------------------------- |
```
hercemer42: Remove the opportunistic formatting changes please

packages/mcp-server/README.md
```diff
@@ -30,9 +30,7 @@ The MCP server is included with the Forest Admin agent. Simply call `mountAiMcpS
 ```typescript
 import { createAgent } from '@forestadmin/agent';
 
-const agent = createAgent(options)
-  .addDataSource(myDataSource)
-  .mountAiMcpServer();
+const agent = createAgent(options).addDataSource(myDataSource).mountAiMcpServer();
```
hercemer42: Opportunistic, remove

packages/mcp-server/README.md
```diff
@@ -57,15 +55,16 @@ yarn start:dev       # Development (loads .env file automatically)
...
+| `FOREST_AGENT_URL`                     | No       | your environment's back-end URL | URL the MCP server uses to reach the back-end's data layer. Set it when the server runs next to a self-hosted back-end at an internal address (e.g. `http://localhost:3310`), instead of the public URL registered in Forest |
+| `FOREST_MCP_ACCESS_TOKEN_TTL_SECONDS`  | No       | `3600` (1 hour)                 | Maximum lifetime of the OAuth access tokens the server issues (`tokenTtl.accessTokenSeconds`). Minimum `60`                                                                                                                  |
+| `FOREST_MCP_REFRESH_TOKEN_TTL_SECONDS` | No       | unbounded                       | Maximum time between two interactive logins (`tokenTtl.refreshTokenSeconds`). Unset, a client that keeps refreshing never signs in again. Minimum `60`                                                                       |
```
hercemer42: Remove opportunistic formatting, make insert with original formatting.

packages/mcp-server/README.md
```diff
@@ -111,9 +110,32 @@ When `enabledTools` is not set, all tools are enabled by default.
...
+
+Native desktop clients (Claude Desktop, MCP Inspector, ...) register loopback (`localhost`) redirect URIs, so they are always rejected when the allowlist is set. There is deliberately no loopback exemption: allowing `localhost` would allow every local application. Omit the option in environments that need native clients (e.g. development).
+
 ## Shorten Token Lifetimes
 
-Forest grants **1 hour** (3600s) for an access token and **8 days** (691200s) for a refresh token — but it re-grants those 8 days on *every* refresh, so without `refreshTokenSeconds` a client that keeps working is never asked to sign in again.
+Forest grants **1 hour** (3600s) for an access token and **8 days** (691200s) for a refresh token — but it re-grants those 8 days on _every_ refresh, so without `refreshTokenSeconds` a client that keeps working is never asked to sign in again.
```
hercemer42: Remove opportunistic refactor

packages/mcp-server/README.md
```diff
@@ -142,12 +164,12 @@ The minimum for either value is 60 seconds; anything lower is raised to it. An i
...
-| POST | `/mcp` | Main MCP protocol endpoint (requires Bearer token) |
-| POST | `/oauth/authorize` | OAuth 2.0 authorization |
-| POST | `/oauth/token` | OAuth 2.0 token exchange |
-| GET | `/.well-known/oauth-protected-resource/mcp` | OAuth metadata discovery |
+| Method | Path                                        | Description                                        |
+| ------ | ------------------------------------------- | -------------------------------------------------- |
+| POST   | `/mcp`                                      | Main MCP protocol endpoint (requires Bearer token) |
+| POST   | `/oauth/authorize`                          | OAuth 2.0 authorization                            |
+| POST   | `/oauth/token`                              | OAuth 2.0 token exchange                           |
+| GET    | `/.well-known/oauth-protected-resource/mcp` | OAuth metadata discovery                           |
```
hercemer42: Undo

packages/mcp-server/README.md
```diff
@@ -194,10 +216,10 @@ yarn clean
...
 These are only needed by Forest Admin developers (e.g. to point to a local or staging server):
 
-| Variable | Default | Description |
-|----------|---------|-------------|
-| `FOREST_SERVER_URL` | `https://api.forestadmin.com` | Forest Admin API URL |
-| `FOREST_APP_URL` | `https://app.forestadmin.com` | Forest Admin application URL |
+| Variable            | Default                       | Description                  |
+| ------------------- | ----------------------------- | ---------------------------- |
+| `FOREST_SERVER_URL` | `https://api.forestadmin.com` | Forest Admin API URL         |
+| `FOREST_APP_URL`    | `https://app.forestadmin.com` | Forest Admin application URL |
```
hercemer42: Undo

packages/mcp-server/src/server.ts
hercemer42: Claude Fable 5 (claude-fable-5): Agreed — shortened to two lines in a6923851 and dropped the native-clients phrase; that detail lives in the README's "Restrict OAuth Clients" section where the trade-off is explained in full.

packages/mcp-server/test/utils/parse-domain-list.test.ts
hercemer42: Claude Fable 5 (claude-fable-5): The parser is a dumb splitter — entries pass through verbatim, so `eu.dust.tt` is a perfectly valid allowlist value (and matching then also covers *its* subdomains). The subdomain semantics live in the provider's matcher, not here. Added a verbatim pass-through test with `eu.dust.tt,mcp.front.eu.dust.tt` in a6923851 to pin that.

packages/mcp-server/test/forest-oauth-provider.test.ts
hercemer42: Claude Fable 5 (claude-fable-5): Yes — matching is `hostname === domain || hostname.endsWith('.' + domain)`, which is depth-agnostic: any label count left of an allowed domain matches. Extended this test with a 5-label URI (`mcp.front.eu.dust.tt`) in a6923851 to pin it.

packages/mcp-server/README.md
hercemer42: Claude Fable 5 (claude-fable-5): Done in a6923851 — the reformatting came from running prettier --write on the whole file; restored the original formatting throughout. The README diff is now insertions-only (+24): the env-table row and the "Restrict OAuth Clients" section, both in the file's original style.

packages/mcp-server/README.md
hercemer42: Claude Fable 5 (claude-fable-5): Done in a6923851 — original formatting restored.

packages/mcp-server/README.md
hercemer42: Claude Fable 5 (claude-fable-5): Done in a6923851 — original formatting restored.

packages/mcp-server/README.md
hercemer42: Claude Fable 5 (claude-fable-5): Done in a6923851 — original formatting restored.

packages/mcp-server/README.md
hercemer42: Claude Fable 5 (claude-fable-5): Done in a6923851 — original formatting restored.

packages/mcp-server/README.md
hercemer42: Claude Fable 5 (claude-fable-5): Done in a6923851 — original formatting restored.

packages/mcp-server/src/cli.ts
```diff
@@ -12,6 +13,7 @@ const server = new ForestMCPServer({
   envSecret: process.env.FOREST_ENV_SECRET,
   authSecret: process.env.FOREST_AUTH_SECRET,
   enabledTools: parseToolList(process.env.FOREST_MCP_ENABLED_TOOLS),
+  allowedOAuthClients: parseDomainList(process.env.FOREST_MCP_ALLOWED_OAUTH_CLIENTS),
```
Tonours, line 16: Blocker, worth a decision. `parseDomainList` returns `undefined` for a set-but-unparseable value, e.g. `FOREST_MCP_ALLOWED_OAUTH_CLIENTS=" , ,"` from broken quoting in compose or k8s. `undefined` means "not configured", so every client is accepted. Misconfigured env fails open. The programmatic API fails closed (constructor throws on `[]`). Wrong direction for a security allowlist. Reproduced e2e: with `" , ,"` set, an off-domain client got a token. Fix: throw at startup when the var is set but parses to zero domains, like the constructor guard.

packages/mcp-server/src/utils/parse-domain-list.ts
```diff
@@ -0,0 +1,11 @@
+export default function parseDomainList(envValue?: string): string[] | undefined {
+  if (!envValue) return undefined;
+
+  const domains = envValue
+    .split(',')
+    .map(domain => domain.trim())
+    .filter(Boolean);
+
+  // A variable containing no domains means "not configured", never "block everyone".
+  return domains.length > 0 ? domains : undefined;
```
Tonours: Allowlist entries aren't validated. A typo like `https://dust.tt` or `dust.tt:443` matches no hostname, so every client is rejected with no startup signal. Fail-closed footgun, found only when the integration breaks. Same for a unicode entry like `bücher.de`, since `URL` yields punycode. Suggest rejecting entries with `/`, `:` or whitespace here. Reproduced e2e: `https://dust.tt` rejected the legit `eu.dust.tt` client, no warning logged.

packages/mcp-server/src/server.ts
```diff
@@ -186,6 +192,15 @@ export default class ForestMCPServer {
     this.agentDispatcher = options?.agentDispatcher;
     this.tokenTtl = normalizeTokenTtl(options?.tokenTtl, this.logger);
 
+    if (options?.allowedOAuthClients && options.allowedOAuthClients.length === 0) {
```
Tonours: Guard checks array length, not content. `allowedOAuthClients: ["  "]` passes, then matches no hostname, so it's a silent deny-all, the exact footgun this guard prevents. Both shipped callers filter blanks upstream, so it's only reachable from a custom caller. Worth a `.map(trim).filter(Boolean)` here to close it?

packages/mcp-server/src/cli.ts
hercemer42: Claude Fable 5 (claude-fable-5): Agreed on the direction — you're right that my "a value with no domains means not configured" parser comment was the wrong call for a security control, and your e2e repro settled the decision. Fixed in 23eac8b4: `parseDomainList` now returns an empty list (not `undefined`) for any set, non-blank value that yields no domains, and the server's startup normalization throws on it — `FOREST_MCP_ALLOWED_OAUTH_CLIENTS=" , ,"` now crashes the boot with an explicit message instead of accepting every client (re-ran your repro against the built CLI). Unset or empty-string still means "not configured". The example agent's inline parsing got the same fail-closed treatment.

packages/mcp-server/src/utils/parse-domain-list.ts
hercemer42: Claude Fable 5 (claude-fable-5): Agreed — fixed in 23eac8b4. Entries are now validated at startup by a shared `normalizeDomainList`: anything containing a scheme, port, path, or whitespace throws with a message naming the offending entry (`Invalid allowedOAuthClients entry "https://dust.tt": list bare domains…`), and unicode entries are punycode-normalized via `new URL('https://'+entry).hostname` so they match what the provider extracts from redirect URIs. Your `https://dust.tt` repro now fails the boot instead of silently rejecting the legit client.

packages/mcp-server/src/server.ts
hercemer42: Claude Fable 5 (claude-fable-5): Agreed, and thanks for flagging the reachability honestly. Fixed in 23eac8b4: the constructor guard is replaced by `normalizeDomainList`, which trims entries and drops blanks before the emptiness check — `['  ']` now throws the same explicit startup error as `[]`. Covered by a construction test.

packages/mcp-server/src/forest-oauth-provider.ts
```diff
@@ -165,12 +169,63 @@ export default class ForestOAuthProvider implements OAuthServerProvider {
...
+      ({ hostname, protocol } = new URL(redirectUri));
+    } catch {
+      return false;
+    }
+
+    // A custom scheme (attacker-app://dust.tt/cb) dispatches the callback to whatever
+    // application registered it, so its hostname says nothing about where the code lands.
+    if (protocol !== 'https:' && protocol !== 'http:') return false;
+
+    return allowedDomains.some(domain => {
```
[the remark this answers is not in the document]
hercemer42: Claude Fable 5 (claude-fable-5): Confirmed and fixed in 23eac8b4, with one severity note. The bypass is real — a literal '' entry turns the suffix check into `endsWith('.')`, which any trailing-dot FQDN satisfies — but reaching it requires a malformed *programmatic* config (the env parser filters empties) plus an attacker registering a trailing-dot redirect URI, so it's more a latent footgun than an immediately exploitable High. Fixed at the root regardless: startup normalization (`normalizeDomainList`) trims and drops blank entries before matching, and throws if nothing survives, so '' can never reach the matcher from either entry point.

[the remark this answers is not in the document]
hercemer42: Claude Fable 5 (claude-fable-5): Confirmed and fixed in 23eac8b4 — note the fix lands at construction rather than at this line: configured entries are now canonicalized once at startup via `new URL('https://'+entry).hostname` (lowercase punycode), the same form the matcher extracts from redirect URIs, so `allowedOAuthClients: ['bücher.de']` matches `https://bücher.de/callback` (hostname `xn--bcher-kva.de`). Covered by a normalization unit test and a supertest through the real express app. One calibration: this failure rejected legit clients (fail-closed availability), it never admitted a wrong one.

packages/mcp-server/src/utils/normalize-domain-list.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  if (entries.length === 0) {
+    throw new Error(
+      'Invalid allowedOAuthClients: no domains to allow. List at least one domain ' +
+        '(e.g. dust.tt), or omit the option to accept any registered client.',
+    );
+  }
+
+  return entries.map(entry => {
+    if (/[/:\s]/.test(entry)) {
```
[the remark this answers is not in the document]
hercemer42: Claude Fable 5 (claude-fable-5): Confirmed and fixed in ad46ed3b — good catch, this was a real hole in the validation I added: `approved.example@attacker.example` passed the old regex and URL parsed the lead as userinfo, normalizing to the attacker's hostname. The regex now rejects `@`, `?`, and `#` alongside `/`, `:`, and whitespace, with a test pinning the exact userinfo example (plus `?`/`#` variants, which normalized to the *right* domain but deserve a loud failure rather than silent truncation).

packages/mcp-server/README.md
```diff
@@ -111,6 +112,31 @@ When `enabledTools` is not set, all tools are enabled by default.
...
+
+List bare domains only — unicode domains are matched through their punycode form. An entry with a scheme, port, path, or spaces fails at startup, as does a configured value containing no domains at all: the allowlist never silently falls back to accepting or rejecting everyone on a malformed configuration.
+
+Every other client is rejected with a standard `invalid_client` error telling the user to contact their administrator; the response does not reveal the allowed domains. Registration itself still succeeds — it happens on the Forest Admin server — the client just cannot use it against this server. Access tokens issued before you enabled the option stay valid until they expire (1 hour at most); refreshes are blocked immediately.
+
+Native desktop clients (Claude Desktop, MCP Inspector, ...) register loopback (`localhost`) redirect URIs, so they are always rejected when the allowlist is set. There is deliberately no loopback exemption: allowing `localhost` would allow every local application. Omit the option in environments that need native clients (e.g. development).
```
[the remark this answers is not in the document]
hercemer42: Claude Fable 5 (claude-fable-5): Fair on the wording — fixed in ad46ed3b. The README now says loopback URIs "are rejected unless you explicitly list `localhost`, which would admit every local application and defeats the restriction", instead of "always rejected". Pushing back on the implied behavior change though: rejecting `localhost` as an *entry* would mean second-guessing explicit admin configuration, and we deliberately don't (same call as allowing public-suffix entries like `co.uk` — explicit config wins). The guarantee that matters — native clients never pass a *domain* allowlist — still holds; severity here is documentation precision, since exploiting it requires the administrator to configure it on purpose.

packages/_example/src/forest/agent.ts
```diff
@@ -36,6 +36,15 @@ export default function makeAgent() {
...
   };
 
+  const rawAllowedOAuthClients = process.env.FOREST_MCP_ALLOWED_OAUTH_CLIENTS;
+  // A set-but-blank value stays an empty list so the agent fails closed at startup.
+  const allowedOAuthClients = rawAllowedOAuthClients
+    ? rawAllowedOAuthClients
+        .split(',')
+        .map(domain => domain.trim())
+        .filter(Boolean)
+    : undefined;
```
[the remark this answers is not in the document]
hercemer42: Claude Fable 5 (claude-fable-5): Partial pushback — the behavior is deliberate and specced; the comment above it was misleading, and that's what ad46ed3b fixes. `''` (e.g. compose's `FOO=`) means "not configured" by design: the acceptance criteria say "an unset or empty variable leaves behaviour unchanged", and the standalone CLI's `parseDomainList` treats `''` identically (`if (!envValue) return undefined`), so distinguishing `''` from unset here would make the two entry points diverge. The fail-closed case is a *set, non-blank* value that parses to zero domains (`" , ,"`), which this code does fail closed on. The comment now spells out both halves. Given the empty-string case is symmetric across entry points and specced, I'd rate this one below High.

packages/mcp-server/src/utils/normalize-domain-list.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (entries.length === 0) {
+    throw new Error(
+      'Invalid allowedOAuthClients: no domains to allow. List at least one domain ' +
+        '(e.g. dust.tt), or omit the option to accept any registered client.',
+    );
+  }
+
+  return entries.map(entry => {
+    // '@' would make URL parse everything before it as userinfo and return the wrong hostname.
+    if (/[/:@?#\s]/.test(entry)) {
```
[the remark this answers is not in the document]
hercemer42: Claude Fable 5 (claude-fable-5): Confirmed and fixed in 63446455 — `\` is indeed a path separator in WHATWG URL parsing for special schemes, so `dust.tt\unexpected` silently truncated to `dust.tt` instead of failing like the other delimiters. Added `\` to the rejected set with a test on your exact example. (Your suggested diff had a stray `((`, applied the intent.)

packages/mcp-server/src/utils/normalize-domain-list.ts
```diff
@@ -0,0 +1,31 @@
...
+  const entries = domains.map(domain => domain.trim()).filter(Boolean);
+
+  if (entries.length === 0) {
+    throw new Error(
+      'Invalid allowedOAuthClients: no domains to allow. List at least one domain ' +
+        '(e.g. dust.tt), or omit the option to accept any registered client.',
+    );
+  }
+
+  return entries.map(entry => {
```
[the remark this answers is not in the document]
hercemer42: Claude Fable 5 (claude-fable-5): Confirmed and fixed — `URL.hostname` does preserve the terminal dot, so an FQDN-style entry could never match a real redirect hostname. The root dot is now stripped after canonicalization, with a test on `dust.tt.`. Note the redirect-URI side deliberately keeps its behavior: an attacker-registered trailing-dot URI (`https://dust.tt./cb`) still fails to match, which is the safe direction. Severity-wise this one is a config-typo footgun (fail-closed, silent), not a High — but at one line it wasn't worth arguing.

Tonours (approved):

PR #1798, opened 2026-08-04
fix(ai-proxy): bound mcp tool loading with a per-server 15s timeout

hercemer42:
`[fable-5]`

## Problem

`McpClient.loadToolsWithFailures` awaits each server's `getTools()` inside an unbounded `Promise.all`. A server that rejects is handled, but one that hangs pending (e.g. an SSE endpoint that accepts the connection and never responds) never settles — the whole `ai-query` request hangs until the 60s ALB idle timeout and returns a 504, also blocking the healthy MCP servers and integration tools (Zendesk/Kolar/Snowflake) loaded in the same request. Datadog shows this recurring in prod (dozens of 504s/day pegged at exactly 60.00s).

## Fix

Race each server's `getTools()` against a 15s per-server timeout inside the existing try/catch. A timed-out server rejects with `McpLoadTimeoutError` (message deliberately contains "Timeout" so `classifyMcpLoadError` files it as a `connection` failure), lands in the existing `failures` channel, and the rest of the request proceeds. The timer is cleared as soon as the server answers. Also adds a per-server Debug timing log so the timeout value can be tuned on real per-phase data.

**Why 15s**: prod ALB logs (21 days, ~94k successful ai-query 200s) show total request time p50 2.2s / p99 9.5s *including LLM generation* — only 0.42% of successful requests exceed 15s total, a hard upper bound on what the timeout could clip, and 15s leaves 45s of the 60s ALB budget for the model call.

## Notes for review

- Known minor: the abandoned hung connect is not aborted (mcp-adapters only registers connections after `connect()` resolves, so `close()` skips it) — leaks one pending socket per timed-out attempt, bounded by undici's ~300s default header/body timeouts.
- Out of scope (same hang class, separate surfaces): `checkConnection()` used by tool-config validation, and mid-call `invoke-remote-tool` hangs.

fixes PRD-863

> [!NOTE]
> ### Bound MCP tool loading with a 15s per-server timeout in `McpClient.loadToolsWithFailures`
> - Wraps each server's `getTools()` call in a `loadWithTimeout` helper that races against a 15s timeout (`LOAD_TOOLS_TIMEOUT_MS`), rejecting with a new `McpLoadTimeoutError` on expiry.
> - Timed-out servers are classified as `'connection'` failures via `classifyMcpLoadError` and included in the failures list, while responsive servers still return their tools normally.
> - `McpLoadTimeoutError` uses an instanceof check in classification to prevent user-controlled server names from influencing the failure kind via text matching.
> - Behavioral Change: `loadToolsWithFailures` no longer hangs indefinitely when a server accepts a connection but stops responding — it now returns after 15s with a per-server failure.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized b3b88e3.</sup>
> 
>


packages/ai-proxy/src/errors.ts
```diff
@@ -139,6 +139,15 @@ export class McpConnectionError extends McpError {
   }
 }
 
+export class McpLoadTimeoutError extends McpError {
+  constructor(serverName: string, timeoutMs: number) {
+    // 'Timeout' must appear in the message so classifyMcpLoadError files this as a connection
+    // failure rather than an unknown one.
+    super(`Timeout: MCP server "${serverName}" did not return its tools within ${timeoutMs}ms`);
```
hercemer42, line 144: Claude Fable 5 (claude-fable-5): **Should fix** — `classifyMcpLoadError` sees only the message text, and this message interpolates the user-controlled server name. A server named e.g. `unauthorized-api` makes `AUTH_PATTERN` (`/\b401\b|unauthorized/i`) match first, so the timeout is classified `auth` — and workflow-executor's remote-tool-fetcher then does a token refresh plus a second full 15s load. Conversely, a future `CONNECTION_PATTERN` copy tweak silently degrades timeouts to `unknown`. Classify structurally instead: `if (error instanceof McpLoadTimeoutError) return 'connection'` at the top of `classifyMcpLoadError` (the message can stay as-is for logs).

packages/ai-proxy/test/mcp-client.test.ts
```diff
@@ -562,6 +562,149 @@ describe('McpClient.loadToolsWithFailures', () => {
...
+      expect.objectContaining({ server: 'github', kind: 'connection' }),
+    ]);
+    expect(logger).toHaveBeenCalledWith(
+      'Error',
+      expect.stringContaining('Failed to load tools from 2/2'),
+    );
+  });
+
+  it('keeps a server that answers just under the timeout and leaves no timer behind', async () => {
+    getToolsMock.mockImplementation(
```
hercemer42, line 679: Claude Fable 5 (claude-fable-5): **Should fix** — this persistent `mockImplementation` (and the `mockReturnValue(neverSettles())` at L634) survive `jest.clearAllMocks()`, which clears calls but not implementations, and the repo jest config sets neither `resetMocks` nor `restoreMocks`. Verified empirically: a test appended after this describe with no queued `*Once` value inherits the 14 999 ms `setTimeout` under real timers and blows jest's 5s default timeout (suite time 3.5s → 15.1s). Nothing fails today — every current test queues its own values — but it's a trap for the next test added to this file. Use `getToolsMock.mockReset()` in this describe's `beforeEach`, or switch to `mockImplementationOnce`/`mockReturnValueOnce`.

packages/ai-proxy/src/mcp-client.ts
```diff
@@ -3,10 +3,26 @@ import type { Logger } from '@forestadmin/datasource-toolkit';
...
 import { MultiServerMCPClient } from '@langchain/mcp-adapters';
 
-import { McpConnectionError } from './errors';
+import { McpConnectionError, McpLoadTimeoutError } from './errors';
 import { type McpLoadFailureKind, classifyMcpLoadError } from './mcp-auth-error';
 import McpServerRemoteTool from './mcp-server-remote-tool';
 
+// A server that accepts the connection but never answers would otherwise keep getTools() pending
+// forever and hang the whole load (and the caller's request) until the infrastructure cuts it.
+const LOAD_TOOLS_TIMEOUT_MS = 15_000;
```
hercemer42: Claude Fable 5 (claude-fable-5): **Preferential** — `LOAD_TOOLS_TIMEOUT_MS` isn't exported, so the test keeps its own `15_000` copy. Drift can't ship silently (the hung-server tests still fail), but it fails as an opaque jest timeout rather than a clear assertion. Exporting the constant and importing it in the test makes the failure mode legible.

packages/ai-proxy/src/mcp-client.ts
```diff
@@ -3,10 +3,26 @@ import type { Logger } from '@forestadmin/datasource-toolkit';
...
+function loadWithTimeout<T>(loading: Promise<T>, serverName: string): Promise<T> {
+  let timer: NodeJS.Timeout | undefined;
+  const timeout = new Promise<never>((_, reject) => {
+    timer = setTimeout(
+      () => reject(new McpLoadTimeoutError(serverName, LOAD_TOOLS_TIMEOUT_MS)),
+      LOAD_TOOLS_TIMEOUT_MS,
+    );
+  });
+
+  return Promise.race([loading, timeout]).finally(() => clearTimeout(timer));
```
hercemer42, line 37: Claude Fable 5 (claude-fable-5): **Preferential** — once the timeout wins the race, the losing `getTools()` promise's eventual settlement is discarded: a server that answers (or rejects with the real root cause, e.g. a late 401) after 15s leaves no trace. Threading the logger into `loadWithTimeout` and attaching `loading.catch(err => logger?.('Debug', `MCP server "${name}" settled after timeout`, err))` would make the true cause of a timed-out server diagnosable in production.

hercemer42:
Claude Fable 5 (claude-fable-5): **Review resolution** — all 4 findings from the pending validator review are accepted and addressed in 00339efd. (The review is pending, so this is a regular comment rather than threaded replies.)

- **Should fix — structural classification** (`errors.ts`): `classifyMcpLoadError` now matches `McpLoadTimeoutError` by instance before any text pattern runs. Honest caveat: the misclassification needed a server literally named with 'unauthorized'/'401' in it — low probability — but the structural check costs nothing and also removes the hazard of a future `CONNECTION_PATTERN` edit silently degrading timeouts to `unknown`. New test pins `classifyMcpLoadError(new McpLoadTimeoutError('unauthorized-api', …)) === 'connection'`.
- **Should fix — mock leakage** (`mcp-client.test.ts`): `getToolsMock.mockReset()` added to the timeout describe's `beforeEach` — `clearAllMocks()` keeps implementations, and the leak was empirically confirmed (a probe test with nothing queued inherited the 14 999 ms delay).
- **Preferential — shared constant**: `LOAD_TOOLS_TIMEOUT_MS` is now exported and imported by the test; no more private `15_000` duplicate.
- **Preferential — late-settlement tracing**: the losing promise's eventual outcome is logged at Debug ("returned its tools / failed after the timeout"), with a test covering the late-rejection path.

No pushback warranted: nothing in the filed review presents intentional behaviour as a defect — the candidates that did (15s clipping slow cold-start servers; `checkConnection()`'s unbounded await) were already excluded at the validator's skeptic gate as ticket-documented scope decisions.

Checks after resolution: 63 tests green across the three touched suites, eslint 0 errors, tsc build clean.

hercemer42 (commented):
## Validator review — PR #1798 (PRD-863: per-server 15s timeout on MCP tool loading)
4 findings — Should fix ×2, Preferential ×2. No Must fix. · 4 files · pending.

**Spec (PRD-863):** conforms
**Notes:**
- No ADRs exist under `docs/adr/**` org-wide for the touched areas.
- Registry `docs` lists are empty for areas `agent` and `testing`, so docs coverage was limited to a product-term search (no contradiction found).

packages/ai-proxy/src/errors.ts
hercemer42: Claude Fable 5 (claude-fable-5): Fixed in 00339efd — `classifyMcpLoadError` now matches `McpLoadTimeoutError` by instance before any text pattern runs; new test pins `unauthorized-api` → `connection`.

packages/ai-proxy/test/mcp-client.test.ts
hercemer42: Claude Fable 5 (claude-fable-5): Fixed in 00339efd — `getToolsMock.mockReset()` added to this describe's `beforeEach` (clearAllMocks keeps implementations).

packages/ai-proxy/src/mcp-client.ts
hercemer42: Claude Fable 5 (claude-fable-5): Fixed in 00339efd — `LOAD_TOOLS_TIMEOUT_MS` is exported and the test imports it; the private `15_000` duplicate is gone.

packages/ai-proxy/src/mcp-client.ts
hercemer42: Claude Fable 5 (claude-fable-5): Fixed in 00339efd — logger threaded into `loadWithTimeout`; the losing promise's late outcome is logged at Debug (guarded by a hasTimedOut flag so pre-timeout rejections aren't mislabeled), with a test for the late-rejection path.

packages/ai-proxy/src/errors.ts
```diff
@@ -139,6 +139,14 @@ export class McpConnectionError extends McpError {
   }
 }
 
+export class McpLoadTimeoutError extends McpError {
+  // classifyMcpLoadError matches this class structurally, so the message is free-form.
```
hercemer42: What's the point of this comment ?

packages/ai-proxy/src/mcp-auth-error.ts
```diff
@@ -51,6 +53,10 @@ export function isMcpAuthError(error: unknown): boolean {
 }
 
 export function classifyMcpLoadError(error: unknown): McpLoadFailureKind {
+  // Structural check first: the timeout message interpolates the user-controlled server name,
+  // which could otherwise trip the auth/connection text patterns below.
```
hercemer42: Isn't that stating the obvious ?

packages/ai-proxy/src/mcp-client.ts
```diff
@@ -3,10 +3,42 @@ import type { Logger } from '@forestadmin/datasource-toolkit';
...
+  let hasTimedOut = false;
+  const timeout = new Promise<never>((_, reject) => {
+    timer = setTimeout(() => {
+      hasTimedOut = true;
+      reject(new McpLoadTimeoutError(serverName, LOAD_TOOLS_TIMEOUT_MS));
+    }, LOAD_TOOLS_TIMEOUT_MS);
+  });
+
+  // The race abandons the losing promise; trace its eventual outcome so the real cause of a
+  // timed-out server stays diagnosable.
```
hercemer42: Isn't that obvious ?

packages/ai-proxy/src/errors.ts
hercemer42: Claude Fable 5 (claude-fable-5): Agreed — it documented a non-constraint (the message being free-form); the classifier side carries the real information. Removed in 0c932b39.

packages/ai-proxy/src/mcp-client.ts
hercemer42: Claude Fable 5 (claude-fable-5): Agreed — the Debug log strings below already say exactly this. Removed in 0c932b39.

packages/ai-proxy/src/mcp-auth-error.ts
hercemer42: Claude Fable 5 (claude-fable-5): Partial pushback: the `instanceof` check is obvious, but the *ordering* is load-bearing and that is not — if this line moves below `isMcpAuthError`, a server named e.g. `unauthorized-api` classifies as a refreshable auth failure again (the test pins it, but only a failing CI run would tell the refactorer why). Trimmed to a single line keeping only the ordering constraint in 0c932b39; happy to drop it entirely if you still think the test is enough.

Tonours (approved):

PR #1801, opened 2026-08-04
feat(agent-bff): validate relation filters against the foreign collection

Tonours:
fixes PRD-677

## Problem

Relation `list` and `count` requests were forwarded without validating filters, projections, and sorts against the capabilities of the foreign collection.

## Fix

- Validate relation list filters, projections, and sorts against the foreign collection.
- Validate relation count filters against the foreign collection.
- Re-check the parent collection, relation target, and foreign collection after capability reads.
- Return a local `404` when a foreign collection disappears during a failed capability read.
- Preserve the `filter_too_deep` error type and `maxDepth` details for deeply nested filters.
- Add regression coverage for list/count validation, schema refreshes, stale targets, and error contracts.

## Scope and safety

- Changes are limited to `packages/agent-bff`.
- No change to `agent`, `agent-client`, or MCP packages.
- Requests without constrained inputs still skip the capabilities fetch.
- The agent remains the authority for access control.

## How to test

```bash
```

Verified: 733 tests passing across 54 suites.

### General

### Security

### DIFF
--- a/packages/agent-bff/src/data/data-routes-middleware.ts
+++ b/packages/agent-bff/src/data/data-routes-middleware.ts
@@ -151,29 +151,44 @@ interface RelationHandlerDeps extends RequestHandlerDeps {
 
 type RelationListHandlerDeps = RelationHandlerDeps & { primaryKeys: PrimaryKeyField[] };
 
-// The route resolved parent, relation and foreign collection against an earlier generation; a schema
-// refresh during the capabilities fetch can invalidate any of the three, so all three are re-checked
-// against the generation the capabilities belong to. The foreign name must come out unchanged: the
-// capabilities are cached under it, the agent query projects under it and the response is stamped
-// with it, so a re-targeted relation would validate one collection and serve another.
-async function resolveRelationCapabilities(
-  deps: RelationHandlerDeps,
-): Promise<{ capabilities: CapabilitiesResult; primaryKeys: PrimaryKeyField[] }> {
-  const { capabilities, readModel } = await resolveCapabilities(deps, deps.foreignCollection);
-
+function assertRelationStillExposed(readModel: ReadModel, deps: RelationHandlerDeps): void {
   assertCollectionStillAllowed(readModel, deps.collection);
 
-  const foreignCollection = resolveForeignCollection(
+  const stillTargets = resolveForeignCollection(
     readModel.getRelationTarget(deps.collection, deps.relation),
   );
 
-  if (foreignCollection !== deps.foreignCollection) {
+  if (stillTargets !== deps.foreignCollection) {
     throw unknownRelation(`Unknown relation: ${deps.collection}.${deps.relation}`);
   }
 
-  assertCollectionStillAllowed(readModel, foreignCollection);
+  assertCollectionStillAllowed(readModel, deps.foreignCollection);
+}
 
-  return { capabilities, primaryKeys: readModel.getPrimaryKeys(foreignCollection) };
+async function resolveForeignCapabilitiesAndReassertRelationIsStillExposed(
+  deps: RelationHandlerDeps,
+): Promise<{ capabilities: CapabilitiesResult; readModel: ReadModel }> {
+  assertRelationStillExposed(await resolveReadModel(deps.store), deps);
+
+  let result: { capabilities: CapabilitiesResult; readModel: ReadModel };
+
+  try {
+    result = await resolveCapabilities(deps, deps.foreignCollection);
+  } catch (error) {
+    deps.logger('Warn', 'Foreign capabilities lookup failed; re-checking relation exposure', {
+      collection: deps.collection,
+      relation: deps.relation,
+      foreignCollection: deps.foreignCollection,
+      cause: error instanceof Error ? `${error.name}: ${error.message}` : String(error),
+    });
+
+    assertRelationStillExposed(await resolveReadModel(deps.store), deps);
+    throw error;
+  }
+
+  assertRelationStillExposed(result.readModel, deps);
+
+  return result;
 }
 
 async function handleRelationList(
@@ -181,9 +196,6 @@ async function handleRelationList(
   body: RelationListRequestBody,
   deps: RelationListHandlerDeps,
 ) {
-  // The nested-relation guard IS wired here: the agent's list-related asserts browse only on the
-  // immediate foreign collection, so a nested `:`-path would traverse to a third collection whose
-  // browse is never checked. Plain foreign fields (no `:`) are unaffected.
   assertNoRelationFieldPaths(collectListFieldPaths(body));
 
   const validationInput = {
@@ -195,9 +207,10 @@ async function handleRelationList(
   let { primaryKeys } = deps;
 
   if (hasCapabilityConstrainedInput(validationInput)) {
-    const resolved = await resolveRelationCapabilities(deps);
-    assertValidAgainstCapabilities(validationInput, resolved.capabilities);
-    primaryKeys = resolved.primaryKeys;
+    const { capabilities, readModel } =
+      await resolveForeignCapabilitiesAndReassertRelationIsStillExposed(deps);
+    assertValidAgainstCapabilities(validationInput, capabilities);
+    primaryKeys = readModel.getPrimaryKeys(deps.foreignCollection);
   }
 
   const query = buildListAgentQuery(deps.foreignCollection, deps.timezone, body);
@@ -218,7 +231,9 @@ async function handleRelationCount(
   assertNoRelationFieldPaths(collectCountFieldPaths(body));
 
   if (body.filter !== undefined) {
-    const { capabilities } = await resolveRelationCapabilities(deps);
+    const { capabilities } = await resolveForeignCapabilitiesAndReassertRelationIsStillExposed(
+      deps,
+    );
     assertValidAgainstCapabilities({ filter: body.filter }, capabilities);
   }
 
--- a/packages/agent-bff/src/data/data-routes-middleware.ts
+++ b/packages/agent-bff/src/data/data-routes-middleware.ts
@@ -79,6 +79,14 @@ function assertCollectionStillAllowed(readModel: ReadModel, collection: string):
   if (!readModel.isCollectionAllowed(collection)) throw unknownCollection();
 }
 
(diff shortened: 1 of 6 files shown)


packages/agent-bff/src/data/data-routes-middleware.ts
```diff
@@ -150,24 +151,76 @@ interface RelationHandlerDeps extends RequestHandlerDeps {
...
+    throw unknownRelation(`Unknown relation: ${deps.collection}.${deps.relation}`);
+  }
+
+  assertCollectionStillAllowed(readModel, deps.foreignCollection);
+}
+
+async function resolveForeignCapabilitiesAndReassertRelationIsStillExposed(
+  deps: RelationHandlerDeps,
+): Promise<{ capabilities: CapabilitiesResult; readModel: ReadModel }> {
+  assertRelationStillExposed(await resolveReadModel(deps.store), deps);
```
nbouliol: handleRelation already resolved the target and the allow-list a few lines earlier — is this pre-flight pass needed on top of the post-fetch one?

packages/agent-bff/src/data/data-routes-middleware.ts
```diff
@@ -150,24 +151,76 @@ interface RelationHandlerDeps extends RequestHandlerDeps {
...
+    result = await resolveCapabilities(deps, deps.foreignCollection);
+  } catch (error) {
+    deps.logger('Warn', 'Foreign capabilities lookup failed; re-checking relation exposure', {
+      collection: deps.collection,
+      relation: deps.relation,
+      foreignCollection: deps.foreignCollection,
+      cause: error instanceof Error ? `${error.name}: ${error.message}` : String(error),
+    });
+
+    assertRelationStillExposed(await resolveReadModel(deps.store), deps);
```
nbouliol: resolveReadModel here can throw too, swapping the original error for a schema one — worth guarding?

packages/agent-bff/test/data/data-routes-middleware.test.ts
```diff
@@ -991,4 +993,418 @@ describe('data routes middleware', () => {
...
+        };
+        const refreshedRelationReadModel = new ReadModel([
+          collection('users', [column('id'), relation('posts', 'HasMany', 'posts.id')]),
+        ]);
+        let readModelReads = 0;
+        const refreshedStore = {
+          getReadModel: async () => {
+            readModelReads += 1;
+
+            return readModelReads < 3 ? relationReadModel : refreshedRelationReadModel;
```
nbouliol: The 3 pins the internal call count — could the stub switch on what it's asked for rather than how many times?

packages/agent-bff/src/data/data-routes-middleware.ts
```diff
@@ -150,24 +151,76 @@ interface RelationHandlerDeps extends RequestHandlerDeps {
...
   ctx: Context,
   body: RelationListRequestBody,
   deps: RelationListHandlerDeps,
 ) {
-  // The nested-relation guard IS wired here: the agent's list-related asserts browse only on the
-  // immediate foreign collection, so a nested `:`-path would traverse to a third collection whose
-  // browse is never checked. Plain foreign fields (no `:`) are unaffected.
   assertNoRelationFieldPaths(collectListFieldPaths(body));
 
+  const validationInput = {
```
nbouliol: Same block as handleList:103 — extract a toValidationInput(body)?

packages/agent-bff/src/data/data-routes-middleware.ts
```diff
@@ -150,24 +151,76 @@ interface RelationHandlerDeps extends RequestHandlerDeps {
...
+  );
+
+  if (stillTargets !== deps.foreignCollection) {
+    throw unknownRelation(`Unknown relation: ${deps.collection}.${deps.relation}`);
+  }
+
+  assertCollectionStillAllowed(readModel, deps.foreignCollection);
+}
+
+async function resolveForeignCapabilitiesAndReassertRelationIsStillExposed(
```
nbouliol: The And in the name is doing a lot of work here — split the fetch from the re-checks?

packages/agent-bff/src/data/data-routes-middleware.ts
Tonours: I tried removing it and five tests went red, so it stays. The difference is which read model each check runs against: handleRelation checks the one resolved once at middleware entry, the pre-flight re-reads the store, and a schema refresh can land between the two. Its payoff is asserted by the existing tests — relation already gone means 404 with `getCapabilities` never called. Without it we still 404, but only after a wasted capabilities round-trip to the agent. Your question did shake out two real problems though, see the other two threads.

packages/agent-bff/src/data/data-routes-middleware.ts
Tonours: Fixed, extracted into `rethrowUnlessRelationVanished`. If the schema re-read itself fails, the original capabilities error is rethrown instead of being swapped for a schema one; if the re-read succeeds and the relation really vanished, the 404 still wins. Pinned by a test where both the capabilities fetch and the re-read fail, asserting `agent_unavailable` rather than `schema_unavailable`.

packages/agent-bff/test/data/data-routes-middleware.test.ts
Tonours: Done. The stub now flips a `foreignCollectionDropped` flag inside `getCapabilities` itself, so it models the event rather than a call index. Adding or removing a read-model read no longer breaks the fixture.

packages/agent-bff/src/data/data-routes-middleware.ts
Tonours: Extracted as `toValidationInput(body)`, used by both `handleList` and `handleRelationList`.

packages/agent-bff/src/data/data-routes-middleware.ts
Tonours: Renamed to `resolveForeignCapabilities`, and the error-path re-check moved out into `rethrowUnlessRelationVanished` — that inline block was what the `And` was covering for. The pre-flight and post-fetch asserts stay in the function: both callers need the same bundle, and the pre-flight turned out to be load-bearing (see the first thread). Happy to rename again if you would rather the name carried the asserting contract, e.g. `resolveExposedRelationCapabilities`.

packages/agent-bff/src/data/data-routes-middleware.ts
```diff
@@ -150,24 +155,87 @@ interface RelationHandlerDeps extends RequestHandlerDeps {
...
+  let result: { capabilities: CapabilitiesResult; readModel: ReadModel };
+
+  try {
+    result = await resolveCapabilities(deps, deps.foreignCollection);
+  } catch (error) {
+    deps.logger('Warn', 'Foreign capabilities lookup failed; re-checking relation exposure', {
+      collection: deps.collection,
+      relation: deps.relation,
+      foreignCollection: deps.foreignCollection,
+      cause: error instanceof Error ? `${error.name}: ${error.message}` : String(error),
```
nbouliol: `cause` here is always the already-mapped `BffHttpError` — `callAgent` flattens any 5xx to a fixed "The agent is unavailable", so the real reason is gone. Log it inside `resolveCapabilities` instead?

packages/agent-bff/src/data/data-routes-middleware.ts
```diff
@@ -150,24 +155,87 @@ interface RelationHandlerDeps extends RequestHandlerDeps {
...
+  } catch {
+    throw original;
+  }
+
+  assertRelationStillExposed(readModel, deps);
+
+  throw original;
+}
+
+async function resolveForeignCapabilities(
```
nbouliol: Yes to `resolveExposedRelationCapabilities` — it asserts twice and reads the store, the current name says it only fetches.

packages/agent-bff/src/data/data-routes-middleware.ts
```diff
@@ -150,24 +155,87 @@ interface RelationHandlerDeps extends RequestHandlerDeps {
...
+  );
+
+  if (stillTargets !== deps.foreignCollection) {
+    throw unknownRelation(`Unknown relation: ${deps.collection}.${deps.relation}`);
+  }
+
+  assertCollectionStillAllowed(readModel, deps.foreignCollection);
+}
+
+async function rethrowUnlessRelationVanished(
```
nbouliol: This 404-on-vanish only exists for relations — the same race on `/list` and `/count` still returns 503 `agent_unavailable`. Asymmetry intentional?

packages/agent-bff/src/data/data-routes-middleware.ts
Tonours: Right — `callAgent` already flattened it, so that `cause` was always the string `BffHttpError: The agent is unavailable`. A test was even pinning that value, which made the log look useful while carrying nothing.

Moved into the mapper: the 5xx branch now logs the agent detail (`errors[0].detail`, else `responseText`) plus the upstream status right where the message is replaced by the generic one. Every caller gets it — `/list`, `/count` and the action routes had the same blind spot, not just relations. The relation-side log is down to which collection is being re-checked. The test now asserts `cause: "Service Unavailable"` instead of the flattened placeholder.

packages/agent-bff/src/data/data-routes-middleware.ts
Tonours: Renamed to `resolveExposedRelationCapabilities`.

packages/agent-bff/src/data/data-routes-middleware.ts
Tonours: Not intentional — the relation path got the re-check because that is where the race was noticed, and `/list` and `/count` were left behind. Same window, same schema refresh, so they now behave the same.

Extracted `resolveCapabilitiesOrRethrowUnlessVanished(deps, collection, assertStillExposed)`: on a failed capabilities read it re-reads the store and runs the caller assert, so a genuinely vanished target wins as a 404 and everything else rethrows untouched. Relations pass `assertRelationStillExposed`, `/list` and `/count` pass the allow-list check through a matching `resolveOwnCapabilities`. If the re-read itself fails, the original error is still what surfaces.

Four tests added: collection dropped mid-failure gives 404 on both routes, and an unreadable schema during the re-check keeps `agent_unavailable`.

nbouliol (approved):

PR #1803, opened 2026-08-04
feat(agent-bff): make the agent request timeout configurable

Tonours:
fixes PRD-871

## Problem

The BFF always used the agent client's default 10-second timeout. `BFF_AGENT_TIMEOUT_MS` was not configurable at the BFF boundary.

## Fix

- Parse `BFF_AGENT_TIMEOUT_MS` with a 10-second default.
- Reject zero, negative, fractional, non-numeric, and values above Node's timer ceiling.
- Apply the configured timeout to data, action, and capabilities agent requests.
- Keep timeout failures mapped to the existing `502 network_error` contract.

## Scope and safety

- Changes are limited to `packages/agent-bff` in the final PR diff.
- `agent-client` has no net change.
- SaaS-facing timeouts, retries, and rate limiting are unchanged.

## How to test

```bash
```

### General

### Security

### DIFF
--- a/packages/agent-bff/src/action/action-routes-middleware.ts
+++ b/packages/agent-bff/src/action/action-routes-middleware.ts
@@ -67,7 +67,7 @@ function describePayloadShape(raw: unknown): string {
 export interface ActionRoutesMiddlewareOptions {
   store: ReadModelStore;
   agentUrl: string;
-  agentTimeoutMs?: number;
+  timeoutMs?: number;
   logger: Logger;
   createClient?: (options: AgentActionClientOptions) => AgentActionClient;
 }
@@ -161,7 +161,7 @@ async function handleExecute(
 export default function createActionRoutesMiddleware({
   store,
   agentUrl,
-  agentTimeoutMs,
+  timeoutMs,
   logger,
   createClient = defaultCreateAgentActionClient,
 }: ActionRoutesMiddlewareOptions): Middleware {
@@ -197,7 +197,7 @@ export default function createActionRoutesMiddleware({
       agentUrl,
       token,
       actionEndpoints: readModel.getActionEndpoints(),
-      timeoutMs: agentTimeoutMs,
+      timeoutMs,
     });
 
     if (verb === 'execute') {
--- a/packages/agent-bff/src/action/action-routes-middleware.ts
+++ b/packages/agent-bff/src/action/action-routes-middleware.ts
@@ -67,6 +67,7 @@ function describePayloadShape(raw: unknown): string {
 export interface ActionRoutesMiddlewareOptions {
   store: ReadModelStore;
   agentUrl: string;
+  agentTimeoutMs?: number;
   logger: Logger;
   createClient?: (options: AgentActionClientOptions) => AgentActionClient;
 }
@@ -160,6 +161,7 @@ async function handleExecute(
 export default function createActionRoutesMiddleware({
   store,
   agentUrl,
+  agentTimeoutMs,
   logger,
   createClient = defaultCreateAgentActionClient,
 }: ActionRoutesMiddlewareOptions): Middleware {
@@ -195,6 +197,7 @@ export default function createActionRoutesMiddleware({
       agentUrl,
       token,
       actionEndpoints: readModel.getActionEndpoints(),
+      timeoutMs: agentTimeoutMs,
     });
 
     if (verb === 'execute') {
--- a/packages/agent-bff/src/agent/create-agent-http-requester.ts
+++ b/packages/agent-bff/src/agent/create-agent-http-requester.ts
@@ -1,0 +1,24 @@
+import { HttpRequester } from '@forestadmin/agent-client';
+
+type QueryOptions = Parameters<HttpRequester['query']>[0];
+
+export default function createAgentHttpRequester(
+  token: string,
+  agentUrl: string,
+  timeoutMs?: number,
+): HttpRequester {
+  const requester = new HttpRequester(token, { url: agentUrl });
+
+  if (timeoutMs === undefined) {
+    return requester;
+  }
+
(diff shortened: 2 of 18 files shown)


packages/agent-bff/src/agent/create-agent-http-requester.ts
```diff
@@ -0,0 +1,24 @@
...
+  timeoutMs?: number,
+): HttpRequester {
+  const requester = new HttpRequester(token, { url: agentUrl });
+
+  if (timeoutMs === undefined) {
+    return requester;
+  }
+
+  const query = requester.query.bind(requester);
+  requester.query = (<Data = unknown>(options: QueryOptions): Promise<Data> =>
```
ShohanRahman: **Latent gap:** this wraps only `query`, not `stream`. `HttpRequester.stream` (used by `exportCsv`/segment export) independently defaults `maxTimeAllowed ?? 10_000`, so it won't pick up `BFF_AGENT_TIMEOUT_MS`.

Not exploitable today — no `agent-bff` source calls `stream`/`exportCsv` (confirmed by grep) — but the factory returns a full `HttpRequester`, so a future CSV/export caller wired through here would silently get the hardcoded 10s, contradicting the feature's contract.

Suggest either wrapping `stream` with the same `?? timeoutMs` pattern, or adding a one-line comment scoping this factory as query-only until export is needed.

packages/agent-bff/src/agent/create-agent-http-requester.ts
```diff
@@ -0,0 +1,24 @@
...
+    query({
+      ...options,
+      maxTimeAllowed: options.maxTimeAllowed ?? timeoutMs,
+    })) as HttpRequester['query'];
```
ShohanRahman: **Nit:** the `as HttpRequester['query']` cast is unnecessary — the wrapper's inferred signature is already structurally assignable, so a plain reassignment typechecks. The cast suppresses future signature-drift detection: if `HttpRequester.query`'s signature changes, this won't fail to compile.

Suggest dropping the cast; keep it only if a real mismatch forces it, with a comment noting what's being suppressed.

packages/agent-bff/src/action/action-routes-middleware.ts
```diff
@@ -195,6 +197,7 @@ export default function createActionRoutesMiddleware({
       agentUrl,
       token,
       actionEndpoints: readModel.getActionEndpoints(),
+      timeoutMs: agentTimeoutMs,
```
ShohanRahman: **Missing test (asymmetry):** `data-routes-middleware.test.ts` gained a test asserting `agentTimeoutMs` is forwarded to the data client, but the action side has no equivalent. The existing action-middleware tests use `createClient: () => client`, which ignores the options object, so nothing verifies `timeoutMs` reaches the action client here.

Suggest mirroring the data-middleware test: assert `createClient` receives `expect.objectContaining({ timeoutMs: … })`.

ShohanRahman (commented):
Automated review (Claude Code). Three inline notes — none blocking. The core design is sound; these are one latent gap and two small robustness nits.

packages/agent-bff/src/agent/create-agent-http-requester.ts
Tonours: Wrapped `stream` with the same `?? timeoutMs` pattern in 9e9c0cd8 rather than scoping the factory to query-only: the feature contract says the agent timeout is configurable, so carving out export would document the inconsistency instead of fixing it. Two tests cover it (configured default + explicit per-call override).

packages/agent-bff/src/agent/create-agent-http-requester.ts
Tonours: Dropped in 9e9c0cd8. Verified empirically: removed the cast, `tsc` passes, so the wrapper was already structurally assignable. Signature-drift detection is back.

packages/agent-bff/src/action/action-routes-middleware.ts
Tonours: Added in 9e9c0cd8, mirroring the data-middleware test: one asserting `createClient` receives `objectContaining({ timeoutMs: 2500 })`, one asserting it stays `undefined` when no timeout is configured. Both build the app with a real `createClient` mock instead of `() => client`, which was what swallowed the options object.

ShohanRahman:
**Naming inconsistency: `agentTimeoutMs` ↔ `timeoutMs`**

The timeout value is called `agentTimeoutMs` in the config/middleware layers (`BFFConfig`, `ActionRoutesMiddlewareOptions`, `DataRoutesMiddlewareOptions`) but renamed to `timeoutMs` at the client boundary (`AgentActionClientOptions`, `AgentDataClientOptions`, `AgentCapabilitiesFetcherOptions`, and `createAgentHttpRequester`).

Since these are separate type declarations rather than one reused field, TypeScript can't catch a mismatch, and a reader tracing the value through `cli-core.ts` → middleware → client → requester pays a "these are the same field" mental hop at each layer.

There's only one timeout concept in each interface, so collapsing to a single name throughout (either always `timeoutMs` or always `agentTimeoutMs`) would remove the mapping cost with no loss of clarity. Non-blocking.

Tonours:
Renamed to \`timeoutMs\` in both middlewares and \`RequestHandlerDeps\` in d1a6aaec. Kept \`agentTimeoutMs\` in \`BFFConfig\` only: that interface holds every BFF setting side by side (\`httpPort\`, \`defaultTimezone\`, \`agentUrl\`), so the \`agent\` prefix carries real information there — a bare \`timeoutMs\` next to \`httpPort\` would be ambiguous. Everything downstream is already agent-scoped, so the mapping now happens once, in the \`cli-core.ts\` destructuring, instead of at every layer boundary.

ShohanRahman (approved):

PR #1804, opened 2026-08-05
feat(agent-bff): serve the permissions display endpoint

Tonours:
fixes PRD-683

## What

`GET /agent/v1/permissions?collections=a,b` returns, per collection, the CRUD booleans and the per-action visibility and approval hints for the calling user.

Display hints only. It never grants access. The agent stays the authority, and `finality: "display_hint"` says so on every action entry. Zendesk currently gets this from `probeAccessibleActions`, a fail-open fan-out that probes actions one by one; this replaces it with one call.

The permission evaluation itself comes from `forestadmin-client` (#1802, now merged into `main`). What stays local is `can-role-perform.ts`, the three-line predicate that upstream keeps private inside `ActionPermissionService`.

## Decisions

**`?collections=` is optional.** Absent means every exposed collection. Requiring it would force the caller to fetch the schema first just to build the list. Present but empty scopes to nothing, so a client joining an empty array does not over-fetch.

**No new HTTP client.** `ForestHttpApi.getEnvironmentPermissions()` and `.getUsers()` were already public, `forest-secret-key` included.

**One cache entry, holding the evaluated permissions.** `PermissionsCache` stores `{ actionPermissions, users }` — the output of `generateActionsFromPermissions` — under a 15-minute TTL, mirroring how upstream's `ActionPermissionService` caches. Hints are rebuilt per request from the current read model, so nothing schema-derived is cached and no invalidation on schema refresh is needed. Deriving per request instead would be a regression: `generateActionsFromPermissions` builds its accumulators by object spread, so it is quadratic.

**Concurrent fetches collapse into one.** `fetchPermissions` issues two SaaS calls and nothing deduplicated concurrent misses, so N tabs hitting an expired TTL meant 2N calls. Same shared-promise pattern as `schema-cache.ts` and `capabilities-cache.ts`. A failed fetch is not memoized, the next caller retries.

**A caller the SaaS does not cover is remembered, not re-probed.** An authenticated user missing from `/permissions/users` would otherwise trigger a fresh SaaS fetch on every request. The rejected user ids live on the cache entry they were derived from: they expire with it, and are dropped when a new fetch returns a different `users` payload. A newly added user is therefore still served on their first call — only the repeats short-circuit. A rejection is recorded only when it came from a fresh fetch, never from the stale-cache fallback.

## Fail-closed

| Situation | Response |
| -- | -- |
| No caller identity | `401 unauthorized` |
| Caller absent from `/permissions/users` | `403 forest_identity_not_allowed` |
| SaaS fails, fresh cache entry covering the caller | `200` from cache |
| SaaS fails, fresh cache entry not covering the caller | `403 forest_identity_not_allowed` |
| SaaS fails, no fresh entry | `503 permissions_unavailable` + `Retry-After` |

An expired entry is never served. A missing CRUD descriptor or action-event flag is `false`.

## Scope

Untouched: data endpoints, action endpoints, the agent, the OAuth flow. No SaaS or DB write, cache in-memory per process. `roleId` and raw role arrays never reach the response.

Two additive changes outside `src/permissions/`:
- `BffHttpError` takes an options object for the trailing `details` / `retryAfter` pair. `error-middleware.ts` already read `retryAfter`, nothing could set it.
- `ReadModel.getAllowedCollections()` returns a fresh copy per call.

## How to test

```bash
```

913 tests, lint and typecheck clean. Manually verified against a live SaaS in normal mode: 401 unauthenticated, 403 unknown caller, fallback to all collections, unknown collection omitted, cache hit, 503 + `Retry-After` with the SaaS down and no cached entry, 200 from cache with the SaaS down and a fresh one.

That manual pass predates the review fixes that followed it (single-entry cache, negative caching, `visibleActions` shape), which are covered by unit tests only. Dev mode (roles disabled) is also unit-tested only, the QA environment runs in normal mode.

### General

### Security

### DIFF
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -165,7 +165,10 @@ function buildAgentRouteMiddlewares(config: BFFConfig, logger: Logger): Middlewa
   const apiKeyConfig = resolveApiKeyConfig(config);
 
   if (!apiKeyConfig) {
-    logger('Warn', 'Data endpoints disabled: AGENT_URL or read-model configuration is missing');
+    logger(
+      'Warn',
+      'Data, action and permissions endpoints disabled: FOREST_SERVER_URL, FOREST_ENV_SECRET or FOREST_AUTH_SECRET is missing',
+    );
 
     return [createAgentStubMiddleware()];
   }
@@ -176,8 +179,6 @@ function buildAgentRouteMiddlewares(config: BFFConfig, logger: Logger): Middlewa
     logger,
   });
   const { agentUrl, agentTimeoutMs: timeoutMs } = config;
-  const permissionsCache = new PermissionsCache();
-  store.registerGenerationScopedCache(permissionsCache);
 
   const permissionsMiddleware = createPermissionsRoutesMiddleware({
     store,
@@ -185,13 +186,12 @@ function buildAgentRouteMiddlewares(config: BFFConfig, logger: Logger): Middlewa
       forestServerUrl: apiKeyConfig.forestServerUrl,
       envSecret: apiKeyConfig.forestEnvSecret,
     }),
-    cache: permissionsCache,
-    envSecret: apiKeyConfig.forestEnvSecret,
+    cache: new PermissionsCache(),
     logger,
   });
 
   if (!agentUrl) {
-    logger('Warn', 'Data endpoints disabled: AGENT_URL or read-model configuration is missing');
+    logger('Warn', 'Data and action endpoints disabled: AGENT_URL is missing');
 
     return [permissionsMiddleware, createAgentStubMiddleware()];
   }
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -26,6 +26,9 @@ import createOAuthRoutes from './oauth/oauth-routes';
 import createInMemorySessionStore from './oauth/session-store';
 import createTokenCipher from './oauth/token-cipher';
 import createOpenApiRoutes from './openapi/openapi-routes';
+import PermissionsCache from './permissions/permissions-cache';
+import PermissionsClient from './permissions/permissions-client';
+import createPermissionsRoutesMiddleware from './permissions/permissions-routes-middleware';
 import createReadModel from './read-model/create-read-model';
 import createTimezoneMiddleware from './timezone/timezone-middleware';
 import version from './version';
@@ -173,8 +176,20 @@ function buildAgentRouteMiddlewares(config: BFFConfig, logger: Logger): Middlewa
     logger,
   });
   const { agentUrl, agentTimeoutMs: timeoutMs } = config;
+  const permissionsCache = new PermissionsCache();
+  store.registerGenerationScopedCache(permissionsCache);
 
   return [
+    createPermissionsRoutesMiddleware({
+      store,
+      client: new PermissionsClient({
+        forestServerUrl: apiKeyConfig.forestServerUrl,
+        envSecret: apiKeyConfig.forestEnvSecret,
+      }),
+      cache: permissionsCache,
+      envSecret: apiKeyConfig.forestEnvSecret,
+      logger,
+    }),
     createDataRoutesMiddleware({ store, agentUrl, timeoutMs, logger }),
     createActionRoutesMiddleware({ store, agentUrl, timeoutMs, logger }),
   ];
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -164,7 +164,7 @@ function buildApiKeyMiddleware(config: BFFConfig, logger: Logger): Middleware | 
 function buildAgentRouteMiddlewares(config: BFFConfig, logger: Logger): Middleware[] {
   const apiKeyConfig = resolveApiKeyConfig(config);
 
-  if (!apiKeyConfig || !config.agentUrl) {
+  if (!apiKeyConfig) {
     logger('Warn', 'Data endpoints disabled: AGENT_URL or read-model configuration is missing');
 
     return [createAgentStubMiddleware()];
@@ -179,17 +179,25 @@ function buildAgentRouteMiddlewares(config: BFFConfig, logger: Logger): Middlewa
   const permissionsCache = new PermissionsCache();
   store.registerGenerationScopedCache(permissionsCache);
 
-  return [
-    createPermissionsRoutesMiddleware({
-      store,
-      client: new PermissionsClient({
-        forestServerUrl: apiKeyConfig.forestServerUrl,
-        envSecret: apiKeyConfig.forestEnvSecret,
-      }),
-      cache: permissionsCache,
+  const permissionsMiddleware = createPermissionsRoutesMiddleware({
+    store,
+    client: new PermissionsClient({
+      forestServerUrl: apiKeyConfig.forestServerUrl,
       envSecret: apiKeyConfig.forestEnvSecret,
-      logger,
     }),
+    cache: permissionsCache,
+    envSecret: apiKeyConfig.forestEnvSecret,
+    logger,
+  });
+
+  if (!agentUrl) {
+    logger('Warn', 'Data endpoints disabled: AGENT_URL or read-model configuration is missing');
+
+    return [permissionsMiddleware, createAgentStubMiddleware()];
+  }
+
+  return [
+    permissionsMiddleware,
     createDataRoutesMiddleware({ store, agentUrl, timeoutMs, logger }),
     createActionRoutesMiddleware({ store, agentUrl, timeoutMs, logger }),
   ];
--- a/packages/agent-bff/src/http/bff-http-error.ts
+++ b/packages/agent-bff/src/http/bff-http-error.ts
@@ -1,22 +1,21 @@
+export interface BffHttpErrorOptions {
+  details?: unknown;
+  retryAfter?: number;
+}
+
 export class BffHttpError extends Error {
   readonly status: number;
   readonly type: string;
   readonly details?: unknown;
   readonly retryAfter?: number;
 
-  constructor(
-    status: number,
-    type: string,
-    message: string,
-    details?: unknown,
-    retryAfter?: number,
-  ) {
+  constructor(status: number, type: string, message: string, options: BffHttpErrorOptions = {}) {
     super(message);
     this.name = 'BffHttpError';
     this.status = status;
     this.type = type;
-    this.details = details;
-    this.retryAfter = retryAfter;
+    this.details = options.details;
+    this.retryAfter = options.retryAfter;
   }
 }
 
@@ -93,6 +92,6 @@ export function relationFieldNotSupported(fields: string[]): BffHttpError {
     422,
     'relation_field_not_supported',
     'Nested relation field paths are not supported on top-level list and count',
-    { fields },
+    { details: { fields } },
   );
 }
--- a/packages/agent-bff/src/http/bff-http-error.ts
+++ b/packages/agent-bff/src/http/bff-http-error.ts
@@ -2,13 +2,21 @@ export class BffHttpError extends Error {
   readonly status: number;
   readonly type: string;
   readonly details?: unknown;
+  readonly retryAfter?: number;
 
-  constructor(status: number, type: string, message: string, details?: unknown) {
+  constructor(
+    status: number,
+    type: string,
+    message: string,
+    details?: unknown,
+    retryAfter?: number,
+  ) {
     super(message);
     this.name = 'BffHttpError';
     this.status = status;
     this.type = type;
     this.details = details;
+    this.retryAfter = retryAfter;
   }
 }
 
--- a/packages/agent-bff/src/permissions/build-permission-hints.ts
+++ b/packages/agent-bff/src/permissions/build-permission-hints.ts
@@ -19,7 +19,6 @@ export interface CrudHints {
 export interface ActionHints {
   collection: string;
   name: string;
-  qualifiedName: string;
   visible: boolean;
   requiresApprovalHint: boolean;
   canApproveHint: boolean;
@@ -27,15 +26,20 @@ export interface ActionHints {
   finality: typeof DISPLAY_HINT_FINALITY;
 }
 
+export interface VisibleActionRef {
+  collection: string;
+  name: string;
+}
+
 export interface CollectionHints {
   crud: CrudHints;
   actions: Record<string, ActionHints>;
-  visibleActions: string[];
+  visibleActions: VisibleActionRef[];
 }
 
 export interface PermissionHints {
   collections: Record<string, CollectionHints>;
-  visibleActions: string[];
+  visibleActions: VisibleActionRef[];
 }
 
 export interface BuildPermissionHintsParams {
@@ -43,10 +47,6 @@ export interface BuildPermissionHintsParams {
   roleId: number;
   readModel: ReadModel;
   collections: string[];
-}
-
-function qualify(collection: string, action: string): string {
-  return `${collection}.${action}`;
 }
 
 function devModeCrud(): CrudHints {
@@ -75,7 +75,6 @@ function devModeAction(collection: string, name: string): ActionHints {
   return {
     collection,
     name,
-    qualifiedName: qualify(collection, name),
     visible: true,
     requiresApprovalHint: false,
     canApproveHint: false,
@@ -102,7 +101,6 @@ function normalModeAction(
   return {
     collection,
     name,
-    qualifiedName: qualify(collection, name),
     visible: can(CustomActionEvent.Trigger),
     requiresApprovalHint: can(CustomActionEvent.RequireApproval),
     canApproveHint: can(CustomActionEvent.Approve),
@@ -120,7 +118,7 @@ function buildCollectionHints(
   const schemaExposedActionNames = Object.keys(readModel.getActionEndpoints()[collection] ?? {});
 
   const actions: Record<string, ActionHints> = Object.create(null);
-  const visibleActions: string[] = [];
+  const visibleActions: VisibleActionRef[] = [];
 
   for (const name of schemaExposedActionNames) {
     const hints = isDevelopment
@@ -128,7 +126,7 @@ function buildCollectionHints(
       : normalModeAction(permissions, roleId, collection, name);
 
     actions[name] = hints;
-    if (hints.visible) visibleActions.push(hints.qualifiedName);
+    if (hints.visible) visibleActions.push({ collection, name });
   }
 
   return {
@@ -141,7 +139,7 @@ function buildCollectionHints(
 export default function buildPermissionHints(params: BuildPermissionHintsParams): PermissionHints {
   const permissions = params.actionPermissions;
   const collections: Record<string, CollectionHints> = Object.create(null);
-  const visibleActions: string[] = [];
+  const visibleActions: VisibleActionRef[] = [];
 
   for (const collection of params.collections) {
     const hints = buildCollectionHints({ ...params, permissions }, collection);
--- a/packages/agent-bff/src/permissions/build-permission-hints.ts
+++ b/packages/agent-bff/src/permissions/build-permission-hints.ts
@@ -1,0 +1,157 @@
+import type { ActionPermissions } from './action-permissions';
+import type ReadModel from '../read-model/read-model';
+import type { EnvironmentPermissionsV4 } from '@forestadmin/forestadmin-client';
+
+import { CollectionActionEvent, CustomActionEvent } from './action-identifiers';
+import {
+  buildActionPermissions,
+  canRolePerformCollectionAction,
+  canRolePerformCustomAction,
+} from './action-permissions';
+
+export const DISPLAY_HINT_FINALITY = 'display_hint';
+
+export interface CrudHints {
+  browse: boolean;
+  read: boolean;
+  edit: boolean;
+  add: boolean;
+  delete: boolean;
+  export: boolean;
+}
+
+export interface ActionHints {
+  collection: string;
+  name: string;
+  qualifiedName: string;
+  visible: boolean;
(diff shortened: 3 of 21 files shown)


packages/agent-bff/src/permissions/build-permission-hints.ts
```diff
@@ -0,0 +1,157 @@
...
+  return {
+    crud: isDevelopment ? devModeCrud() : normalModeCrud(permissions, roleId, collection),
+    actions,
+    visibleActions,
+  };
+}
+
+export default function buildPermissionHints(params: BuildPermissionHintsParams): PermissionHints {
+  const permissions = buildActionPermissions(params.environmentPermissions);
+  const collections: Record<string, CollectionHints> = {};
```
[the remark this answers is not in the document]
Tonours: Applied, but the rationale is wrong: the read model does not use `Object.create(null)`, it uses `new Map()` — immune by construction. The bug is real though, a collection named `__proto__` silently vanishes from the response.

Also fixed the same thing on the `actions` map in `buildCollectionHints`, which had the same hole and wasn't flagged. Both now use `Object.create(null)`, covered by a test that fails without the fix.

packages/agent-bff/src/permissions/build-permission-hints.ts
```diff
@@ -0,0 +1,157 @@
...
+}
+
+export interface BuildPermissionHintsParams {
+  environmentPermissions: EnvironmentPermissionsV4;
+  roleId: number;
+  readModel: ReadModel;
+  collections: string[];
+}
+
+function qualify(collection: string, action: string): string {
```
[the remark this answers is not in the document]
Tonours: Not changing the delimiter. `qualifiedName` is an output contract the frontend reads (`users.Block user`), swapping it would break consumers for something the ticket never asked for. `ActionHints` already carries `collection` and `name` separately, so a consumer can always disambiguate.

Digging into it, the real bug was worse than the duplicate you describe: `actions` was keyed by `qualifiedName`, so with collection `User` + action `address.reset` and collection `User.address` + action `reset`, the two hints overwrite each other across collections. Fixed by keying `actions` on the action name — unique per collection, and the map already lives under `collections[collection]` so the prefix was redundant. Test added for the collision case.

packages/agent-bff/src/permissions/permissions-cache.ts
```diff
@@ -0,0 +1,88 @@
...
+  collections: string[];
+}
+
+export function buildCacheKey({ envSecret, callerId, collections }: CacheKeyParts): string {
+  const scope = JSON.stringify([...collections].sort());
+
+  return `${envSecret}|${callerId}|${scope}`;
+}
+
+export default class PermissionsCache {
```
nbouliol, line 29: Upstream `ActionPermissionService` caches the raw permissions in one `TTLCache` entry and derives decisions per call. Caching the derived hints per caller x collection-set instead is what forces the 1000-entry cap, the eviction loop, the collection set in the key, and the generation guard - would caching `EnvironmentAndUserPermissions` once collapse all of that?

packages/agent-bff/src/permissions/permissions-routes-middleware.ts
```diff
@@ -0,0 +1,143 @@
...
+        ctx.body = stillFresh;
+
+        return;
+      }
+
+      throw permissionsUnavailable(UNAVAILABLE_RETRY_AFTER_SECONDS);
+    }
+
+    const roleId = findCallerRoleId(resolved.users, caller.userId);
+    const refreshedDuringFetch = cache.currentGeneration !== generationBeforeRead;
```
nbouliol: `ReadModelStore.getCapabilities` already handles a mid-fetch generation bump with a bounded retry - could this reuse that instead of the five ternaries?

packages/agent-bff/src/cli-core.ts
```diff
@@ -173,8 +176,28 @@ function buildAgentRouteMiddlewares(config: BFFConfig, logger: Logger): Middlewa
...
+      forestServerUrl: apiKeyConfig.forestServerUrl,
+      envSecret: apiKeyConfig.forestEnvSecret,
+    }),
+    cache: permissionsCache,
+    envSecret: apiKeyConfig.forestEnvSecret,
+    logger,
+  });
+
+  if (!agentUrl) {
+    logger('Warn', 'Data endpoints disabled: AGENT_URL or read-model configuration is missing');
```
nbouliol: Same message as the earlier branch, but there the permissions endpoint is disabled too - worth differentiating the two?

packages/agent-bff/src/http/bff-http-error.ts
```diff
@@ -2,13 +2,21 @@ export class BffHttpError extends Error {
...
   readonly details?: unknown;
+  readonly retryAfter?: number;
 
-  constructor(status: number, type: string, message: string, details?: unknown) {
+  constructor(
+    status: number,
+    type: string,
+    message: string,
+    details?: unknown,
+    retryAfter?: number,
```
nbouliol: Fifth positional param forces `new BffHttpError(503, ..., undefined, retryAfter)` - options object for the trailing pair?

packages/agent-bff/src/permissions/permissions-routes-middleware.ts
```diff
@@ -0,0 +1,143 @@
...
+
+function resolveRequestedCollections(ctx: Context, readModel: ReadModel): string[] {
+  const raw = ctx.query.collections;
+  const value = Array.isArray(raw) ? raw.join(',') : raw;
+  const requested = (value ?? '')
+    .split(',')
+    .map(name => name.trim())
+    .filter(name => name !== '');
+
+  if (requested.length === 0) return readModel.getAllowedCollections();
```
nbouliol: `?collections=` with an empty value falls back to every collection - a client joining an empty array would over-fetch.

packages/agent-bff/src/permissions/permissions-routes-middleware.ts
```diff
@@ -0,0 +1,158 @@
...
+  logger: Logger;
+}): Promise<{ permissions: EvaluatedPermissions; roleId: number }> {
+  const cached = cache.getFresh();
+  const cachedRoleId = cached && findCallerRoleId(cached.users, caller.userId);
+
+  if (cached && cachedRoleId !== undefined) {
+    return { permissions: cached, roleId: cachedRoleId };
+  }
+
+  if (cached) cache.clear();
```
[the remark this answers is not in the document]
Tonours: Good catch, fixed. The `cache.clear()` is gone: a caller-specific miss now just falls through to the refetch, and a successful fetch overwrites the entry anyway. The stale-serve fallback survives, so a SaaS outage no longer turns a 403 into a 503 for the unknown caller, and other callers keep being served from cache. Pinned by a test that fails with the clear restored (503 instead of 403).

packages/agent-bff/src/permissions/permissions-cache.ts
Tonours: Done, and it collapsed all four. The cache now holds one entry: the derived `ActionPermissions` plus `users`. Gone: the 1000 cap, the eviction loop, `buildCacheKey`, the generation guard and `registerGenerationScopedCache`. One nuance worth flagging: upstream caches the *derived* index, not the raw payload, and `generateActionsFromPermissions` builds its accumulators by object spread, so it is quadratic. Deriving per request would have been a real regression. It now runs once per fetch and each request only does Set/Map lookups on the cached index.

packages/agent-bff/src/permissions/permissions-routes-middleware.ts
Tonours: The five ternaries are gone, without needing the retry. Since the cached permissions are schema-independent, resolving the read model *after* the fetch is enough: it is now the last await before the response is built, so nothing can supersede it. The generation guard and `registerGenerationScopedCache` went with it.

packages/agent-bff/src/cli-core.ts
Tonours: Differentiated, and each now names the env vars an operator can act on. Missing api-key config: "Data, action and permissions endpoints disabled: FOREST_SERVER_URL, FOREST_ENV_SECRET or FOREST_AUTH_SECRET is missing". Missing agent URL: "Data and action endpoints disabled: AGENT_URL is missing" — permissions stay up in that branch.

packages/agent-bff/src/http/bff-http-error.ts
Tonours: Switched to an options object: `new BffHttpError(status, type, message, { details?, retryAfter? })`. The `undefined` placeholder is gone. All call sites updated — TypeScript caught two test files that were passing details positionally, which is the point.

packages/agent-bff/src/permissions/permissions-routes-middleware.ts
Tonours: Fixed. `?collections=` absent still means every exposed collection, but a present-but-empty value now scopes to nothing, so a client joining an empty array gets an empty response instead of over-fetching. Pinned by a test.

packages/agent-bff/src/permissions/build-permission-hints.ts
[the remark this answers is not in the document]
Tonours: Fixed in a000cc26c.

`qualifiedName` is gone and `visibleActions` now carries `{ collection, name }` pairs instead of flat strings. No delimiter is safe here: collection names legitimately contain dots (mongoose nested collections like `User.address`, see the comment in `read-model.ts:22`) and action names are free-form (`Block user`), so the collection boundary can only be unambiguous if it is structural. The nested `collections[...].actions[...]` map already disambiguated; the flat top-level list was the only lossy spot.

Test updated to assert both `User` / `address.reset` and `User.address` / `reset` stay distinct in `visibleActions`.

packages/agent-bff/src/permissions/permissions-routes-middleware.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  const permissions = await resolvePermissions({ cache, client, logger });
+  const roleId = findCallerRoleId(permissions.users, caller.userId);
+
+  if (roleId === undefined) throw forestIdentityNotAllowed();
+
+  return { permissions, roleId };
+}
+
+export default function createPermissionsRoutesMiddleware({
```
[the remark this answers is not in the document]
Tonours: I don't think this one holds — the cache does not store anything schema-derived.

`PermissionsCache` holds `{ actionPermissions, users }`, i.e. the output of `generateActionsFromPermissions(environmentPermissions)`. That structure is keyed by SaaS-side identifiers (`generateCollectionActionIdentifier` / `generateCustomActionIdentifier`), so it is a projection of the Forest permission payload, not of the agent schema. No collection list, action list or generation is captured in it.

The hints are rebuilt on every request, and the read model is resolved *after* the permissions (`permissions-routes-middleware.ts:138` then `:145`), so the response always pairs the freshest schema with the freshest read model. `buildCollectionHints` iterates `readModel.getActionEndpoints()[collection]` — the new schema — and queries the cached permissions per identifier.

For an action the schema exposes but the cached permissions don't describe yet, `isActionIdentifierAllowedForRole` resolves `actionsByRole.get(id)?.allowedRoles.has(roleId)` to `undefined` and returns `false`. So a newly deployed action reads `visible: false` until the next permissions fetch: fail-closed, never a spurious grant. Invalidating the permissions cache on every schema refresh would only flip that hint from `false` to `false` while adding a SaaS round-trip.

Added a test in a000cc26c (`when the schema exposes an action the cached permissions do not describe yet`) to lock the fail-closed behaviour in.

Happy to reopen if you were pointing at a case where stale permissions can *grant* something — I couldn't construct one.

nbouliol (commented):
The description has drifted from the diff: it still documents `registerGenerationScopedCache`, the 1000-entry cap with its eviction loop, and the `generationBeforeRead`/`refreshedDuringFetch` fixes, none of which survive in the code. The stacking note is stale too now that the base is `main`. And the fail-closed table says a fresh cache entry serves a 200 when the SaaS fails, but an unknown caller gets a 403 in that case (tested at `permissions-routes-middleware.test.ts:222`) - worth reflecting before merge.

packages/agent-bff/src/permissions/permissions-routes-middleware.ts
```diff
@@ -0,0 +1,156 @@
...
+  logger: Logger;
+}): Promise<{ permissions: EvaluatedPermissions; roleId: number }> {
+  const cached = cache.getFresh();
+  const cachedRoleId = cached && findCallerRoleId(cached.users, caller.userId);
+
+  if (cached && cachedRoleId !== undefined) {
+    return { permissions: cached, roleId: cachedRoleId };
+  }
+
+  const permissions = await resolvePermissions({ cache, client, logger });
```
nbouliol: An authenticated caller absent from `/permissions/users` refetches on every request, and `inFlight` only collapses the concurrent ones - worth a short cool-down before retrying?

packages/agent-bff/src/permissions/build-permission-hints.ts
```diff
@@ -0,0 +1,151 @@
...
+    const hints = isDevelopment
+      ? devModeAction(collection, name)
+      : normalModeAction(permissions, roleId, collection, name);
+
+    actions[name] = hints;
+    if (hints.visible) visibleActions.push({ collection, name });
+  }
+
+  return {
+    crud: isDevelopment ? devModeCrud() : normalModeCrud(permissions, roleId, collection),
```
nbouliol: `normalModeCrud` already returns all-true in dev, since `isActionIdentifierAllowedForRole` short-circuits on `isDevelopment` - only the action branch needs the split?

packages/agent-bff/src/permissions/build-permission-hints.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export interface CollectionHints {
+  crud: CrudHints;
+  actions: Record<string, ActionHints>;
+  visibleActions: VisibleActionRef[];
+}
+
+export interface PermissionHints {
+  collections: Record<string, CollectionHints>;
+  visibleActions: VisibleActionRef[];
```
nbouliol, line 41: Now that each ref carries its collection, this is derivable from `collections[x].actions` filtered on `visible` - does the caller need all three shapes?

packages/agent-bff/test/permissions/permissions-routes-middleware.test.ts
```diff
@@ -0,0 +1,512 @@
...
+        .query({ collections: 'users' });
+
+      expect(response.status).toBe(200);
+      expect(response.body.visibleActions).toEqual([{ collection: 'users', name: 'Block user' }]);
+      expect(client.fetchPermissions).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe('when the same collection is requested twice in the query', () => {
+    it('should deduplicate it in visibleActions and reuse the single-name cache entry', async () => {
```
nbouliol: The cache key no longer includes the collection set - this name still describes the old one.

packages/agent-bff/src/permissions/build-permission-hints.ts
Tonours: Right, `normalModeCrud` was already returning all-true in dev through the short-circuit, so `devModeCrud` was dead code. Dropped it, and `normalModeCrud` is now just `buildCrudHints` since there is nothing to contrast it with.

The action branch does still need the split, for the reason the CRUD one does not: the short-circuit would make `requiresApprovalHint`, `canApproveHint` and `canSelfApproveHint` all true in dev, and those must read false. Renamed `devModeAction` to `visibleWithoutApprovalHints` so the name says what it is for rather than which mode reaches it. The existing dev-mode test already pins both halves (CRUD all true, approval hints all false).

packages/agent-bff/src/permissions/build-permission-hints.ts
Tonours: No, two is enough. Dropped the per-collection `visibleActions`: it was derivable from the sibling `actions` map right next to it, and nothing read it — not the tests, not Zendesk. The top-level list is now built by filtering `hints.actions` on `visible`, which is the derivation you describe.

Kept the top-level one: it is the only cross-collection aggregate, so deriving it client-side means walking every collection. Zendesk needs "what can this user trigger" without caring where each action lives.

packages/agent-bff/src/permissions/permissions-routes-middleware.ts
Tonours: Added, but keyed on the identity rather than on a clock. A flat cool-down would have broken the case the test above it pins: a user who was just added to the project must be served on their first call, not made to wait out a timer.

`PermissionsCache` now records the user ids a *fresh* fetch failed to cover. First call from an unknown caller still refetches — that is the newly-added user — and only the repeats short-circuit to 403 with no SaaS call. `set()` clears the set, so the next successful fetch gives every rejected id another chance; a revoked identity therefore costs one fetch per TTL window instead of one per request.

The fresh-fetch part matters: on a SaaS outage `resolvePermissions` falls back to the stale cache, and a caller missing from *that* was never actually refuted, so nothing is recorded and the retry still happens once the SaaS is back. `resolvePermissions` returns `fromFreshFetch` to keep the two apart.

Three tests: repeats collapse to one fetch, a refreshed payload re-enables the refetch, and an outage does not record a rejection. The id set is bounded by real revoked users and comes from the signed token, not the request.

packages/agent-bff/test/permissions/permissions-routes-middleware.test.ts
Tonours: Renamed to `should deduplicate it in visibleActions and fetch the permissions once` — the cache is a single entry now, so there is no per-name entry to reuse.

nbouliol (commented):
The `visibleActions` derivation and the `buildCrudHints` collapse both landed cleanly, and `visibleWithoutApprovalHints` is a better name than the mode it replaced.

Two follow-ups on the negative cache though, both reproduced against this branch - details inline.

packages/agent-bff/src/permissions/permissions-cache.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  rememberRejected(userId: number): void {
+    this.rejectedUserIds.add(userId);
+  }
+
+  getFresh(): EvaluatedPermissions | undefined {
+    if (!this.entry) return undefined;
+
+    if (this.now() - this.entry.storedAt >= this.ttlMs) {
+      this.entry = undefined;
```
nbouliol: This drops the entry but keeps `rejectedUserIds`, and nothing in `src/` ever calls `clear()` - so once the TTL lapses a caller the SaaS now covers stays 403 with no refetch, until some other caller happens to repopulate the cache. Clearing the rejections in this branch would tie them to the entry they were derived from.

Repro:
1. Unknown caller `U` hits the route - the fetch runs, `U` is rejected and remembered.
2. Advance the clock by `PERMISSIONS_CACHE_TTL_MS` and add `U` to the users payload.
3. `U` hits the route again.

Observed `status=403 fetches=0`, expected `200`.

packages/agent-bff/src/permissions/permissions-cache.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      return undefined;
+    }
+
+    return this.entry.permissions;
+  }
+
+  set(permissions: EvaluatedPermissions): void {
+    this.entry = { permissions, storedAt: this.now() };
+    this.rejectedUserIds.clear();
```
nbouliol: This clear runs on the very fetch that is about to record a rejection, so each new rejection wipes the previous one and the set never holds more than one at a time - two rejected callers alternating still refetch on every request, which is the amplification this was meant to remove. Clearing only when the `users` list actually changed, or giving the rejections their own short TTL, would hold both.

Repro: alternate two unknown callers `A, B, A, B, A` against one cache.

Observed `fetches=5`, expected `2`.

packages/agent-bff/src/permissions/permissions-cache.ts
Tonours: Confirmed and fixed — I reproduced both of these before touching anything, and both went red on the previous commit.

Root cause was the same for the two: the rejection set was a field on the cache, living in parallel to the entry it was derived from. It is now a field *of* the entry, so it expires with the payload it came from. Your case 2 no longer happens: the TTL lapses, the entry goes, and the rejections go with it.

Test added — unknown caller rejected, clock advanced by `PERMISSIONS_CACHE_TTL_MS`, users payload now covering them: 200, one fetch. It fails with `403 fetches=0` on the previous commit, exactly as you observed.

packages/agent-bff/src/permissions/permissions-cache.ts
Tonours: Also confirmed. Went with your first option: `set()` now keeps the previous rejection set when the incoming `users` payload is unchanged, and starts a fresh one otherwise. `sameUsers` compares the id/roleId pairs, so a role change also drops the rejections rather than carrying a stale verdict.

Preferred that over a separate TTL — it ties the invalidation to the thing the rejection was actually derived from instead of adding a second clock to reason about.

Your alternating repro is now a test: `A, B, A, B, A` against one cache gives `fetches=2`. It fails with `fetches=5` on the previous commit.

Tonours:
Description rewritten — all four were right.

Dropped: the stacking note (base is `main`, #1802 is merged), `registerGenerationScopedCache`, the 1000-entry cap and its eviction loop, and the `generationBeforeRead` / `refreshedDuringFetch` section. None of that survives in the diff.

Fail-closed table: split the SaaS-failure row in two, since the answer depends on whether the fresh entry covers the caller — 200 when it does, 403 when it does not, which is the case pinned at `permissions-routes-middleware.test.ts:222`.

Added the parts that had no entry: the single-entry cache and why hints stay schema-independent, the present-but-empty `?collections=`, the `BffHttpError` options object, and the negative caching of rejected callers. Test count is 913. Also flagged that the manual SaaS pass predates the review fixes that followed it.

nbouliol (approved):
Both cache defects are fixed, and tying the rejected ids to the entry they came from is the right seam - the two repros flip cleanly. Description reads accurate now too.

PR #1806, opened 2026-08-05
fix(workflow-executor): surface the cause when MCP tool loading fails

hercemer42:
## What

A workflow MCP step whose tool load fails told the customer *that* it failed, never *why*. Two gaps, both fixed here.

**1. The executor's own line had no cause.** `errorOnPartialLoadFailure` inferred failure by diffing config ids against loaded tool ids, so it could name the server but never the reason — a revoked token, an unreachable host and a 15s timeout all logged identically. The providers already classify each failure and carry its error, but the main load path called the tools-only `loadRemoteTools` and dropped them. It now reads the failures channel:

```
{"level":"Error","message":"MCP servers failed to load tools","requestedMcpServerId":"39","mcpServerName":"acme-crm",
 "failures":[{"server":"acme-crm","kind":"connection","error":"connect ECONNREFUSED 10.0.4.12:8080"}]}
```

**2. ai-proxy's own diagnostics went nowhere.** `AiClient` holds an optional host logger and no-ops every emit when none is passed; `workflow-executor` never passed one. Both construction sites now receive the logger the executor already builds, through `to-ai-proxy-logger.ts`, which bridges the two logger contracts — ai-proxy's third parameter is an `Error`, the executor's is a context object.

## Behaviour changes worth a reviewer's eye

**A healthy server exposing no tools is no longer a failure.** Reading the failures channel removes a false positive the id-diff could not avoid: such a server contributed no ids, so it was reported as failed and `GET /list-mcp-tools` answered `503 The MCP server could not be reached to list its tools`. It now answers `200` with an empty list. A server that genuinely errors still sets `loadFailed` and still 503s. The frontend counterpart is ForestAdmin/forestadmin#9892, which drops the matching "empty list means the load failed" inference; without it this change never reaches the user.

**A recovered OAuth load now says so.** When a cached token is rejected and the forced refresh succeeds, the provider has already emitted its rejection at `Error`, so at the default level the run read as a pure failure with nothing saying it continued. It now states the recovery at `Info`. Observed on the rig before it was fixed — the retry was `Debug`-only, so two `Error` lines were all an operator saw.

**Pretty output no longer prints absent keys.** `formatContext` rendered `JSON.stringify(undefined)` as the literal `undefined`, so every bridged line trailed `cause=undefined` — most MCP errors don't chain, since `@langchain/mcp-adapters` interpolates the nested text into its own message. It now skips undefined values, which also cleans up `base-step-executor`'s existing lines.

## Why the cause is flattened rather than passed through

An `Error`'s own properties are non-enumerable, so handing it to the executor's logger as the context object drops the cause from the line that gets emitted. All three consumers agree: `console-logger` spreads it into `JSON.stringify`, `pretty-logger` iterates `Object.entries`, and the agent-embedded `formatLog` takes its message-only branch when `Object.keys(context).length === 0`. The bridge flattens to `{ error, cause, stack }` — the shape `base-step-executor.ts:94` already uses — and carries the cause chain, so a wrapped `fetch failed` still names the `ECONNREFUSED` underneath it.

The bridge also guards its own body: ai-proxy logs from inside its per-server `catch` **before** recording the failure it caught (`mcp-client.ts:113`, `failures.push` at `:114`), so a host logger that threw would reject the whole `Promise.all` — discarding tools from healthy servers and skipping the OAuth reauth pause. `ExecutorOptions.logger` is host-supplied, and `agent/src/embedded-workflow-executor.ts:12` already states that invariant for the neighbouring edge.

## Scope

The `agent-nodejs` half of PRD-876. The `forestadmin-server` half (`make-ai-router-service.ts`, the `POST /api/ai-proxy/ai-query` path — AC#1–#3) ships separately; neither PR blocks the other. No ai-proxy file is touched: the timeout, classification and `failures` channel stay as PRD-863 left them.

AC#4 is implemented as reworded in the ticket comment — the existing executor log line carries the failure `kind` and cause, sourced from `loadRemoteToolsWithFailures` rather than inferred from absent tools — **and** the logger threading it originally called for, which is what makes ai-proxy's other diagnostics (per-server load errors with stacks, `Unsupported integration:`, connection-cleanup failures) visible at all.

## Tests

- `to-ai-proxy-logger.test.ts` — level/message pass-through for all four levels, cause flattening, the cause chain, the cause surviving `createConsoleLogger()` at its **default** level (business rule 3), a throwing host logger, stackless / empty-message / non-`Error` / `null` causes, independence across calls. 100% coverage.
- `remote-tool-fetcher.test.ts` — the log line names server, kind and cause; `loadFailed` follows the reported failures, including when the post-refresh retry fails for a non-auth reason; a healthy server exposing no tools is not a failure; the recovery line fires only when the retry actually recovered.
- Both AI adapter suites assert the logger reaching `AiClient`; `build-workflow-executor` asserts it reaching both adapters; `runner` and the integration suite cover the dispatch path.
- The 8 changed test files run 225 tests green. ai-proxy's own suite runs unchanged and green (38 suites / 429 tests) — AC#5.

## Verified on a live rig

Driven through `GET /list-mcp-tools`, which runs the whole changed path (`fetch` → `loadRemoteToolsWithFailures` → `errorOnPartialLoadFailure` → ai-proxy's `McpClient`), against a standalone executor and the frontend:

- **Unreachable server** → 503, with ai-proxy's per-server `Error loading tools for …` carrying `error`/`stack`, its `Failed to load tools from 1/1 …` summary, and the executor's line with `kind:"connection"`.
- **401 from the server** → same shape with `kind:"auth"`, so a rejected credential and an unreachable host are one grep apart. This is the support-deflection claim, and it holds.
- **Healthy server** → `Loaded 10 tools from MCP server "data-gouv" in 925ms` at `Debug`, absent at `Info`, while the failure lines stay visible there (business rules 2 and 3).
- **A/B on the 503→200 change** — `origin/main`'s fetcher answered `503` and logged `failedConfigNames` for a reachable zero-tool server; this branch answers `200 {"tools":[]}` and logs nothing at `Error`.
- **OAuth retry**, against a live OAuth MCP server with its access revoked mid-session: attempt 1 logs the auth failure, attempt 2 succeeds, 200 with 3 tools.

## Notes for the reviewer

- Neither log line carries `runId` / `stepId`: `RemoteToolFetcher` and the `AiClient` are both built once per executor with the process-level logger. Concurrent runs against the same connector are still distinguished only by server name and timestamp. Separate ticket.
- Two manual checks are still open, both needing a harness that doesn't exist yet — the example app never calls `addWorkflowExecutor`: the embedded-agent rendering of the same failure, and a host logger that throws passed via `addWorkflowExecutor({ logger })`. The latter is covered by unit test, not end to end.
- The package `CLAUDE.md` gained the Logging clause and the failures-channel invariant, per that file's own keep-current instruction; the README gained a "When an MCP step fails to load its tools" section.

fixes PRD-876

> [!NOTE]
> ### Surface MCP tool load failure causes in workflow executor logs
> - Switches `RemoteToolFetcher` from `loadRemoteTools` to `loadRemoteToolsWithFailures`, so failures are determined from provider-reported data rather than inferred from an absent tool list; an empty tool list from a healthy server is no longer treated as a failure.
> - Logs structured failure details (server, kind, error message) via `errorOnPartialLoadFailure()`.
> - Adds a new [`to-ai-proxy-logger`](https://github.com/ForestAdmin/agent-nodejs/pull/1806/files#diff-f0f54109ed0ebf5b452a7a1dea54b24f446431a56f32467b602cab19b399ed24) bridge that flattens `Error` objects into readable fields and guards against host logger throws before delegating to the executor logger.
> - Wires the executor logger into `AiClientAdapter` and `ServerAiAdapter` so ai-proxy internal logs (including tool load diagnostics) are forwarded by default.
> - Behavioral Change: OAuth servers with a rejected cached token now trigger a forced token refresh and retry; persistent auth failures throw `OAuthReauthRequiredError`; a successful retry logs an Info-level recovery message.
>
> 
> #### Changes since #1806 opened
>
> - Added `loadToolsWithFailures` method to `ForestIntegrationClient` class in `ai-proxy` package [14ee7b0]
> - Added test coverage for `loadToolsWithFailures` method in `ai-proxy` package [14ee7b0]
> - Modified `RemoteToolFetcher.logFailures` to extract and log error causes using utility functions [14ee7b0]
> - Updated test expectations for failure logging to verify cause extraction [14ee7b0]
> - Fixed `toAiProxyLogger` function to omit nullish cause values instead of stringifying them [14ee7b0]
> - Added test coverage for `toAiProxyLogger` null cause handling [14ee7b0]
> - Updated MCP server failure documentation in `workflow-executor` package README [14ee7b0]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 1c2aabc.</sup>
> 
>


packages/workflow-executor/src/adapters/to-ai-proxy-logger.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { extractErrorMessage } from '../errors';
+
+// ai-proxy hands the cause as an Error, the executor's logger expects a context object.
+export type AiProxyLogger = (level: LoggerLevel, message: string, error?: Error) => void;
+
+// An Error's own properties are non-enumerable, so forwarding it as the context would emit the line
+// with the cause silently stripped — flatten it the way the rest of the executor logs causes.
+export default function toAiProxyLogger(logger: Logger): AiProxyLogger {
+  return (level, message, error) => {
```
hercemer42, line 11: Claude Opus 5 (claude-opus-5[1m]): **Should fix**

**Applies to:** the failure mode lives at `packages/ai-proxy/src/mcp-client.ts:113` (not in this diff); the fix belongs here, in the wrapper body.

A host logger that throws now aborts the entire tool load, and this PR is what makes that reachable. `mcp-client.ts:113` calls `this.logger?.(...)` as the **first** statement of its per-server `catch`, before `failures.push` at `:114`. A throw there rejects the `Promise.all` at `:92`, so `loadRemoteToolsWithFailures` rejects and three things break at once: tools already collected from healthy servers are discarded (defeating the per-server resilience the class exists for), `remote-tool-fetcher.ts:105` never evaluates `hasAuthFailure` so a revoked token that should pause the step with `OAuthReauthRequiredError` becomes a generic step error, and the original MCP error is replaced by the logger's own error as the `AiModelPortError` cause.

Before this PR `this.logger` was undefined on this path, so the throw was impossible. `ExecutorOptions.logger` (`build-workflow-executor.ts:67`) is host-supplied public API and `ports/logger-port.ts` documents no must-not-throw contract. The codebase already guards the neighbouring edge for exactly this reason — `agent/src/embedded-workflow-executor.ts:12`: *"Never throws — logging must not crash the executor"*, with the try/catch at `:17`.

No in-repo logger can trigger it (`console-logger` / `pretty-logger` only stringify two strings), so this is severity-not-likelihood. Wrapping this closure's body in try/catch covers every ai-proxy emit site at once and keeps "logging never changes control flow" true at the boundary where an untrusted callback enters.

packages/workflow-executor/test/adapters/server-ai-adapter.test.ts
```diff
@@ -152,4 +155,54 @@ describe('ServerAiAdapter', () => {
...
+
+      aiProxyLoggerGivenToLatestClient()?.('Error', 'Error loading tools for notion', cause);
+
+      expect(executorLogger).toHaveBeenCalledWith('Error', 'Error loading tools for notion', {
+        error: '401 Unauthorized',
+        stack: cause.stack,
+      });
+    });
+
+    it('also gives the per-call AiClient built by getModel the executor logger', () => {
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): **Should fix**

This test's scenario cannot happen. The per-call client is built from a single config (`buildProxyConfiguration` always names it `forest-server`) and `getModel` destructures only `{ userId }`, so `client.getModel()` is always called with no `aiName` → `get-ai-configuration.ts:11` `if (aiName)` is false and the fallback warning is never emitted. That client is discarded right after `getModel`, so its other two logger sites (`createToolProviders`, `disposeToolProviders`) are unreachable too — `logger: this.aiProxyLogger` at `src/adapters/server-ai-adapter.ts:41` is inert today.

It is not vacuous (delete `:41` and it fails), but it reports coverage for a production path that does not exist. Either reword it as an explicit wiring check, or drop `:41` and this test and let the constructor client — the one that actually reaches `createToolProviders`, and therefore every MCP diagnostic this ticket is about — carry the logger alone.

packages/workflow-executor/src/build-workflow-executor.ts
```diff
@@ -119,9 +119,9 @@ function buildCommonDependencies(options: ExecutorOptions) {
   if (forceAiError) {
     aiModelPort = new AlwaysErrorAiModelPort();
   } else if (options.aiConfigurations?.length) {
-    aiModelPort = new AiClientAdapter(options.aiConfigurations);
+    aiModelPort = new AiClientAdapter(options.aiConfigurations, logger);
```
hercemer42, line 122: Claude Opus 5 (claude-opus-5[1m]): **Preferential**

**Applies to:** `packages/workflow-executor/src/remote-tool-fetcher.ts:105-124` (not in this diff). Anchored here because this is the line that turns those diagnostics on.

A run that self-heals will now log at `Error`. On the OAuth path `attemptLoad(false)` runs with the cached token; if the server rejects it (revoked credential, scope change, clock skew) `mcp-client.ts:113` and `:127` emit two `Error` lines, then `remote-tool-fetcher.ts:122` takes `reloadWithFreshAuth()`, the refresh succeeds and the step goes green. The executor deliberately does **not** call `errorOnPartialLoadFailure` for that first attempt — it treats it as expected and recoverable — while ai-proxy one layer down now shouts about it, and nothing marks the line as recovered.

Not necessarily wrong: business rule 3 does want tool-load failures at `Error`. But it is a decision worth making explicitly rather than inheriting — accept the noise, or stamp an `attempt` / `recoverable` key into the context so alerting can filter runs that recovered.

packages/workflow-executor/src/adapters/to-ai-proxy-logger.ts
```diff
@@ -0,0 +1,19 @@
...
+    if (error === undefined || error === null) return logger(level, message);
+
+    return logger(level, message, {
+      error: extractErrorMessage(error),
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): **Preferential**

The cause chain stops at the outer message. `extractErrorMessage` returns on the first non-empty `.message` (`errors.ts:18`) and `Error.prototype.stack` does not render `[cause]`, so a `TypeError: fetch failed` logs as `error: "fetch failed"` with the ECONNREFUSED / ENOTFOUND / TLS detail gone — which is exactly the "unreachable server vs 401" distinction this ticket is buying. `errors.ts:7` already exports `causeMessage` for this and nothing here uses it, and `base-step-executor.ts:84-85` logs `cause` alongside `error` / `stack` for the same reason.

`cause: causeMessage(error)` would match that house shape for one more key. Mitigating: on the main `getTools()` path `@langchain/mcp-adapters` interpolates the nested text into its own message, so the detail often survives anyway — this bites hardest on the dispose / cleanup path, where `mcp-client.ts:174` also passes an unnormalised rejection `reason` (a plain object there degrades to `"[object Object]"` with no stack, unlike its siblings at `ai-client.ts:85` and `router.ts:116` which normalise first).

packages/workflow-executor/src/build-workflow-executor.ts
```diff
@@ -119,9 +119,9 @@ function buildCommonDependencies(options: ExecutorOptions) {
...
   } else {
-    aiModelPort = new ServerAiAdapter({ forestServerUrl, envSecret: options.envSecret });
+    aiModelPort = new ServerAiAdapter({ forestServerUrl, envSecret: options.envSecret, logger });
```
hercemer42, line 124: Claude Opus 5 (claude-opus-5[1m]): **Preferential**

**Applies to:** the PR as a whole — the `AiClient` is constructed once per executor, here and at `:122`.

The bridged lines carry no `runId` / `stepId`. `BaseStepExecutor` stamps `logCtx` on every other executor line, but this logger is the process-level one, so `Error loading tools for notion` lands with `{ error, stack }` and nothing else, while the executor's own `MCP servers failed to load tools` (`remote-tool-fetcher.ts:174`) carries the run context and no cause. Two concurrent runs against the same connector — one user's OAuth token revoked, the other's fine — produce two contextless cause lines and two causeless context lines, and joining them means matching on server name plus timestamp.

Out of scope to fix here, and it does match the neighbouring `RemoteToolFetcher`, which also takes the bare logger. Noting it because per-run attribution is what turns these lines into a support answer: the cheap follow-up is not re-plumbing the logger but logging at `remote-tool-fetcher.ts:105`, which already receives the classified `failures` (each with `server`, `kind`, `error`) and keeps only `.kind`.

packages/workflow-executor/CLAUDE.md
```diff
@@ -53,7 +53,7 @@ Front  ◀──▶  Orchestrator  ◀──pull/push──▶  Executor (this p
...
 - **Graceful shutdown** — `stop()` drains in-flight steps (`idle → running → draining → stopped`), `stopTimeoutMs` default 30s, HTTP stays up during drain. Signal handling is the consumer's job.
-- **Logging** — `Logger = (level, message, context?) => void`. `BaseStepExecutor` stamps `logCtx` (runId/stepId/stepIndex/stepType); type-specific ids via `getExtraLogContext()`. `createConsoleLogger`/`createPrettyLogger(minLevel)` factories; CLI level from `LOG_LEVEL` (default `Info`).
+- **Logging** — `Logger = (level, message, context?) => void`. `BaseStepExecutor` stamps `logCtx` (runId/stepId/stepIndex/stepType); type-specific ids via `getExtraLogContext()`. `createConsoleLogger`/`createPrettyLogger(minLevel)` factories; CLI level from `LOG_LEVEL` (default `Info`). ai-proxy's logger takes the cause as a third `Error` argument instead of a context object, so both AI adapters bridge it with `toAiProxyLogger` (flattens to `{ error, stack }` — an `Error`'s own properties are non-enumerable and would vanish from the emitted line).
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): **Preferential**

**Applies to:** the ecosystem pack, `FOREST-ARCHITECTURE.md:546` (outside this repo). Anchored here because this is the documentation line the PR does update.

The pack still tells readers that ai-proxy's *"remote-tool load failures are silently swallowed per-server — a misconfigured MCP server surfaces as an empty tool list (`NoMcpToolsError`), not a loud error."* The control-flow half stays true (still caught per server, still a possibly-empty tool list, never rethrown), but the "silent" half was always conditional on no logger being injected, and this PR injects one on the workflow-executor path. Worth correcting while it is fresh: *"…and silent only when no `logger` is injected — the workflow-executor path now injects the host logger, so load failures surface in its logs."*

Same section, if you are editing anyway: L133 lists ai-proxy as used by "agent and SaaS" and omits workflow-executor as a third in-process consumer, and the Key contracts table has no row for the cross-package logger signature — `datasource-toolkit`'s `(level, message, error?)` versus this package's `(level, message, context?)`, which is the mismatch this PR had to bridge.

hercemer42 (commented):
Automated validator pass (Claude Opus 5, `claude-opus-5[1m]`) — 6 inline findings: 2 Should fix, 4 Preferential. No approval implied; the spec check against PRD-876 conforms for this PR's half (AC#4, AC#5). Submitted rather than left pending so the findings are visible in the workflow.

packages/workflow-executor/src/adapters/to-ai-proxy-logger.ts
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Agreed, and fixed in b7b165c2.

Verified the ordering before acting on it: `mcp-client.ts:113` is the first statement of the per-server `catch`, `failures.push` is `:114`, and the `Promise.all` is `:92` — so a throw did abort the whole load, and `remote-tool-fetcher.ts:105` never reached `hasAuthFailure`. The wrapper body is now inside a try/catch with the throw swallowed on purpose, which puts the guard at the boundary where the host callback enters ai-proxy and covers every emit site at once instead of per call site.

New test: *"keeps the throw away from ai-proxy, with and without a cause"* — it asserts both branches and that the logger was still attempted twice. File coverage stays 100%.

packages/workflow-executor/test/adapters/server-ai-adapter.test.ts
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Half agreed — the test is fixed in b7b165c2, the src line stays.

You are right about reachability, and I confirmed it rather than taking it on trust: `buildProxyConfiguration` always names the single config `forest-server`, `getModel` destructures only `{ userId }`, so `client.getModel()` runs with no `aiName` and `get-ai-configuration.ts:11` never enters the branch that logs. The test now reads *"wires the same logger into the per-call AiClient built by getModel"*, uses a neutral message, and carries a one-line comment saying it is a wiring check because that client reaches none of `AiClient`'s own emit sites. No more implied coverage of a path that does not exist.

Pushing back on dropping `server-ai-adapter.ts:41`. "Inert today" is not the same as wrong: the alternative leaves one `new AiClient` in this class that gets a logger and one that does not, which is the asymmetry that becomes a silent gap the moment `getModel` forwards an `aiConfigName` or that client is reused for anything else. It costs one property reference, and it is the same value the constructor client already holds.

packages/workflow-executor/src/build-workflow-executor.ts
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Confirmed the path, declining to act on it here.

`remote-tool-fetcher.ts:122` does return into `reloadWithFreshAuth()` before `errorOnPartialLoadFailure` runs, so yes — a rejected cached token produces two ai-proxy `Error` lines on a run that then goes green.

Two reasons not to change it in this PR. First, it is not clearly wrong: a rejected or revoked token **is** a tool-load failure, business rule 3 asks for those at `Error`, and "token rejected at 14:02, refreshed, continued" is information support wants — the silence there is what this ticket exists to fix. Second, neither fix is reachable from here: demoting the level means touching `mcp-client.ts`, which business rule 4 rules out, and stamping an `attempt` / `recoverable` key needs the retry state, which lives in `RemoteToolFetcher` and never reaches a logger built once per executor. That belongs with the follow-up in the sibling comment on `:124`.

packages/workflow-executor/src/adapters/to-ai-proxy-logger.ts
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Agreed, and fixed in b7b165c2.

The payload now carries `cause`, read the same way `base-step-executor.ts:94` reads it — `extractErrorMessage` on the destructured `.cause` rather than `causeMessage`, so a non-`Error` cause and an empty-message cause still resolve. The new test wraps `fetch failed` over an `ECONNREFUSED` cause, which is exactly the unreachable-server-versus-rejected-token distinction this ticket is buying. `.cause` is read through a cast because the repo targets ES2020, where `ErrorOptions` is not in the lib — the same cast the base executor uses.

Left the dispose-path `"[object Object]"` note alone: that one originates at `mcp-client.ts:174` passing an unnormalised rejection reason while its siblings normalise first, so it is ai-proxy's to fix, and `String(err)` is this package's convention everywhere else.

packages/workflow-executor/src/build-workflow-executor.ts
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Accurate, and deliberately not fixed here.

The `AiClient` is built once in `buildCommonDependencies`, so its logger is the process-level one and structurally cannot carry the per-step `logCtx` that `BaseStepExecutor` stamps. Fixing that properly means either an `AiClient` per step or a contextual logger — both real design changes, neither in this ticket, and both a worse first move than your own suggestion: `remote-tool-fetcher.ts:105` already receives `failures` with `server` / `kind` / `error` and keeps only `.kind`, so logging them there puts cause and run context on one line for a couple of lines of code.

Worth its own ticket rather than widening this one. Recording the consequence honestly: until then, attributing a cause to a run means matching on server name plus timestamp.

packages/workflow-executor/CLAUDE.md
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Agreed the pack line is stale, but it cannot be fixed from this PR.

`FOREST-ARCHITECTURE.md` lives in the `claudine` repo (installed at `~/.claude/skills/ecosystem/`), not in agent-nodejs, so the correction — plus the two adjacent gaps, L133 omitting workflow-executor as an in-process ai-proxy consumer and the missing Key-contracts row for the cross-package logger signature — needs its own change there. Flagging it to the pack owner rather than leaving it silently: the "silently swallowed" wording was already conditional on no logger being injected, so it was imprecise for the agent path before this PR too.

hercemer42:
**Claude Opus 5 (`claude-opus-5[1m]`): review resolution** — 3 of 6 findings applied in b7b165c2, 3 declined with reasoning in the threads.

**Applied**
- *A throwing host logger could abort a whole tool load.* The bridge now guards its body, so logging cannot become control flow at the boundary where a host callback enters ai-proxy. New test covers the cause and no-cause paths.
- *The cause chain stopped at the outer message.* The payload carries `cause`, read the way `base-step-executor.ts:94` does. New test wraps `fetch failed` over `ECONNREFUSED` — the unreachable-server-vs-rejected-token distinction this ticket is buying.
- *A test asserted a fallback warning the per-call client cannot emit.* Reworded as an explicit wiring check with a neutral message and a comment saying so.

**Declined**
- *Drop `logger` from the `getModel` client (`server-ai-adapter.ts:41`).* Inert today, but removing it leaves one `AiClient` in the class with a logger and one without — the asymmetry that becomes a silent gap the moment that client is reused.
- *Recovered-OAuth `Error` noise.* Real and confirmed, but a rejected token is a genuine tool-load failure that business rule 3 wants at `Error`, and neither fix is reachable without touching `mcp-client.ts` (business rule 4 forbids it) or the retry state that never reaches this logger.
- *Missing `runId`/`stepId`, and the stale ecosystem-pack line.* Both accurate, neither fixable here: the first needs a design change or the `remote-tool-fetcher.ts:105` follow-up, the second lives in the `claudine` repo.

**Checks after the fix**: 4 suites / 78 tests green, `to-ai-proxy-logger.ts` still 100% covered, eslint + prettier + tsc clean.

Paused for a human call on the three declines before this leaves draft.

packages/workflow-executor/src/build-workflow-executor.ts
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Half of this is now done in 16f90896, after the ticket author flagged the same gap from the other direction.

The `remote-tool-fetcher.ts` suggestion is implemented: the main load path reads `loadRemoteToolsWithFailures`, and `MCP servers failed to load tools` now carries `failures: [{ server, kind, error }]` instead of an id-diff. That also killed the false positive the diff could not avoid — a healthy server exposing no tools was reported as failed and answered 503 on the tool-listing endpoint.

Still open, and unchanged by that commit: neither line carries `runId` / `stepId`, since `RemoteToolFetcher` and the `AiClient` are both built once per executor with the process-level logger. Concurrent runs against one connector remain distinguishable only by server name and timestamp. Separate ticket, as above.

Correction to my own comment: I wrote that the executor's line "carries the run context". It does not — `runner.ts:73-78` hands `RemoteToolFetcher` the process-level logger, so that line carries *server* context (`requestedMcpServerId`, `mcpServerName`) only.

hercemer42:
**Claude Opus 5 (`claude-opus-5[1m]`): review closed out.** All six threads answered and resolved; CI green on `16f90896` — the whole rollup, after re-running `LLM Integration Tests (ai-proxy)`, which had failed on an upstream OpenAI 500 for `gpt-3.5-turbo` in its live-API tool-support test (identical failure on all three commits of this branch, including the first; no ai-proxy file is touched here). Supersedes my earlier resolution comment, which predates the widened scope.

**Fixed**
- Throwing host logger could abort a whole tool load → the bridge guards its body (`b7b165c2`).
- Cause chain stopped at the outer message → the payload carries `cause`, read the way `base-step-executor.ts:94` does (`b7b165c2`).
- A test asserted a message the per-call `getModel` client cannot emit → reworded as an explicit wiring check (`b7b165c2`).
- Cause and run context on separate lines → half of this is now moot: `remote-tool-fetcher.ts` reads the failures channel, so `MCP servers failed to load tools` carries `{ server, kind, error }` instead of an id-diff (`16f90896`).

**Declined, reasoning in the threads**
- Dropping `logger` from the `getModel` client: inert today, but removing it leaves one `AiClient` in the class with a logger and one without.
- Recovered-OAuth `Error` noise: a rejected token is a genuine tool-load failure that business rule 3 wants at `Error`, and neither fix is reachable without editing `mcp-client.ts` (business rule 4) or the retry state.
- Stale ecosystem-pack line: accurate, but it lives in the `claudine` repo.

**Still open, deliberately**
- Neither log line carries `runId` / `stepId` — `RemoteToolFetcher` and the `AiClient` are both built once per executor. Concurrent runs against one connector are distinguished only by server name and timestamp. Own ticket.

**Checked and dismissed**: whether the failures-channel switch could regress a config map holding two entries under one server id (a partial failure would now 503 where it used to return the working subset). It cannot happen — `ai-service.ts:773` merges two disjoint queries on `aiMcpConfigs` (`isForestConnector` false/true), each entry's `id` is its own row id, and a name collision overwrites rather than duplicates. No guard added.

hercemer42:
**Claude Opus 5 (`claude-opus-5[1m]`)**: qlty's *Function with many returns (count = 4): `fetchOAuthTools`* is fixed in b1a1fe97 — the two tail exits differed only in which attempt produced the tools, so the rejected-token case is named once and the retry's result read once. Three returns now; qlty reports **All good ✅**.

Worth recording that it wasn't introduced here: `origin/main` has the same four returns in that function. It surfaced as a new issue because this PR edited the body, and the analyser attributes findings to changed lines. Fixed rather than argued, since it's two lines either way.

hercemer42:
**Claude Opus 5 (`claude-opus-5[1m]`): manual testing on a live rig found one gap, fixed in db5ba656.**

Ran the executor from this branch (pm2 `agent-stack`, `tsx watch cli.ts --pretty`) against the OAuth sim, with a credential deposited through `POST /mcp-oauth-credentials` and the server's access tokens revoked mid-flight. The run recovers — HTTP 200, 3 tools — but the log read as a pure failure:

```
error Error loading tools for sim-mcp b  error="Authentication failed for HTTP server…"
error Failed to load tools from 1/1 MCP server(s): sim-mcp b: Authentication failed…
debug Loaded 3 tools from MCP server "sim-mcp b" in 26ms      ← the recovery, invisible at Info
```

The retry that fixes it only logged at `Debug`, so at the production default an operator saw the two `Error` lines and nothing saying it continued. `RemoteToolFetcher` now states the recovery at `Info` after a forced refresh succeeds:

```
info  MCP tools loaded after refreshing the credential requestedMcpServerId="34" mcpServerName="sim-mcp b"
```

This reverses my earlier decline on the recovered-OAuth thread. I argued there that "token rejected at 14:02, refreshed, continued" was information support wants — that was wrong: support only ever saw the rejection. The noise itself stays (the failure is real, and business rule 3 wants it at `Error`), but it is no longer misleading.

packages/workflow-executor/src/remote-tool-fetcher.ts
```diff
@@ -102,33 +106,33 @@ export default class RemoteToolFetcher {
...
+      this.errorOnPartialLoadFailure(attempt.failures, mcpServerId, mcpServerName);
+      // The rejected credential was already logged at Error. Without this the default level shows
+      // the failure and never says it recovered.
+      this.logger('Info', 'MCP tools loaded after refreshing the credential', {
+        requestedMcpServerId: mcpServerId,
+        mcpServerName,
+      });
 
       return attempt.tools;
     };
```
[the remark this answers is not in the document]
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Confirmed and fixed in 71e1e25e.

`errorOnPartialLoadFailure`'s return value was discarded inside `reloadWithFreshAuth`, and `loadFailed` was hard-coded `undefined` on the rejected-token path — so a retry that failed to connect surfaced as `200` with an empty tool list instead of `503`. It also made the recovery line lie: it claimed the refresh had worked while the retry's own failure had just been logged at `Error`.

The reload hook is handed to the caller for its own post-401 retry, so its `Promise<RemoteTool[]>` signature stays; it now reports the retry's outcome through the enclosing scope, and the recovery line is emitted only when the retry really had no failures.

Test added — first attempt an auth failure, retry a `connection` failure → `loadFailed === true` and no recovery line. 168 tests green across the fetcher, runner and HTTP server suites.

The behaviour predates this PR (`origin/main` also omitted `loadFailed` there), but it contradicts exactly what this PR is for, so it belongs here rather than in a follow-up.

PMerlet (commented):
Adversarial pass on the fetcher + logger bridge. The OAuth retry logic is sound (and fixes the `loadFailed`-left-`undefined` hole `main` had on the rejected-token path), tests assert full payloads, CI is green. Three inline points below: one contract gap worth addressing before merge (or an explicit follow-up ticket, since the fix lives in ai-proxy), two one-liners in files already touched.

packages/workflow-executor/src/remote-tool-fetcher.ts
```diff
@@ -155,26 +166,23 @@ export default class RemoteToolFetcher {
...
     mcpServerId: string,
     mcpServerName: string | undefined,
   ): boolean {
-    const loadedMcpServerIds = new Set(tools.map(t => t.mcpServerId));
-    const failedConfigNames = Object.entries(scoped)
-      .filter(([, cfg]) => !loadedMcpServerIds.has(cfg.id))
-      .map(([name]) => name);
-
-    if (failedConfigNames.length === 0) return false;
+    if (failures.length === 0) return false;
```
PMerlet, line 176: **The `failures` channel is only populated by `McpClient`** — in `AiClient.loadRemoteToolsWithFailures`, a provider without `loadToolsWithFailures` contributes `failures: []`, and `ForestIntegrationClient` doesn't implement it. Worse, its `loadTools` swallows an unsupported `integrationName` with a `Warn` and returns no tools.

So for a Forest connector whose integration name isn't recognized (version drift: the orchestrator advertises an integration this ai-proxy build doesn't support yet), the old id-diff answered `loadFailed=true` → 503 + Error log, while this now reads as a healthy empty server → `200 {"tools":[]}`. That's a different case from the false positive this PR legitimately removes — and it's exactly the drift scenario where you'd want the failure surfaced. (Mitigation: the `Warn` is at least visible now that the logger is wired — but at `Warn`, and without the 503.)

Note the unit test that covers this (`makeFailure('zendesk-prod', 'unknown', 'Unsupported integration: Zendesk')`) mocks a failure the real ai-proxy can never emit for a Forest connector, so it encodes a contract the upstream implementation doesn't honor. The new CLAUDE.md invariant ("never infer failure from absent tools") codifies a rule with this known hole.

Suggested fix: give `ForestIntegrationClient` a `loadToolsWithFailures` that pushes a `kind: 'unknown'` failure from the `default:` branch of its switch — or a follow-up ticket, since this PR deliberately doesn't touch ai-proxy.

packages/workflow-executor/src/remote-tool-fetcher.ts
```diff
@@ -155,26 +166,23 @@ export default class RemoteToolFetcher {
...
 
     this.logger('Error', 'MCP servers failed to load tools', {
       requestedMcpServerId: mcpServerId,
       mcpServerName,
-      failedConfigNames,
+      failures: failures.map(failure => ({
+        server: failure.server,
+        kind: failure.kind,
+        error: failure.error.message,
```
PMerlet: `failure.error.message` drops the cause chain: Node's `fetch failed` errors carry the actual reason (`ECONNREFUSED`, …) in `.cause`, and wrapped infra errors can have an empty `.message` — the exact cases `extractErrorMessage`/`causeMessage` in `errors.ts` exist for. The bridged ai-proxy line does name the cause, but this structured summary is the greppable line the PR body presents as the deliverable, and here it would say `error: "fetch failed"` alone.

```suggestion
        error: extractErrorMessage(failure.error),
        cause: causeMessage(failure.error),
```

packages/workflow-executor/src/adapters/to-ai-proxy-logger.ts
```diff
@@ -0,0 +1,32 @@
...
+        logger(level, message);
+
+        return;
+      }
+
+      const { cause } = error as { cause?: unknown };
+
+      logger(level, message, {
+        error: extractErrorMessage(error),
+        cause: extractErrorMessage(cause),
```
PMerlet: `extractErrorMessage` only returns `undefined` for `undefined`; for `null` it falls through to `String(null)` → the string `"null"`. So `new Error('x', { cause: null })` emits `cause="null"` in pretty output — the same artifact this PR removes for `undefined`, in another guise (the pretty-logger filter also only tests `!== undefined`).

```suggestion
        cause: cause == null ? undefined : extractErrorMessage(cause),
```

packages/workflow-executor/src/remote-tool-fetcher.ts
hercemer42: Confirmed, and fixed here rather than deferred — `8f00c79d`.

You're right on every link of it: `ai-client.ts:57` hands back `failures: []` for any provider without `loadToolsWithFailures`, `ForestIntegrationClient` didn't implement it, and its `default:` branch only warned. So an integration name this build doesn't recognise answered `200 {"tools":[]}` where main's id-diff answered 503 — and that's the version-drift case, not the false positive this PR set out to remove.

I'd originally have taken your follow-up-ticket option. Talking it through with the author, that didn't hold up: this PR *created* the regression, so deferring means shipping a known downgrade plus a `CLAUDE.md` invariant that is false the day it lands. The business-rule-4 objection doesn't cover it either — PRD-863's timeout, classification and channel are all untouched, `loadToolsWithFailures` is already optional on `ToolProvider` with `ai-client.ts:57` branching on its presence, so this is additive and no existing call path changes. AC#5 holds literally: no existing ai-proxy test needed an edit.

`ForestIntegrationClient` now mirrors `McpClient`'s shape — the switch moved into `loadToolsWithFailures`, the `default:` branch pushes `{ server: integrationName, mcpServerId, kind: 'unknown', error }` alongside the existing `Warn`, and `loadTools` delegates to it.

One thing I deliberately did **not** do: wrap the `getZendeskTools`/`getKolarTools`/`getSnowflakeTools` calls in a try/catch. Those can throw today and that propagates to the caller; catching would be a real behaviour change on an existing path and could quietly invalidate tests that assert it. Only the `default:` branch changed.

Good catch on the test too — `makeFailure('zendesk-prod', 'unknown', …)` was mocking a failure the real provider could never emit, so it read as coverage for exactly this hole while proving nothing. It now asserts the shape the provider really produces (`server` is the integration name, not the Record key), and there are three new ai-proxy tests covering the unsupported case, the all-supported case, and a supported connector still contributing its tools alongside a broken one.

packages/workflow-executor/src/remote-tool-fetcher.ts
hercemer42: Agreed — taken as suggested, in `14ee7b07`.

The asymmetry was the tell: the bridge this PR adds reads causes with `extractErrorMessage`, and then the structured line the PR body presents as the deliverable read `.message` raw. A Node `fetch failed` reduced to `error: "fetch failed"` on the one line I was asking people to grep.

Now `error: extractErrorMessage(failure.error)` and `cause: causeMessage(failure.error)` — which also gives `causeMessage` its first caller. Added a test that feeds a `fetch failed` wrapping an `ECONNREFUSED` and asserts both fields land separately.

Worth noting for anyone reading later: in practice most MCP failures still won't populate `cause`, because `@langchain/mcp-adapters` interpolates the nested text into its own message rather than chaining. That's why this looked adequate in the rig logs, and it's exactly why it shouldn't be relied on — the cases where the chain *is* real are the infra ones you most need it for.

packages/workflow-executor/src/adapters/to-ai-proxy-logger.ts
hercemer42: Correct, and taken — `14ee7b07`.

Verified the mechanism: `extractErrorMessage` only short-circuits on `undefined`, so `null` falls to `String(null)` and emits `cause="null"`; and `743d48ab`'s pretty-logger filter tests `!== undefined`, so it wouldn't catch it either. Same artifact, second guise, as you say.

Pushing back mildly on severity rather than on the fix: reachability is low. It needs someone to set `cause: null` explicitly — neither Node's `fetch` nor `@langchain/mcp-adapters` does, they either omit `cause` or set an `Error`. So this is a cost argument, not a risk one: the guard is one token and the test is five lines, which is cheaper than leaving a known-wrong branch in a file whose whole job is not mangling causes.

Used `cause == null ? undefined : extractErrorMessage(cause)` as suggested, with a test asserting the key is omitted rather than stringified.

PMerlet (approved):
LGTM

PR #1810, opened 2026-08-06
feat(agent-bff): serve the static OpenAPI document behind the agent gate

Tonours:
**Stacked on #1809** (`chore/dedupe-zod-single-v4-resolution`). Base is that branch, not `main`; #1809 must be reviewed and merged first. Hard prerequisite: `@asteasolutions/zod-to-openapi` takes zod as a peer dependency, so without the dedupe yarn can bind it to a different zod copy than the one agent-bff builds its schemas from.

fixes PRD-886

## What

`GET /agent/openapi.json` serves an OpenAPI 3.1 document describing the six generic BFF routes (list, count, relation list/count, action form/execute). Generated from Zod, built once at boot, static: path placeholders only, no tenant data. The only environment-derived value is the package version.

## Why this mount point

The BFF has no deny-by-default gate: auth applies only to paths under `/agent`. Mounting the route inside the `agentScoped` chain, right after the auth middlewares, inherits the existing gate, so an unauthenticated request gets a 401 before path matching. Root mounting would have made the document public. The route sits before `buildAgentRouteMiddlewares`, so it answers even when `AGENT_URL` is unset and never calls the agent.

## How

- New `src/openapi/` module: `extendZodWithOpenApi` loaded once, request/response/error schemas, registry plus `OpenApiGeneratorV31`, and a `serializeOpenApi` shared with the future CLI subcommand.
- Schemas written against the code, not prior docs: `/agent/v1` prefix, recursive condition-tree filter, PascalCase operators (required), `sort[].direction` optional, `page.offset` must be a multiple of `limit`, `parentId` string or number, `recordIds` required even when empty.
- Every status a route actually returns is declared, including 502 (transport failure), 413 (16kb body cap), 429 (agent rate-limiting). 501 is declared only on execute and points at a separate schema because that body carries no `message` field.
- Additive: the only edit outside `src/openapi/` is one line in the `cli-core.ts` middleware chain. No existing route, handler, or schema changed.

Known limits: action `200` responses are typed `unknown` (belongs with the action work); relation list/count skip capabilities validation, so their 422 field errors are unreachable, and the `ConditionTree` description says so (#1801 fixes the validation gap). Per-collection unfolding is PRD-684/PRD-685.

## How to test

```bash
```

Against a running agent:

```bash
FOREST_AUTH_SECRET=... FOREST_ENV_SECRET=... FOREST_SERVER_URL=... \
FOREST_APP_URL=... AGENT_URL=http://localhost:3009/forest HTTP_PORT=4599 \
node packages/agent-bff/dist/cli.js

curl -i localhost:4599/agent/openapi.json                        # 401, no document
curl -i localhost:4599/openapi.json                              # 404
curl -s -H "X-Forest-Bff-Key: <session>" \
  localhost:4599/agent/openapi.json | npx @redocly/cli lint -    # valid
```

Ran on this branch: 767 tests pass (41 added), lint and build clean. `redocly lint` runs inside the suite, so the document cannot go invalid unnoticed. An end-to-end pass against a live agent covered the 401 gate (no leak in the body), the root 404, the authenticated 200, redocly validation over HTTP, and eight unauthenticated path variants (trailing slash, case change, `%6f`, dot segments, `..`, doubled slash): none serves the document.

### General

### Security

### DIFF
--- a/packages/agent-bff/src/openapi/openapi-routes.ts
+++ b/packages/agent-bff/src/openapi/openapi-routes.ts
@@ -1,0 +1,27 @@
+import type { Middleware } from 'koa';
+
+import { generateOpenApiDocument, serializeOpenApi } from './openapi-document';
+
+export const OPENAPI_PATH = '/agent/openapi.json';
+
+const READ_METHODS = new Set(['GET', 'HEAD']);
+
+export interface OpenApiRoutesOptions {
+  version: string;
+}
+
+export default function createOpenApiRoutes({ version }: OpenApiRoutesOptions): Middleware {
+  const document = serializeOpenApi(generateOpenApiDocument(version));
+
+  return async function openApiRoutes(ctx, next) {
+    if (!READ_METHODS.has(ctx.method) || ctx.path !== OPENAPI_PATH) {
+      await next();
+
+      return;
+    }
+
+    ctx.status = 200;
+    ctx.type = 'application/json';
+    ctx.body = document;
+  };
+}
--- a/packages/agent-bff/src/openapi/openapi-routes.ts
+++ b/packages/agent-bff/src/openapi/openapi-routes.ts
@@ -22,6 +22,7 @@ export default function createOpenApiRoutes({ version }: OpenApiRoutesOptions): 
 
     ctx.status = 200;
     ctx.type = 'application/json';
+    ctx.set('Cache-Control', 'no-store');
     ctx.body = document;
   };
 }
--- a/packages/agent-bff/package.json
+++ b/packages/agent-bff/package.json
@@ -31,6 +31,7 @@
     "test": "jest"
   },
   "dependencies": {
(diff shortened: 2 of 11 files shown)


packages/agent-bff/src/openapi/openapi-routes.ts
```diff
@@ -0,0 +1,27 @@
...
+  return async function openApiRoutes(ctx, next) {
+    if (!READ_METHODS.has(ctx.method) || ctx.path !== OPENAPI_PATH) {
+      await next();
+
+      return;
+    }
+
+    ctx.status = 200;
+    ctx.type = 'application/json';
+    ctx.body = document;
```
nbouliol, line 26: The API key path sets `Cache-Control: no-store`, this one doesn't — a shared cache could hold the gated document. Set it here too?

nbouliol (approved):
Clean additive change, the gate placement and test coverage are solid.

PR #1811, opened 2026-08-06
feat(agent-bff): emit the OpenAPI document from a forest-bff openapi subcommand

Tonours:
**Stacked on #1810** (`feature/prd-886-openapi-static-document`), which is itself stacked on #1809. The base is #1810's branch, not `main`: both parents must be reviewed and merged first, in order (#1809, then #1810, then this). The diff below is only this branch's commit.

Adds `forest-bff openapi`: writes the OpenAPI document to stdout, or to a file with `--output`, without booting the server.

fixes PRD-887

## Why a CLI at all

`GET /agent/openapi.json` is auth-gated, so a CI job, a codegen step, or a client dev without credentials has no HTTP path to the document. The CLI is that path. It needs no configuration:

```bash
forest-bff openapi > openapi.json          # stdout
forest-bff openapi --output                # writes ./openapi.json
forest-bff openapi --output docs/api.json  # writes that path, creating docs/
```

## argv contract

Previously `forest-bff <anything>` ignored argv and booted the server. Now argv is parsed:

| Invocation | Before | After |
|---|---|---|
| `forest-bff` | boots the server | boots the server, unchanged |
| `forest-bff openapi` | booted the server | document on stdout, exit 0 |
| `forest-bff openapi --output [path]` | booted the server | writes the file, exit 0 |
| `forest-bff --help` / `-h` | booted the server | usage on stdout, exit 0 |
| `forest-bff --version` / `-v` | booted the server | bare version on stdout, exit 0 |
| `forest-bff bogus` | booted the server | exit 1, reason on stderr |
| `forest-bff openapi extra` | booted the server | exit 1, names the extra |

`--help` and `--version` print to stdout, exit 0, and ignore what follows them: GNU tools treat a help request as a success, never an error. `openapi` is the strict one, because a typo there would otherwise produce a wrong export instead of a message.

A deployment that passed a stray flag and silently got a server will now fail on upgrade. The repo's own `start` scripts pass no arguments.

## --output parsing

The token after `--output` is the destination only when it exists, is non-empty, and does not start with `-`. Otherwise the default `openapi.json` applies and the leftover token is rejected as an extra. So `openapi --output --help` exits 1 and writes nothing, rather than creating a file named `--help`.

Extras are validated before any write, so an invalid command line never leaves an artifact behind. `--output=path` and a repeated `--output` are rejected the same way.

A missing parent directory is created. An existing file is overwritten, like `>` and every codegen exporter. An unwritable destination exits 1 with `Cannot write <path>: <reason>`, no stack trace.

The success line goes to **stderr**, not stdout: `--output` must leave stdout empty so the no-flag form stays pipeable, and so a CI job can keep redirecting without catching a log line.

## Stdout purity

An export piped to a file must contain only the document. This is easy to break silently: `createConsoleLogger` sends `Info` to `console.info`, which writes to **stdout**. Any future log on the export path would corrupt every pipe without failing anything.

Verified rather than assumed: a test pins one single `process.stdout.write` call and asserts `console.info` is never reached. The export path never calls `parseConfig`, never constructs a logger, never binds a socket, and `process.exitCode` is set instead of calling `process.exit()`, so a slow pipe cannot truncate the write.

## Scope and safety

The no-argv path is untouched: same `runCli`, same config parsing, same middleware chain. Hand-rolled argv switch in its own `cli-dispatch.ts`, no arg-parsing dependency.

`openapi.json` is added to `.gitignore`: `--output` has a default filename, and a dev running it from the package would otherwise commit the artifact by accident.

## How to test

```bash

cd packages/agent-bff
env -i PATH="$PATH" node dist/cli.js openapi | npx @redocly/cli lint -   # valid, no config
node dist/cli.js openapi --output && cat openapi.json | head -3          # 0, file written
node dist/cli.js openapi --output --help; echo $?                        # 1, no file created
node dist/cli.js --output; echo $?                                       # 1, points at openapi
```

Ran on this branch: 809 tests pass (777 before, 32 added), lint and build clean.

An end-to-end script drove the packaged binary through 40 checks, all passing: every argv path with its exit code and stream, the export under an empty environment, byte equality between the CLI export and the body the route serves, no socket bound, no truncation through a real pipe and a real redirect, `--output` writing the default name and an explicit path, a created parent directory, a dash token refused as a filename, and a directory destination failing with its path named.

## Known limitations

`DispatchOutcome.exitCode` is typed `number` and the type admits `{exitCode: 1, server}`, a combination the code never produces. Harmless today since the only consumer reads `exitCode` alone.

`--output` writes with `writeFileSync`, which truncates before writing. A mid-write failure such as `ENOSPC` leaves a partial file, same as `>` and `curl -o`.

### General

### Security

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,9 @@ lerna-debug.log
 .env.executors
 .forestadmin-schema.json
 
+# forest-bff openapi --output default destination
+openapi.json
+
 # yarn
 yarn-error.log
 .vscode/settings.json
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -5,8 +5,8 @@ agent from a browser without learning MCP or JSON:API.
 
 It is a bootable Koa 3 server with a `/health` endpoint, a version header, env-driven config
 validation, OAuth (Mode 1) + API-key (Mode 2) auth, and a hardened request edge (timezone, CORS,
-auth-mode precedence, structured error contract). The data-endpoint proxy and OpenAPI generation
-land in later slices.
+auth-mode precedence, structured error contract), and it serves and exports its own OpenAPI
+document. The data-endpoint proxy lands in a later slice.
 
 ## Usage
 
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -16,6 +16,25 @@ Packaged / production — run the bin:
 forest-bff
 ```
 
+Export the OpenAPI document without booting the server. Needs no configuration, so
+it works in CI to commit the document, diff it, or generate a client:
+
+```bash
+forest-bff openapi > openapi.json          # stdout, redirected
+forest-bff openapi --output                # writes ./openapi.json
+forest-bff openapi --output docs/api.json  # writes that path
+```
+
+`--output` takes the next argument as the destination unless it is empty or starts with
+`-`; the default `openapi.json` is written only when `--output` is the last argument, and
+a leftover token is rejected like any other extra. A missing parent directory is created,
+an existing file is overwritten, and a destination that cannot be written exits 1 with the
+path on stderr. The path is confirmed on stderr, so stdout stays empty and pipeable.
+
+`forest-bff --help` and `forest-bff --version` print to stdout and exit 0, ignoring
+anything that follows. An unknown command, or an argument other than `--output` after
+`openapi`, exits 1 with the reason on stderr.
+
 Local development — copy the env template, fill it in, and start with it loaded:
 
 ```bash
--- a/packages/agent-bff/src/cli-dispatch.ts
+++ b/packages/agent-bff/src/cli-dispatch.ts
@@ -1,0 +1,136 @@
+import type BFFHttpServer from './http/bff-http-server';
+import type { Logger } from './ports/logger-port';
+
+import { mkdirSync, writeFileSync } from 'fs';
+import path from 'path';
+
+import runCli from './cli-core';
+import { extractErrorMessage } from './errors';
+import { generateOpenApiDocument, serializeOpenApi } from './openapi/openapi-document';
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -15,6 +15,9 @@ lerna-debug.log
 .env.executors
 .forestadmin-schema.json
 
+# forest-bff openapi --output default destination
+openapi.json
```
nbouliol, line 19: This ignores any openapi.json in the monorepo, while the README suggests committing the exported document — scope it to packages/agent-bff/?

packages/agent-bff/README.md
```diff
@@ -16,6 +16,25 @@ Packaged / production — run the bin:
 forest-bff
 ```
 
+Export the OpenAPI document without booting the server. Needs no configuration, so
```
nbouliol, line 19: The intro above (line 8) still says OpenAPI generation lands in a later slice, just above this section.

packages/agent-bff/src/cli-dispatch.ts
```diff
@@ -0,0 +1,136 @@
...
+function writeOpenApiFile(file: string): DispatchOutcome {
+  const asDirectory = file.endsWith('/') || file.endsWith(path.sep);
+  const destination = path.resolve(
+    process.cwd(),
+    asDirectory ? path.join(file, DEFAULT_OUTPUT_FILE) : file,
+  );
+
+  try {
+    mkdirSync(path.dirname(destination), { recursive: true });
+    printOpenApi(chunk => writeFileSync(destination, chunk));
```
nbouliol: writeFileSync per chunk would silently keep only the last one if printOpenApi ever splits its output — what about a renderOpenApi() returning the string, used by both paths?

nbouliol (approved):

PR #1819, opened 2026-08-11
fix(agent): authorize create with linked one-to-one on the foreign collection

PMerlet:
## Problem

When a record is created with a linked `OneToOne` relation, the create route asserted `edit` on the **parent** collection, then performed two `update` calls on the **foreign** collection (null out the old owner's FK, write the linked record's FK).

A user with `add` + `edit` on the parent but no `edit` on the foreign collection could therefore rewrite the FK of an arbitrary foreign record and break another record's existing relation. The writes were already scope-bounded, but the collection-level `edit` permission on the foreign collection was bypassed entirely.

Same bug class as #1818 (authorize on the wrong side of the relation), on the write side.

## Fix

In the one-to-one linking branch of `create.ts`, assert `edit` on the foreign collection, mirroring `updateOneToOne` in `update-relation.ts`. The parent-side `assertCanAdd` at the start of the route is unchanged.

fixes PRD-916

> [!NOTE]
> ### Authorize edit on linked OneToOne foreign collections before creating parent record
> - Adds `assertCanEditLinkedOneToOneRelations` in [`create.ts`](https://github.com/ForestAdmin/agent-nodejs/pull/1819/files#diff-e88a452638f94fc3d1f83d5cb2c21a332ecde1ab6b0df16ca7eea27e653ee8dc) that checks edit permission on each linked OneToOne foreign collection before the parent record is created.
> - Moves authorization out of `linkOneToOneRelations` (where it incorrectly checked the parent collection) and into the earlier pre-creation step.
> - If any foreign collection edit check fails, neither the parent create nor the foreign update are executed.
>
> 
> #### Changes since #1819 opened
>
> - Added authorization checks in `CreateRoute` to verify edit permissions on foreign collections when creating records with linked one-to-one relations [fa336e0]
> - Refactored `CreateRoute` to extract one-to-one relation filtering logic into a reusable helper method [fa336e0]
> - Updated `CreateRoute.linkOneToOneRelations` method to process filtered one-to-one relations [fa336e0]
> - Extended test data source schema to include multiple relation types for authorization testing [fa336e0]
> - Updated and expanded authorization tests to verify edit permission checks across multiple one-to-one relations [fa336e0]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 21b0044.</sup>
> 
>

### DIFF
--- a/packages/agent/src/routes/modification/create.ts
+++ b/packages/agent/src/routes/modification/create.ts
@@ -1,4 +1,9 @@
-import type { CompositeId, ManyToOneSchema, RecordData } from '@forestadmin/datasource-toolkit';
+import type {
+  CompositeId,
+  ManyToOneSchema,
+  OneToOneSchema,
+  RecordData,
+} from '@forestadmin/datasource-toolkit';
 import type Router from '@koa/router';
 import type { Context } from 'koa';
 
@@ -80,18 +85,29 @@ export default class CreateRoute extends CollectionRoute {
     return record;
   }
 
+  private getLinkedOneToOneRelations(
+    relations: Record<string, RecordData>,
+  ): Array<{ linked: RecordData; relation: OneToOneSchema }> {
+    return Object.entries(relations)
+      .map(([field, linked]) => ({
+        linked,
+        relation: SchemaUtils.getRelation(this.collection.schema, field, this.collection.name),
+      }))
+      .filter(
+        (entry): entry is { linked: RecordData; relation: OneToOneSchema } =>
+          entry.linked !== null && entry.relation.type === 'OneToOne',
+      );
+  }
+
   private async assertCanEditLinkedOneToOneRelations(
     context: Context,
     relations: Record<string, RecordData>,
   ): Promise<void> {
-    const promises = Object.entries(relations).map(async ([field, linked]) => {
-      const relation = SchemaUtils.getRelation(this.collection.schema, field, this.collection.name);
-      if (linked === null || relation.type !== 'OneToOne') return;
-
-      await this.services.authorization.assertCanEdit(context, relation.foreignCollection);
-    });
-
-    await Promise.all(promises);
+    await Promise.all(
+      this.getLinkedOneToOneRelations(relations).map(({ relation }) =>
+        this.services.authorization.assertCanEdit(context, relation.foreignCollection),
+      ),
+    );
   }
 
   private async linkOneToOneRelations(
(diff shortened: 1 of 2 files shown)


packages/agent/src/routes/modification/create.ts
```diff
@@ -91,7 +91,7 @@ export default class CreateRoute extends CollectionRoute {
       // Permissions
       const foreignCollection = this.dataSource.getCollection(relation.foreignCollection);
       const scope = await this.services.authorization.getScope(foreignCollection, context);
-      await this.services.authorization.assertCanEdit(context, this.collection.name);
+      await this.services.authorization.assertCanEdit(context, foreignCollection.name);
```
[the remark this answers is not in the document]
PMerlet: Good catch — fixed in the follow-up commit. The one-to-one `edit` assertion is now hoisted into a pre-create pass (`assertCanEditOneToOneRelations`) that runs before `createRecord`, so a denied request writes nothing at all instead of leaving an orphan parent. As a bonus this also makes the multi-relation case atomic: a denial on one linked relation no longer lets the writes on the others proceed. Added a regression test asserting neither `persons.create` nor `passports.update` is called when the foreign edit permission is denied.

packages/agent/test/routes/modification/create.test.ts
```diff
@@ -190,6 +192,62 @@ describe('CreateRoute', () => {
...
+            jsonapi: { version: '1.0' },
+          },
+        });
+
+        const assertCanEdit = services.authorization.assertCanEdit as jest.Mock;
+        assertCanEdit.mockClear();
+
+        await create.handleCreate(context);
+
+        expect(assertCanEdit).toHaveBeenCalledTimes(1);
```
Scra3: issue (non-blocking): The `persons` fixture declares a single OneToOne, so this count cannot distinguish "asserts once per linked relation" from "asserts once, ever" — replacing `Object.entries(relations)` with `Object.entries(relations).slice(0, 1)` in `assertCanEditLinkedOneToOneRelations` keeps all 9 tests green while reintroducing the exact bypass this PR closes. Adding 2 OneToOne + 1 ManyToOne to the fixture and asserting on the collected collection names instead of a bare count also covers the `relation.type !== 'OneToOne'` and `linked === null` guards, which are currently unpinned too.

packages/agent/test/routes/modification/create.test.ts
```diff
@@ -190,6 +192,62 @@ describe('CreateRoute', () => {
...
+                  data: { type: 'passports', id: '1d162304-78bf-599e-b197-93590ac3d314' },
+                },
+              },
+            },
+            jsonapi: { version: '1.0' },
+          },
+        });
+
+        const error = new Error('Forbidden');
+        (services.authorization.assertCanEdit as jest.Mock).mockRejectedValueOnce(error);
```
Scra3: suggestion: `mockClear()` does not drain the `Once` queue, so once the fixture carries two OneToOne relations this rejects only one of the two `Promise.all` branches — key a `mockImplementation` on the collection name instead.

packages/agent/src/routes/modification/create.ts
```diff
@@ -77,6 +80,20 @@ export default class CreateRoute extends CollectionRoute {
     return record;
   }
 
+  private async assertCanEditLinkedOneToOneRelations(
+    context: Context,
+    relations: Record<string, RecordData>,
+  ): Promise<void> {
+    const promises = Object.entries(relations).map(async ([field, linked]) => {
+      const relation = SchemaUtils.getRelation(this.collection.schema, field, this.collection.name);
+      if (linked === null || relation.type !== 'OneToOne') return;
```
Scra3: suggestion: This guard is duplicated verbatim in `linkOneToOneRelations`, and it is now the definition of "which writes need edit on the foreign collection" — widening it there without widening it here silently drops the permission check on the write. Extracting the filtered relation list once and iterating it in both methods removes that drift risk.

packages/agent/test/routes/modification/create.test.ts
```diff
@@ -190,6 +192,62 @@ describe('CreateRoute', () => {
...
+        const create$ = jest.spyOn(dataSource.getCollection('persons'), 'create').mockClear();
+        const update$ = jest.spyOn(dataSource.getCollection('passports'), 'update').mockClear();
+
+        await expect(create.handleCreate(context)).rejects.toThrow(error);
+
+        expect(create$).not.toHaveBeenCalled();
```
Scra3: praise: Asserting on `create` and not just `update` is the load-bearing part — it pins the ordering, which is the actual security property, since asserting after `createRecord` would leave an orphaned parent record behind on a permission failure.

Scra3 (commented):
Fix is correct — `assertCanEdit` on `relation.foreignCollection` matches `updateOneToOne` in `update-relation.ts:97`, and asserting before `createRecord` avoids an orphaned parent on denial. Non-blocking: the single-relation fixture leaves per-relation enforcement unpinned.

packages/agent/src/routes/modification/create.ts
PMerlet: Done in fa336e0c. Extracted `getLinkedOneToOneRelations` — a single filtered list now feeds both the permission pass and the linking pass, so the guard has one definition and can't drift between them.

packages/agent/test/routes/modification/create.test.ts
PMerlet: Done in fa336e0c. Fixture now has two OneToOne + a null OneToOne + a ManyToOne, and the test asserts the collected collection names (`['passports','visas']`) plus explicit `not.toHaveBeenCalledWith` for the null and many-to-one relations. Verified the `.slice(0,1)` mutant you described now fails the suite.

packages/agent/test/routes/modification/create.test.ts
PMerlet: Done in fa336e0c. Switched to `mockImplementation((_, name) => name === 'passports' ? reject : resolve)` so the denial is deterministic across both `Promise.all` branches, and restore the mock in a `finally`.

Scra3 (approved):

PR #1190, opened 2024-10-08
fix(permissions): properly check permissions when dissociating or deleting related resources

Thenkei:
…

### General

- [ ] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)
- [ ] Test manually the implemented changes
- [ ] Validate the code quality (indentation, syntax, style, simplicity, readability)

### Security

- [ ] Consider the security impact of the changes made

PR #1820, opened 2026-08-11
fix(agent): authorize associate-related on the foreign collection and scope the many-to-many link

PMerlet:
## Problem

The associate-related route asserted `edit` on the **parent** collection, then performed the write on the **foreign** collection (OneToMany — sets the foreign record's FK) or created a row in the **through** collection (ManyToMany). A caller with `edit` on the parent but no `edit` on the foreign collection could reassign/hijack a foreign record or link an arbitrary one. Same bug class as #1818 (PRD-898) and #1819 (PRD-916), on the write side. The mirror route `dissociate-delete-related` was already fixed for this in #1190, leaving associate/dissociate asymmetric.

Second, distinct defect in the same route: `associateManyToMany` fetched the foreign scope but only the OneToMany branch used it, so a foreign record hidden by the caller's scope could still be linked by id.

## Fix

1. **Permission** — assert `edit` on `this.foreignCollection.name` (one assertion, shared by both branches), mirroring `dissociate-delete-related`.
2. **Scope (ManyToMany)** — resolve the target through a scope-intersected `list` and skip the link when it returns nothing, consistent with the OneToMany branch's scoped write. Out-of-scope targets are left untouched rather than 403'd, matching how the OneToMany branch and `update-relation` already behave.

Incidental correctness improvement: the ManyToMany branch now stores the resolved record's `foreignKeyTarget` value in the through row instead of the raw requested id. When `foreignKeyTarget` is the PK (common case) this is identical; when it is a non-PK column the old code stored the wrong value. It also no longer creates a dangling through row for a nonexistent foreign id.

## ⚠️ Behavioral change (permission tightening)

A role with `edit` on the **parent** but **not** on the target/foreign collection could previously associate; it now receives a 403. This aligns associate with `dissociate-delete-related`, one-to-one update, and count-related, which already require edit on the target collection. Projects whose roles were configured with parent-edit-only on a related collection will need edit on the target collection to associate. Shipped as a `fix(` (patch), consistent with the sibling PRD-898/916 security fixes.

## Tests

- Both branches assert the edit check targets the foreign collection (`bookPersons` / `libraries`), exactly once, and never the parent (`books`).
- Denial tests: when `assertCanEdit` rejects, no foreign update / through create happens.
- ManyToMany: asserts the `list` is called with the scope-intersected filter + `Projection`, and that no through row is created when the scoped lookup returns nothing. The fixture uses a **non-PK `foreignKeyTarget`** so the test pins that the through row stores the scope-resolved record's value, not the raw requested id.
- Empirically verified the mutants (revert to parent-side permission; drop the scope guard; drop scope from the filter; link the raw id) each fail the suite.

## Adversarial review

Three independent review passes (refutation, test-attack, parity/regression) confirmed: the permission target is correct under renames/customizer, the scope intersection handles composite PKs and null scope, there is no existence oracle (out-of-scope and nonexistent both yield an empty scoped list), and associate is now symmetric with dissociate on both permission and ManyToMany scope. Residual items surfaced are all pre-existing and separately tracked: parent-side scope on the origin resolution (PRD-909), unscoped to-one reads (PRD-907/917).

fixes PRD-922

> [!NOTE]
> ### Fix authorization and scope handling in `AssociateRelatedRoute` for many-to-many links
> - Changes the edit permission check in `handleAssociateRelatedRoute` to assert on the foreign collection instead of the origin collection.
> - Scopes the foreign record lookup in `associateManyToMany` by intersecting the target ID filter with the resolved scope, using only the `foreignKeyTarget` field via `Projection`.
> - If no foreign record is found within scope, the join record is not created.
> - The through record is now written using the `foreignKeyTarget` field value rather than the foreign record's primary key.
> - Behavioral Change: many-to-many association requests that previously succeeded may now be rejected or silently skipped if the foreign record is outside the user's scope.
>
> 
> #### Changes since #1820 opened
>
> - Changed `AssociateRelatedRoute.associateOneToMany` to retrieve the origin value using `relation.originKeyTarget` instead of the collection's primary key [c6095b0]
> - Updated many-to-many association test to validate origin key resolution using `originKeyTarget` [c6095b0]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized f577f38.</sup>
> 
>

### DIFF
--- a/packages/agent/src/routes/modification/associate-related.ts
+++ b/packages/agent/src/routes/modification/associate-related.ts
@@ -12,6 +12,8 @@ import {
   CollectionUtils,
   ConditionTreeFactory,
   ConditionTreeLeaf,
+  Filter,
+  Projection,
   SchemaUtils,
 } from '@forestadmin/datasource-toolkit';
 
@@ -30,7 +32,7 @@ export default class AssociateRelatedRoute extends RelationRoute {
   }
 
   public async handleAssociateRelatedRoute(context: Context): Promise<void> {
-    await this.services.authorization.assertCanEdit(context, this.collection.name);
+    await this.services.authorization.assertCanEdit(context, this.foreignCollection.name);
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     const body = context.request.body as any;
@@ -43,7 +45,7 @@ export default class AssociateRelatedRoute extends RelationRoute {
     if (relation.type === 'OneToMany') {
       await this.associateOneToMany(caller, scope, relation, parentId, targetedRelationId, context);
     } else {
-      await this.associateManyToMany(caller, relation, parentId, targetedRelationId);
+      await this.associateManyToMany(caller, scope, relation, parentId, targetedRelationId);
     }
 
     context.response.status = HttpCode.NoContent;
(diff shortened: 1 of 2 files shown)


christophebrun-forest (approved):
LGTM
Non-blocking follow-up: the origin side still uses the parent's PK instead of originKeyTarget — twin of the bug fixed here; fine as a one-liner here or in a ticket.

PMerlet (commented):
Thanks! Done in c6095b0c — `associateManyToMany` now resolves the origin value from `relation.originKeyTarget` (matching `associateOneToMany`) instead of the parent PK. The fixture's ManyToMany `originKeyTarget` is now a non-PK column so the through row's origin value is pinned to the resolved record, and I verified the revert-to-PK mutant fails the suite.

christophebrun-forest (approved):
thank for taking into account my comment.
LGTM

PR #1564, opened 2026-04-20
feat(executor): introduce workflow executor

Scra3:
MAIN BRANCH TO INTRODUCE WORKFLOW EXECUTOR.
 

image

> [!NOTE]
> ### Add workflow executor package with step executors, HTTP API, and CLI
> - Introduces a new `@forestadmin/workflow-executor` package that runs Forest Admin workflow steps (condition, read-record, update-record, trigger-action, load-related-record, mcp, guidance) against agent data with AI assistance.
> - Adds a `Runner` with start/stop lifecycle, polling, in-flight run tracking, and graceful drain-on-shutdown; step execution is handled by per-type executors derived from `BaseStepExecutor` with timeout enforcement and idempotency checks.
> - Exposes a Koa HTTP server (`ExecutorHttpServer`) with JWT auth, `/health`, `GET /runs/:runId`, and `POST /runs/:runId/trigger` endpoints; the agent gains a `workflowExecutorUrl` option and a proxy route (`WorkflowExecutorProxyRoute`) that forwards `/_internal/workflow-executions/*` to the executor.
> - Persistence uses either a SQLite/Postgres-backed `DatabaseStore` (with Umzug migrations) or an `InMemoryStore`; activity log transitions are tracked and drained via `ActivityLogDrainer`.
> - Extends `@forestadmin/ai-proxy` with `AiClient`, `createBaseChatModel`, `mcpServerId` propagation across `RemoteTool` subclasses, and `validateAiConfigurations`; fixes `QuerySerializer.formatFields` to emit comma-separated strings instead of arrays to prevent field loss in Rack-based agents.
> - Risk: the `QuerySerializer` change alters serialized query parameter format for field projections, which may affect any consumers relying on the previous array format.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6aa8faa.</sup>
> 
>

PR #1821, opened 2026-08-11
fix(workflow-executor): acknowledge a trigger once the run is claimed, not once the chain ends

Scra3:
fixes PRD-923

> [!IMPORTANT]
> **Merge after the frontend guard ships** (forestadmin#9897). See *Sequencing*.

## Problem

`POST /runs/:runId/trigger` awaits the confirmed step **and its entire auto-chain** before responding — by design up to `stepTimeoutS × maxChainDepth` (5 min × 50). Nothing cancels when the caller gives up, so any edge in front of the executor eventually cuts the request while the run keeps going.

The operator confirms a step, waits ~30 s, and is shown a failure — banner *and* toast, carrying the raw internal message:

```
Request timeout while fetching POST …/runs/133289/trigger DURATION: 30245ms
```

The run was fine. **202 occurrences in 30 days**, one customer, still daily. Run 133289 is representative: 70 steps, three human pauses, and the confirmation at step 15 triggers **54 chained steps** — a run that takes 5 min 20 in total.

The 30 s is not ours: every timeout Forest owns on that path is 120 s (`workflow-executor-proxy.ts:34`) and the frontend imposes none. Not a regression either — the blocking `await` dates from the executor's first commit (#1564, 9 June).

## Fix

`triggerPoll` returns as soon as the run is claimed and validated, and lets the chain run detached.

This loses no information, because **the response never carried the chain's outcome**:

- Every error a caller can act on is raised *before* execution: `RunAlreadyInFlightError` (400), `RunNotFoundError` (404), `UserMismatchError` (403). The acknowledgement sits right after them.
- `doExecuteStep` never rejects on a step failure: each path reports a `status: 'error'` outcome through `updateStepExecution`, then returns. Failures already reach the panel via orchestrator → subscription — which is why the operator sees the steps keep scrolling *while* the error banner is up.
- The frontend never reads the body (`executor.ts:50-57` discards it).

`inFlightRuns.track()` still holds the detached promise — registered synchronously before the response is written — so trigger dedup and the shutdown drain are unchanged.

## Behavioural changes

**`200 {triggered: true}` now means "chain started"**, not "chain finished".

**A deploy mid-chain now truncates trigger-started chains.** Previously `server.close()` could not complete while the response was held open, giving them an effectively unbounded grace period; they now fall under the same `stopTimeoutS` drain as polled chains. That alignment is the intent, and a truncated run is re-dispatched by the poll cycle (`getAvailableRun` is a plain GET, no server-side claim). Documented in the package `CLAUDE.md`.

**`RunAlreadyInFlightError` gets a `userMessage`.** That path becomes common — a client can now re-trigger a run it was just told 200 about — and it was surfacing its raw technical string, runId included, to the operator.

## Sequencing

The frontend's `isSubmittingPendingData` lock is currently the only same-tick guard against a double confirm on this path; the other gates arrive over the WebSocket. Releasing the HTTP call earlier releases that lock earlier, so forestadmin#9897 must ship first.

## Tests

`describe('trigger acknowledgement')`, 1507 pass overall:

- **resolves while the chain is still executing** — blocks `executor.execute`, asserts `triggerPoll` resolves with the run still in `inFlightRuns` and `updateStepExecution` not yet called. **Red against `main`**, where it times out.
- **resolves even when the chain later reports an error outcome** — gates `updateStepExecution` so it proves the ack precedes the outcome. Also red against `main`.
- **survives a host logger that throws from the detached chain handler** — without the guard, the test run *dies*: the rejection escapes and takes the process down, where `main` only ever returned a 500.
- **drains a chain started by a trigger on `stop()`** — via the public API, and it pins that registration precedes the ack.
- **still rejects when the run cannot be claimed / when the bearer user does not own the run** — the acceptance criterion that genuine failures stay failures.

The 33 existing call sites that assumed "await = execution finished" go through a `triggerAndDrain` helper.

## Known gaps, filed separately

Both share a root cause — an unguarded host logger — and want one pass:

- When the detached chain rejects, no outcome is posted. The run is not lost (the sweeper returns it to `pending` and the poll re-dispatches), but a deterministic cause means a silent retry loop. The repair belongs in `doExecuteStep`'s chain loop, which can see `currentStep`.
- `runPollCycle` has the same throwing-logger exposure and there it is already fatal today: `schedulePoll`'s `setTimeout` discards the rejected promise, so it exits the process on the first poll cycle — before this handler could ever matter.

> [!NOTE]
> fixes PRD-923
>
> > [!IMPORTANT]
> > **Merge after the frontend guard ships** (forestadmin#9897). See *Sequencing*.
>
> ## Problem
>
> `POST /runs/:runId/trigger` awaits the confirmed step **and its entire auto-chain** before responding — by design up to `stepTimeoutS × maxChainDepth` (5 min × 50). Nothing cancels when the caller gives up, so any edge in front of the executor eventually cuts the request while the run keeps going.
>
> The operator confirms a step, waits ~30 s, and is shown a failure — banner *and* toast, carrying the raw internal message:
>
> ```
> Request timeout while fetching POST …/runs/133289/trigger DURATION: 30245ms
> ```
>
> The run was fine. **202 occurrences in 30 days**, one customer, still daily. Run 133289 is representative: 70 steps, three human pauses, and the confirmation at step 15 triggers **54 chained steps** — a run that takes 5 min 20 in total.
>
> The 30 s is not ours: every timeout Forest owns on that path is 120 s (`workflow-executor-proxy.ts:34`) and the frontend imposes none. Not a regression either — the blocking `await` dates from the executor's first commit (#1564, 9 June).
>
> ## Fix
>
> `triggerPoll` returns as soon as the run is claimed and validated, and lets the chain run detached.
>
> This loses no information, because **the response never carried the chain's outcome**:
>
> - Every error a caller can act on is raised *before* execution: `RunAlreadyInFlightError` (400), `RunNotFoundError` (404), `UserMismatchError` (403). The acknowledgement sits right after them.
> - `doExecuteStep` never rejects on a step failure: each path reports a `status: 'error'` outcome through `updateStepExecution`, then returns. Failures already reach the panel via orchestrator → subscription — which is why the operator sees the steps keep scrolling *while* the error banner is up.
> - The frontend never reads the body (`executor.ts:50-57` discards it).
>
> `inFlightRuns.track()` still holds the detached promise — registered synchronously before the response is written — so trigger dedup and the shutdown drain are unchanged.
>
> ## Behavioural changes
>
> **`200 {triggered: true}` now means "chain started"**, not "chain finished".
>
> **A deploy mid-chain now truncates trigger-started chains.** Previously `server.close()` could not complete while the response was held open, giving them an effectively unbounded grace period; they now fall under the same `stopTimeoutS` drain as polled chains. That alignment is the intent, and a truncated run is re-dispatched by the poll cycle (`getAvailableRun` is a plain GET, no server-side claim). Documented in the package `CLAUDE.md`.
>
> **`RunAlreadyInFlightError` gets a `userMessage`.** That path becomes common — a client can now re-trigger a run it was just told 200 about — and it was surfacing its raw technical string, runId included, to the operator.
>
> ## Sequencing
>
> The frontend's `isSubmittingPendingData` lock is currently the only same-tick guard against a double confirm on this path; the other gates arrive over the WebSocket. Releasing the HTTP call earlier releases that lock earlier, so forestadmin#9897 must ship first.
>
> ## Tests
>
> `describe('trigger acknowledgement')`, 1507 pass overall:
>
> - **resolves while the chain is still executing** — blocks `executor.execute`, asserts `triggerPoll` resolves with the run still in `inFlightRuns` and `updateStepExecution` not yet called. **Red against `main`**, where it times out.
> - **resolves even when the chain later reports an error outcome** — gates `updateStepExecution` so it proves the ack precedes the outcome. Also red against `main`.
> - **survives a host logger that throws from the detached chain handler** — without the guard, the test run *dies*: the rejection escapes and takes the process down, where `main` only ever returned a 500.
> - **drains a chain started by a trigger on `stop()`** — via the public API, and it pins that registration precedes the ack.
> - **still rejects when the run cannot be claimed / when the bearer user does not own the run** — the acceptance criterion that genuine failures stay failures.
>
> The 33 existing call sites that assumed "await = execution finished" go through a `triggerAndDrain` helper.
>
> ## Known gaps, filed separately
>
> Both share a root cause — an unguarded host logger — and want one pass:
>
> - When the detached chain rejects, no outcome is posted. The run is not lost (the sweeper returns it to `pending` and the poll re-dispatches), but a deterministic cause means a silent retry loop. The repair belongs in `doExecuteStep`'s chain loop, which can see `currentStep`.
> - `runPollCycle` has the same throwing-logger exposure and there it is already fatal today: `schedulePoll`'s `setTimeout` discards the rejected promise, so it exits the process on the first poll cycle — before this handler could ever matter.
>
> #### Changes since #1821 opened
>
> - Made `causeMessage` utility function null-safe by guarding against null or undefined inputs [b481cea]
> 
>


packages/workflow-executor/src/runner.ts
```diff
@@ -200,7 +200,22 @@ export default class Runner {
       throw new UserMismatchError(runId, options.bearerUserId, step.user.id);
     }
 
-    await this.executeStep(step, auth.forestServerToken, options?.pendingData);
+    // Not awaited: the chain's outcome travels through updateStepExecution, never through this
+    // response.
+    void this.executeStep(step, auth.forestServerToken, options?.pendingData).catch(error => {
```
hercemer42, line 205: Claude Opus 5 (claude-opus-5): **Should fix**

When the detached chain rejects here, the operator has already been told `200 {triggered: true}` and **no outcome is ever posted** — the panel's subscription receives nothing and the run sits claimed, spinning, for good. The PR body's "it remains visible through the subscription and the step's error context" is true of step failures (those go through `updateStepExecution`), but not of the rejection class this handler exists to catch.

`doExecuteStep` can only reject at its bare `this.logger(...)` calls outside any try, and the first of them (`runner.ts:325`, `Run started executing`) fires before an executor is even constructed — nothing has been reported at that point. In embedded mode that logger is the customer's own, passed straight through, which is the threat model the guard just below was added for. The poll cycle then re-dispatches the run, hits the same throw, and `Promise.allSettled` at `runner.ts:251` discards it without a line.

The repair can't live in this handler — it only holds the initial `step`, and the chain may have advanced several `stepIndex` values. Wrapping the `while` body in `doExecuteStep`, where `currentStep` is in scope, lets it fall back to the synthetic-error-outcome path at `runner.ts:362-378` — the one whose own comment explains that without it the step "loops forever".

packages/workflow-executor/src/runner.ts
```diff
@@ -200,7 +200,22 @@ export default class Runner {
...
+      // A throwing host logger must not reject this handler — nothing awaits it, so that would
+      // take the process down instead of the one run.
+      try {
+        this.logger('Error', 'FATAL: in-flight chain rejected — outcome not reported', {
```
hercemer42: Claude Opus 5 (claude-opus-5): **Should fix**

Whoever gets paged by this line will hunt a stuck run that isn't stuck: `outcome not reported` is false at 4 of the 8 reachable rejection sites. `runner.ts:404`, `:420`, `:433` and `:444` all run *after* `updateStepExecution` returned successfully, so the outcome was reported and the run is in a fine state. The context also carries only `runId`, where every sibling error log in this file carries `stepId` and `stepIndex` — with a chain up to `maxChainDepth` deep there is no way to tell which step died.

Narrower, same lines: the payload is built *inside* the `try`, and `causeMessage` destructures its argument (`src/errors.ts:7-9`), so a rejection value of `null`/`undefined` makes this log call itself throw and the empty catch converts the one FATAL line into no output at all — precisely the outcome the guard was added to prevent. Computing `error`/`cause`/`stack` before the `try` closes that.

packages/workflow-executor/src/runner.ts
```diff
@@ -200,7 +200,22 @@ export default class Runner {
       throw new UserMismatchError(runId, options.bearerUserId, step.user.id);
     }
 
-    await this.executeStep(step, auth.forestServerToken, options?.pendingData);
+    // Not awaited: the chain's outcome travels through updateStepExecution, never through this
```
hercemer42, line 203: Claude Opus 5 (claude-opus-5): **Should fix**

**Applies to:** `src/build-workflow-executor.ts:174-183` and `runner.ts:118-154` (not in this diff).

A deploy during a trigger-started chain now truncates it, where it used to let it finish. Shutdown is `await server.stop()` then `await runner.stop()`: while the trigger response was held open, `server.close()` could not complete until the chain had finished, so trigger chains got an effectively unbounded grace period. Now the response is written immediately, `server.stop()` returns at once, and the chain falls under the `stopTimeoutS` drain race (30s default) — after which `aiModelPort.closeConnections()` and `runStore.close()` (`runner.ts:151-154`) run underneath a still-executing step whose own cap is `stepTimeoutS` (300s). That step's next `runStore` write and `updateStepExecution` fail against closed resources, and the run is left in `started` with no outcome.

This aligns trigger chains with poll-cycle chains, which were always bounded, so it may well be the intended trade — but it is a behavioural change the PR body doesn't mention, and the new `stop()` test only covers the chain that finishes inside the grace period.

packages/workflow-executor/src/runner.ts
```diff
@@ -200,7 +200,22 @@ export default class Runner {
...
+          cause: causeMessage(error),
+          stack: error instanceof Error ? error.stack : undefined,
+        });
+      } catch {
+        /* empty */
+      }
+    });
   }
 
   private assertRunNotInFlight(runId: string): void {
```
hercemer42, line 225: Claude Opus 5 (claude-opus-5): **Should fix**

**Applies to:** `src/errors.ts:459-463` (not in this diff); this guard is the reason the path is now reachable.

This 400 is newly reachable from a client that just received a 200, and its body is a raw internal string. `RunAlreadyInFlightError` passes no `userMessage`, `WorkflowExecutorError` defaults `userMessage` to the technical `message` (`errors.ts:36`), and `toHttpError` (`src/http/http-errors.ts:82-84`) returns it verbatim — so the operator is shown `Run "133289" is already being processed`, echoing the runId, on a run that is progressing normally. That is the same failure class AC 1 of PRD-923 exists to remove, arriving through a different door, and `log:false` means nothing records it server-side. The package `CLAUDE.md` is explicit that a `WorkflowExecutorError` subclass must set a distinct, jargon-free `userMessage`; this one predates the PR but was unreachable from a well-behaved client until now.

One line in `errors.ts` (`'This run is already running.'`) makes the newly-reachable path harmless regardless of how completely the frontend guard in forestadmin#9897 covers re-entry (second tab, continue/summary path) — which I can't observe from this repo.

packages/workflow-executor/test/runner.test.ts
```diff
@@ -2007,7 +2224,7 @@ describe('error handling', () => {
     });
 
     runner = new Runner(createRunnerConfig({ workflowPort, logger: mockLogger }));
-    await runner.triggerPoll('run-1');
+    await triggerAndDrain(runner, 'run-1');
```
hercemer42, line 2238: Claude Opus 5 (claude-opus-5): **Should fix**

**Applies to:** `test/runner.test.ts:2188-2206` (not in this diff) — the sibling test just above this conversion.

`does not re-throw if updateStepExecution fails after a construction error` was left un-converted, and its single assertion `await expect(runner.triggerPoll('run-1')).resolves.toBeUndefined()` can no longer fail: `triggerPoll` now resolves before `getModel` throws, and the new `.catch` swallows whatever the chain does. It passes for any valid dispatch and no longer exercises the swallowed-`updateStepExecution`-failure path at `runner.ts:370-378` it was written for — green with nothing behind it, which is what this repo's CLAUDE.md calls out as fake coverage.

`await triggerAndDrain(runner, 'run-1')` plus an assertion on the `FATAL: also failed to report synthetic error outcome` log (the shape used at :2279) restores the behaviour it claims to cover.

packages/workflow-executor/src/runner.ts
```diff
@@ -200,7 +200,22 @@ export default class Runner {
...
+      // A throwing host logger must not reject this handler — nothing awaits it, so that would
+      // take the process down instead of the one run.
+      try {
+        this.logger('Error', 'FATAL: in-flight chain rejected — outcome not reported', {
+          runId,
+          error: extractErrorMessage(error),
+          cause: causeMessage(error),
+          stack: error instanceof Error ? error.stack : undefined,
+        });
+      } catch {
```
hercemer42, line 219: Claude Opus 5 (claude-opus-5): **Preferential**

**Applies to:** `runner.ts:229-262` (not in this diff).

This guard hardens the on-demand path against a throwing host logger while the continuously-running one stays open, and the open one is fatal: `runPollCycle` logs at `:246` inside the try, its catch logs again at `:255`, and a throw there escapes the catch — `finally` still reschedules, but `schedulePoll`'s `setTimeout(() => this.runPollCycle(), …)` at `:231` discards the rejected promise, so it surfaces as an unhandled rejection and exits the process on Node ≥15. A logger that throws consistently therefore takes the executor down on the first poll cycle, before this handler can ever matter.

Same one-line wrap there, and while you're in it, inspecting the `rejected` entries of `Promise.allSettled` at `:251` would give polled chains the FATAL line trigger-started ones now get — today that rejection is marked handled and vanishes.

packages/workflow-executor/src/runner.ts
```diff
@@ -200,7 +200,22 @@ export default class Runner {
       throw new UserMismatchError(runId, options.bearerUserId, step.user.id);
     }
 
-    await this.executeStep(step, auth.forestServerToken, options?.pendingData);
+    // Not awaited: the chain's outcome travels through updateStepExecution, never through this
+    // response.
```
hercemer42, line 204: Claude Opus 5 (claude-opus-5): **Should fix**

**Applies to:** `packages/workflow-executor/CLAUDE.md` (not in this diff).

This comment is currently the only place the new contract is written down. The package `CLAUDE.md` carries the Runner's invariants — auto-chain, graceful shutdown, fetched-steps-must-execute — and asks to be updated when one changes; nothing in it says a trigger returns before its chain, so the next reader has no documented reason not to restore the `await`, and the only structural guard is a test reaching into the private `inFlightRuns` map. Worth one line under Invariants: the trigger acknowledges once the run is claimed, outcomes travel only via `updateStepExecution`, and the drain — not the response — bounds the chain.

hercemer42 (changes requested):

packages/workflow-executor/src/runner.ts
Scra3: Both taken, in a3b49d654 — and the second one is the sharper of the two.

**The null-rejection trap is real and I reproduced the reasoning:** `causeMessage` destructures its argument (`errors.ts:7-8`), the payload was built inside the try, so a `null` rejection threw there and the empty catch turned the single FATAL line into no output whatsoever — the exact outcome the guard was added to prevent. The payload is now built before the try, with an explicit null check before `causeMessage`.

**The message was lying at half the sites.** Dropped 'outcome not reported': it is false at every rejection that fires after `updateStepExecution` already returned. And the context now carries `stepId`/`stepIndex`, matching the sibling logs — the initial step's, since as you note the handler cannot see how far the chain advanced.

packages/workflow-executor/src/runner.ts
Scra3: Taken, in a3b49d654 — one line in `errors.ts`: `'This step is already running.'`

You are right that this PR is what makes the path common: a client that was just told 200 can now re-trigger while the chain runs, where before the request was still open. Showing it `Run "133289" is already being processed` — raw internals, runId echoed — on a run that is progressing normally is the same failure class PRD-923 exists to remove, arriving through a different door.

Two HTTP tests asserted the old body; updated.

packages/workflow-executor/src/runner.ts
Scra3: Taken, in a3b49d654. Two entries under Invariants: one for the acknowledgement contract (returns at claim, the three pre-ack errors, outcomes only via `updateStepExecution`), and an addition to **Graceful shutdown** noting that a trigger chain is now bounded by the drain rather than by the caller.

packages/workflow-executor/test/runner.test.ts
Scra3: Taken, in a3b49d654. It now drains and asserts the failure is logged.

One correction to the target: the assertion is on `'Failed to report step outcome'`, not the synthetic-outcome message. I ran it — a `getModel` throw is caught by `StepExecutorFactory.create` (`step-executor-factory.ts:133`), which logs `'Step execution failed unexpectedly'` and returns an executor producing an error outcome; it is the POST of *that* outcome which then fails, landing in the runner's `Failed to report step outcome` branch. Same path you identified, different log line at the end of it.

packages/workflow-executor/src/runner.ts
Scra3: Confirmed, and documented rather than changed — it is the intended trade, but you are right that the PR body hid it.

Before, `server.close()` could not finish while the trigger response was held open, so trigger chains had an effectively unbounded grace period. Now they fall under the same `stopTimeoutS` drain as polled chains, and past it `aiModelPort.closeConnections()` and `runStore.close()` run under a step whose own cap is 300 s.

I have added it to the package `CLAUDE.md` under Graceful shutdown ('a deploy mid-chain truncates them at `stopTimeoutS` — they used to be held open by the unanswered request') and it goes in the PR body.

Not raising the drain timeout here: aligning trigger chains with polled ones is the point, and a run truncated mid-chain is recovered by the poll cycle — `getAvailableRun` is a plain GET with no server-side claim, so the step stays pending. The gap that deserves attention is the one you flag next door: nothing reports an outcome when that happens.

packages/workflow-executor/src/runner.ts
Scra3: Real, and I am declining the repair in this PR while agreeing with the analysis — the reasoning matters more than the verdict here.

**Where I agree.** `doExecuteStep` can only reject at its bare `this.logger(...)` calls, and the first (`runner.ts:325`) fires before an executor exists, so nothing has been reported. In embedded mode that logger is the customer's own, which is the threat model the guard below was added for. And the poll cycle re-dispatching into the same deterministic throw, with `Promise.allSettled` discarding it, is a hot loop nobody is paged for.

**Where it differs from 'sits claimed for good'.** `getAvailableRun` is a plain GET with no server-side claim (`forest-server-workflow-port.ts:63`), and the sweeper returns an expired `loading` run to `pending` after 60 s, so the run is re-dispatched rather than stuck. That makes it a silent retry loop, not a permanently frozen run — bad, but a different severity.

**Why not here.** As you say, the repair cannot live in this handler: it holds only the initial `step`, while the chain may have advanced several `stepIndex` values. Wrapping the `while` body in `doExecuteStep` is the right shape, and it changes the error contract of the chain loop — every existing catch in there posts a synthetic outcome and returns, so adding an outer net means deciding what happens when *that* post fails too. That is a design pass, not a rider on a two-line fix, and it applies equally to the polled path this PR does not touch.

Filing it with your analysis, alongside your `runPollCycle` finding — the two share a root cause (an unguarded host logger) and should be fixed together.

packages/workflow-executor/src/runner.ts
Scra3: Good catch, and it inverts the priority I had assumed.

You are right that `runPollCycle` is the fatal one: its `:246` log sits inside the try, its catch logs again at `:255`, and a throw from that catch escapes — `schedulePoll`'s `setTimeout` discards the rejected promise, so it exits the process on Node ≥15. A consistently throwing host logger therefore takes the executor down on the first poll cycle, long before the on-demand path this PR guards ever matters.

Which means my guard is the narrower half of the problem, not the fix for it. Same one-line wrap belongs there, and inspecting the `rejected` entries of the `allSettled` at `:251` would give polled chains the FATAL line trigger-started ones now get.

Both are outside this diff and neither is caused by it, so I am filing them together with the sibling finding on the detached-chain repair — same root cause, and they want one pass rather than three.

hercemer42 (approved):

PR #1822, opened 2026-08-11
feat(agent): accept the Forest-Projection header on list, relationship and csv routes

PMerlet:
## What

The following routes now read the projection from the `Forest-Projection` header (same format as get-one: `field,relation:subField`), taking precedence over the `fields[...]` query params:

- list (`GET /forest/:collection`)
- relationship list (`GET /forest/:collection/:id/relationships/:relation`)
- csv exports (the `.csv` variants of both)

The count variants accept and ignore the header — they never parse a projection — and tests lock that behavior in.

Relationship list/count URLs observed in production routinely approach the 2KB AWS WAF `SizeRestrictions_QUERYSTRING` default, and header-based projections are a prerequisite for projecting workspaces. Exports of the same wide views carry the same oversized query strings, hence the csv routes.

Unlike the other routes, csv keeps NOT adding pks to the projection, so the exported columns are exactly the requested ones.

## New capability

The agent announces `canUseProjectionViaHeaderOnList`, covering every record-listing route (list, relationship list, csv exports).

`canUseProjectionViaHeader` alone cannot be used to gate this: it is already released in 1.93.0–1.93.2, where the header only works on get-one. Without a second flag the frontend could not tell those versions apart from this one, and would have to keep `fields[...]` in the URL and switch to the header only above the WAF threshold — real logic on the front to work around a missing signal. `canUseMultipleFieldsProjectionOnRelation` and query-string parsing stay untouched, so frontend and agent deployments remain decoupled.

## Also fixed

On the csv routes, `CsvRouteContext.buildResponse` ran *before* the projection was parsed, so a request with an invalid projection returned a 400 wearing `Content-Disposition: attachment` — a browser-triggered export downloaded an error file instead of showing the error. It now runs once nothing can throw anymore. Pre-existing with an invalid `fields[...]` param; the header added one more trigger and these lines were being touched anyway.

## Notes

- CORS: nothing to do, the header is already allow-listed since #1813, and the agent's `@koa/cors` echoes `Access-Control-Request-Headers` on every route.
- `parseProjectionWithPks` lost its last two callers and is removed. It is not exported from the package's public surface.
- Out of scope: `agent-bff` and `agent-client` still serialize projections into `fields[...]` on their own hop to the agent, so a large projection can hit the same limit there.

fixes PRD-928

### DIFF
--- a/packages/agent/src/routes/access/csv-related.ts
+++ b/packages/agent/src/routes/access/csv-related.ts
@@ -25,7 +25,6 @@ export default class CsvRelatedRoute extends RelationRoute {
     await this.services.authorization.assertCanExport(context, this.foreignCollection.name);
 
     const { header } = context.request.query as Record<string, string>;
-    CsvRouteContext.buildResponse(context);
 
     const projection =
       QueryStringParser.parseProjectionFromHeader(this.foreignCollection, context) ??
@@ -34,6 +33,10 @@ export default class CsvRelatedRoute extends RelationRoute {
     const caller = QueryStringParser.parseCaller(context);
     const filter = ContextFilterFactory.buildPaginated(this.foreignCollection, context, scope);
     const parentId = IdUtils.unpackId(this.collection.schema, context.params.parentId);
+
+    // Set the download headers only once nothing can throw anymore: a 400 wearing a
+    // Content-Disposition: attachment would download an error file instead of showing it.
+    CsvRouteContext.buildResponse(context);
 
     const gen = CsvGenerator.generate(
       caller,
--- a/packages/agent/src/routes/access/csv-related.ts
+++ b/packages/agent/src/routes/access/csv-related.ts
@@ -26,9 +26,10 @@ export default class CsvRelatedRoute extends RelationRoute {
 
     const { header } = context.request.query as Record<string, string>;
 
-    const projection =
-      QueryStringParser.parseProjectionFromHeader(this.foreignCollection, context) ??
-      QueryStringParser.parseProjection(this.foreignCollection, context);
+    const projection = QueryStringParser.parseProjectionFromHeaderOrQuery(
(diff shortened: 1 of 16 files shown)


packages/agent/src/routes/access/csv-related.ts
```diff
@@ -25,14 +25,17 @@ export default class CsvRelatedRoute extends RelationRoute {
     await this.services.authorization.assertCanExport(context, this.foreignCollection.name);
 
     const { header } = context.request.query as Record<string, string>;
-    CsvRouteContext.buildResponse(context);
 
-    const projection = QueryStringParser.parseProjection(this.foreignCollection, context);
+    const projection =
```
christophebrun-forest: Code duplicated here and in 4 other files; it could be centralized in a helper.

christophebrun-forest (commented):
LGTM
Just one comment on dupplicate.
A helper to centralize precedence rules would be a quick and good solution.

packages/agent/src/routes/access/csv-related.ts
PMerlet: Good catch, done in b31f350b.

The header-then-query-string fallback now lives in a single place, `QueryStringParser.parseProjectionFromHeaderOrQuery`, and the five routes that parse a projection (`get`, `list`, `list-related`, `csv`, `csv-related`) each call it in one line. A new route can no longer forget the header, and the precedence rule has one owner.

I kept `parseProjection` and `parseProjectionFromHeader` public: they are covered by their own tests and the header one carries its own error message (`Invalid Forest-Projection header: ...`). Added four tests on the new helper, including the one that matters most — an invalid header throws instead of silently falling back to `fields[...]`.

christophebrun-forest (approved):
Ok for me with the centralized helper.  thanks
LGTM

PR #1823, opened 2026-08-12
feat(agent-bff): unfold the OpenAPI document per collection, relation and action

nbouliol:
Fixes PRD-684

## What

The document used to describe six generic paths with `{collection}`, `{relation}` and `{action}` as path parameters, so a codegen tool produced six methods where everything is a `string`. It now emits **one path per exposed collection, to-many relation and action**, each carrying that collection's real field set.

The **runtime routes are untouched** — they stay the generic regexes. A documented path is the generic route with its segments already filled in.

```
POST /agent/v1/customers/list
POST /agent/v1/customers/relations/orders/list      # fields of `orders`, the foreign side
POST /agent/v1/customers/actions/Mark%20as%20paid/execute
```

## Decisions

**Field names come from `capabilities()`, not from the schema fields.** The schema carries every column, but `capabilities-validator.ts` answers 422 for any projection/sort/filter field absent from capabilities — a document built from schema fields would advertise JSON columns and to-many relations the runtime rejects. Cost: one capabilities call per collection, through the existing 24h per-collection cache shared with the data routes, capped at 10 in flight.

**Two field enums per collection, not one.** The agent reports a `ManyToOne` without operators, so it is projectable and sortable but not filterable (422 `field_not_filterable`).

**The generic paths are replaced, not kept alongside.** Coexistence is lint-clean (checked), but a document describing each route twice, once typed and once as `string`, is not the concrete surface this is for.

**`operationId` is sanitized; the exact name lives in prose.** `redocly lint` errors on `operation-operationId-url-safe` for any operationId containing a space or a slash, so "the operationId carries the exact action name" and "redocly lint green" cannot both hold for an action named `Mark as paid`. Non-`[A-Za-z0-9_]` becomes `_`, collisions get a deterministic numeric suffix, and the **path segment** keeps the exact name URL-encoded — which is what the middleware decodes. `%2F` and `%20` in a path key are lint-clean (checked).

**Action `values` is typed but indicative.** Properties come from the schema's static form fields, all optional, additional properties allowed. `read-model.ts` defaults `hooks`/`fields`, which destroys the absent-vs-empty distinction `agent-client` documents as load-bearing (`domains/action.ts:101-107`) — so `hooks.load === false` cannot be trusted to mean "static form", and strict typing would reject valid calls on a legacy schema.

**`parentId` is typed from the parent's key.** `Number` → number, anything else → string, composite → the `|`-packed string, falling back to the opaque union when the parent exposes no key metadata.

**A collection whose `capabilities()` fails keeps its paths, untyped.** Free-form `string` field schemas plus the reason in the request description and a `Warn` log — never an empty enum, which would forbid every valid call. Dropping the collection would hide something the runtime serves.

**The listable-relation rule moved into the read-model.** It used to live privately in `data-routes-middleware.ts` (twice, after #1801); the unfolding needs the same predicate, and copies are how the document and the runtime drift apart. The foreign-collection allow-list check **stays** in the middleware: moving it would collapse `unknown_collection` into `unknown_relation`.

**A relation request describes the foreign collection's fields**, and since #1801 the runtime validates them against exactly that collection's capabilities — so the enum and the 422 now agree instead of the document being stricter than the runtime.

**With no `AGENT_URL`, the document stays generic**, even though the store can still read the schema (#1804 made that state real: the permissions endpoint needs the store without the agent). Concrete paths would be enumerable, but every data path answers 501 in that configuration, so unfolding would advertise a dead surface.

**Collect and generate are separate.** All I/O and the fan-out live in `collect-unfolding.ts`; `generateOpenApiDocument(version, unfolding?)` stays pure and synchronous. That is what makes the route/CLI comparison one snapshot fed through two call sites rather than two stubbed network chains.

## `forest-bff openapi` is now hybrid

- Configured to be inspected (`FOREST_SERVER_URL`, `FOREST_ENV_SECRET`, `FOREST_AUTH_SECRET`, `AGENT_URL`) → unfolded. It signs its own short-lived agent token: the agent verifies only the signature, and the capabilities route runs no permission check.
- Not configured → the generic document, so the command keeps working with no configuration at all, as documented.
- Configured but the schema cannot be read → **exits 1**, rather than a generic document that would look like a complete answer. The route answers 503 in the same situation, like the data routes.
- Configured, schema readable, but **not one** collection could be described → **exits 1** too: the paths would be real with every field schema free-form, which is not the document this command advertises. One degraded collection out of many stays a warning.

Each capabilities call signs its own agent token: one token lives 5 minutes, and a fan-out over hundreds of collections against a slow agent can outlive that.

`info.description` states which of the two forms the document is.

## Size

Unfolding multiplies the per-path cost, so it was measured first. The 12 error responses were inlined on every path — 2 847 of the ~3 800 bytes of a path. They are now `components/responses` refs, and the timezone header is a shared `components/parameters` ref.

| | before | after |
| -- | -- | -- |
| Generic document | 54 KB | **36 KB** |
| Per path | 3 775 B | **1 394 B** |

Measured on the unfolded document itself, 5 fields per collection and no relation or action, so the per-path figure above is not the whole story — each collection also registers its own `Fields_*`, `FilterableFields_*`, `Filter_*`, `FilterLeaf_*`, `SortClause_*`, `ListRequest_*` and `CountRequest_*` components:

| Collections | Paths | Document |
| -- | -- | -- |
| 10 | 20 | 100 KB |
| 100 | 200 | 899 KB |
| 200 | 400 | 1.79 MB |

So ~600 paths lands near 2.6 MB, not the ~1.4 MB a naive per-path multiplication suggests (thanks @Tonours for catching that). Without the error-response and header factoring the same document would be several times that. No arbitrary cap on the number of paths: a measured case would have to overrun this first.

## Security

The document is **not filtered per caller** — the allow-list is schema-derived and the agent's capabilities route runs no permission check, so any authenticated caller reads the name of every exposed collection, relation and field. That is stated in `info.description` and in the README next to `BFF_OPENAPI_ENABLED`, which already exists to close that surface. Per-role documents were rejected on purpose: a generated client would then depend on who ran the command. The route still requires an agent token before it reads anything, so an unauthenticated caller cannot make the BFF fetch a schema.

## Scope

Untouched: the data routes' behaviour, the action routes' behaviour, the agent, the MCP server, the OAuth flow. No SaaS or DB write. Per-field operator metadata is PRD-685, verifying a generated client is PRD-687.

Two additive changes outside `src/openapi/`:
- `ReadModel` gains `getAllowedCollections()`, `getListableRelations()` and `getListableRelationTarget()`.
- `CapabilitiesResult.type` is widened from `string` to the real column type: the agent returns the raw `columnType`, so an array-of-primitive column arrives as `['String']` and the old declaration was wrong. Nothing read it before.

**Stop condition #2 is resolved.** This branch was written before #1804 landed and added `getAllowedCollections()` with the same signature on purpose, so the rebase conflict was textual — both bodies were identical. Rebased onto `main` at `788ae0f3c`; the only other resolution was `cli-core.ts`, where #1804 split the store's lifetime from the agent's.

## How to test

```bash
yarn workspace @forestadmin/agent-bff test   # 1018 tests
```

`redocly lint` runs inside the suite, on the generic **and** the unfolded document, and both must come out with the single deliberate warning (the unused session scheme).

Against a real deployment:

```bash
cd packages/agent-bff && yarn build
node --env-file=.env dist/cli.js openapi --output /tmp/openapi.json
jq -r '.paths | keys[]' /tmp/openapi.json
env -u AGENT_URL node dist/cli.js openapi | jq -r '.info.description'   # the generic form
```

The fixture used by the tests covers a collection name with a space, a dotted name, an action name with a slash, two action names that collapse when sanitized, a `ManyToOne`, a composite key, a parent with no key metadata, and a collection whose capabilities failed. To-one, polymorphic and hidden-target relations are covered on the collect side, where they are filtered out.

### General

### Security

> [!NOTE]
> Fixes PRD-684
>
> ## What
>
> The document used to describe six generic paths with `{collection}`, `{relation}` and `{action}` as path parameters, so a codegen tool produced six methods where everything is a `string`. It now emits **one path per exposed collection, to-many relation and action**, each carrying that collection's real field set.
>
> The **runtime routes are untouched** — they stay the generic regexes. A documented path is the generic route with its segments already filled in.
>
> ```
> POST /agent/v1/customers/list
> POST /agent/v1/customers/relations/orders/list      # fields of `orders`, the foreign side
> POST /agent/v1/customers/actions/Mark%20as%20paid/execute
> ```
>
> ## Decisions
>
> **Field names come from `capabilities()`, not from the schema fields.** The schema carries every column, but `capabilities-validator.ts` answers 422 for any projection/sort/filter field absent from capabilities — a document built from schema fields would advertise JSON columns and to-many relations the runtime rejects. Cost: one capabilities call per collection, through the existing 24h per-collection cache shared with the data routes, capped at 10 in flight.
>
> **Two field enums per collection, not one.** The agent reports a `ManyToOne` without operators, so it is projectable and sortable but not filterable (422 `field_not_filterable`).
>
> **The generic paths are replaced, not kept alongside.** Coexistence is lint-clean (checked), but a document describing each route twice, once typed and once as `string`, is not the concrete surface this is for.
>
> **`operationId` is sanitized; the exact name lives in prose.** `redocly lint` errors on `operation-operationId-url-safe` for any operationId containing a space or a slash, so "the operationId carries the exact action name" and "redocly lint green" cannot both hold for an action named `Mark as paid`. Non-`[A-Za-z0-9_]` becomes `_`, collisions get a deterministic numeric suffix, and the **path segment** keeps the exact name URL-encoded — which is what the middleware decodes. `%2F` and `%20` in a path key are lint-clean (checked).
>
> **Action `values` is typed but indicative.** Properties come from the schema's static form fields, all optional, additional properties allowed. `read-model.ts` defaults `hooks`/`fields`, which destroys the absent-vs-empty distinction `agent-client` documents as load-bearing (`domains/action.ts:101-107`) — so `hooks.load === false` cannot be trusted to mean "static form", and strict typing would reject valid calls on a legacy schema.
>
> **`parentId` is typed from the parent's key.** `Number` → number, anything else → string, composite → the `|`-packed string, falling back to the opaque union when the parent exposes no key metadata.
>
> **A collection whose `capabilities()` fails keeps its paths, untyped.** Free-form `string` field schemas plus the reason in the request description and a `Warn` log — never an empty enum, which would forbid every valid call. Dropping the collection would hide something the runtime serves.
>
> **The listable-relation rule moved into the read-model.** It used to live privately in `data-routes-middleware.ts` (twice, after #1801); the unfolding needs the same predicate, and copies are how the document and the runtime drift apart. The foreign-collection allow-list check **stays** in the middleware: moving it would collapse `unknown_collection` into `unknown_relation`.
>
> **A relation request describes the foreign collection's fields**, and since #1801 the runtime validates them against exactly that collection's capabilities — so the enum and the 422 now agree instead of the document being stricter than the runtime.
>
> **With no `AGENT_URL`, the document stays generic**, even though the store can still read the schema (#1804 made that state real: the permissions endpoint needs the store without the agent). Concrete paths would be enumerable, but every data path answers 501 in that configuration, so unfolding would advertise a dead surface.
>
> **Collect and generate are separate.** All I/O and the fan-out live in `collect-unfolding.ts`; `generateOpenApiDocument(version, unfolding?)` stays pure and synchronous. That is what makes the route/CLI comparison one snapshot fed through two call sites rather than two stubbed network chains.
>
> ## `forest-bff openapi` is now hybrid
>
> - Configured to be inspected (`FOREST_SERVER_URL`, `FOREST_ENV_SECRET`, `FOREST_AUTH_SECRET`, `AGENT_URL`) → unfolded. It signs its own short-lived agent token: the agent verifies only the signature, and the capabilities route runs no permission check.
> - Not configured → the generic document, so the command keeps working with no configuration at all, as documented.
> - Configured but the schema cannot be read → **exits 1**, rather than a generic document that would look like a complete answer. The route answers 503 in the same situation, like the data routes.
> - Configured, schema readable, but **not one** collection could be described → **exits 1** too: the paths would be real with every field schema free-form, which is not the document this command advertises. One degraded collection out of many stays a warning.
>
> Each capabilities call signs its own agent token: one token lives 5 minutes, and a fan-out over hundreds of collections against a slow agent can outlive that.
>
> `info.description` states which of the two forms the document is.
>
> ## Size
>
> Unfolding multiplies the per-path cost, so it was measured first. The 12 error responses were inlined on every path — 2 847 of the ~3 800 bytes of a path. They are now `components/responses` refs, and the timezone header is a shared `components/parameters` ref.
>
> | | before | after |
> | -- | -- | -- |
> | Generic document | 54 KB | **36 KB** |
> | Per path | 3 775 B | **1 394 B** |
>
> Measured on the unfolded document itself, 5 fields per collection and no relation or action, so the per-path figure above is not the whole story — each collection also registers its own `Fields_*`, `FilterableFields_*`, `Filter_*`, `FilterLeaf_*`, `SortClause_*`, `ListRequest_*` and `CountRequest_*` components:
>
> | Collections | Paths | Document |
> | -- | -- | -- |
> | 10 | 20 | 100 KB |
> | 100 | 200 | 899 KB |
> | 200 | 400 | 1.79 MB |
>
> So ~600 paths lands near 2.6 MB, not the ~1.4 MB a naive per-path multiplication suggests (thanks @Tonours for catching that). Without the error-response and header factoring the same document would be several times that. No arbitrary cap on the number of paths: a measured case would have to overrun this first.
>
> ## Security
>
> The document is **not filtered per caller** — the allow-list is schema-derived and the agent's capabilities route runs no permission check, so any authenticated caller reads the name of every exposed collection, relation and field. That is stated in `info.description` and in the README next to `BFF_OPENAPI_ENABLED`, which already exists to close that surface. Per-role documents were rejected on purpose: a generated client would then depend on who ran the command. The route still requires an agent token before it reads anything, so an unauthenticated caller cannot make the BFF fetch a schema.
>
> ## Scope
>
> Untouched: the data routes' behaviour, the action routes' behaviour, the agent, the MCP server, the OAuth flow. No SaaS or DB write. Per-field operator metadata is PRD-685, verifying a generated client is PRD-687.
>
> Two additive changes outside `src/openapi/`:
> - `ReadModel` gains `getAllowedCollections()`, `getListableRelations()` and `getListableRelationTarget()`.
> - `CapabilitiesResult.type` is widened from `string` to the real column type: the agent returns the raw `columnType`, so an array-of-primitive column arrives as `['String']` and the old declaration was wrong. Nothing read it before.
>
> **Stop condition #2 is resolved.** This branch was written before #1804 landed and added `getAllowedCollections()` with the same signature on purpose, so the rebase conflict was textual — both bodies were identical. Rebased onto `main` at `788ae0f3c`; the only other resolution was `cli-core.ts`, where #1804 split the store's lifetime from the agent's.
>
> ## How to test
>
> ```bash
> yarn workspace @forestadmin/agent-bff test   # 1018 tests
> yarn workspace @forestadmin/agent-bff lint
> ```
>
> `redocly lint` runs inside the suite, on the generic **and** the unfolded document, and both must come out with the single deliberate warning (the unused session scheme).
>
> Against a real deployment:
>
> ```bash
> cd packages/agent-bff && yarn build
> node --env-file=.env dist/cli.js openapi --output /tmp/openapi.json
> jq -r '.paths | keys[]' /tmp/openapi.json
> env -u AGENT_URL node dist/cli.js openapi | jq -r '.info.description'   # the generic form
> ```
>
> The fixture used by the tests covers a collection name with a space, a dotted name, an action name with a slash, two action names that collapse when sanitized, a `ManyToOne`, a composite key, a parent with no key metadata, and a collection whose capabilities failed. To-one, polymorphic and hidden-target relations are covered on the collect side, where they are filtered out.
>
> ## Definition of Done
>
> ### General
>
> - [x] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)
> - [ ] Test manually the implemented changes
> - [x] Validate the code quality (indentation, syntax, style, simplicity, readability)
>
> ### Security
>
> - [x] Consider the security impact of the changes made
>
> 
> #### Changes since #1823 opened
>
> - Modified `buildUnfoldedDocument`, `createAgentCapabilitiesFetcher`, and CLI `renderOpenApi` command to accept token as either a string or factory function, enabling fresh token generation per capabilities fetch to avoid token expiration during long fan-outs [0e05324]
> - Modified `createOpenApiRoutes` middleware and `renderOpenApi` CLI command to detect degraded unfold results and handle them differently: middleware no longer memoizes documents with any degraded collection, and CLI aborts with error when no collections were successfully described [0e05324]
> - Adjusted `filterSchema` utility to specify types in mutual exclusion `not` clauses for leaf and branch filter discriminators, permitting leaf nodes with non-array `conditions` and branch nodes with non-string `field` properties [0e05324]
> - Changed `collectRelations` utility to sort relation names by code-unit comparison instead of locale-aware `localeCompare`, ensuring deterministic ordering [0e05324]
> - Introduced `toUnfoldSource` helper function to construct `UnfoldSource` from `ReadModelBundle`, config, and logger, replacing inline construction in `resolveUnfoldSource` and `buildAgentMiddlewares` [0e05324]
> - Modified `renderOpenApi` function in `agent-bff` to conditionally compute an unfoldable OpenAPI document only when `FOREST_AUTH_SECRET` is present and the environment requests unfolding, deferring config parsing and source resolution into a guarded branch with optional chaining fallback to the generic document [f578bbb]
> - Implemented `issueOpenApiAgentToken` function in `agent-bff` to generate signed JWT tokens for agent authentication with 5-minute expiry and identity claims including snake_case aliases [f578bbb]
> - Added null safety check in `toFieldSchema` function to prevent accessing `fields` property on null or malformed composite type values [f578bbb]
> - Added test coverage for OpenAPI unfolding defaulting behavior, field schema mappings, identifier sanitization, CLI token signing, and non-registered collection handling [f578bbb]
> - Added validation and filtering to drop malformed field entries lacking valid string field names from OpenAPI schema generation [64149d6]
> - Introduced a no-op `Metrics` implementation (`UNMEASURED`) and threaded an optional `metrics` parameter through `resolveReadModelBundle` to `createReadModel`, with `resolveUnfoldSource` passing `UNMEASURED` to suppress metric emissions during CLI openapi export [ca2540f]
> - Added test coverage to verify that `console.warn` is invoked when no logger is provided to the CLI and that stdout remains valid JSON during the openapi command with unfolding enabled [ca2540f]
> 
>

### DIFF
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -186,13 +186,16 @@ function resolveReadModelBundle(config: BFFConfig, logger: Logger): ReadModelBun
 }
 
 /**
- * What the OpenAPI document needs to unfold. It also needs the AGENT_URL the store does not: the
- * field set of a collection comes from the agent capabilities. Silent on purpose — the CLI and the
- * server report a missing configuration differently.
+ * When unfolding is possible, in ONE place: the document needs the AGENT_URL the store does not,
+ * because a collection's field set comes from the agent capabilities. The server passes the bundle it
+ * already built for the data routes; the CLI goes through `resolveUnfoldSource`. Silent on purpose —
+ * the two report a missing configuration differently.
  */
-export function resolveUnfoldSource(config: BFFConfig, logger: Logger): UnfoldSource | undefined {
-  const bundle = resolveReadModelBundle(config, logger);
-
+function toUnfoldSource(
+  bundle: ReadModelBundle | undefined,
+  config: BFFConfig,
+  logger: Logger,
+): UnfoldSource | undefined {
   if (!bundle || !config.agentUrl) return undefined;
 
   return {
@@ -201,6 +204,10 @@ export function resolveUnfoldSource(config: BFFConfig, logger: Logger): UnfoldSo
     timeoutMs: config.agentTimeoutMs,
     logger,
   };
+}
+
+export function resolveUnfoldSource(config: BFFConfig, logger: Logger): UnfoldSource | undefined {
+  return toUnfoldSource(resolveReadModelBundle(config, logger), config, logger);
 }
 
 // The data middleware falls through to the action middleware on a non-data path.
@@ -257,15 +264,7 @@ function buildAgentMiddlewares(config: BFFConfig, logger: Logger): Middleware[] 
   // One store for the whole edge. The document only unfolds when the agent is reachable too: with no
   // AGENT_URL every data path answers 501, so concrete paths would advertise a dead surface.
   const bundle = resolveReadModelBundle(config, logger);
-  const source =
-    bundle && config.agentUrl
-      ? {
-          store: bundle.store,
-          agentUrl: config.agentUrl,
-          timeoutMs: config.agentTimeoutMs,
-          logger,
-        }
-      : undefined;
+  const source = toUnfoldSource(bundle, config, logger);
 
   const chain: Middleware[] = [
     createAuthModeMiddleware({ authSecret: forestAuthSecret }),
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -1,6 +1,7 @@
 import type { BFFConfig } from './config/env-config';
 import type { UnfoldSource } from './openapi/unfolded-document';
 import type { Logger } from './ports/logger-port';
+import type { Metrics } from './ports/metrics-port';
 import type ReadModelStore from './read-model/read-model-store';
 import type { Middleware } from 'koa';
 
@@ -171,7 +172,11 @@ interface ReadModelBundle {
  * unfolding all share — one cache, one schema fetch. Needs no agent: the permissions endpoint and the
  * schema itself come from the SaaS.
  */
-function resolveReadModelBundle(config: BFFConfig, logger: Logger): ReadModelBundle | undefined {
+function resolveReadModelBundle(
+  config: BFFConfig,
+  logger: Logger,
+  metrics?: Metrics,
+): ReadModelBundle | undefined {
   const apiKeyConfig = resolveApiKeyConfig(config);
 
   if (!apiKeyConfig) return undefined;
@@ -180,6 +185,7 @@ function resolveReadModelBundle(config: BFFConfig, logger: Logger): ReadModelBun
     forestServerUrl: apiKeyConfig.forestServerUrl,
     envSecret: apiKeyConfig.forestEnvSecret,
     logger,
+    metrics,
   });
 
   return { store, apiKeyConfig };
@@ -206,8 +212,16 @@ function toUnfoldSource(
   };
 }
 
+/**
+ * Metrics are dropped rather than logged: without an explicit sink `createReadModel` builds a console
+ * one, which reports gauges at `Info`, and `createConsoleLogger` sends `Info` to `console.info` — the
+ * stdout the document is written to. The server keeps its metrics; the export has no sink for them
+ * anyway.
+ */
+const UNMEASURED: Metrics = { increment: () => undefined, gauge: () => undefined };
+
 export function resolveUnfoldSource(config: BFFConfig, logger: Logger): UnfoldSource | undefined {
-  return toUnfoldSource(resolveReadModelBundle(config, logger), config, logger);
+  return toUnfoldSource(resolveReadModelBundle(config, logger, UNMEASURED), config, logger);
 }
 
 // The data middleware falls through to the action middleware on a non-data path.
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -1,5 +1,7 @@
 import type { BFFConfig } from './config/env-config';
+import type { UnfoldSource } from './openapi/unfolded-document';
 import type { Logger } from './ports/logger-port';
+import type ReadModelStore from './read-model/read-model-store';
 import type { Middleware } from 'koa';
 
 import { bodyParser } from '@koa/bodyparser';
@@ -159,12 +161,55 @@ function buildApiKeyMiddleware(config: BFFConfig, logger: Logger): Middleware | 
   return createApiKeyMiddleware({ authenticator, logger });
 }
 
-// Data and action routes share one read-model store (a single cache, a single schema fetch). The
-// data middleware falls through to the action middleware on a non-data path.
-function buildAgentRouteMiddlewares(config: BFFConfig, logger: Logger): Middleware[] {
+interface ReadModelBundle {
+  store: ReadModelStore;
+  apiKeyConfig: ResolvedApiKeyConfig;
+}
+
+/**
+ * The read-model store the data routes, the action routes, the permissions endpoint and the OpenAPI
+ * unfolding all share — one cache, one schema fetch. Needs no agent: the permissions endpoint and the
+ * schema itself come from the SaaS.
+ */
+function resolveReadModelBundle(config: BFFConfig, logger: Logger): ReadModelBundle | undefined {
   const apiKeyConfig = resolveApiKeyConfig(config);
 
-  if (!apiKeyConfig) {
+  if (!apiKeyConfig) return undefined;
+
+  const { store } = createReadModel({
+    forestServerUrl: apiKeyConfig.forestServerUrl,
+    envSecret: apiKeyConfig.forestEnvSecret,
+    logger,
+  });
+
+  return { store, apiKeyConfig };
+}
+
+/**
+ * What the OpenAPI document needs to unfold. It also needs the AGENT_URL the store does not: the
+ * field set of a collection comes from the agent capabilities. Silent on purpose — the CLI and the
+ * server report a missing configuration differently.
+ */
+export function resolveUnfoldSource(config: BFFConfig, logger: Logger): UnfoldSource | undefined {
+  const bundle = resolveReadModelBundle(config, logger);
+
+  if (!bundle || !config.agentUrl) return undefined;
+
+  return {
+    store: bundle.store,
+    agentUrl: config.agentUrl,
+    timeoutMs: config.agentTimeoutMs,
+    logger,
+  };
+}
+
+// The data middleware falls through to the action middleware on a non-data path.
+function buildAgentRouteMiddlewares(
+  bundle: ReadModelBundle | undefined,
+  config: BFFConfig,
+  logger: Logger,
+): Middleware[] {
+  if (!bundle) {
     logger(
       'Warn',
       'Data, action and permissions endpoints disabled: FOREST_SERVER_URL, FOREST_ENV_SECRET or FOREST_AUTH_SECRET is missing',
@@ -173,11 +218,7 @@ function buildAgentRouteMiddlewares(config: BFFConfig, logger: Logger): Middlewa
     return [createAgentStubMiddleware()];
   }
 
-  const { store } = createReadModel({
-    forestServerUrl: apiKeyConfig.forestServerUrl,
-    envSecret: apiKeyConfig.forestEnvSecret,
-    logger,
-  });
+  const { store, apiKeyConfig } = bundle;
   const { agentUrl, agentTimeoutMs: timeoutMs } = config;
 
   const permissionsMiddleware = createPermissionsRoutesMiddleware({
@@ -213,14 +254,26 @@ function buildAgentMiddlewares(config: BFFConfig, logger: Logger): Middleware[] 
   }
 
   const apiKeyStep = buildApiKeyMiddleware(config, logger) ?? createApiKeyUnavailableGuard(logger);
+  // One store for the whole edge. The document only unfolds when the agent is reachable too: with no
+  // AGENT_URL every data path answers 501, so concrete paths would advertise a dead surface.
+  const bundle = resolveReadModelBundle(config, logger);
+  const source =
+    bundle && config.agentUrl
+      ? {
+          store: bundle.store,
+          agentUrl: config.agentUrl,
+          timeoutMs: config.agentTimeoutMs,
+          logger,
+        }
+      : undefined;
 
   const chain: Middleware[] = [
     createAuthModeMiddleware({ authSecret: forestAuthSecret }),
     apiKeyStep,
     createPerKeyOriginMiddleware(),
-    createOpenApiRoutes({ version, enabled: config.openapiEnabled }),
+    createOpenApiRoutes({ version, enabled: config.openapiEnabled, source }),
     createTimezoneMiddleware({ defaultTimezone }),
-    ...buildAgentRouteMiddlewares(config, logger),
+    ...buildAgentRouteMiddlewares(bundle, config, logger),
   ];
 
   return chain.map(agentScoped);
--- a/packages/agent-bff/src/cli-dispatch.ts
+++ b/packages/agent-bff/src/cli-dispatch.ts
@@ -69,25 +69,23 @@ export async function renderOpenApi(env: NodeJS.ProcessEnv, logger: Logger): Pro
   // to do with (HTTP_PORT, the OAuth keys, the default timezone). A deployment that never asked for an
   // unfolded document must not see its export die on one of those, so the config is parsed only once
   // the four variables unfolding needs are all present — and from there a bad value is a real failure.
-  if (!authSecret || !wantsUnfolding(env)) {
+  const unfoldable =
+    authSecret && wantsUnfolding(env)
+      ? { source: resolveUnfoldSource(parseConfig(env), logger), authSecret }
+      : undefined;
+
+  if (!unfoldable?.source) {
     logger('Warn', `Emitting the generic OpenAPI document: ${NOTHING_TO_UNFOLD}`);
 
     return `${serializeOpenApi(generateOpenApiDocument(version))}\n`;
   }
 
-  const source = resolveUnfoldSource(parseConfig(env), logger);
-
-  if (!source) {
-    logger('Warn', `Emitting the generic OpenAPI document: ${NOTHING_TO_UNFOLD}`);
-
-    return `${serializeOpenApi(generateOpenApiDocument(version))}\n`;
-  }
-
+  const { source } = unfoldable;
   const readModel = await source.store.getReadModel();
   const { document, unfolding } = await buildUnfoldedDocument(
     source,
     readModel,
-    () => issueOpenApiAgentToken(authSecret),
+    () => issueOpenApiAgentToken(unfoldable.authSecret),
     version,
   );
 
--- a/packages/agent-bff/src/cli-dispatch.ts
+++ b/packages/agent-bff/src/cli-dispatch.ts
@@ -42,6 +42,19 @@ export interface DispatchOutcome {
   server?: BFFHttpServer;
 }
 
+const UNFOLD_VARS = [
+  'FOREST_SERVER_URL',
+  'FOREST_ENV_SECRET',
+  'FOREST_AUTH_SECRET',
+  'AGENT_URL',
+] as const;
+
+const NOTHING_TO_UNFOLD = `${UNFOLD_VARS.join(', ')} must all be set to unfold it`;
+
+function wantsUnfolding(env: NodeJS.ProcessEnv): boolean {
+  return UNFOLD_VARS.every(name => (env[name] ?? '').trim() !== '');
+}
+
 /**
  * Unfolds when the deployment is configured to be inspected, and emits the generic document when it
  * is not configured at all — the command keeps working without configuration. A deployment that IS
@@ -49,27 +62,30 @@ export interface DispatchOutcome {
  * unfolded document, and a generic one would look like a complete answer.
  */
 export async function renderOpenApi(env: NodeJS.ProcessEnv, logger: Logger): Promise<string> {
-  const source = resolveUnfoldSource(parseConfig(env), logger);
   const authSecret = env.FOREST_AUTH_SECRET;
 
-  if (source && authSecret) {
-    const readModel = await source.store.getReadModel();
+  // `parseConfig` validates the WHOLE server configuration, including settings the export has nothing
+  // to do with (HTTP_PORT, the OAuth keys, the default timezone). A deployment that never asked for an
+  // unfolded document must not see its export die on one of those, so the config is parsed only once
+  // the four variables unfolding needs are all present — and from there a bad value is a real failure.
+  if (!authSecret || !wantsUnfolding(env)) {
+    logger('Warn', `Emitting the generic OpenAPI document: ${NOTHING_TO_UNFOLD}`);
 
-    return `${await buildUnfoldedDocument(
-      source,
-      readModel,
-      issueOpenApiAgentToken(authSecret),
-      version,
-    )}\n`;
+    return `${serializeOpenApi(generateOpenApiDocument(version))}\n`;
   }
 
-  logger(
-    'Warn',
-    'Emitting the generic OpenAPI document: FOREST_SERVER_URL, FOREST_ENV_SECRET, ' +
-      'FOREST_AUTH_SECRET or AGENT_URL is missing, so there is nothing to unfold against',
-  );
+  const source = resolveUnfoldSource(parseConfig(env), logger);
 
-  return `${serializeOpenApi(generateOpenApiDocument(version))}\n`;
+  if (!source) {
+    logger('Warn', `Emitting the generic OpenAPI document: ${NOTHING_TO_UNFOLD}`);
+
+    return `${serializeOpenApi(generateOpenApiDocument(version))}\n`;
+  }
+
+  const readModel = await source.store.getReadModel();
+  const token = issueOpenApiAgentToken(authSecret);
+
+  return `${await buildUnfoldedDocument(source, readModel, token, version)}\n`;
 }
 
 function rejectCli(reason: string): DispatchOutcome {
--- a/packages/agent-bff/src/cli-dispatch.ts
+++ b/packages/agent-bff/src/cli-dispatch.ts
@@ -10,6 +10,7 @@ import { parseConfig } from './config/env-config';
 import { extractErrorMessage } from './errors';
 import { generateOpenApiDocument, serializeOpenApi } from './openapi/openapi-document';
 import buildUnfoldedDocument, { issueOpenApiAgentToken } from './openapi/unfolded-document';
+import { isFullyDegraded } from './openapi/unfolding';
 import version from './version';
 
 export const DEFAULT_OUTPUT_FILE = 'openapi.json';
@@ -83,9 +84,25 @@ export async function renderOpenApi(env: NodeJS.ProcessEnv, logger: Logger): Pro
   }
 
   const readModel = await source.store.getReadModel();
-  const token = issueOpenApiAgentToken(authSecret);
+  const { document, unfolding } = await buildUnfoldedDocument(
+    source,
+    readModel,
+    () => issueOpenApiAgentToken(authSecret),
+    version,
+  );
 
-  return `${await buildUnfoldedDocument(source, readModel, token, version)}\n`;
+  // Not one collection came back with its field set, so the agent was unreachable throughout. The
+  // paths are real but every field schema is free-form, which is not the document this command
+  // promises — a CI job regenerating it needs something to branch on, and the degraded notes buried
+  // in the descriptions are not it. One odd collection stays a warning, not a failure.
+  if (isFullyDegraded(unfolding)) {
+    throw new Error(
+      'No collection could be described: the agent answered no capabilities call, so every field ' +
+        'schema in the document would be free-form. Check AGENT_URL and the agent, then retry.',
+    );
+  }
+
+  return `${document}\n`;
 }
 
 function rejectCli(reason: string): DispatchOutcome {
--- a/packages/agent-bff/src/cli-dispatch.ts
+++ b/packages/agent-bff/src/cli-dispatch.ts
@@ -4,9 +4,12 @@ import type { Logger } from './ports/logger-port';
 import { mkdirSync, writeFileSync } from 'fs';
 import path from 'path';
 
-import runCli from './cli-core';
+import createConsoleLogger from './adapters/console-logger';
+import runCli, { resolveUnfoldSource } from './cli-core';
+import { parseConfig } from './config/env-config';
 import { extractErrorMessage } from './errors';
 import { generateOpenApiDocument, serializeOpenApi } from './openapi/openapi-document';
+import buildUnfoldedDocument, { issueOpenApiAgentToken } from './openapi/unfolded-document';
 import version from './version';
 
 export const DEFAULT_OUTPUT_FILE = 'openapi.json';
@@ -17,7 +20,11 @@ export const USAGE = `Usage: forest-bff [command]
 
 Commands:
   (none)      Start the BFF server, configured from the environment.
-  openapi     Write the OpenAPI document to stdout. Needs no configuration.
+  openapi     Write the OpenAPI document to stdout. Needs no configuration, but
+              a deployment configured to reach its Forest schema and its agent
+              (FOREST_SERVER_URL, FOREST_ENV_SECRET, FOREST_AUTH_SECRET,
+              AGENT_URL) unfolds one path per collection, relation and action
+              instead of the generic ones.
               --output [file]  Write to a file instead of stdout, defaulting
                                to ${DEFAULT_OUTPUT_FILE} in the current directory.
 
@@ -35,7 +42,33 @@ export interface DispatchOutcome {
   server?: BFFHttpServer;
 }
 
-export function renderOpenApi(): string {
+/**
+ * Unfolds when the deployment is configured to be inspected, and emits the generic document when it
+ * is not configured at all — the command keeps working without configuration. A deployment that IS
+ * configured but whose schema cannot be read fails instead of quietly degrading: it asked for the
+ * unfolded document, and a generic one would look like a complete answer.
+ */
+export async function renderOpenApi(env: NodeJS.ProcessEnv, logger: Logger): Promise<string> {
+  const source = resolveUnfoldSource(parseConfig(env), logger);
+  const authSecret = env.FOREST_AUTH_SECRET;
+
+  if (source && authSecret) {
+    const readModel = await source.store.getReadModel();
+
+    return `${await buildUnfoldedDocument(
+      source,
+      readModel,
+      issueOpenApiAgentToken(authSecret),
+      version,
+    )}\n`;
+  }
+
+  logger(
+    'Warn',
+    'Emitting the generic OpenAPI document: FOREST_SERVER_URL, FOREST_ENV_SECRET, ' +
+      'FOREST_AUTH_SECRET or AGENT_URL is missing, so there is nothing to unfold against',
+  );
+
   return `${serializeOpenApi(generateOpenApiDocument(version))}\n`;
 }
 
@@ -64,7 +97,7 @@ function parseOutputOption(rest: string[]): OutputOption {
   };
 }
 
-function writeOpenApiFile(file: string): DispatchOutcome {
+function writeOpenApiFile(file: string, document: string): DispatchOutcome {
   const asDirectory = file.endsWith('/') || file.endsWith(path.sep);
   const destination = path.resolve(
     process.cwd(),
@@ -73,7 +106,7 @@ function writeOpenApiFile(file: string): DispatchOutcome {
 
   try {
     mkdirSync(path.dirname(destination), { recursive: true });
-    writeFileSync(destination, renderOpenApi());
+    writeFileSync(destination, document);
   } catch (error) {
     process.stderr.write(`Cannot write ${destination}: ${extractErrorMessage(error)}\n`);
 
@@ -124,11 +157,13 @@ export default async function dispatchCli(
     );
   }
 
+  const document = await renderOpenApi(env, logger ?? createConsoleLogger());
+
   if (file !== undefined) {
-    return writeOpenApiFile(file);
+    return writeOpenApiFile(file, document);
   }
 
-  process.stdout.write(renderOpenApi());
+  process.stdout.write(document);
 
   return { exitCode: 0 };
 }
--- a/packages/agent-bff/src/openapi/collect-unfolding.ts
+++ b/packages/agent-bff/src/openapi/collect-unfolding.ts
@@ -1,0 +1,143 @@
+import type {
+  CollectionFields,
+  UnfoldedAction,
+  UnfoldedCollection,
+  UnfoldedRelation,
+  Unfolding,
+} from './unfolding';
+import type { Logger } from '../ports/logger-port';
+import type { CapabilitiesFetcher } from '../read-model/capabilities-cache';
+import type ReadModel from '../read-model/read-model';
+import type ReadModelStore from '../read-model/read-model-store';
+
+import { extractErrorMessage } from '../errors';
+
+// A cold document costs one capabilities call per collection. They are capped rather than fired all
+// at once so a 200-collection deployment cannot open 200 sockets to the agent at the same time; a
+// warm BFF pays nothing, the per-collection cache is shared with the data routes.
+export const CAPABILITIES_CONCURRENCY = 10;
+
+export interface CollectUnfoldingOptions {
+  readModel: ReadModel;
+  store: ReadModelStore;
+  capabilitiesFetcher: CapabilitiesFetcher;
+  logger: Logger;
+  concurrency?: number;
+}
+
+async function mapWithConcurrency<T, R>(
+  items: T[],
+  limit: number,
+  run: (item: T) => Promise<R>,
+): Promise<R[]> {
+  const results = new Array<R>(items.length);
+  let cursor = 0;
+
+  // Recursive rather than a while loop: each worker takes the next item only once the previous one
+  // settled, which is what caps the concurrency.
+  const worker = async (): Promise<void> => {
+    if (cursor >= items.length) return;
+
+    const index = cursor;
+    cursor += 1;
+    results[index] = await run(items[index]);
+
+    await worker();
+  };
+
+  await Promise.all(Array.from({ length: Math.min(limit, items.length) }, worker));
+
+  return results;
+}
+
+const UNTYPED = (degraded: CollectionFields['degraded']): CollectionFields => ({
+  projectable: [],
+  filterable: [],
+  degraded,
+});
+
+async function collectFields(
+  collection: string,
+  { store, capabilitiesFetcher, logger }: CollectUnfoldingOptions,
+): Promise<CollectionFields> {
+  try {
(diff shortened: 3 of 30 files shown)


packages/agent-bff/src/openapi/openapi-routes.ts
```diff
@@ -10,31 +14,53 @@ const READ_METHODS = new Set(['GET', 'HEAD']);
...
-      return;
+  async function resolveDocument(ctx: Parameters<Middleware>[0]): Promise<string> {
+    if (!source) return generic as string;
+
+    // Before the schema is even read, so a caller with no agent credentials cannot make the BFF
+    // fetch anything: the unfolded document needs a token to ask the agent for capabilities.
+    const token = requireAgentToken(ctx);
+    const readModel = await resolveReadModel(source.store);
+
+    if (unfolded?.readModel !== readModel) {
```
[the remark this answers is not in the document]
nbouliol: Fixed in 5a1338dd5.

You are right, and the note I had left in `collect-unfolding.ts` was inconsistent with the rest of the package: it argued the memo is self-healing (it is — the next request sees a new read-model identity and rebuilds), but the request that loses the race is still served a document mixing one generation's structure with another's field sets. `ReadModelStore` goes to some length to avoid exactly that split-brain, so accepting it here was the wrong call.

`resolveDocument` now re-reads the read-model after the build and discards the document when the identity changed, retrying up to 3 times before failing — the same bounded pattern as `ReadModelStore.getCapabilities`, and driven by the 24h schema TTL rather than request volume. Two tests pin it: the document served is the new generation's, and an endlessly changing schema throws instead of looping.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -0,0 +1,475 @@
...
+}
+
+// A name reaches the runtime through decodeURIComponent, so the document must carry it encoded —
+// including the separators that would otherwise change the route, such as a slash inside an action
+// name (`a/b` is reachable only as `a%2Fb`).
+function segment(name: string): string {
+  return encodeURIComponent(name);
+}
+
+function fieldsEnum(
```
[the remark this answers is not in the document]
nbouliol: Not applying this one, deliberately — the premise is right but the remedy was ruled out before implementation.

The premise holds: with `capabilities.fields = []`, `capabilities-validator.ts` indexes nothing, so every field in `projection`, `sort` or `filter` answers 422. A free-form string is permissive where nothing is accepted.

Two reasons not to emit an empty enum:

1. It was an explicit product decision on PRD-684: `no_fields` and `capabilities_unavailable` get the same degraded path with different descriptions, and never an empty schema. The ticket's own stop condition asks for "a degraded path or skip the collection, rather than emitting a path with an empty schema".
2. `enum: []` is a JSON Schema that nothing satisfies. Codegen tools handle it badly (an enum type with no member), and the collection is still perfectly callable without `projection`, `sort` or `filter` — `hasCapabilityConstrainedInput` skips the capabilities check entirely in that case, so a plain list works. A document that made those properties unsatisfiable would be precise about the fields and misleading about the path.

What did change in 5a1338dd5: the `no_fields` description now states the consequence instead of implying it — "Any field sent in `projection`, `sort` or `filter` is therefore rejected with 422 unknown_field — call this path without them." The contract is expressed in prose rather than in the schema, which is the trade-off that was chosen.

If we want it in the schema later, `maxItems: 0` on `projection`/`sort` is the tool-safe form; it does not cover `filter`.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -0,0 +1,475 @@
...
+    });
+  });
+}
+
+/**
+ * Emits one path per exposed collection, to-many relation and action. The runtime routes stay the
+ * generic regexes — only the document unfolds.
+ */
+export default function registerUnfoldedPaths(deps: Deps, unfolding: Unfolding): void {
+  const namer = createNamer();
```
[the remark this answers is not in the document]
nbouliol: Fixed in 3bd6d13d0. Real bug, and nothing in the suite caught it.

Confirmed the collision: collection `A_B` with action `C` and collection `A` with action `B_C` both composed `executeAction_A_B_C` and `ActionRequest_A_B_C`, and since `ComponentPool.add` re-registers the same name, the second definition won — one action really did end up pointing at the other's request schema. Relations had the same hole.

Uniqueness is now enforced on the whole composed identifier: one namer per identifier family (collections, relations, actions), each fed `${collectionKey}_${childName}` rather than the child name alone, so the second composition gets the deterministic `_2` suffix. Component names dropped the now-redundant collection prefix (`RelationListRequest_${relationKey}`), since the composed key already carries it.

I kept `_` as the separator rather than picking one that cannot appear in a sanitized key: a non-`[A-Za-z0-9_]` separator is exactly what `operation-operationId-url-safe` and codegen method names reject, so the delimiter route trades this bug for that one.

Three tests pin it, on a fixture built for the collision: operationIds are all unique, the second composition is suffixed, and the two actions reference two distinct request schemas.

packages/agent-bff/src/cli-dispatch.ts
```diff
@@ -35,7 +42,33 @@ export interface DispatchOutcome {
...
 
-export function renderOpenApi(): string {
+/**
+ * Unfolds when the deployment is configured to be inspected, and emits the generic document when it
+ * is not configured at all — the command keeps working without configuration. A deployment that IS
+ * configured but whose schema cannot be read fails instead of quietly degrading: it asked for the
+ * unfolded document, and a generic one would look like a complete answer.
+ */
+export async function renderOpenApi(env: NodeJS.ProcessEnv, logger: Logger): Promise<string> {
+  const source = resolveUnfoldSource(parseConfig(env), logger);
```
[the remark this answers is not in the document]
nbouliol: Fixed in 3bd6d13d0. Verified the regression first: `parseConfig({ HTTP_PORT: "nope" })` throws `Invalid configuration: HTTP_PORT must be an integer between 0 and 65535.`, so `forest-bff openapi` did inherit validation of settings it never reads.

`renderOpenApi` now checks the four variables unfolding needs (`FOREST_SERVER_URL`, `FOREST_ENV_SECRET`, `FOREST_AUTH_SECRET`, `AGENT_URL`) directly on `env` and returns the generic document without touching `parseConfig` when any is missing. The configuration-independent export is back to being configuration-independent.

I did not make the failure non-fatal in the other direction: once all four are set, the deployment has asked for an unfolded document, and a malformed value stays an exit 1 with the offending variable named. Silently emitting a generic document there would look like a complete answer — the same reasoning as the 503 on an unreadable schema.

Two tests: a broken `HTTP_PORT` alone still yields the six generic paths; a broken `HTTP_PORT` alongside the four unfold variables rejects with a message naming `HTTP_PORT`.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -0,0 +1,489 @@
...
+        `The composite id of the parent ${quoted(parent)} record: the values of ` +
+        `${primaryKeys.map(key => key.name).join(', ')} joined by ` +
+        `${quoted(PACKED_ID_SEPARATOR)}, in that order.`,
+    };
+  }
+
+  const [key] = primaryKeys;
+  const schema = key.type === 'Number' ? { type: 'number' as const } : { type: 'string' as const };
+
+  return { ...schema, description: `The parent ${quoted(parent)} record id (${key.name}).` };
```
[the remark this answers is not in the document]
nbouliol: Fixed in c8355ca3c — you are right, and this was the same mistake I pushed back on in the two other threads: the document must never be stricter than the runtime.

Confirmed against `parseParentId` (`agent-query.ts:175-187`): it takes a non-blank string or a finite number whatever the key type is, then forwards it to the agent as a string. Narrowing a numeric key to `number` made a generated client reject `{"parentId": "123"}` on an endpoint that accepts it.

The shape is now `anyOf: [{ type: string, pattern: "\\S" }, { type: number }]` for every single key — the `pattern` kept because `parseParentId` rejects a blank string. What the read-model knows moved to the description: `The parent "X" record id (id, Number). A number or its string form are both accepted; the BFF forwards it to the agent as a string.` That preserves the point of typing it from the key (the client is told which column the id belongs to) without constraining the wire form.

One deliberate difference from the diff: a composite key stays `type: string`. `parseParentId` does accept a number there too, but a number can never carry the `|` separator, so it fails immediately afterwards in `unpackPrimaryKey` — documenting the only form that can work is not a false restriction.

packages/agent-bff/src/openapi/openapi-routes.ts
```diff
@@ -1,40 +1,82 @@
...
+    // A schema refresh landing during the capabilities fan-out mixes the new generation's field sets
+    // into this generation's collections, relations and actions. Such a document must not be served
+    // nor cached — a consumer would generate a client from a self-contradictory schema. Bounded like
+    // `ReadModelStore.getCapabilities`, and driven by the 24h schema TTL rather than request volume.
+    if ((await resolveReadModel(source.store)) !== readModel) {
+      return resolveDocument(ctx, attemptsLeft - 1);
     }
 
-    if (!READ_METHODS.has(ctx.method)) {
+    unfolded = { readModel, document };
```
Tonours: A document built while the agent was down stays memoized until the next schema revision: the memo is keyed on read-model identity, which only changes on a 24h TTL bump, so every later request gets the untyped document and no capabilities retry ever runs. `CapabilitiesCache` caches successes only, on purpose, and this line makes the transient failure sticky anyway. A 10s agent restart at the wrong moment costs a day of field-less spec, and a failed schema refresh serves stale without bumping the revision, so it can outlast 24h. Surface `degraded` from `buildUnfoldedDocument` and skip the memo write when any collection carries it.

packages/agent-bff/src/openapi/collect-unfolding.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// A to-many relation pointing at a collection absent from the allow-list is dropped: the runtime
+// answers 404 on it (data-routes-middleware checks the foreign collection), so documenting it would
+// promise a dead path.
+function collectRelations(readModel: ReadModel, collection: string): UnfoldedRelation[] {
+  return readModel
+    .getListableRelations(collection)
+    .filter(relation => readModel.isCollectionAllowed(relation.foreignCollection))
+    .sort((left, right) => left.name.localeCompare(right.name));
```
Tonours: Relations sort with `localeCompare` while collections (:131) and actions (:107) sort by code units, and `localeCompare` follows the process ICU locale. A CLI export and the HTTP route running under different `LANG` then order relations differently for one schema, which breaks the byte-identity the docstring above promises: `ärenden`/`orders`/`zulu` come out in three different orders under code units, `en-US` and `sv-SE`. Use `.sort()` on the names like the other two levels.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -0,0 +1,501 @@
...
+
+  const projectable = fieldsEnum(
+    pool,
+    `Fields_${plan.key}`,
+    fields.projectable,
+    `A field of ${quoted(name)}.`,
+  );
+
+  // Filterable is a strict subset of projectable: the agent reports a ManyToOne without operators,
+  // so projecting or sorting on it works while filtering on it answers 400 field_not_filterable.
```
Tonours: `field_not_filterable` is a 422, not 400 (`validation-errors.ts:10`). Same fix in `unfolding.ts:30`.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -0,0 +1,501 @@
...
+  const leaf = pool.add(`FilterLeaf_${plan.key}`, {
+    type: 'object',
+    description: `A single condition on ${quoted(name)}.`,
+    properties: {
+      field: filterable,
+      operator: { type: 'string', enum: OPERATORS },
+      value: {},
+    },
+    required: ['field', 'operator'],
+    not: { properties: { conditions: {} }, required: ['conditions'] },
```
Tonours: `not: { required: ['conditions'] }` also rejects a leaf carrying a non-array `conditions`, which the runtime accepts: `isBranch` needs `Array.isArray(conditions)` and `isLeaf` only needs a string `field` (`capabilities-validator.ts:24-38`), so `{field, operator, conditions: "x"}` is a valid leaf the document refuses. `not: { properties: { conditions: { type: 'array' } }, required: ['conditions'] }` matches the runtime exactly.

packages/agent-bff/src/cli-dispatch.ts
```diff
@@ -35,8 +42,50 @@ export interface DispatchOutcome {
...
+  if (!source) {
+    logger('Warn', `Emitting the generic OpenAPI document: ${NOTHING_TO_UNFOLD}`);
+
+    return `${serializeOpenApi(generateOpenApiDocument(version))}\n`;
+  }
+
+  const readModel = await source.store.getReadModel();
+  const token = issueOpenApiAgentToken(authSecret);
+
+  return `${await buildUnfoldedDocument(source, readModel, token, version)}\n`;
```
Tonours: With the agent unreachable every collection degrades and this still exits 0, while the README sells this form as "each carrying the collection's real field set". The schema path is right and covered, this is the capabilities one: the document is still unfolded and the degraded notes are there, so it is not silent, but a CI job regenerating `openapi.json` has nothing to branch on. What about exiting non-zero when every collection is degraded?

packages/agent-bff/src/cli-dispatch.ts
```diff
@@ -35,8 +42,50 @@ export interface DispatchOutcome {
...
+  const source = resolveUnfoldSource(parseConfig(env), logger);
+
+  if (!source) {
+    logger('Warn', `Emitting the generic OpenAPI document: ${NOTHING_TO_UNFOLD}`);
+
+    return `${serializeOpenApi(generateOpenApiDocument(version))}\n`;
+  }
+
+  const readModel = await source.store.getReadModel();
+  const token = issueOpenApiAgentToken(authSecret);
```
Tonours: This 5m token is signed once before the whole fan-out (`AGENT_TOKEN_EXPIRES_IN`, `agent-token.ts:8`). At the 10s default agent timeout and concurrency 10, a ~300-collection deployment crosses the TTL and the tail degrades to `capabilities_unavailable` with exit 0. Passing a token factory to `createAgentCapabilitiesFetcher` so each fetch signs fresh would close it.

packages/agent-bff/src/cli-core.ts
```diff
@@ -213,14 +254,26 @@ function buildAgentMiddlewares(config: BFFConfig, logger: Logger): Middleware[]
   }
 
   const apiKeyStep = buildApiKeyMiddleware(config, logger) ?? createApiKeyUnavailableGuard(logger);
+  // One store for the whole edge. The document only unfolds when the agent is reachable too: with no
+  // AGENT_URL every data path answers 501, so concrete paths would advertise a dead surface.
+  const bundle = resolveReadModelBundle(config, logger);
+  const source =
```
Tonours: This block re-encodes the `bundle && config.agentUrl` rule that `resolveUnfoldSource` (:193) already holds, with the same object built twice; the copy exists only because `resolveUnfoldSource` builds its own bundle. Extract a `toUnfoldSource(bundle, config, logger)` both call, so the server and the CLI cannot drift on when unfolding is possible.

Tonours (commented):
Reviewed the unfolding end to end against a local deployment (suite + lint green, 44 QA checks green). One blocking item, the rest are suggestions.

Two notes on the PR body rather than the code:

- The size table undercounts. The 1394 B/path figure covers `paths` only, but each collection also registers its own `Fields_*`, `FilterableFields_*`, `Filter_*`, `FilterLeaf_*`, `ListRequest_*`, `CountRequest_*` and `SortClause_*` components. Measured on this branch: 98 KB for 10 collections, 883 KB for 100, 1.76 MB for 200 (400 paths). So ~600 paths lands near 2.6 MB, not ~1.4 MB. The conclusion still holds, the number does not.
- `field_not_filterable` is a 422, not a 400 (`validation-errors.ts:10`). The 400 appears in the PR body, in the Linear design comment, and in two code comments.

Tonours (commented):
Two more, both outside the diff hunks so they land here.

`read-model.ts` — dropping `resolveForeignCollection` from the data middleware leaves `getRelationTarget` with no production caller, and `isRelationAllowed` was already test-only. Both are now exercised only by `read-model.test.ts`. Delete them or fold them into the listable API, otherwise the next reader has to work out which of the two relation APIs is the live one.

`test/openapi/openapi-cli.test.ts` — the agent-unreachable export is untested. The schema-failure path is pinned at :93, but the case where the schema reads fine and every capabilities call fails (unfolded document, no enums, degraded notes, exit 0) is not. That is the contract the comment above questions, so a test would make the exit-0 behaviour deliberate rather than incidental.

packages/agent-bff/src/openapi/openapi-routes.ts
nbouliol: Fixed in 0e05324e8 — this was the worst of the batch and I had not seen it. `CapabilitiesCache` caches successes only, on purpose, and my memo undid that guarantee one level up.

`buildUnfoldedDocument` now hands back the `Unfolding` alongside the document, and the route only writes the memo when `hasDegradedCollection` is false. A degraded document is served but never cached, so the fan-out re-runs until the agent answers again — the doc route is not a hot path, and the per-collection cache still absorbs the successes.

Two tests: two consecutive requests with capabilities failing run the fan-out twice (4 calls for 2 collections) instead of serving the memo, and the enumerated fields appear on the first request after the agent recovers.

packages/agent-bff/src/openapi/collect-unfolding.ts
nbouliol: Fixed in 0e05324e8. This one directly broke the acceptance criterion the PR is built around — `localeCompare` reads the process ICU locale, so the same schema could order relations differently between the CLI export and the HTTP route, and the "same document" comparison only held because both run under the same `LANG` in CI.

Relations now sort by code unit like the other two levels. Regression test uses your example: `zulu`, `ärenden`, `orders` come out `orders`, `zulu`, `ärenden` — the order `localeCompare` would not produce under `en-US`.

packages/agent-bff/src/openapi/unfolded-paths.ts
nbouliol: Fixed in 0e05324e8 — `validation-errors.ts:9-13` indeed builds it with 422, same as `unknown_field`.

Corrected in both code comments (`unfolded-paths.ts` and `unfolding.ts`) and in the PR body. The Linear design comment is next.

packages/agent-bff/src/openapi/unfolded-paths.ts
nbouliol: Fixed in 0e05324e8, and the symmetric case too.

You are right on the mechanism: `properties: { conditions: {} }` matched on the mere presence of the key, while `isBranch` needs `Array.isArray(conditions)` — so `{field, operator, conditions: "x"}` is a plain leaf the runtime accepts and the document refused. That is the same "stricter than the runtime" mistake I pushed back on in two other threads.

The branch alternative had it as well: `not: { properties: { field: {} } }` rejected `{aggregator, conditions, field: 42}`, which `isLeaf` does not read as a leaf either (it needs a STRING `field`). Both now carry the type:

- leaf: `not: { properties: { conditions: { type: "array" } }, required: ["conditions"] }`
- branch: `not: { properties: { field: { type: "string" } }, required: ["field"] }`

Pinned by a test that asserts the two types explicitly, with the reason in a comment.

packages/agent-bff/src/cli-dispatch.ts
nbouliol: Taken, in 0e05324e8 — with the threshold set at "not one collection could be described" rather than "any".

`isFullyDegraded` (non-empty collections, all degraded) now throws, so `forest-bff openapi` exits non-zero with `No collection could be described: the agent answered no capabilities call...`. That state means the agent was unreachable throughout, and the document is useless for the codegen the command exists for — same reasoning as the exit 1 on an unreadable schema.

I deliberately did not fail on ANY degraded collection: one odd collection out of two hundred would break every CI regeneration, and the degraded note plus the `Warn` log are the right signal at that scale. An empty schema is not treated as degraded either — there was nothing to enumerate.

Two tests: all-degraded rejects, one-degraded still emits the document with the other collection enumerated.

packages/agent-bff/src/cli-dispatch.ts
nbouliol: Fixed in 0e05324e8. Confirmed the arithmetic: `AGENT_TOKEN_EXPIRES_IN` is `5m` (`agent-token.ts:8`), and at the 10s default timeout with concurrency 10 a ~300-collection deployment reaches 300s in the worst case, so the tail would fail on an expired token and silently degrade.

`createAgentCapabilitiesFetcher` now accepts `token: string | (() => string)`: a factory builds the client per fetch so every call signs fresh, a plain string keeps the single client. The CLI passes `() => issueOpenApiAgentToken(authSecret)`; the route keeps passing its string, since a borrowed caller token cannot be re-minted — worth noting the same TTL applies there, and only the caller can do something about it.

Two tests in `agent-capabilities-fetcher.test.ts`: a string token builds one client for two fetches, a factory mints two distinct tokens.

packages/agent-bff/src/cli-core.ts
nbouliol: Fixed in 0e05324e8. Extracted `toUnfoldSource(bundle, config, logger)`; `resolveUnfoldSource` is now a one-liner over it for the CLI, and `buildAgentMiddlewares` calls it with the bundle it already built. The `bundle && config.agentUrl` rule and the object shape exist once.

nbouliol:
Thanks — this was a genuinely useful review, and the blocking item was the one I would have shipped without noticing. All seven are addressed in 0e05324e8, none were pushed back on. 1018 tests, lint and typecheck green.

On the two body notes:

**The size table.** You are right and I reproduced it before rewriting: 100 KB for 10 collections, 899 KB for 100, 1.79 MB for 200 (5 fields each, no relation or action). My 1394 B/path only covered `paths`, and multiplying it ignored the seven components each collection registers. The body now carries the measured table and ~2.6 MB for 600 paths, with the naive figure called out as wrong.

**`field_not_filterable`.** 422, not 400 (`validation-errors.ts:9-13`). Fixed in the two code comments and in the PR body; the Linear design comment is corrected too.

One thing your review changed beyond the code: the CLI now has two distinct failure modes rather than one — unreadable schema (exit 1, already there) and no collection describable (exit 1, new). Both are stated in the body under the hybrid section.

packages/agent-bff/src/openapi/field-schemas.ts
```diff
@@ -0,0 +1,46 @@
...
+  // Null-checked because the capabilities payload is cast from untyped JSON, so nothing guarantees a
+  // column type is one of the three shapes the type declares.
+  if (typeof type === 'object' && type !== null && Array.isArray(type.fields)) {
+    return {
+      type: 'object',
+      properties: Object.fromEntries(
+        type.fields.map(field => [field.field, toFieldSchema(field.type)]),
+      ),
+    };
+  }
```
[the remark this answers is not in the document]
nbouliol: Fixed in 64149d6bb. Same class as the null column type from the previous round, and reachable for the same reason: the schema an action field type comes from is cast from untyped JSON, so nothing guarantees the declared shape.

I filtered on the NAME rather than on the element being an object, which also covers your case and one more: `{ field: 42, type: "String" }` is an object, so your predicate keeps it, and `Object.fromEntries` would then coerce the key to `"42"`. The filter is now `typeof field?.field === "string"`, so a null entry, a nameless one and a numerically-named one are all dropped while their valid siblings survive.

While verifying it I found the same hole one layer up, not in the diff you flagged: `read-model.ts` turns a malformed entry into `{}` via `{ ...field }` (spreading null does not throw), so `collectActions` mapped it to `name: undefined` and the action `values` schema ended up carrying a property literally called `"undefined"`. No crash, just a lie in the document. Same filter applied in `collect-unfolding.ts`.

Four tests on the composite path (null entry, no name, non-string name, valid sibling preserved) and one on the action path.

packages/agent-bff/src/cli-dispatch.ts
```diff
@@ -124,11 +188,13 @@ export default async function dispatchCli(
     );
   }
 
+  const document = await renderOpenApi(env, logger ?? createConsoleLogger());
```
Tonours, line 191: This is where the regression enters: the default console logger reaches `createReadModel`, which builds a `createConsoleMetrics(logger)` when no `metrics` is passed. Those metrics log at `Info`, and `console-logger.ts:19` sends `Info` to `console.info` — stdout, the same stream as the document three lines down.

On `main` `renderOpenApi()` was synchronous and read nothing, so the question never came up.

Repro:

```
$ node --env-file=.env dist/cli.js openapi 2>/dev/null | head -1
{"level":"Info","message":"metric.gauge","timestamp":"...","metric":"schema_cache_age_seconds","value":0}
```

packages/agent-bff/src/cli-core.ts
```diff
@@ -159,12 +161,62 @@ function buildApiKeyMiddleware(config: BFFConfig, logger: Logger): Middleware |
...
+ * unfolding all share — one cache, one schema fetch. Needs no agent: the permissions endpoint and the
+ * schema itself come from the SaaS.
+ */
+function resolveReadModelBundle(config: BFFConfig, logger: Logger): ReadModelBundle | undefined {
   const apiKeyConfig = resolveApiKeyConfig(config);
 
-  if (!apiKeyConfig) {
+  if (!apiKeyConfig) return undefined;
+
+  const { store } = createReadModel({
```
Tonours, line 184: `metrics` is left out here, so `create-read-model.ts:33` falls back to `createConsoleMetrics(logger)` and the export starts emitting gauges on stdout.

Fine for the server, which wants its metrics. Not for the export, whose stdout is the document.

Passing a no-op `metrics` on the export path is enough — the option already exists, and it keeps the server's behaviour untouched.

Tonours (commented):
Ran the unfolded export against a live agent (16 collections, 56 paths). The document itself holds up: deterministic across runs and across locales, valid per redocly, unique operationIds, and the fully-degraded guard correctly exits 1 when the agent is unreachable.

One regression though, on stdout.

`forest-bff openapi > openapi.json` now produces a file `JSON.parse` rejects. 17 `{"level":"Info","message":"metric.gauge",...}` lines land on stdout ahead of the document:

```
$ node --env-file=.env dist/cli.js openapi > out.json
$ node -e 'JSON.parse(require("fs").readFileSync("out.json","utf8"))'
SyntaxError: Unexpected non-whitespace character after JSON at position 127
```

`jq` hides it — it parses the first object and stops. `JSON.parse` does not.

The chain:

```
cli-dispatch.ts:191     renderOpenApi(env, logger ?? createConsoleLogger())
cli-core.ts:210 -> 179  createReadModel({ ..., logger })      // no `metrics` passed
create-read-model.ts:33 safeMetrics(metrics ?? createConsoleMetrics(logger))
console-metrics.ts:12   logger('Info', 'metric.gauge', ...)
console-logger.ts:19    Error -> console.error, Warn -> console.warn, rest -> console.info
```

`console.info` writes to stdout in Node, so the metrics share the document's stream.

17 = 1 read-model read + 16 collections. `read-model-store.ts:55` calls `getReadModel()` before each `getCapabilities`, and the warm-cache branch (`schema-cache.ts:49`) re-emits the gauge every time.

This is new here. On `main` `renderOpenApi()` is synchronous and touches neither the schema nor the cache, so no metric is ever emitted. `createConsoleMetrics` predates the PR (#1732) — the CLI path just never reached it before.

Only the unfolded path is affected; the generic one still writes a clean stdout, which is why the existing tests miss it. `--output` writes a valid file, but still leaves the 17 lines on stdout, against what the README promises at line 46: "The path is confirmed on stderr, so stdout stays empty and pipeable."

Smallest fix I see: pass a no-op `metrics` in `resolveReadModelBundle` for the export path. `createReadModel` already takes the option, so it is an argument to supply, not an abstraction to build. Routing metrics to stderr would work too, but changes the server's behaviour, not just the export's.

Worth a regression test on the unfolded path asserting stdout parses — the current suite only covers the generic document there.

packages/agent-bff/src/cli-dispatch.ts
nbouliol: Fixed in ca2540fde. Reproduced against the live agent first, byte for byte what you got:

```
$ node --env-file=.env dist/cli.js openapi > repro.json
$ node -e "JSON.parse(require(\"fs\").readFileSync(\"repro.json\",\"utf8\"))"
SyntaxError: Unexpected non-whitespace character after JSON at position 127
```

17 lines, all `metric.gauge`, exactly the chain you traced. Your diagnosis was complete — I had nothing to add to it.

`resolveUnfoldSource` now passes a no-op `metrics` to `createReadModel`, so the console fallback is never built on the export path. The server path is untouched: `buildAgentMiddlewares` still calls `resolveReadModelBundle` without a sink and keeps its gauges.

After the fix, against the same deployment: 82 paths, stdout parses, and stderr carries nothing but the Node TLS warning that comes from my `.env`.

On why the suite missed it — the test that claims to cover exactly this (`should write nothing to stdout but the document, since console.info also targets stdout`) passes a noop logger, so it could never see it. Added a test that runs the REAL console logger through `dispatchCli(["openapi"], VALID_ENV)`, collects every `process.stdout.write`, and asserts the concatenation parses. It fails on the parent commit and passes on this one.

I considered routing the export logger to stderr wholesale instead, so that no future Info log can corrupt stdout regardless of its source. I did not: it adds a logging concept for a class of bug the new test now catches at the boundary that matters. If it ever fires again, that is the moment.

packages/agent-bff/src/cli-core.ts
nbouliol: Done in ca2540fde, exactly as you suggested — the option was already there.

`resolveReadModelBundle` takes an optional `metrics`, and `resolveUnfoldSource` (the CLI entry point) passes a no-op one. The server keeps calling it without a sink, so `createConsoleMetrics(logger)` still applies there and its gauges are unchanged.

Dropping them rather than sending them to stderr is deliberate: the export has no metrics sink to send them to, and 17 gauge lines per run would be noise in a CI log. The comment on the constant says why it exists, so nobody re-adds the default by accident.

Tonours (approved):

packages/agent-bff/src/cli-dispatch.ts
nbouliol: Re-verified today on the built binary rather than on jest, since the suite had already proved itself blind here — the pre-existing stdout tests passed a `noopLogger`, which is exactly what hid this. Stub Forest server on `/liana/forest-schema` plus a capabilities stub, then the same two invocations from your export, run against `ca2540fde` and against `64149d6bb` as a control:

| | fixed | control (64149d6bb) |
| -- | -- | -- |
| `openapi > file` | 20463 B, `JSON.parse` OK, stderr empty | `metric.gauge` on line 2, **parse fails** at pos 127 |
| `openapi --output f` | stdout **0 B** | stdout **254 B** |

The control reproduces both of your FAILs, so the harness can fail. The 2 gauge lines I get with one collection are the same mechanism as your 17: one gauge per schema-cache read.

Also checked the test the fix adds is a real guard, not decoration — restored the broken `cli-core.ts` and it fails on `expect(() => JSON.parse(written.join(''))).not.toThrow()`.

And `#1824` is rebased onto this tip, so the stacked branch does not carry a CLI that pollutes stdout when someone builds from it.

PR #1824, opened 2026-08-13
feat(agent-bff): document the real operator set of each filterable field

nbouliol:
Fixes PRD-685

Stacked on #1823 — review that one first.

## Why

The unfolded document enumerated the filterable fields but offered every toolkit operator on all of them. A generated client, or the Zendesk filter UI, could not tell that `Contains` works on a String and not on a Boolean until the agent answered 400.

## What

Each filter leaf now pairs a field with the operators its capabilities report, as a JSON `enum` a codegen can read:

```json
"FilterLeaf_My_Coll_1": {
  "properties": {
    "field":    { "type": "string", "enum": ["id", "tags"] },
    "operator": { "type": "string", "enum": ["Equal", "NotEqual", "In"] },
    "value": {}
  }
}
```

Fields are grouped by the operator set they share rather than emitted one leaf each. One leaf per field would be as exact but would multiply the document by the field count; the number of distinct operator sets is bounded by the column types a collection really uses, and the cross product of a group is precisely the set of valid `(field, operator)` pairs. Two fields of **different** types sharing an operator set land in one group, so nothing here is a type table.

`FilterableFields_<key>` is gone: each leaf carries its own field subset, and their union is exactly the filterable set. A field the agent reports without operators — a `ManyToOne` — is in no leaf at all, which is what the runtime's 422 `field_not_filterable` means.

## Notable decisions

- **Operators are normalized and canonically ordered at collect time.** The agent serializes them from a set, so their order is not contractual; a canonical order is what lets two fields be recognized as sharing a set, and it keeps the document stable across the route's and the CLI's independent capabilities calls.
- **An unmappable operator keeps the whole operator set on that field, and logs the skew.** Documenting less than the runtime accepts would make a generated client refuse a call the agent honours; throwing like the runtime does would cost the entire document for one field. No raw snake_case ever reaches the public document.
- **A collection whose capabilities could not be read keeps its permissive leaf** — free-form field, every operator. Unchanged behavior.
- **Relations needed no wiring.** A relation request composes the foreign collection request, so the operators come from the foreign collection by construction. Covered by tests both ways: a rich parent pointing at a degraded foreign, and a degraded parent pointing at a rich foreign.

## Cost

No extra network call — the collector already had the operators in hand and threw them away.

## Tests

1073 pass, lint clean. New coverage: per-field operator enums, cross-type grouping, canonical ordering, the version-skew fallback and its log, the relation operator source in both directions, and the absence of `FilterableFields_*`. The existing route/CLI parity test and the redocly spec-validity test cover the new schema shape.

> [!NOTE]
> ### Document per-field operator sets in filterable field schemas for agent-bff OpenAPI spec
> - The `filterable` field in `CollectionFields` changes from `string[]` to `FilterableField[]` (a `{ name, operators }` pair), giving each field its own canonical operator set.
> - Operators are normalized to PascalCase and deduplicated in canonical order via `toCanonicalOperatorSet`; unknown operators trigger a `Warn` log and fall back to the full `allOperators` set.
> - Filter leaf schemas in the OpenAPI output are now grouped by shared operator set, producing one leaf alternative per unique operator group instead of a single global leaf with a separate filterable-fields enum.
> - Collections with no known filterable fields get a permissive leaf (free-form field string + all operators); collections with known fields get scoped field enums paired with their operator enums.
> - Behavioral Change: `FilterableFields_*` schemas are removed from the OpenAPI output; filter schemas now use grouped leaf `$ref` alternatives.
>
> 
> #### Changes since #1824 opened
>
> - Changed `documentedOperators` function to return `null` when encountering operators with no canonical mapping from `normalizeOperator`, instead of returning all operators from `@forestadmin/datasource-toolkit` [eae3def]
> - Updated `collectFilterableFields` to exclude fields with unmappable operators by filtering out entries where `documentedOperators` returns `null` [eae3def]
> - Added guard in `groupByOperators` to skip fields with empty operator lists, preventing creation of unsatisfiable filter leaf schemas with `enum: []` for operators [eae3def]
> - Changed filter leaf component naming scheme from `FilterLeaf_${plan.key}_${index + 1}` to `FilterLeaf_${plan.key}-${index + 1}` to prevent namespace collisions after identifier sanitization [eae3def]
> - Added test coverage verifying that fields with unmappable operators are dropped from filterable fields, log appropriate warnings, remain projectable, do not degrade collection quality, and that empty operator sets fall back to generic leaves [eae3def]
> 
>
> 
>
> <details>
> <summary>📊 <a href="https://app.macroscope.com">Macroscope</a> summarized e4a8538. 4 files reviewed, 0 issues evaluated, 0 issues filtered, 0 comments posted</summary>
>
> ### 🗂️ Filtered Issues
> No issues evaluated.
>
>
> </details>
>

### DIFF
--- a/packages/agent-bff/src/openapi/collect-unfolding.ts
+++ b/packages/agent-bff/src/openapi/collect-unfolding.ts
@@ -12,8 +12,6 @@ import type ReadModel from '../read-model/read-model';
 import type ReadModelStore from '../read-model/read-model-store';
 import type { Operator } from '@forestadmin/datasource-toolkit';
 
-import { allOperators } from '@forestadmin/datasource-toolkit';
-
 import { extractErrorMessage } from '../errors';
 import { normalizeOperator, toCanonicalOperatorSet } from '../validation/operator-normalizer';
 
@@ -62,31 +60,35 @@ const UNTYPED = (degraded: CollectionFields['degraded']): CollectionFields => ({
 });
 
 /**
- * The operators to document for one field. An operator with no canonical mapping means the agent
- * runs a newer operator set than this package: the field then keeps the WHOLE operator set rather
- * than the mappable part of it. Documenting less than the runtime accepts would make a generated
- * client refuse a call the agent honours, and a raw snake_case operator must never reach the public
- * document — so the skew is surfaced in the log instead.
+ * The operators to document for one field, or null to leave the field out of the filterable set.
+ *
+ * An operator with no canonical mapping means the agent runs a newer operator set than this package.
+ * The field is dropped rather than widened, because `capabilities-validator` normalizes a field's
+ * WHOLE operator list before comparing the one that was sent: a skewed field answers 500
+ * `mapping_error` on ANY filter, even a plain `Equal`. Documenting it as filterable would promise a
+ * call the BFF never honours, and a raw snake_case operator must never reach the public document. The
+ * skew is surfaced in the log instead — dropping the field keeps the document truthful, so it stays
+ * cacheable.
  */
 function documentedOperators(
   collection: string,
   field: string,
   operators: string[],
   logger: Logger,
-): Operator[] {
+): Operator[] | null {
   const normalized: Operator[] = [];
 
   for (const operator of operators) {
     const canonical = normalizeOperator(operator);
 
     if (!canonical) {
-      logger('Warn', 'Documenting a field with every operator: one of its own has no mapping', {
+      logger('Warn', 'Documenting a field as not filterable: one of its operators has no mapping', {
         collection,
         field,
         operator,
       });
 
-      return [...allOperators];
+      return null;
     }
 
     normalized.push(canonical);
@@ -100,12 +102,18 @@ function collectFilterableFields(
   capabilities: CapabilitiesResult,
   logger: Logger,
 ): FilterableField[] {
-  return capabilities.fields
-    .filter(field => (field.operators?.length ?? 0) > 0)
-    .map(field => ({
-      name: field.name,
-      operators: documentedOperators(collection, field.name, field.operators as string[], logger),
-    }));
+  return capabilities.fields.flatMap(field => {
+    if ((field.operators?.length ?? 0) === 0) return [];
+
+    const operators = documentedOperators(
+      collection,
+      field.name,
+      field.operators as string[],
+      logger,
+    );
+
+    return operators === null ? [] : [{ name: field.name, operators }];
+  });
 }
 
 async function collectFields(
--- a/packages/agent-bff/src/openapi/collect-unfolding.ts
+++ b/packages/agent-bff/src/openapi/collect-unfolding.ts
@@ -1,16 +1,21 @@
 import type {
   CollectionFields,
+  FilterableField,
   UnfoldedAction,
   UnfoldedCollection,
   UnfoldedRelation,
   Unfolding,
 } from './unfolding';
 import type { Logger } from '../ports/logger-port';
-import type { CapabilitiesFetcher } from '../read-model/capabilities-cache';
+import type { CapabilitiesFetcher, CapabilitiesResult } from '../read-model/capabilities-cache';
 import type ReadModel from '../read-model/read-model';
 import type ReadModelStore from '../read-model/read-model-store';
+import type { Operator } from '@forestadmin/datasource-toolkit';
+
+import { allOperators } from '@forestadmin/datasource-toolkit';
 
 import { extractErrorMessage } from '../errors';
+import { normalizeOperator, toCanonicalOperatorSet } from '../validation/operator-normalizer';
 
 // A cold document costs one capabilities call per collection. They are capped rather than fired all
 // at once so a 200-collection deployment cannot open 200 sockets to the agent at the same time; a
@@ -56,6 +61,53 @@ const UNTYPED = (degraded: CollectionFields['degraded']): CollectionFields => ({
   degraded,
 });
 
+/**
+ * The operators to document for one field. An operator with no canonical mapping means the agent
+ * runs a newer operator set than this package: the field then keeps the WHOLE operator set rather
+ * than the mappable part of it. Documenting less than the runtime accepts would make a generated
+ * client refuse a call the agent honours, and a raw snake_case operator must never reach the public
+ * document — so the skew is surfaced in the log instead.
+ */
+function documentedOperators(
+  collection: string,
+  field: string,
+  operators: string[],
+  logger: Logger,
+): Operator[] {
+  const normalized: Operator[] = [];
+
+  for (const operator of operators) {
+    const canonical = normalizeOperator(operator);
+
+    if (!canonical) {
+      logger('Warn', 'Documenting a field with every operator: one of its own has no mapping', {
+        collection,
+        field,
+        operator,
+      });
+
+      return [...allOperators];
+    }
+
+    normalized.push(canonical);
+  }
+
+  return toCanonicalOperatorSet(normalized);
+}
+
+function collectFilterableFields(
+  collection: string,
+  capabilities: CapabilitiesResult,
+  logger: Logger,
+): FilterableField[] {
+  return capabilities.fields
+    .filter(field => (field.operators?.length ?? 0) > 0)
+    .map(field => ({
+      name: field.name,
+      operators: documentedOperators(collection, field.name, field.operators as string[], logger),
+    }));
+}
+
 async function collectFields(
   collection: string,
   { store, capabilitiesFetcher, logger }: CollectUnfoldingOptions,
@@ -73,9 +125,7 @@ async function collectFields(
 
     return {
       projectable: capabilities.fields.map(field => field.name),
-      filterable: capabilities.fields
-        .filter(field => (field.operators?.length ?? 0) > 0)
-        .map(field => field.name),
+      filterable: collectFilterableFields(collection, capabilities, logger),
       degraded: null,
     };
   } catch (error) {
--- a/packages/agent-bff/src/openapi/unfolded-paths.ts
+++ b/packages/agent-bff/src/openapi/unfolded-paths.ts
@@ -100,6 +100,11 @@ function groupByOperators(filterable: FilterableField[]): OperatorGroup[] {
   const groups = new Map<string, OperatorGroup>();
 
   filterable.forEach(field => {
+    // An empty set would emit `enum: []`, which forbids every value and makes the leaf unsatisfiable —
+    // the trap `fieldsEnum` guards above. `collectFilterableFields` cannot produce one, but `Unfolding`
+    // is hand-constructible plain data, so the generator holds the invariant rather than assume it.
+    if (field.operators.length === 0) return;
+
     // Safe as a key because the operators are normalized to one canonical order at collect time.
     const signature = field.operators.join(',');
     const group = groups.get(signature);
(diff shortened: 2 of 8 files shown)


packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -82,40 +83,113 @@ function fieldsEnum(
...
+ * for instance — is in none of them: filtering on it answers 422 field_not_filterable, so it must not
+ * be documented as filterable at all.
+ */
+function filterLeaves(pool: ComponentPool, plan: Pick<CollectionPlan, 'key' | 'collection'>) {
+  const { name, fields } = plan.collection;
+  const groups = groupByOperators(fields.filterable);
+
+  // No known filterable field: the field stays free-form and every operator stays allowed, because a
+  // collection whose capabilities could not be read still accepts whatever it really exposes.
+  if (groups.length === 0) {
```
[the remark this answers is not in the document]
nbouliol: The diagnosis is right — a healthy collection with no filterable field is documented as accepting a filter the runtime rejects with 422 `field_not_filterable`. Not taking the suggested fix, for two reasons.

**The fix produces a filter tree with no terminal.** Returning no leaf leaves `anyOf: [branch]`, and that branch's `conditions` are trees that can themselves only be branches. The only document satisfying it is `{aggregator: "And", conditions: []}`. It also drops `no_fields` from the permissive path, which would harden `filter` while `projection` and `sort` stay free-form on the same collection (`fieldsEnum` returns a plain string on an empty list), and it leaves `requestDescription` contradicting the schema.

The right fix, if this case were worth fixing, is to omit `filter` from `ListRequest_<key>` / `CountRequest_<key>` entirely and say so in the description — the truth of the runtime, rather than a degenerate shape nobody would write.

**But the case is not reachable through any shipped code path.** Emptying `filterable` on a healthy collection requires EVERY capabilities-emitted field to carry no operator. The agent (`routes/capabilities.ts`) emits every primitive column with its `filterOperators`, and a collection always has a primitive primary key, which carries operators. The only sources of an empty operator set are computed fields (`computed/collection.ts:97`) and hand-written datasources: the three `datasource-zendesk` collections are all mixed — 4/9 non-filterable fields on `organization`, 4/17 on `ticket`, 2/12 on `user`. None is at 100%.

So this is a documentation inaccuracy on a branch only a custom datasource that makes every one of its fields non-filterable could reach. Keeping the permissive leaf, which is also the pre-existing behavior of #1823 rather than something this PR introduces.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -82,40 +83,113 @@ function fieldsEnum(
...
+      pool.add(
+        `FilterLeaf_${plan.key}`,
+        leafShape({ type: 'string' }, OPERATORS, `A single condition on ${quoted(name)}.`),
+      ),
+    ];
+  }
+
+  return groups.map((group, index) =>
+    pool.add(
+      `FilterLeaf_${plan.key}_${index + 1}`,
```
Tonours: These two names step on each other: `FilterLeaf_${key}_${index + 1}` here, and the bare `FilterLeaf_${key}` at :152. `ComponentPool.add` does not dedup — it says "the caller owns name uniqueness" and hands off to `registerComponent`, which settles a duplicate with `rawComponents[type][name] = component`. Silent last-wins, no throw.

And collection keys already carry `_<n>` suffixes: `createNamer` mints them for names that collapse under sanitizing, and a customer can just call a collection `Invoice 1`. Ran the generator on `"Invoice 1"` (capabilities unavailable, so zero groups) + `"Invoice"` (one operator group on `total`):

```
schemas -> [ "FilterLeaf_Invoice_1" ]                       // one component, not two
Filter_Invoice_1.anyOf[0] -> #/.../FilterLeaf_Invoice_1
that leaf -> field: {enum:["total"]}, operator: {enum:["Equal","GreaterThan"]}
description -> 'A single condition on "Invoice".'
```

So `Invoice 1` documents `Invoice`'s fields and operators, and names the wrong collection in its own prose. Registration order picks the winner: flip the two and it's `Invoice` that loses its exact leaf, which hands a generated client operators the runtime kills with 400 `invalid_filter_operator`.

Same `_`-is-both-separator-and-sanitizer-output trap the comment at :552 documents for action and relation keys, except those go through a namer and these don't. `3bd6d13d0` closed that class, the new leaf namespace reopens it.

Two one-liners, either one: a separator sanitizing can't produce (`FilterLeaf_${key}-${index + 1}`, `-` is legal in a component key), or you run the leaf names through a namer. The fixture can't catch it, none of `My Coll` / `orders` / `users.address` sanitizes to a trailing `_<digit>`.

packages/agent-bff/src/openapi/collect-unfolding.ts
```diff
@@ -56,6 +61,53 @@ const UNTYPED = (degraded: CollectionFields['degraded']): CollectionFields => ({
...
+    const canonical = normalizeOperator(operator);
+
+    if (!canonical) {
+      logger('Warn', 'Documenting a field with every operator: one of its own has no mapping', {
+        collection,
+        field,
+        operator,
+      });
+
+      return [...allOperators];
```
Tonours: The docstring says documenting less than the runtime accepts would make a generated client refuse "a call the agent honours". The BFF honours none of them. `capabilities-validator.ts:60-68` normalizes the field's **whole** operator list through this same `normalizeOperator` and throws `mappingError` on the first miss, i.e. a **500 `mapping_error`** (`bff-local-errors.ts:31-33`).

So on a skewed field, `POST /agent/v1/:collection/list` with `{filter:{field:'id',operator:'Equal',value:1}}` answers 500. Not 400, not 422: a plain `Equal` is enough, the validator walks the field's full list, not the operator you sent. The document promises 44 operators on a field where every filter blows up, and the two modules contradict each other while sharing the helper.

Second half: `degraded` stays `null` here, so nobody sees the skew. `openapi-routes.ts:68` memoizes on `!hasDegradedCollection(unfolding)`, so a skewed document **is** cached for the schema TTL (24h, longer if a refresh fails without bumping the revision). `isFullyDegraded` never fires, `forest-bff openapi` exits 0. And `DEGRADED_NOTE` is keyed on `fields.degraded`, so no note at all — instead the leaf asserts "These fields share one operator set, as the collection's capabilities report it.", the one thing that's false right there. #1823 added that memo guard precisely to keep a transient degradation out of the document; this adds a third degradation mode the guard can't see.

My take: drop the skewed field from the leaves, same treatment as `operators: []`, keep the `Warn`. Smallest diff, and the document then states only what the BFF honours. If you want to keep the widening, then make it visible — `'operator_skew'` in `DegradedReason` (`unfolding.ts:53`) plus one `DEGRADED_NOTE` entry, and `hasDegradedCollection` pulls it out of the memo. But keeping the widening **and** `degraded: null`, no.

Either way the "the agent honours" line needs fixing: it describes the agent, not the BFF this document is the contract for.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -82,40 +83,113 @@ function fieldsEnum(
...
-  // A node readable as BOTH a leaf and a branch is rejected with 400 (`agent-query.ts`
-  // `assertNoNodeReadableAsBothLeafAndBranch`), so each alternative excludes the other's
-  // discriminator. The exclusion carries the TYPE the runtime looks for — `isBranch` needs an ARRAY
-  // `conditions` and `isLeaf` a STRING `field` — because `{field, operator, conditions: "x"}` is a
-  // plain leaf the runtime accepts. Expressed as `not: { required }` rather than
-  // `additionalProperties: false`, which would also forbid an unknown extra key the runtime strips.
-  const leaf = pool.add(`FilterLeaf_${plan.key}`, {
+  filterable.forEach(field => {
+    // Safe as a key because the operators are normalized to one canonical order at collect time.
+    const signature = field.operators.join(',');
```
Tonours, line 109: No guard on the empty set: a `FilterableField` with `operators: []` takes signature `''`, forms a group, lands in `leafShape`, which emits `operator: {type:'string', enum: []}`. An empty enum forbids every value, so the leaf is unsatisfiable. Tested:

```
{ name: 'E', filterable: [{ name: 'x', operators: [] }] }
  -> FilterLeaf_E_1.properties.operator = {"type":"string","enum":[]}
```

`collectFilterableFields` filters `(field.operators?.length ?? 0) > 0`, so the collect path can't produce it. But `Unfolding` is hand-constructible plain data on purpose — that's what its docblock argues, and the tests build three by hand. The generator doesn't hold the invariant its input type only documents.

`fieldsEnum` already guards exactly this risk 20 lines up: "An empty enum would forbid every value". Reuse it: skip empty groups here, or carry that guard into `leafShape`.

Tonours (commented):
Superseded — see the inline comments on this PR.

packages/agent-bff/src/openapi/unfolded-paths.ts
nbouliol: Fixed in the last commit, with your dash. Reproduced your run first, byte for byte:

```
FilterLeaf* components: [ 'FilterLeaf_Invoice_1' ]      <- one, not two
Filter_Invoice_1.anyOf[0] -> #/.../FilterLeaf_Invoice_1
Filter_Invoice.anyOf[0]   -> #/.../FilterLeaf_Invoice_1
that leaf: field {enum:["total"]}, desc 'A single condition on "Invoice"'
```

After:

```
FilterLeaf* components: [ 'FilterLeaf_Invoice_1', 'FilterLeaf_Invoice-1' ]
Filter_Invoice_1.anyOf[0] -> #/.../FilterLeaf_Invoice_1   field {type:'string'}, desc '"Invoice 1"'
Filter_Invoice.anyOf[0]   -> #/.../FilterLeaf_Invoice-1   field {enum:['total']}, desc '"Invoice"'
```

Took the separator over the namer for the reason you imply: `sanitizeIdentifier` maps everything outside `[A-Za-z0-9_]` to `_`, so a dash is unreachable and the two namespaces are disjoint by construction — no state, no dependence on registration order. `-` is legal in a component key and redocly lints the emitted document clean.

You were right that the fixture can't catch it, so the regression test builds the snapshot by hand: two collections, `Invoice 1` degraded and `Invoice` with one group, asserting two distinct leaf components and that the bare one names `"Invoice 1"` in its own prose with a free-form field. Checked it fails with the underscore restored.

And yes — same class as `3bd6d13d0`, which I closed for the action and relation keys and then reopened one namespace over.

packages/agent-bff/src/openapi/collect-unfolding.ts
nbouliol: You are right and the decision is reversed: the skewed field is now dropped from `filterable`, `Warn` kept, docstring rewritten. Took your option (a).

The premise I built the widening on was simply false. Confirmed your reading in `capabilities-validator.ts:60-72`: `normalizeFieldOperators` maps the field's whole list and throws `mappingError` on the first miss, and `bff-local-errors.ts:31-33` builds that as a 500. So a plain `Equal` on a skewed field answers 500 — the widening documented 44 operators on a field where nothing works, and the leaf description asserted "as the collection's capabilities report it", the one false statement in the block.

On your second half: dropping the field is what makes `degraded: null` correct rather than a hole. The document now states only what the BFF honours, so there is nothing to pull out of the memo and no note to add — which is why I did not take option (b). A test pins that: a collection whose only problem is a skewed field stays `degraded: null`.

The field stays projectable — only filtering on it fails — with a test for that too, and one that the surviving fields keep their own operators rather than the collection being written off.

Worth noting the 500 is itself questionable: an agent/BFF version skew answering `mapping_error` rather than something in the 4xx family is PRD-640's own stop condition. Out of scope here, and dropping the field is right either way — the field is unusable for filtering today whatever status the runtime settles on.

packages/agent-bff/src/openapi/unfolded-paths.ts
nbouliol: Fixed. `groupByOperators` skips an empty set, so the field forms no group; a collection left with no group falls back to the permissive leaf, same as a degraded one. Reproduced your case first (`FilterLeaf_E_1.properties.operator = {"type":"string","enum":[]}`), and after the fix `FilterLeaf_E_1` does not exist at all — only the bare `FilterLeaf_E` with a free-form field.

Guarded in the generator rather than leaning on `collectFilterableFields`, for the reason you give: `Unfolding` is hand-constructible plain data by design, so the generator has to hold the invariant its input type only documents. Same argument as the orphaned-relation guard already there. Regression test added, and it fails with the guard removed.

Tonours (approved):

PR #1827, opened 2026-08-14
test(agent-bff): pin the OpenAPI mount point under /agent and cover mode 2

nbouliol:
## Why

The BFF has no deny-by-default gate: auth applies only to paths under `/agent`, and `/health` is a hardcoded public exact-match ahead of the chain. `GET /agent/openapi.json` is therefore gated by construction, not by a mechanism — which means the next route added elsewhere could serve the document unauthenticated, silently. An unauthenticated OpenAPI document hands over the full API surface (collections, fields, filter operators, actions) — OWASP API8:2023.

PRD-886 already covers 401-without-credential, 200-with-a-session-token and 404-at-the-root. This PR adds what unit tests did not yet pin: the invariant itself, and the Mode 2 path.

## What

`test/openapi/openapi-mount-invariant.test.ts` (static, no server boot):

- `OPENAPI_PATH` must start with `/agent/`.
- No file in `src/` outside `src/openapi/` may import `src/openapi/*`, except the known mount points: `cli-core.ts` -> `openapi-routes`, `unfolded-document`, and `cli-dispatch.ts` -> `openapi-document`, `unfolded-document`, `unfolding`. Asserted as an exact map, so any new importer fails — the first CI run on this branch demonstrated it by catching the importers PRD-684 added on `main`.
- `src/index.ts` exports nothing matching `/openapi/i`, so a package consumer cannot mount the document itself.

`test/openapi/openapi-routes.test.ts`: the missing Mode 2 coverage — a valid API key gets 200 and the very same unfolded document a session caller gets, and 404 `openapi_disabled` with a valid API key when `BFF_OPENAPI_ENABLED=false`.

Test-only. No production code touched.

## Verification

- `yarn workspace @forestadmin/agent-bff test` — 1084 tests green; lint clean.
- Mutation-tested: moving `OPENAPI_PATH` to the root, adding an `openapi-document` import in another module, and exporting `createOpenApiRoutes` from `index.ts` each fail the suite. The Mode 2 cases fail if key resolution fails, so the 200 does not pass for the wrong reason.
- Manual, local BFF against a real Forest server: no credential 401, Mode 1 session token 200, Mode 2 API key 200, root `/openapi.json` 404. Logged on the ticket.

## Out of scope

A generic deny-by-default gate for the whole BFF, and pinning its public HTTP surface. Recorded as still-open in a ticket comment.

Fixes PRD-686

### DIFF
--- a/packages/agent-bff/test/openapi/openapi-mount-invariant.test.ts
+++ b/packages/agent-bff/test/openapi/openapi-mount-invariant.test.ts
@@ -1,0 +1,64 @@
+import { readFileSync, readdirSync } from 'fs';
+import path from 'path';
+
+import * as publicApi from '../../src/index';
+import { OPENAPI_PATH } from '../../src/openapi/openapi-routes';
+
+const SRC_DIR = path.join(__dirname, '..', '..', 'src');
+const OPENAPI_MODULE_PREFIX = 'openapi/';
+
+const ALLOWED_OPENAPI_IMPORTS: Record<string, string[]> = {
+  'cli-core.ts': ['openapi/openapi-routes', 'openapi/unfolded-document'],
+  'cli-dispatch.ts': ['openapi/openapi-document', 'openapi/unfolded-document', 'openapi/unfolding'],
+};
+
+function sourceFiles(): string[] {
+  return readdirSync(SRC_DIR, { recursive: true, encoding: 'utf8' })
+    .map(entry => entry.split(path.sep).join('/'))
+    .filter(entry => entry.endsWith('.ts'));
+}
+
+function relativeImportsOf(file: string): string[] {
+  const source = readFileSync(path.join(SRC_DIR, file), 'utf8');
+
+  return [...source.matchAll(/from '(\.[^']*)'/g)].map(match =>
+    path.posix.normalize(path.posix.join(path.posix.dirname(file), match[1])),
+  );
+}
+
+function openapiImportsOutsideTheOpenapiDir(): Record<string, string[]> {
+  const byFile: Record<string, string[]> = {};
+  const outsiders = sourceFiles().filter(file => !file.startsWith(OPENAPI_MODULE_PREFIX));
+
+  for (const file of outsiders) {
+    const targets = [
+      ...new Set(
+        relativeImportsOf(file).filter(target => target.startsWith(OPENAPI_MODULE_PREFIX)),
+      ),
+    ].sort();
+
+    if (targets.length > 0) byFile[file] = targets;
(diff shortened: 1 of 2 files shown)


packages/agent-bff/test/openapi/openapi-mount-invariant.test.ts
```diff
@@ -0,0 +1,64 @@
...
+function sourceFiles(): string[] {
+  return readdirSync(SRC_DIR, { recursive: true, encoding: 'utf8' })
+    .map(entry => entry.split(path.sep).join('/'))
+    .filter(entry => entry.endsWith('.ts'));
+}
+
+function relativeImportsOf(file: string): string[] {
+  const source = readFileSync(path.join(SRC_DIR, file), 'utf8');
+
+  return [...source.matchAll(/from '(\.[^']*)'/g)].map(match =>
```
Tonours, line 24: The scanner only matches static single-quoted `from '...'` imports, so `await import('../openapi/openapi-document')` slips through unflagged and nothing in the lint config forbids it. What about widening the regex to `/(?:from|import\s*\(|require\s*\()\s*'(\.[^']*)'/g`?

packages/agent-bff/test/openapi/openapi-mount-invariant.test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  describe('when a module outside src/openapi reaches into it', () => {
+    it('should only be the two known mount points, since any other one could serve the document off /agent', () => {
+      expect(openapiImportsOutsideTheOpenapiDir()).toEqual(ALLOWED_OPENAPI_IMPORTS);
+    });
+  });
+
+  describe('when the package public surface is read', () => {
+    it('should expose nothing OpenAPI-related, since a consumer could mount it off /agent', () => {
+      expect(Object.keys(publicApi).filter(name => /openapi/i.test(name))).toEqual([]);
```
Tonours, line 61: This pins the main entry only: with no `exports` map in package.json and `files: dist/**/*.js`, a consumer can still deep-import `@forestadmin/agent-bff/dist/openapi/openapi-document` and serve the document ungated. Adding `"exports": { ".": "./dist/index.js" }` in a follow-up would close it, or we narrow the test name to the main entry.

Tonours (approved):
Verified the mutation-testing claim locally: adding an openapi import outside the allowlist fails the invariant and names the file, and moving OPENAPI_PATH to the root breaks 18 tests. Two non-blocking suggestions inline on the scanner regex and the deep-import surface.

PR #1828, opened 2026-08-14
test(agent-bff): drive a generated client through list, count, form and execute

nbouliol:
## What

Proves the dynamic OpenAPI document is correct enough for codegen, not just schema-valid: a client generated from it can actually call list, count, action form and action execute.

The test:

1. emits the document through the CLI command (`dispatchCli(['openapi', '--output', …])`, the body of `forest-bff openapi`) from a fixture schema and fixture capabilities;
2. runs `@hey-api/openapi-ts` on that file, as a consumer would from the command line;
3. drives the generated client over HTTP against the **real** data and action middlewares (agent clients faked through the existing `createClient` seam), asserting a 2xx and the exact response body for each of the six operations.

One fixture feeds both the document and the runtime read-model. Two sources would let the test go green while the document described a surface the runtime does not have.

## Why the operator check has teeth

The filter operator is **read from the emitted enum** (`components.schemas.FilterLeaf_users-1.properties.operator.enum`) and injected into the call, instead of being hardcoded — so the assertion follows the document rather than a human's memory of it.

A negative control sends an operator the document leaves out and asserts `400 invalid_filter_operator`. Without it, a permissive stub (or a regression that stopped validating) would keep the check green. The fixture reports a single capability operator (`equal` → `Equal`) so the out-of-set operator lands unambiguously on 400 rather than on 422 `field_not_filterable` or 500 `mapping_error`.

## What it found

Nothing broken. Worth recording, since these were the risks:

- the percent-encoded action segments survive codegen and the round trip, including a non-ASCII name (`Générer la facture` → `G%C3%A9n%C3%A9rer%20la%20facture`);
- the recursive `anyOf` filter tree becomes a clean recursive TS union, and the operator enum reaches the generated types as a string literal;
- `not` is dropped by the codegen — a tooling quirk, not a document defect: the union already encodes the leaf/branch exclusivity structurally, and the runtime 400 stays documented in prose. The document is unchanged.

## Notes

- `@hey-api/openapi-ts` over `openapi-generator-cli`: the document is OpenAPI **3.1**, and openapi-generator converts 3.1 → 3.0 internally, which would make "document defect vs tooling quirk" undecidable on exactly the constructs at risk. It also avoids a Java toolchain and a jar download inside a CI test.
- Auth stays out of scope, as the ticket states — but the generated client does send the documented `X-Forest-Bff-Key` scheme, and the test asserts the credential arrives under that header.
- The generated client lands in a gitignored `test/openapi/.generated/`, wiped before and after the run. ESLint ignores dot-directories by default, so no lint config change was needed.
- Runtime: the whole file takes ~2.5 s; the package suite stays at ~10 s (1093 tests green).

Fixes PRD-687

> [!NOTE]
> ### Add generated-client test suite for `agent-bff` OpenAPI operations
> - Adds a Jest test in [openapi-generated-client.test.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1828/files#diff-3e1323a123d38645d81b8e8c73f430e37c5c8a0a63eb072613589257c2526c71) that runs the CLI to emit an OpenAPI doc, invokes `@hey-api/openapi-ts` to generate a TypeScript client, then drives the client against an in-memory Koa server through list, count, form, and execute flows.
> - Verifies the generated client exports documented operationIds, propagates security and timezone headers, accepts documented filter operators (Equal), and rejects undocumented ones (GreaterThan) with a 400.
> - Adds `@hey-api/openapi-ts` 0.99.0 to [package.json](https://github.com/ForestAdmin/agent-nodejs/pull/1828/files#diff-37a99046a348d45ffb67bbaf6761234b41b2c6c46fd28bb6e9c87c5ef4ed9f60) and ignores the `.generated` output directory in [.gitignore](https://github.com/ForestAdmin/agent-nodejs/pull/1828/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947).
>
> 
> #### Changes since #1828 opened
>
> - Added validation for OpenAPI client generation with action names containing path separators [1bf8e77]
> - Modified the `beforeAll` hook in the OpenAPI generated client test to capture stderr before spy restoration and clear inherited Node options during codegen [1bf8e77]
> - Moved the ignore rule for `test/openapi/.generated` from root `.gitignore` to `packages/agent-bff/.gitignore` [1bf8e77]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 90b37ef.</sup>
> 
>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,8 @@ lerna-debug.log
 # forest-bff openapi --output default destination
 openapi.json
 
+# document and client written by the OpenAPI codegen verification test
+packages/agent-bff/test/openapi/.generated
 
 # yarn
 yarn-error.log
--- a/.gitignore
+++ b/.gitignore
@@ -18,9 +18,6 @@ lerna-debug.log
 # forest-bff openapi --output default destination
 openapi.json
 
-# document and client written by the OpenAPI codegen verification test
-packages/agent-bff/test/openapi/.generated
-
 # yarn
 yarn-error.log
 .vscode/settings.json
--- a/packages/agent-bff/test/openapi/openapi-generated-client.test.ts
+++ b/packages/agent-bff/test/openapi/openapi-generated-client.test.ts
@@ -20,6 +20,8 @@ import { action, collection, column, relation } from '../read-model/fixtures';
 
 const MARK_AS_PAID = 'Mark as paid';
 const GENERATE_INVOICE = 'Générer la facture';
+// A name carrying the path separator: raw in the document it would not resolve to any route.
+const SPLIT_INVOICE = 'Facture 50/50';
 
 // One fixture for both sides: the document is generated from this schema and these capabilities, and
 // the chain the generated client calls runs on the very same ones. Two sources would let the test go
@@ -31,6 +33,7 @@ const SCHEMA = [
     [
       action(MARK_AS_PAID, '/forest/users/actions/mark-as-paid'),
       action(GENERATE_INVOICE, '/forest/users/actions/generer-la-facture'),
+      action(SPLIT_INVOICE, '/forest/users/actions/facture-50-50'),
     ],
   ),
   collection('orders', [column('id')]),
@@ -110,6 +113,8 @@ const OPERATIONS = [
   'executeActionUsersMarkAsPaid',
   'getActionFormUsersGNRerLaFacture',
   'executeActionUsersGNRerLaFacture',
+  'getActionFormUsersFacture5050',
+  'executeActionUsersFacture5050',
 ] as const;
 
 const dataClient: AgentDataClient = {
@@ -201,17 +206,21 @@ describe('a client generated from the emitted OpenAPI document', () => {
 
     // The command writes its own progress to stderr. Silenced so the run stays readable, but kept:
     // it is the only account of why an emission failed.
+    // Read out before restoring, not after: `mockRestore` clears `mock.calls`, so a message built
+    // from the spy afterwards would always be empty — which is the whole account of a failure here.
     const stderr = jest.spyOn(process.stderr, 'write').mockReturnValue(true);
+    let written = '';
     const emitted = await dispatchCli(
       ['openapi', '--output', DOCUMENT_FILE],
       ENV,
       noopLogger,
-    ).finally(() => stderr.mockRestore());
+    ).finally(() => {
+      written = stderr.mock.calls.flat().join('');
+      stderr.mockRestore();
+    });
 
     if (emitted.exitCode !== 0) {
-      throw new Error(
-        `The CLI could not emit the document: ${stderr.mock.calls.flat().join('')}`.trim(),
-      );
+      throw new Error(`The CLI could not emit the document: ${written}`.trim());
     }
 
     // Timed out rather than left to the `beforeAll` deadline: `spawnSync` blocks the event loop, so a
@@ -220,7 +229,13 @@ describe('a client generated from the emitted OpenAPI document', () => {
     const result = spawnSync(
       process.execPath,
       [CODEGEN_BIN, '--input', DOCUMENT_FILE, '--output', CLIENT_DIR, '--silent', '--no-log-file'],
-      { encoding: 'utf8', timeout: CODEGEN_TIMEOUT_MS },
+      {
+        encoding: 'utf8',
+        timeout: CODEGEN_TIMEOUT_MS,
+        // Cleared so a Node warning inherited from the parent cannot land in the output this file
+        // asserts on: what is measured here is the document, not the runner's flags.
+        env: { ...process.env, NODE_OPTIONS: '' },
+      },
     );
     codegenOutput = `${result.stdout ?? ''}${result.stderr ?? ''}`;
 
@@ -282,9 +297,12 @@ describe('a client generated from the emitted OpenAPI document', () => {
     it('should carry the documented operator set into the generated types', () => {
       const types = readFileSync(path.join(CLIENT_DIR, 'types.gen.ts'), 'utf8');
 
-      expect(
-        documentedOperators(document).filter(operator => !types.includes(`'${operator}'`)),
-      ).toEqual([]);
+      const operators = documentedOperators(document);
+
+      // Asserted against the literal first: both sides of the filter below derive from the document,
+      // so an empty enum would satisfy it vacuously.
+      expect(operators).toEqual([DOCUMENTED_OPERATOR]);
+      expect(operators.filter(operator => !types.includes(`'${operator}'`))).toEqual([]);
     });
   });
 
@@ -381,9 +399,20 @@ describe('a client generated from the emitted OpenAPI document', () => {
       });
     });
 
-    it('should reach an action whose name needs URL-encoding beyond spaces', async () => {
+    it('should reach an action whose name carries an accent', async () => {
       const form = await sdk.getActionFormUsersGNRerLaFacture({ body: { recordIds: [RECORD_ID] } });
       const executed = await sdk.executeActionUsersGNRerLaFacture({
+        body: { recordIds: [RECORD_ID] },
+      });
+
+      expect([form.response.status, executed.response.status]).toEqual([200, 200]);
+    });
+
+    // The separator is what makes this more than a formality: a raw name would split the path into
+    // segments the route cannot match, so only a percent-encoded document resolves here.
+    it('should reach an action whose name carries the path separator', async () => {
+      const form = await sdk.getActionFormUsersFacture5050({ body: { recordIds: [RECORD_ID] } });
+      const executed = await sdk.executeActionUsersFacture5050({
         body: { recordIds: [RECORD_ID] },
       });
 
--- a/packages/agent-bff/test/openapi/openapi-generated-client.test.ts
+++ b/packages/agent-bff/test/openapi/openapi-generated-client.test.ts
@@ -6,7 +6,7 @@ import type { Server } from 'http';
 
 import { bodyParser } from '@koa/bodyparser';
 import { spawnSync } from 'child_process';
-import { readFileSync, rmSync } from 'fs';
+import { existsSync, readFileSync, rmSync } from 'fs';
 import Koa from 'koa';
 import path from 'path';
 
@@ -76,11 +76,14 @@ const RECORD_ID = '1';
 const GENERATED_DIR = path.join(__dirname, '.generated');
 const DOCUMENT_FILE = path.join(GENERATED_DIR, 'openapi.json');
 const CLIENT_DIR = path.join(GENERATED_DIR, 'client');
+const SDK_FILE = path.join(CLIENT_DIR, 'sdk.gen.ts');
 
 const CODEGEN_BIN = path.join(
   path.dirname(require.resolve('@hey-api/openapi-ts/package.json')),
   'bin/run.js',
 );
+
+const CODEGEN_TIMEOUT_MS = 60_000;
 
 const noopLogger: Logger = () => undefined;
 
@@ -188,7 +191,7 @@ function documentedOperators(document: {
 }
 
 describe('a client generated from the emitted OpenAPI document', () => {
-  let codegen: { status: number | null; output: string };
+  let codegenOutput: string;
   let document: ReturnType<typeof JSON.parse>;
   let sdk: Record<string, Call>;
   let server: Server;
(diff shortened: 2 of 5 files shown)


packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -0,0 +1,401 @@
...
+    );
+    codegen = {
+      status: result.status,
+      output: `${result.stdout ?? ''}${result.stderr ?? ''}`,
+    };
+
+    document = JSON.parse(readFileSync(DOCUMENT_FILE, 'utf8'));
+
+    // eslint-disable-next-line global-require, import/no-dynamic-require
+    sdk = require(path.join(CLIENT_DIR, 'sdk.gen.ts'));
```
Tonours: When codegen exits non-zero, `sdk.gen.ts` is missing and this `require` fails with MODULE_NOT_FOUND, hiding the `codegen.output` we just captured. Let's throw right after `spawnSync` when `result.status !== 0`, embedding `codegen.output` in the message, mirroring the guard on line 203.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -0,0 +1,401 @@
...
+  beforeAll(async () => {
+    rmSync(GENERATED_DIR, { recursive: true, force: true });
+
+    const stderr = jest.spyOn(process.stderr, 'write').mockReturnValue(true);
+    const emitted = await dispatchCli(['openapi', '--output', DOCUMENT_FILE], ENV, noopLogger);
+    stderr.mockRestore();
+
+    if (emitted.exitCode !== 0) throw new Error('The CLI could not emit the document');
+
+    const result = spawnSync(
```
Tonours, line 229: `spawnSync` blocks the event loop, so the 60_000 `beforeAll` timeout can never fire and a hanging codegen child hangs the whole Jest run until CI kills the job. Let's pass `{ timeout: 60_000 }` to `spawnSync`.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -0,0 +1,401 @@
...
+describe('a client generated from the emitted OpenAPI document', () => {
+  let codegen: { status: number | null; output: string };
+  let document: ReturnType<typeof JSON.parse>;
+  let sdk: Record<string, Call>;
+  let server: Server;
+
+  beforeAll(async () => {
+    rmSync(GENERATED_DIR, { recursive: true, force: true });
+
+    const stderr = jest.spyOn(process.stderr, 'write').mockReturnValue(true);
```
Tonours, line 211: This spy swallows the CLI's own error text, so the throw on line 203 carries no cause, and if `dispatchCli` throws then `mockRestore` never runs and stderr stays mocked for the rest of the file. Let's build the message from `stderr.mock.calls.flat().join('')` and restore the spy in a `try`/`finally`.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -0,0 +1,401 @@
...
+
+    client.setConfig({
+      baseUrl: `http://127.0.0.1:${port}`,
+      auth: () => API_KEY,
+      headers: { [TIMEZONE_HEADER]: TIMEZONE },
+    });
+  }, 60_000);
+
+  afterAll(() => {
+    server?.close();
```
Tonours: `server?.close()` is fire-and-forget, so the port can still be open when the next suite starts: let's await it like `test/http/bff-http-server.test.ts:55-67` does, with `await new Promise(resolve => server.close(resolve))`.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
nbouliol: Done — `{ encoding: "utf8", timeout: CODEGEN_TIMEOUT_MS }`, with the `beforeAll` deadline set to twice that so the child is always the one to give up first. Added `--no-log-file` in the same call while verifying this: a failing run otherwise drops an `openapi-ts-error-*.log` in the working directory, which I hit for real.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
nbouliol: Done — the spy is restored through `.finally()` on the `dispatchCli` promise, so a throw restores it too, and the failure message is built from `stderr.mock.calls.flat().join("")` (the calls survive `mockRestore`).

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
nbouliol: Done — `afterAll` is async and awaits `server.close`, guarded on `server.listening` like `closeServer` in `test/http/bff-http-server.test.ts:55-67`, and rejecting on the close error rather than swallowing it.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
nbouliol: Done, and your comment found more than it aimed at. I probed the tool before trusting the status: `@hey-api/openapi-ts@0.99.0` exits **0** when the input file is missing (it prints a friendly diagnostic and returns success), while a real document defect — I tried a dangling `$ref` — exits 1. So a status guard alone would still have let the `require` fail with MODULE_NOT_FOUND in the case where our own emission wrote nothing.

The guard is therefore `result.status !== 0 || !existsSync(SDK_FILE)`, and the message carries the captured output, the status, `result.error?.message` when the child was killed, and the path that was never written when the output is empty. I verified it fires by pointing the codegen at a missing input: every test in the file now reports `The codegen could not consume the document (status 0): nothing was written to …/sdk.gen.ts` instead of MODULE_NOT_FOUND.

I kept the `codegenOutput === ""` assertion as its own test: under `--silent` any output at all means the document confused the tool, which is worth failing on separately from the hard failure.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -0,0 +1,431 @@
...
+    rmSync(GENERATED_DIR, { recursive: true, force: true });
+
+    // The command writes its own progress to stderr. Silenced so the run stays readable, but kept:
+    // it is the only account of why an emission failed.
+    const stderr = jest.spyOn(process.stderr, 'write').mockReturnValue(true);
+    const emitted = await dispatchCli(
+      ['openapi', '--output', DOCUMENT_FILE],
+      ENV,
+      noopLogger,
+    ).finally(() => stderr.mockRestore());
```
Tonours: The `.finally()` runs before the `if` right below it, and `mockRestore()` clears `mock.calls` on the way out. So `stderr.mock.calls.flat().join('')` always yields `''`: on a real emission failure the message is a bare `The CLI could not emit the document:`. That is exactly what 90b37ef88 set out to fix.

Verified with an isolated probe: `"captured-text"` before, `""` after `mockRestore()`.

Capture before restoring:

```ts
const stderr = jest.spyOn(process.stderr, 'write').mockReturnValue(true);
let written = '';
const emitted = await dispatchCli(['openapi', '--output', DOCUMENT_FILE], ENV, noopLogger).finally(
  () => {
    written = stderr.mock.calls.flat().join('');
    stderr.mockRestore();
  },
);

if (emitted.exitCode !== 0) {
  throw new Error(`The CLI could not emit the document: ${written}`.trim());
}
```

The twin guard below is fine, it reads `codegenOutput` captured right after the `spawnSync`.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -0,0 +1,431 @@
...
+  describe('when a standard codegen reads the document', () => {
+    it('should generate without a warning, since a warning is the document confusing the tool', () => {
+      expect(codegenOutput).toBe('');
+    });
+
+    it('should expose one function per documented operation, named after its operationId', () => {
+      expect(OPERATIONS.filter(name => typeof sdk[name] !== 'function')).toEqual([]);
+    });
+
+    it('should carry the documented operator set into the generated types', () => {
```
Tonours, line 297: This test cannot fail. Both sides derive from the same document: drop the operator enum and `documentedOperators(document)` returns `[]`, and `[].filter(...)` is `[]`. I mutated `operator: { type: 'string', enum: operators }` down to a bare `{ type: 'string' }` in `unfolded-paths.ts` and this test stayed green.

It is the worse of the two, since it is the only test claiming the enum survives into the generated client, which is the codegen half of the ticket.

Assert against the literal, or at least guard non-emptiness:

```ts
const operators = documentedOperators(document);

expect(operators).toEqual([DOCUMENTED_OPERATOR]);
expect(operators.filter(operator => !types.includes(`'${operator}'`))).toEqual([]);
```

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -0,0 +1,431 @@
...
+        body: { recordIds: [RECORD_ID], values: { comment: 'paid' } },
+      });
+
+      expect({ status: response.status, data }).toEqual({
+        status: 200,
+        data: { type: 'success', message: 'Done', invalidated: [], html: null },
+      });
+    });
+
+    it('should reach an action whose name needs URL-encoding beyond spaces', async () => {
```
Tonours: This one cannot fail either. I dropped the encoder (`segment` returning `name` as-is) and the suite stayed green. The paths go out raw and undici percent-encodes them client-side, so the route still resolves 200. The test does not tell an encoded document from a raw one.

It would only bite on a separator, a `/` inside an action name, which the comment names but the fixture does not have. Put an action like `Facture 50/50` in the fixture and the test becomes real.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -0,0 +1,431 @@
...
+      );
+    }
+
+    // Timed out rather than left to the `beforeAll` deadline: `spawnSync` blocks the event loop, so a
+    // codegen child that hangs would hang the whole Jest run until CI kills the job. `--no-log-file`
+    // because a failing run otherwise drops an `openapi-ts-error-*.log` in the working directory.
+    const result = spawnSync(
+      process.execPath,
+      [CODEGEN_BIN, '--input', DOCUMENT_FILE, '--output', CLIENT_DIR, '--silent', '--no-log-file'],
+      { encoding: 'utf8', timeout: CODEGEN_TIMEOUT_MS },
```
Tonours: `spawnSync` passes no `env`, so the child inherits `NODE_OPTIONS` (`--dns-result-order=ipv4first` here) and anything Node writes on its streams lands in `codegenOutput`. A `toBe('')` about *document* quality then goes red on an unrelated Node warning.

I could not trigger it on Node 24.10, so this is latent rather than broken. But the repo already emits `DEP0169` on the parent process, so the class of pollution is there.

```ts
{ encoding: 'utf8', timeout: CODEGEN_TIMEOUT_MS, env: { ...process.env, NODE_OPTIONS: '' } }
```

.gitignore
```diff
@@ -18,6 +18,8 @@ lerna-debug.log
 # forest-bff openapi --output default destination
 openapi.json
 
+# document and client written by the OpenAPI codegen verification test
+packages/agent-bff/test/openapi/.generated
```
Tonours: This is the only package-scoped path in the root `.gitignore`. Four sibling packages ignore their test artifacts locally: `datasource-sql` (`test/ssh-config/*`), `datasource-mongo`, `agent-testing` (`database-test*`, `schema-test.json`), `forest-cloud`.

It belongs in `packages/agent-bff/.gitignore` as `test/openapi/.generated`, otherwise a local test detail forces an edit at the root.

Tonours (commented):
Second pass on the branch head. Five findings: one blocking (the emission-failure message is always empty), two tests that cannot fail, one latent coupling and one gitignore placement.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
nbouliol: You are right, and my earlier reply on the parent thread was wrong: I claimed the calls survive `mockRestore`, they do not. Reproduced your probe — `"captured-text"` before, `""` after — so 90b37ef88 shipped exactly the empty message it set out to remove. Fixed with your shape: `written` is filled inside the `.finally()` before the restore, and the throw reads that.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
nbouliol: Taken — `env: { ...process.env, NODE_OPTIONS: "" }`. Agreed it is latent rather than broken, but the assertion is about the document and anything the parent runtime prints is noise in that channel, so there is no reason to leave it wired in.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
nbouliol: Right — both sides came from the same document, so an empty enum satisfied it. Reproduced your mutation (`operator: { type: "string" }` in `unfolded-paths.ts`) and the test stayed green; with the literal assertion added it now goes red. Took your snippet as written.

.gitignore
nbouliol: Agreed — moved to `packages/agent-bff/.gitignore` as `test/openapi/.generated`, and the root entry is gone. Verified with `git check-ignore -v`, which now points at the package file.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
nbouliol: Confirmed your experiment: with `segment` returning the name as-is the whole suite stayed green, because undici encodes the path on its way out.

Added `Facture 50/50` to the fixture, and it separates the two cases cleanly. Raw, the document emits `/agent/v1/users/actions/Facture 50/50/form`, whose extra separator no longer matches `ACTION_ROUTE` — so with the encoder mutated away the new test fails while the accented one still passes, which is exactly your point about what the accent does and does not prove.

Worth recording that the encoded form works end to end: the client sends `Facture%2050%2F50`, `[^/]+` matches it, `decodeSegment` gives back `Facture 50/50`, and `isActionAllowed` resolves it. I had deliberately kept a separator out of the fixture when planning this, expecting it to turn a document check into a routing hunt — it did not, so the objection was mine to drop.

Tonours (approved):
All five findings are addressed. I verified the two decorative tests now bite, by mutation rather than by reading the diff:

- `segment` returning `name` unchanged: `should reach an action whose name carries the path separator` goes red (14 passed, 1 failed). The `Facture 50/50` fixture is what gives it teeth.
- `operator: { type: 'string', enum: operators }` down to a bare `{ type: 'string' }`: `should carry the documented operator set into the generated types` now goes red too, along with the two operator tests (12 passed, 3 failed). It stayed green before the literal assertion.

The other three read clean: `written` is captured before `mockRestore()`, `NODE_OPTIONS: ''` isolates the child's output from the runner's flags, and the ignore rule moved into `packages/agent-bff/.gitignore` with the root entry gone.

Package suite 1099/1099 green locally, CI green on 1bf8e77.

mrlubos:
@nbouliol feel free to tag me if you have any feedback or run into issues with @hey-api!

PR #1829, opened 2026-08-14
feat(agent-bff): serve the OpenAPI document in a browser through Redoc

nbouliol:
> Stacked on #1828 — base it there, review this diff on its own, and retarget to `main` once #1828 lands.

## What

`GET /docs` renders the BFF's dynamic OpenAPI document in a browser, with Redoc served by the BFF itself. `GET /docs/redoc.standalone.js` serves the viewer bundle.

## Why it is shaped this way

BFF auth is header-only — `Authorization: Bearer <bff_access>` or `X-Forest-Bff-Key` (`src/auth/auth-mode-middleware.ts`), no cookie anywhere in the package. `resolveAuthMode` throws `unauthorized()` when neither header is present (`src/auth/auth-mode.ts:24-30`), and that middleware covers every `/agent/*` path. A browser navigating to a page sends neither header, and Redoc cannot attach one to its own spec fetch. So:

- both routes sit **outside** `/agent`, next to the existing public `/oauth/*` routes;
- the page is an **empty shell** — no schema in it, nothing to leak;
- it prompts for a BFF API key, fetches `/agent/openapi.json` with that header itself, and passes the parsed object to `Redoc.init`.

The document therefore stays exactly as unreachable as before: `test/cli-core.test.ts` asserts `/agent/openapi.json` still answers 401 while `/docs` answers 200, which is the interaction the mount invariant of #1827 cannot see on its own.

The key is held in memory only — passed as an argument, input cleared, never written to `localStorage` or `sessionStorage`, never put in a URL.

## Bundle provenance

`redoc` is a **devDependency**; its `redoc.standalone.js` is copied into `dist` at build time (`build:copy`, the pattern `forest-cloud` already uses for its templates). Consequences:

- no BFF consumer installs redoc's dependency tree;
- `files: dist/**/*.js` already covers the copied asset — no packaging change needed;
- running from `src` (tests, `build:watch`) there is nothing to copy to, so the route falls back to resolving the devDependency. Both paths are covered by the tests.

Not a CDN, deliberately: the page holds a credential in memory, and a third-party script in that page could read it. Pinning an SRI hash would mitigate that at the cost of a manual hash bump per version.

## Disabled state

Same flag as the document, `BFF_OPENAPI_ENABLED`. When it is off both routes **fall through to 404** rather than throw: `/docs` is outside the agent-scoped error middleware (`src/cli-core.ts` wraps `createErrorMiddleware` in `agentScoped`), so a thrown `openapiDisabled()` would surface as a bare Koa 500 instead of the BFF error contract. A 404 also keeps a disabled deployment from advertising a page it does not serve.

## Structural guarantee

`src/docs/` imports nothing from `src/openapi/` — the document path is passed in from `cli-core`. The existing `openapi-mount-invariant.test.ts` enforces this mechanically: any such import would show up in `openapiImportsOutsideTheOpenapiDir()` and fail. That is a stronger statement than the substring assertion on the page body, and both are in place.

## Tests

10 route tests plus 3 full-chain tests: page and bundle public, page carries no schema, page never cached, bundle served as a script, document still 401, both routes 404 when disabled, non-docs paths and writes passed through. Package suite: 71 suites / 1106 tests green, `yarn build` copies the bundle into `dist/docs/`.

## Not in scope

No "try it out" console, no write path from the page, and no OAuth login flow in the page — the BFF's OAuth is a registered-client authorization server, so the page would have to be registered as a client first.

Fixes PRD-965

> [!NOTE]
> ### Add Redoc documentation viewer and OpenAPI collection tags to `agent-bff`
> - Serves a public Redoc viewer at `/docs` and its bundle at `/docs/redoc.standalone.js` via `createDocsRoutes` in [docs-routes.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1829/files#diff-3dcba5d66b2d959c4b9880d5839f46457c1ce973af10e1fef317c2d9603312bf). The viewer renders the gated OpenAPI JSON after the user inputs a BFF API key client-side.
> - Injects per-operation code samples for cURL, JavaScript, and Ruby at runtime via `SAMPLES_SCRIPT` in [docs-samples.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1829/files#diff-df99f9972657ae5a6fd5da1c6cde3856ecca90542fe76d8ce83fb20709bcf61c).
> - Adds collection tags to unfolded OpenAPI documents in [openapi-document.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1829/files#diff-fe205dfba3d9163361b5101e76cdd78dc3bd2c2eb7a03e54d0bdc112b2b5584c) and [unfolded-paths.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1829/files#diff-972dcd53d7455a76a11d3004abd1b7f4be88a350650a3bd8ff57f02fa5814c54).
> - Adds an end-to-end test generating a TypeScript client with `@hey-api/openapi-ts` to validate the generated document's types and operations.
> - Behavioral Change: `OperationOptions` in [unfolded-paths.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1829/files#diff-972dcd53d7455a76a11d3004abd1b7f4be88a350650a3bd8ff57f02fa5814c54) now requires a `tag` property; the OpenAPI `info.title` changes from 'Forest Admin BFF' to 'Forest BFF'.
>
> 
> #### Changes since #1829 opened
>
> - Added depth-bounded traversal to prevent infinite recursion in self-referential OpenAPI schemas [5bf0327]
> - Parameterized and exported the `resolveBundle` function for testability [5bf0327]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8b78321.</sup>
> 
>

### DIFF
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -315,8 +315,13 @@ export default async function runCli(
     bodyParser({ jsonLimit: BODY_LIMIT }),
     ...oauthMiddlewares,
     // Outside the agent-scoped chain on purpose: the viewer is a public page, the document it fetches
-    // is not.
-    createDocsRoutes({ enabled: config.openapiEnabled, documentPath: OPENAPI_PATH, logger }),
+    // is not. Gated on the edge being mounted too, like the error middleware above: with no agent
+    // chain there is no document to fetch, and the page would only ever reach a bare Koa 404.
+    createDocsRoutes({
+      enabled: config.openapiEnabled && agentMiddlewares.length > 0,
+      documentPath: OPENAPI_PATH,
+      logger,
+    }),
     ...agentMiddlewares,
   ];
   const server = new BFFHttpServer({
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -19,6 +19,7 @@ import { parseConfig } from './config/env-config';
 import createCorsMiddleware from './cors/cors-middleware';
 import createPerKeyOriginMiddleware from './cors/per-key-origin';
 import createDataRoutesMiddleware from './data/data-routes-middleware';
+import createDocsRoutes from './docs/docs-routes';
 import { extractErrorMessage } from './errors';
 import { unauthorized } from './http/bff-http-error';
 import BFFHttpServer from './http/bff-http-server';
@@ -28,7 +29,7 @@ import ForestServerClient from './oauth/forest-server-client';
 import createOAuthRoutes from './oauth/oauth-routes';
 import createInMemorySessionStore from './oauth/session-store';
 import createTokenCipher from './oauth/token-cipher';
-import createOpenApiRoutes from './openapi/openapi-routes';
+import createOpenApiRoutes, { OPENAPI_PATH } from './openapi/openapi-routes';
 import PermissionsCache from './permissions/permissions-cache';
 import PermissionsClient from './permissions/permissions-client';
 import createPermissionsRoutesMiddleware from './permissions/permissions-routes-middleware';
@@ -313,6 +314,9 @@ export default async function runCli(
     ...agentErrorMiddleware,
     bodyParser({ jsonLimit: BODY_LIMIT }),
     ...oauthMiddlewares,
+    // Outside the agent-scoped chain on purpose: the viewer is a public page, the document it fetches
+    // is not.
+    createDocsRoutes({ enabled: config.openapiEnabled, documentPath: OPENAPI_PATH, logger }),
     ...agentMiddlewares,
   ];
   const server = new BFFHttpServer({
--- a/packages/agent-bff/src/docs/docs-page.ts
+++ b/packages/agent-bff/src/docs/docs-page.ts
@@ -6,6 +6,13 @@
  *
  * The key is never persisted: it is read from the input, passed down as an argument, and the input is
  * cleared. Once the document is fetched the page has no further use for it.
+ *
+ * Deliberately NOT a `<form>`. A form with no `action` navigates to `/docs?key=<the key>` the moment
+ * its default submit is not prevented — a CSP that blocks this inline script is enough — which would
+ * put the key in the browser history and in every access log on the way. A form submit is also what
+ * Chrome reads as a login, and it then offers to save the key whatever `autocomplete` says. With no
+ * form there is no default action to prevent and no submit to observe: without this script the button
+ * does nothing at all.
  */
 export default function renderDocsPage(documentPath: string, bundlePath: string): string {
   return `<!doctype html>
@@ -26,19 +33,21 @@ export default function renderDocsPage(documentPath: string, bundlePath: string)
     </style>
   </head>
   <body>
-    <form id="unlock">
+    <div id="unlock">
       <label for="key">BFF API key</label>
-      <input id="key" name="key" type="password" autocomplete="off" spellcheck="false" />
-      <button type="submit">Load the API document</button>
-    </form>
+      <input id="key" type="password" autocomplete="off" spellcheck="false" />
+      <button id="load" type="button">Load the API document</button>
+    </div>
     <div id="error"></div>
     <div id="redoc"></div>
     <script src="${bundlePath}"></script>
     <script>
       (function () {
         var DOCUMENT_PATH = ${JSON.stringify(documentPath)};
-        var form = document.getElementById('unlock');
+        var BUNDLE_PATH = ${JSON.stringify(bundlePath)};
+        var unlock = document.getElementById('unlock');
         var input = document.getElementById('key');
+        var button = document.getElementById('load');
         var errorBox = document.getElementById('error');
 
         function show(message) {
@@ -58,6 +67,33 @@ export default function renderDocsPage(documentPath: string, bundlePath: string)
           }
 
           return 'The BFF answered ' + status + ': ' + JSON.stringify(body);
+        }
+
+        /**
+         * Kept out of the fetch chain: a throw from here is a viewer problem, and reporting it as
+         * "could not reach the document" would point the reader at the wrong thing. \`untrustedSpec\`
+         * because the descriptions in the document come from the agent's own schema, and Redoc renders
+         * their markdown as HTML — unsanitized unless it is told the spec is untrusted.
+         */
+        function render(spec) {
+          if (typeof Redoc === 'undefined') {
+            show('The Redoc viewer did not load from ' + BUNDLE_PATH + ', so the document cannot be rendered.');
+
+            return;
+          }
+
+          unlock.style.display = 'none';
+
+          try {
+            Redoc.init(
+              spec,
+              { hideDownloadButton: true, untrustedSpec: true },
+              document.getElementById('redoc'),
+            );
+          } catch (initError) {
+            unlock.style.display = '';
+            show('The Redoc viewer could not render the document: ' + initError);
+          }
         }
 
         function load(key) {
@@ -87,22 +123,24 @@ export default function renderDocsPage(documentPath: string, bundlePath: string)
                 return;
               }
 
-              form.style.display = 'none';
-              Redoc.init(result.body, { hideDownloadButton: true }, document.getElementById('redoc'));
+              render(result.body);
             })
             .catch(function (fetchError) {
               show('Could not reach ' + DOCUMENT_PATH + ': ' + fetchError);
             });
         }
 
-        form.addEventListener('submit', function (event) {
-          event.preventDefault();
-
+        function unlockDocument() {
           var key = input.value.trim();
           input.value = '';
 
           if (key) load(key);
           else show('A BFF API key is required: the document is never served unauthenticated.');
+        }
+
+        button.addEventListener('click', unlockDocument);
+        input.addEventListener('keydown', function (event) {
+          if (event.key === 'Enter') unlockDocument();
         });
       })();
     </script>
--- a/packages/agent-bff/src/docs/docs-page.ts
+++ b/packages/agent-bff/src/docs/docs-page.ts
@@ -1,0 +1,112 @@
+/**
+ * The page is served WITHOUT credentials, so it must carry no schema: it is an empty shell that asks
+ * the caller for a BFF API key, fetches the document with it, and hands the parsed object to Redoc.
+ * That is the only design that is both openable in a browser — which sends no header when it
+ * navigates — and compatible with a document that is never reachable unauthenticated.
+ *
+ * The key is never persisted: it is read from the input, passed down as an argument, and the input is
+ * cleared. Once the document is fetched the page has no further use for it.
+ */
+export default function renderDocsPage(documentPath: string, bundlePath: string): string {
+  return `<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="robots" content="noindex" />
+    <title>Forest Admin BFF API</title>
+    <style>
+      body { margin: 0; font-family: system-ui, sans-serif; }
+      #unlock { display: flex; gap: 8px; flex-wrap: wrap; align-items: center; padding: 16px; border-bottom: 1px solid #e1e1e1; }
+      #unlock label { font-size: 14px; }
+      #unlock input { flex: 1 1 280px; padding: 6px 8px; font: inherit; }
+      #unlock button { padding: 6px 14px; font: inherit; cursor: pointer; }
+      #error { display: none; margin: 16px; padding: 12px; border-left: 3px solid #c00; background: #fff5f5; font-size: 14px; white-space: pre-wrap; }
+      #error[data-shown] { display: block; }
+    </style>
+  </head>
+  <body>
+    <form id="unlock">
+      <label for="key">BFF API key</label>
+      <input id="key" name="key" type="password" autocomplete="off" spellcheck="false" />
+      <button type="submit">Load the API document</button>
+    </form>
+    <div id="error"></div>
+    <div id="redoc"></div>
+    <script src="${bundlePath}"></script>
+    <script>
+      (function () {
+        var DOCUMENT_PATH = ${JSON.stringify(documentPath)};
+        var form = document.getElementById('unlock');
+        var input = document.getElementById('key');
+        var errorBox = document.getElementById('error');
+
+        function show(message) {
+          errorBox.textContent = message;
+          errorBox.setAttribute('data-shown', '');
+        }
+
+        function hide() {
+          errorBox.removeAttribute('data-shown');
+        }
+
+        function describe(status, body) {
+          var error = body && body.error;
+
+          if (error && error.type) {
+            return 'The BFF answered ' + status + ' ' + error.type + ': ' + (error.message || '');
+          }
+
+          return 'The BFF answered ' + status + ': ' + JSON.stringify(body);
+        }
+
+        function load(key) {
+          hide();
+
+          fetch(DOCUMENT_PATH, {
+            cache: 'no-store',
+            headers: { 'X-Forest-Bff-Key': key },
+          })
+            .then(function (response) {
+              return response.text().then(function (text) {
+                var body;
+
+                try {
+                  body = JSON.parse(text);
+                } catch (parseError) {
+                  body = { error: { type: 'unreadable_response', message: text.slice(0, 200) } };
+                }
+
+                return { ok: response.ok, status: response.status, body: body };
+              });
+            })
+            .then(function (result) {
+              if (!result.ok) {
+                show(describe(result.status, result.body));
+
+                return;
+              }
+
+              form.style.display = 'none';
+              Redoc.init(result.body, { hideDownloadButton: true }, document.getElementById('redoc'));
+            })
+            .catch(function (fetchError) {
+              show('Could not reach ' + DOCUMENT_PATH + ': ' + fetchError);
+            });
+        }
+
+        form.addEventListener('submit', function (event) {
+          event.preventDefault();
+
+          var key = input.value.trim();
+          input.value = '';
+
+          if (key) load(key);
+          else show('A BFF API key is required: the document is never served unauthenticated.');
+        });
+      })();
+    </script>
+  </body>
+</html>
+`;
+}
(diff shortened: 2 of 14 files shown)


packages/agent-bff/src/docs/docs-page.ts
```diff
@@ -0,0 +1,112 @@
...
+      #unlock input { flex: 1 1 280px; padding: 6px 8px; font: inherit; }
+      #unlock button { padding: 6px 14px; font: inherit; cursor: pointer; }
+      #error { display: none; margin: 16px; padding: 12px; border-left: 3px solid #c00; background: #fff5f5; font-size: 14px; white-space: pre-wrap; }
+      #error[data-shown] { display: block; }
+    </style>
+  </head>
+  <body>
+    <form id="unlock">
+      <label for="key">BFF API key</label>
+      <input id="key" name="key" type="password" autocomplete="off" spellcheck="false" />
```
Tonours: The form has no `method` and no `action`, so with inline scripts blocked the `preventDefault` at line 99 never runs and the browser navigates to `/docs?key=<secret>`, putting the key in browser history, BFF access logs and any proxy in between: let's drop `name="key"` since the script reads the input by id, and add `method="post"` as a belt.

Tonours: A password input, a form hidden on success and a cleared value are exactly Chrome's successful-login heuristic, which ignores `autocomplete="off"` on password fields and offers to save the key, so the "key is never persisted" claim in the header comment breaks as soon as the user accepts: let's use `autocomplete="new-password"`, or state that limit in the header comment instead of claiming the key is never persisted.

packages/agent-bff/src/docs/docs-page.ts
```diff
@@ -0,0 +1,112 @@
...
+            })
+            .then(function (result) {
+              if (!result.ok) {
+                show(describe(result.status, result.body));
+
+                return;
+              }
+
+              form.style.display = 'none';
+              Redoc.init(result.body, { hideDownloadButton: true }, document.getElementById('redoc'));
```
Tonours: Redoc 2.x renders markdown descriptions as HTML without sanitizing unless `untrustedSpec` is set, and DOMPurify already ships inside redoc's tree, so the option costs nothing on a page that holds a credential in memory: let's pass `untrustedSpec: true` to `Redoc.init`.

packages/agent-bff/src/docs/docs-page.ts
```diff
@@ -0,0 +1,112 @@
...
+            })
+            .catch(function (fetchError) {
```
Tonours, line 142: The `catch` is chained after the second `then`, so a throw from `Redoc.init` at line 91, including the `ReferenceError` when the bundle script did not load, is reported as "Could not reach /agent/openapi.json" and points the reader at the wrong thing: let's check `typeof Redoc === 'undefined'` before init with its own message, or catch around the init separately.

packages/agent-bff/src/cli-core.ts
```diff
@@ -313,6 +314,9 @@ export default async function runCli(
     ...agentErrorMiddleware,
     bodyParser({ jsonLimit: BODY_LIMIT }),
     ...oauthMiddlewares,
+    // Outside the agent-scoped chain on purpose: the viewer is a public page, the document it fetches
+    // is not.
+    createDocsRoutes({ enabled: config.openapiEnabled, documentPath: OPENAPI_PATH, logger }),
```
Tonours: Without `FOREST_AUTH_SECRET`, `buildAgentMiddlewares` returns `[]` so no OpenAPI route is mounted, yet the docs routes are gated on `config.openapiEnabled` alone, so `/docs` serves a page whose fetch can only ever get a bare Koa 404: the exact case `docs-routes.ts:51-52` says the 404 fall-through exists to avoid. Let's gate it like the error middleware two lines above, with `enabled: config.openapiEnabled && agentMiddlewares.length > 0`.

packages/agent-bff/test/docs/docs-routes.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    });
+
+    it('should never be cached, since the page is the entry point to a credential prompt', async () => {
+      const response = await request(buildApp(true).callback()).get(DOCS_PATH);
+
+      expect(response.headers['cache-control']).toBe('no-store');
+    });
+  });
+
+  describe('when the viewer bundle is requested without credentials', () => {
```
Tonours, line 108: The page's `no-store` is asserted at lines 52-56 but the bundle's `public, max-age=3600` is not, and it is the only cache header in the package that diverges from `no-store`: let's assert `expect(response.headers['cache-control']).toBe('public, max-age=3600')` in the bundle describe.

packages/agent-bff/src/docs/docs-page.ts
nbouliol: Done in 749ce16 — differently: the `<form>` is gone entirely rather than neutralised.

`method="post"` still leaves a form that submits. The POST would land on 404 (`READ_METHODS` only holds GET/HEAD), so no key in a URL, but the form element stays and that is what Chrome reads as a login in your next comment. With a plain `<div id="unlock">`, a `<button type="button">` and an explicit `keydown`/Enter listener there is no default action to prevent: without the inline script the button does nothing at all instead of navigating. `name="key"` is gone with it.

nbouliol: Done in 749ce16, via the form removal rather than `autocomplete`.

`new-password` would not have helped — it is the signup/change-password marker, so Chrome still offers to save, and additionally suggests a generated password on a field that is not one. What actually drives the heuristic is the form submit, and there is no form any more.

The header comment now states the reasoning instead of just the claim.

packages/agent-bff/src/docs/docs-page.ts
nbouliol: Done in 749ce16 — `untrustedSpec: true` on `Redoc.init`. The descriptions come from the agent's own schema, which is customer-authored, so it is the right default here regardless of cost.

packages/agent-bff/src/docs/docs-page.ts
nbouliol: Done in 749ce16. The init moved out of the fetch chain into a `render()` with a `typeof Redoc === 'undefined'` guard (its own message, naming the bundle path) and its own `try/catch` around `Redoc.init`. The chain's `.catch` now only ever reports a fetch failure, which is what it says.

packages/agent-bff/src/cli-core.ts
nbouliol: Done in 749ce16 — `enabled: config.openapiEnabled && agentMiddlewares.length > 0`, exactly as the error middleware two lines above.

Covered in `test/cli-core.test.ts`: with `FOREST_AUTH_SECRET` unset, `/docs` and `/agent/openapi.json` both answer 404.

packages/agent-bff/test/docs/docs-routes.test.ts
nbouliol: Done in 749ce16 — `expect(response.headers['cache-control']).toBe('public, max-age=3600')` in the bundle describe.

packages/agent-bff/src/docs/docs-routes.ts
```diff
@@ -0,0 +1,95 @@
...
+    if (ctx.path === DOCS_BUNDLE_PATH) {
+      // Read once and kept in memory: ~1 MB, served on every page load.
+      if (script === undefined) script = readFileSync(bundle, 'utf8');
+
+      ctx.status = 200;
+      ctx.type = 'application/javascript';
+      ctx.set('Cache-Control', 'public, max-age=3600');
+      ctx.body = script;
+
+      return;
```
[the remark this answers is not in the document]
nbouliol: Done in 749ce16 — `readFileSync` wrapped, falling through via `next()` on failure, with a `Warn` naming the file it could not read (a silent 404 on a path that resolved at boot is hard to diagnose otherwise).

Two tests in `test/docs/docs-routes.test.ts` cover it through the `resolveBundlePath` seam pointing at a non-existent file: fall-through instead of a 500, and the log line.

packages/agent-bff/src/docs/docs-page.ts
```diff
@@ -0,0 +1,146 @@
...
+
+          try {
+            Redoc.init(spec, REDOC_OPTIONS, document.getElementById('redoc'));
+          } catch (initError) {
+            unlock.style.display = '';
+            show('The Redoc viewer could not render the document: ' + initError);
+          }
+        }
+
+        function load(key) {
```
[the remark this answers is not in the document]
nbouliol: Real, and fixed in 389d25e.

An `attempt` counter, checked in both `then` and `catch`, so a completion that is not the latest is dropped. Two failure modes it closes: a stale 401 painting an error box over an already rendered document, and a stale document rendering over the one the reader actually asked for.

A counter rather than an `AbortController`: aborting fires the same `catch` that would then need the staleness check anyway, so the check is the whole fix — the abort would only save a request already in flight.

Worth noting what this took to test properly. The page script had no executable test at all, and asserting the guard by substring would prove nothing about ordering. It now runs in a `vm` against a stub DOM and a `fetch` whose responses are resolved by hand, which is what lets the out-of-order cases be driven. Verified to bite: with the two checks removed, the three race tests fail and the other three still pass.

packages/agent-bff/src/docs/docs-page.ts
```diff
@@ -0,0 +1,163 @@
...
+              return response.text().then(function (text) {
+                var body;
+
+                try {
+                  body = JSON.parse(text);
+                } catch (parseError) {
+                  body = { error: { type: 'unreadable_response', message: text.slice(0, 200) } };
+                }
+
+                return { ok: response.ok, status: response.status, body: body };
```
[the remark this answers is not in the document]
nbouliol: Right, and fixed in 25c20fe — the parse failure now sets `ok: false`.

Written as two returns rather than a mutable `ok`, so the invariant is on the surface: a body that will not parse is never a document, whatever the status line said. The sharp part of this one is that the code already built the exact message for it and could never show it — on a 200 the placeholder went to `Redoc.init` as a spec instead.

Covered by a test that drives a 200 carrying `<html>gateway</html>` and asserts the error text plus that `Redoc.init` was never called. Verified to bite: restoring `ok: response.ok` fails that test alone.

While confirming it I also found a flake in the harness I added one commit earlier: `flush()` awaited a single `process.nextTick`, and the nextTick queue runs before the microtask queue, so one tick does not settle a three-hop fetch chain. It uses `setImmediate` now. Not a guess — a bare probe shows a 3-deep chain unsettled after nextTick and settled after setImmediate. Three full suite runs clean since.

packages/agent-bff/src/docs/docs-samples.ts
```diff
@@ -0,0 +1,261 @@
...
+            var item = paths[path] || {};
+
+            Object.keys(item).forEach(function (method) {
+              var operation = item[method];
+
+              if (!operation || typeof operation !== 'object' || !operation.responses) return;
+
+              var body = exampleBody(spec, operation);
+              var headers = headersOf(spec, operation, body);
+              var url = origin + path;
```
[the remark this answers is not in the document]
nbouliol: Correct on the fact, fixed in a21b213 — with one deliberate difference from the suggestion.

The fact: on the GENERIC document every path is a template, and `origin + path` left all three languages (curl included, not just node and ruby) pointing at a literal `/agent/v1/{collection}/list`. That document is reachable whenever the deployment cannot unfold — no `AGENT_URL`, or no read-model configuration. Reproduced before fixing.

The difference: not a "usable sample value". The generic document is served precisely BECAUSE the deployment cannot enumerate its collections, so there is no real name to substitute, and an invented one would read as runnable and answer 404 — worse than an obvious placeholder. Each declared path parameter now becomes the same `<name>` placeholder the bodies already use (`<collection>`, `<action>`, alongside `<parentId>`), resolving a `$ref`d parameter too. One notation across a snippet reads as "replace this", where `{collection}` can be mistaken for syntax the API expects.

Worth noting for the record that the UNFOLDED document was never affected: its segments are the real names, already URL-encoded, and it declares no path parameter. A test pins that it stays untouched, and against a real 82-operation document all 82 still generate three samples with no missing auth or timezone header.

packages/agent-bff/src/docs/docs-samples.ts
```diff
@@ -0,0 +1,287 @@
...
+
+          headers.push({ name: TIMEZONE_HEADER, value: SAMPLE_TIMEZONE });
+
+          if (body !== undefined) headers.push({ name: 'Content-Type', value: 'application/json' });
+
+          return headers;
+        }
+
+        function curlSample(url, method, headers, body) {
+          var lines = ["curl -X " + method + " '" + url + "'"];
```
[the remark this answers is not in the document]
nbouliol: Real, and worse than hypothetical — fixed in 8b02367.

`encodeURIComponent` does not encode `'`, which I checked rather than assumed: `encodeURIComponent("John's orders")` returns `John's%20orders`. So a collection with an apostrophe in its name reaches the samples carrying it, and a field name or enum value can carry one into a body. All three languages broke, curl included — not only node and ruby.

Interpolation now goes through the quoting of the target language: POSIX close-reopen for shell, `JSON.stringify` for every JavaScript literal, backslash escaping inside Ruby single quotes, and `#{` neutralised in the Ruby body, since a JSON literal is double-quoted and Ruby interpolates there — a field named `a#{b}` would otherwise have been evaluated.

Verified with the real parsers, not by eye: 164 samples per language — the 82-operation document plus a copy whose collection and field names carry apostrophes — all pass `bash -n`, JavaScript parsing and `ruby -c`.

packages/agent-bff/src/docs/docs-samples.ts
```diff
@@ -0,0 +1,287 @@
...
+
+          headers.forEach(function (header) {
+            var value = header.secret ? header.prefix + '$' + KEY_VARIABLE : header.value;
+
+            lines.push("  -H '" + header.name + ": " + value + "'");
```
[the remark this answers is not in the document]
nbouliol: Correct, and my own claim that the command "runs as pasted" was wrong for exactly this header. Fixed in 8b02367, along the lines you suggest: the secret header is double-quoted so the shell expands it, the fixed text around the variable is escaped for that context, and every other header keeps single quotes where expansion would be wrong.

Proven by running a generated sample with a stub `curl` that prints the arguments it received, with `BFF_KEY` set:

```
arg: -H
arg: X-Forest-Bff-Key: fbff_realkey_value
```

The value expanded, and the URL arrived as a single argument with its apostrophe intact. A test now pins the double quotes on the secret header and the single quotes on the literal ones, so the two cannot be swapped back.

packages/agent-bff/src/docs/docs-samples.ts
```diff
@@ -0,0 +1,323 @@
...
+            var schemas = (spec.components || {}).schemas || {};
+
+            return schemaOf(spec, schemas[node.$ref.split('/').pop()]);
+          }
+
+          return node;
+        }
+
+        // A relation request is its foreign collection request plus a parent id, expressed as allOf.
+        function flatten(spec, node) {
```
Tonours: `flatten` recurses on `allOf` with no depth bound, so a self-referential `allOf` overflows the stack and drops the samples of every remaining operation: add the `depth > 6` guard `placeholder` already carries.

packages/agent-bff/src/docs/docs-routes.ts
```diff
@@ -0,0 +1,106 @@
...
+  resolveBundlePath?: () => string | undefined;
+}
+
+/**
+ * The bundle is copied next to this module at build time (`build:copy`), which is what a published
+ * install serves. Running from `src` — tests, `build:watch` — there is nothing to copy to, so the
+ * `redoc` devDependency is resolved instead: the same file, from the package that pins its version.
+ */
+function resolveBundle(): string | undefined {
+  const copied = path.join(__dirname, BUNDLE_FILE);
```
Tonours: The description says both bundle lookups are covered by tests, but running from `src` only ever exercises the `require.resolve` fallback: the `__dirname` copy branch, which is what a published install serves, never runs; add a test that points the lookup at a directory holding the file.

Tonours (commented):
`README.md` documents the OpenAPI document but never mentions the new public `GET /docs` page or its bundle route; add them to the `BFF_OPENAPI_ENABLED` row, including that both are unauthenticated and that the flag turns them off.

The row already warns in bold that the served document is not filtered per caller, so it is the line an operator reads to decide: right now it does not say the same flag also exposes an HTML page that needs no credential.

packages/agent-bff/src/docs/docs-samples.ts
nbouliol: Right, and it went one level deeper than the comment says — fixed in 6f54b08.

`schemaOf` had the same hole, and it is the load-bearing one: every cycle has to pass through a `$ref`, so a schema naming itself looped in `schemaOf` before `flatten` even recursed. Bounding `flatten` alone would have left that open, and bounding `schemaOf` alone happens to break the `allOf` cycle too. All three walks now share one `MAX_DEPTH`, which is also what makes the guard obvious rather than incidental.

Your reading of the blast radius is exactly right and worth keeping in the code: the throw leaves `decorateWithSamples` through the loop, so every operation after the offending one loses its samples too. That is now the comment above the constant.

Verified to bite: with the two bounds removed, a document carrying a self-referential `allOf` and a self-referential `$ref` overflows and the test fails; with them, both operations and the one after them still get three samples. A 20-deep `anyOf` chain and a cyclic `allOf` also come out as samples that pass `bash -n`, JavaScript parsing and `ruby -c`. Nothing legitimate is truncated — real shapes sit at depth 2 at most, and the tests asserting the exact relation and action bodies still pass.

packages/agent-bff/src/docs/docs-routes.ts
nbouliol: Correct, and the description was wrong to claim both were covered — fixed in 5bf0327.

The lookup directory is a parameter now, defaulting to `__dirname`, so both branches are reachable: one test points it at a temporary directory holding `redoc.standalone.js` and expects that exact path, the other points it at an empty one and expects `require.resolve("redoc/bundles/redoc.standalone.js")`. The branch a published install actually takes is the first, and it had never executed.

Worth noting what this does not cover: that `build:copy` puts the file there in the first place. That is a build-script assertion rather than a runtime one, and I have left it out rather than shell out to `yarn build` from a test — say the word if you want it pinned too.

Tonours (approved):

PR #1815, opened 2026-08-10
feat: upload files through the MCP server for action File fields

Scra3:
Action forms with **File fields** could not run over MCP. The agent expects a file value as a data uri, so the bytes had to travel through the model's context window — a 2-5 MB document costs hundreds of thousands of tokens and exceeds most MCP clients' payload limits. Two customers hit this: Roundtable in April (document-extraction flow blocked on the upload step) and Qonto, who opened #1814 to implement it themselves.

This PR builds on that contribution. Reviewing it surfaced a deeper gap, so the work splits in three.

fixes PRD-913

## 1. agent-client owns the encoding

`setFields` was a pass-through with no support for `File` / `['File']` fields, so every caller had to hand-craft the data uri. It now accepts a `File` object (`{ buffer, mimeType, name }`) and encodes it; strings pass through untouched, so already-encoded data uris and opaque references still work.

The codec itself was duplicated in the agent and in plugin-aws-s3 (and the `File` type redeclared), so it moves to `datasource-toolkit` and both reuse it — **-90 lines**. Running the agent's and the plugin's existing suites against the shared implementation is what proves encode/decode symmetry.

Three fixes come along for free:

- **The `name` is now percent-encoded.** #1814 built the data uri by hand without it, so a filename like `rapport final;v2.pdf` broke the agent's header parsing (it splits on `;`).
- **`parseDataUri` fails readably.** No comma raised a `TypeError`, a lone `%` a `URIError`, and both surfaced as a generic 500 — from a value a model populates freely. They are now a `ValidationError` carrying the message, which the agent renders as a 400. Only `name` and `charset` are read back from the media types, so a uri carrying `buffer=oops` can no longer replace the decoded bytes with a string.
- `makeDataUri` no longer emits `charset=undefined`.

**No agent change is needed, on any generation.** agent-client hits the standard `/forest/...` routes, so what we send is byte-for-byte what the browser already sends for a File field.

⚠️ **One observable change for agent-client consumers:** `getType()` now answers `'FileList'` where it used to leak the wire array `['File']`. The array form is kept in the payload echoed back to the agent, whose change-hook parser matches only that shape. This is what keeps `getType(): string` and unbreaks `agent-bff` and `workflow-executor`.

## 2. The MCP upload side-channel

[...]

PR #1832, opened 2026-08-17
fix(mcp-server): let a deployed standalone server know its public url

Scra3:
Follow-up to #1815, which shipped in `@forestadmin/mcp-server@1.22.0`.

`run()` hardcodes `baseUrl = http://localhost:<port>`, and that explicit value wins over the environment's `api_endpoint`. A process cannot derive its own public URL — DNS, proxy and TLS termination live outside it — so a **deployed** standalone server today:

- advertises `localhost` as its OAuth issuer and endpoints (pre-existing, not introduced by #1815),
- and, on the in-memory upload store #1815 added, mints `localhost` upload URLs no remote client can reach.

Nobody had seen it because every standalone test so far ran locally. A standalone deployment with a configured storage backend is unaffected — the upload URL comes from the backend.

`FOREST_MCP_SERVER_URL` is the public URL the server is reachable at. Set, it becomes the base URL, fixing the issuer and the upload URLs in one move; unset, behavior is unchanged. A value that does not parse as a URL fails at startup, like every other option.

Verified by test through the real HTTP server `run()` starts: the advertised issuer is the configured URL. The suite's global `fetch` is the Forest mock, so the test reaches it with the kept `originalFetch`.

> [!NOTE]
> ### Let a deployed MCP server advertise its public URL via `FOREST_MCP_SERVER_URL`
> - Adds a new `FOREST_MCP_SERVER_URL` env var to [server.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1832/files#diff-c54d7a276ef5be4d65ba37d55b279e77958ba4496af9d9ca7ec80c23c847d155) that sets the public http(s) origin used for OAuth metadata and file upload URLs; defaults to `http://localhost:<port>` when unset.
> - Validates that `FOREST_MCP_SERVER_URL` is a bare origin (no path, query, fragment, or credentials), failing startup with a clear error on invalid input.
> - Tightens `MCP_SERVER_PORT` parsing to require an integer in [0, 65535]; rejects `MCP_SERVER_PORT=0` unless `FOREST_MCP_SERVER_URL` is set, since the ephemeral port can't be advertised without a known public URL.
> - Logs the advertised URL at startup and warns when advertising a localhost URL because `FOREST_MCP_SERVER_URL` is unset.
> - Behavioral Change: servers deployed behind a public URL must now set `FOREST_MCP_SERVER_URL`; omitting it causes advertised endpoints to resolve to localhost.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 059365a.</sup>
> 
>

### DIFF
--- a/packages/mcp-server/src/server.ts
+++ b/packages/mcp-server/src/server.ts
@@ -788,8 +788,49 @@ export default class ForestMCPServer {
    * Run the MCP server as a standalone HTTP server.
    */
   async run(): Promise<void> {
-    const port = Number(process.env.MCP_SERVER_PORT) || 3931;
-    const baseUrl = new URL(`http://localhost:${port}`);
+    // Parsed before defaulting: `Number(x) || 3931` turns port 0, which means "any free port",
+    // into 3931. A configured FOREST_MCP_SERVER_URL also replaces the default url, so nothing else
+    // parses the port either.
+    const rawPort = process.env.MCP_SERVER_PORT;
+    const port = rawPort ? Number(rawPort) : 3931;
+    const configuredUrl = process.env.FOREST_MCP_SERVER_URL;
+
+    if (!Number.isInteger(port) || port < 0 || port > 65535) {
+      throw new Error(
+        `Invalid MCP_SERVER_PORT "${rawPort}": expected an integer between 0 and 65535.`,
+      );
+    }
+
+    // The url is built here, before listen() picks the port, so 0 cannot appear in it.
+    if (port === 0 && !configuredUrl) {
+      throw new Error(
+        'MCP_SERVER_PORT=0 binds a port chosen by the OS, which cannot be in the url advertised ' +
+          'to clients. Set FOREST_MCP_SERVER_URL to the public url they should use.',
+      );
+    }
+
+    const publicUrl = configuredUrl || `http://localhost:${port}`;
+    const baseUrl = URL.canParse(publicUrl) ? new URL(publicUrl) : undefined;
+
+    // Origin only: the OAuth endpoints are concatenated onto this href, the uploads base resolves
+    // against it.
+    if (
+      !baseUrl ||
+      !['http:', 'https:'].includes(baseUrl.protocol) ||
+      baseUrl.href !== `${baseUrl.origin}/`
+    ) {
+      // Never the raw value: it may carry credentials. `origin` is "null" for an opaque scheme,
+      // which is what a forgotten scheme parses as.
+      const shown =
+        baseUrl && baseUrl.origin !== 'null'
+          ? baseUrl.origin
+          : publicUrl.slice(publicUrl.lastIndexOf('@') + 1);
+
+      throw new Error(
+        `Invalid FOREST_MCP_SERVER_URL "${shown}": expected an http(s) origin with no path, ` +
+          'query, fragment or credentials, e.g. https://mcp.example.com',
+      );
+    }
 
     const app = await this.buildExpressApp(baseUrl);
 
@@ -797,7 +838,18 @@ export default class ForestMCPServer {
     this.httpServer = http.createServer(app);
 
     this.httpServer.listen(port, () => {
-      this.logger('Info', `Forest Admin MCP Server running on http://localhost:${port}`);
+      this.logger(
+        'Info',
+        `Forest Admin MCP Server running on port ${port}, advertising ${baseUrl.href}`,
+      );
+
+      if (!configuredUrl) {
+        this.logger(
+          'Warn',
+          `Advertising http://localhost:${port} to clients. Deployed behind a public url? Set ` +
+            'FOREST_MCP_SERVER_URL, or remote OAuth and file uploads will point at localhost.',
+        );
+      }
     });
   }
 }
--- a/packages/mcp-server/README.md
+++ b/packages/mcp-server/README.md
@@ -63,6 +63,7 @@ yarn start:dev       # Development (loads .env file automatically)
 | `FOREST_ENV_SECRET` | **Yes** | - | Your Forest Admin environment secret |
 | `FOREST_AUTH_SECRET` | **Yes** | - | Your Forest Admin authentication secret (must match your agent) |
 | `MCP_SERVER_PORT` | No | `3931` | Port for the HTTP server |
+| `FOREST_MCP_SERVER_URL` | No | `http://localhost:<port>` | Public URL this server is reachable at — an http(s) **origin only**, no path. **Required for any deployed server**: without it the OAuth metadata advertise `localhost`, and the in-memory upload store mints `localhost` upload URLs no remote client can reach |
 | `FOREST_MCP_ENABLED_TOOLS` | No | - | Comma-separated list of tools to enable (allowlist) |
 | `FOREST_MCP_ALLOWED_OAUTH_CLIENTS` | No | - | Comma-separated domains of the OAuth client applications allowed to connect (`allowedOAuthClients`). Unset, any registered client is accepted |
 | `FOREST_AGENT_URL` | No | your environment's back-end URL | URL the MCP server uses to reach the back-end's data layer. Set it when the server runs next to a self-hosted back-end at an internal address (e.g. `http://localhost:3310`), instead of the public URL registered in Forest |
@@ -90,6 +91,13 @@ Or set the variables inline:
 
 ```bash
 FOREST_ENV_SECRET="your-env-secret" FOREST_AUTH_SECRET="your-auth-secret" npx forest-mcp-server
+```
+
+Deployed behind a public URL, add it — clients are sent wherever this says:
+
+```bash
+FOREST_MCP_SERVER_URL="https://mcp.example.com" \
+  FOREST_ENV_SECRET="your-env-secret" FOREST_AUTH_SECRET="your-auth-secret" npx forest-mcp-server
 ```
 
 ## Restrict Tools
--- a/packages/mcp-server/test/server.test.ts
+++ b/packages/mcp-server/test/server.test.ts
@@ -118,7 +118,155 @@ describe('ForestMCPServer Instance', () => {
 
   describe('run method', () => {
     afterEach(async () => {
+      delete process.env.FOREST_MCP_SERVER_URL;
+      delete process.env.MCP_SERVER_PORT;
       await shutDownHttpServer(server?.httpServer as http.Server);
+    });
+
+    it('advertises FOREST_MCP_SERVER_URL instead of localhost when set', async () => {
+      const testPort = await getAvailablePort();
+      process.env.MCP_SERVER_PORT = testPort.toString();
+      process.env.FOREST_MCP_SERVER_URL = 'https://mcp.example.com';
+
+      server = new ForestMCPServer({
+        authSecret: 'AUTH_SECRET',
+        envSecret: 'ENV_SECRET',
+        forestServerClient: createMockForestServerClient(),
+      });
+      server.run();
+      await new Promise(resolve => {
+        setTimeout(resolve, 500);
+      });
+
+      // global.fetch is the Forest mock here.
+      const response = await originalFetch(
+        `http://localhost:${testPort}/.well-known/oauth-authorization-server`,
+      );
+      const metadata = (await response.json()) as { issuer: string };
+
+      expect(metadata.issuer).toBe('https://mcp.example.com/');
+
+      // The other consumer of the same base url — the half that motivated the option.
+      const { ephemeralStorage } = server as unknown as {
+        ephemeralStorage: { createUploadUrl(p: { key: string }): Promise<{ url: string }> };
+      };
+      const { url } = await ephemeralStorage.createUploadUrl({ key: 'k' });
+
+      expect(url.startsWith('https://mcp.example.com/mcp/uploads/')).toBe(true);
+    });
+
+    // The OAuth endpoints are concatenated onto the href and the uploads base resolves against the
+    // origin, so anything beyond an http(s) origin advertises broken urls instead of failing here.
+    it.each([
+      ['a trailing slash', 'https://mcp.example.com/', 'https://mcp.example.com/'],
+      ['no trailing slash', 'https://mcp.example.com', 'https://mcp.example.com/'],
+      ['an explicit default port', 'https://mcp.example.com:443', 'https://mcp.example.com/'],
+      ['a custom port', 'https://mcp.example.com:8443', 'https://mcp.example.com:8443/'],
+    ])('accepts a FOREST_MCP_SERVER_URL with %s', async (_, value, expected) => {
+      const testPort = await getAvailablePort();
+      process.env.MCP_SERVER_PORT = testPort.toString();
+      process.env.FOREST_MCP_SERVER_URL = value;
+
+      server = new ForestMCPServer({
+        authSecret: 'AUTH_SECRET',
+        envSecret: 'ENV_SECRET',
+        forestServerClient: createMockForestServerClient(),
+      });
+      server.run();
+      await new Promise(resolve => {
+        setTimeout(resolve, 500);
+      });
+
+      const response = await originalFetch(
+        `http://localhost:${testPort}/.well-known/oauth-authorization-server`,
+      );
+
+      expect(((await response.json()) as { issuer: string }).issuer).toBe(expected);
+    });
+
+    it('refuses MCP_SERVER_PORT=0 unless it is told what to advertise', async () => {
+      process.env.MCP_SERVER_PORT = '0';
+
+      server = new ForestMCPServer({
+        authSecret: 'AUTH_SECRET',
+        envSecret: 'ENV_SECRET',
+        forestServerClient: createMockForestServerClient(),
+      });
+
+      await expect(server.run()).rejects.toThrow(/MCP_SERVER_PORT=0 binds a port chosen by the OS/);
+    });
+
+    it('binds an ephemeral port on MCP_SERVER_PORT=0 rather than falling back to 3931', async () => {
+      process.env.MCP_SERVER_PORT = '0';
+      process.env.FOREST_MCP_SERVER_URL = 'https://mcp.example.com';
+
+      server = new ForestMCPServer({
+        authSecret: 'AUTH_SECRET',
+        envSecret: 'ENV_SECRET',
+        forestServerClient: createMockForestServerClient(),
+      });
+      server.run();
+      await new Promise(resolve => {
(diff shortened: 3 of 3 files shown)


packages/mcp-server/src/server.ts
```diff
@@ -789,7 +789,20 @@ export default class ForestMCPServer {
...
     const port = Number(process.env.MCP_SERVER_PORT) || 3931;
-    const baseUrl = new URL(`http://localhost:${port}`);
+    // A process cannot derive its own public url — dns, proxy and tls termination live outside it.
+    // Without this, a deployed standalone server advertises localhost as its OAuth issuer and, on
+    // the in-memory store, mints localhost upload urls no remote client can reach.
+    const publicUrl = process.env.FOREST_MCP_SERVER_URL;
+    let baseUrl: URL;
+
+    try {
+      baseUrl = new URL(publicUrl || `http://localhost:${port}`);
```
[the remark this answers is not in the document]
Scra3: Real, and fixed in 029f626fa with exactly the shape you suggest — reject rather than normalize.

Startup now requires an http(s) origin with no path, query or fragment; anything else fails with a message naming the expected shape. Rejecting a path (rather than normalizing to a directory URL) is deliberate: a trailing slash rescues the concatenated OAuth endpoints but the uploads base still resolves `/mcp/uploads` against the origin and drops the path, so no normalization makes both consumers correct — an origin is the only honest contract, and the README row now says so.

Two shapes beyond your list are covered by the same guard, from the same review round: a host-port pair without a scheme (`mcp.example.com:8080` parses as an opaque scheme and used to crash later with a bare `Invalid URL`), and non-http(s) schemes. Each rejected shape is pinned by an `it.each` case, and the happy-path test now asserts the minted upload URL host as well as the issuer.

packages/mcp-server/src/server.ts
```diff
@@ -789,7 +789,25 @@ export default class ForestMCPServer {
...
     const port = Number(process.env.MCP_SERVER_PORT) || 3931;
-    const baseUrl = new URL(`http://localhost:${port}`);
+    // A process cannot derive its own public url; unset, a deployed server advertises localhost.
+    const publicUrl = process.env.FOREST_MCP_SERVER_URL || `http://localhost:${port}`;
+    const baseUrl = URL.canParse(publicUrl) ? new URL(publicUrl) : undefined;
+
+    // Origin only: the OAuth endpoints are concatenated onto this href and the uploads base
+    // resolves /mcp/uploads against its origin, so a path, query or fragment breaks one or both —
+    // silently, and only once a client follows the advertised urls.
+    if (
```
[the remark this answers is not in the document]
Scra3: Real, and the nastiest of the invalid shapes: everything else the guard rejects yields a broken deployment, this one yields a working deployment that serves its credentials to every client through the OAuth metadata. Fixed in e6b0a5b9d — `username`/`password` join the origin-only guard, the error message names credentials, and `https://user:secret@example.com` is pinned in the `it.each`.

packages/mcp-server/src/server.ts
```diff
@@ -789,15 +789,46 @@ export default class ForestMCPServer {
    */
   async run(): Promise<void> {
     const port = Number(process.env.MCP_SERVER_PORT) || 3931;
-    const baseUrl = new URL(`http://localhost:${port}`);
+    // A process cannot derive its own public url; unset, a deployed server advertises localhost.
+    const configuredUrl = process.env.FOREST_MCP_SERVER_URL;
+    const publicUrl = configuredUrl || `http://localhost:${port}`;
```
hercemer42, line 812: Claude Opus 5 (claude-opus-5): **Should fix**

The operator this PR exists for — deployed, `FOREST_MCP_SERVER_URL` set — is the one who can no longer reach the `MCP_SERVER_PORT` message written for them.

Port validation used to be a side effect of building the default URL: on `main`, `new URL('http://localhost:' + port)` rejected a bad port unconditionally (`new URL('http://localhost:99999')` → `Invalid URL`, confirmed). Here `publicUrl` is the configured URL whenever one is set, so the port never reaches a parser.

`FOREST_MCP_SERVER_URL=https://mcp.example.com` + `MCP_SERVER_PORT=99999` (a typo, or a templated value that didn't render): validation passes, `buildExpressApp` then does a full `/liana/environment` + `/liana/forest-schema` round trip and `oauthProvider.initialize()`, and only then does `listen(port)` at `:820` throw `RangeError: options.port should be >= 0 and < 65536` (confirmed). Generic Node error, after network I/O, and the curated `check MCP_SERVER_PORT="…"` branch at `:811` is unreachable in this configuration — the comment at `:805` ("an out-of-range port lands here too") holds only when the variable is unset.

Validating `port` on its own before `:794` fixes it and makes both messages reachable in both configurations, rather than each depending on which variable happened to build the string.

packages/mcp-server/src/server.ts
```diff
@@ -789,15 +789,46 @@ export default class ForestMCPServer {
...
+    if (
+      !baseUrl ||
+      !['http:', 'https:'].includes(baseUrl.protocol) ||
+      baseUrl.href !== `${baseUrl.origin}/`
+    ) {
+      // The default is built from MCP_SERVER_PORT, so an out-of-range port lands here too. Naming
+      // the url the operator actually set keeps it from chasing a variable it never touched.
+      throw new Error(
+        configuredUrl
+          ? `Invalid FOREST_MCP_SERVER_URL "${publicUrl}": expected an http(s) origin with no ` +
```
hercemer42: Claude Opus 5 (claude-opus-5): **Should fix**

The credentials this branch exists to keep away from clients get written to the logs by the rejection itself.

`publicUrl` is interpolated verbatim, so `FOREST_MCP_SERVER_URL=https://svc:p4ssw0rd@mcp.example.com` — the exact input the userinfo check catches, and the case covered at `test/server.test.ts:208` — crashes startup with `Invalid FOREST_MCP_SERVER_URL "https://svc:p4ssw0rd@mcp.example.com": …` on stderr, into container logs and whatever aggregates them. The comment at `:796-798` names userinfo as the thing that must not be handed out ("serves the credentials back to every client"); this hands it to a different audience instead, and log read access is usually broader than access to the environment itself.

`baseUrl ? baseUrl.origin : publicUrl` keeps the message diagnostic — the operator still sees `"https://mcp.example.com"`, which is what they need in order to see what was read — while dropping the secret. The raw string is only needed in the branch where parsing failed, and userinfo always parses.

packages/mcp-server/src/server.ts
```diff
@@ -789,15 +789,46 @@ export default class ForestMCPServer {
...
+    }
 
     const app = await this.buildExpressApp(baseUrl);
 
     // Create HTTP server from Express app
     this.httpServer = http.createServer(app);
 
     this.httpServer.listen(port, () => {
-      this.logger('Info', `Forest Admin MCP Server running on http://localhost:${port}`);
+      this.logger('Info', `Forest Admin MCP Server running on ${baseUrl.href}`);
```
hercemer42: Claude Opus 5 (claude-opus-5): **Should fix**

With the option set there is no longer any log line stating the port the process actually bound, and the line that looks like one is the advertised URL instead.

`FOREST_MCP_SERVER_URL=https://mcp.example.com` + `MCP_SERVER_PORT=8080` logs `running on https://mcp.example.com/` while the socket is on `:8080`. Before this change the line was `running on http://localhost:${port}`, which was the true bind — so the information is lost precisely in the deployed configuration the PR adds, and that is where someone debugging a reverse proxy or a container healthcheck goes looking for it first.

The two facts are different and both matter now that they can differ. `running on port ${port}, advertising ${baseUrl.href}` carries both. Nothing consumes this string (single occurrence repo-wide), so it is free to change.

hercemer42 (changes requested):

packages/mcp-server/src/server.ts
Scra3: Confirmed and fixed in 20cd30ada. Reproduced exactly as described before touching anything: with `FOREST_MCP_SERVER_URL` set and `MCP_SERVER_PORT=99999`, `run()` completes validation, does the environment and schema round trip, and only then throws `RangeError: options.port should be >= 0 and < 65536` from `listen()`.

Your diagnosis of *why* is the part I had missed: port validation used to be a side effect of building the default url, and the configured url removes that side effect. So my own fix for the misattribution created this hole — the curated branch became unreachable in precisely the configuration this PR exists for.

`port` is now validated on its own, before the url is built, so both messages are reachable in both configurations. Fractional and negative values are covered too (`Number('3931.5') || 3931` is truthy, so they used to survive as well). Pinned by `refuses an %s MCP_SERVER_PORT before doing any work`.

One thing your finding surfaced indirectly: `MCP_SERVER_PORT` was never cleaned in this `describe`'s `afterEach`. Harmless until now, but with the port validated first, a bad value leaking out of one test failed the *next* test on its port rather than on its own subject. Cleaned up in the same commit.

packages/mcp-server/src/server.ts
Scra3: Right, and the irony is the point — the check exists to keep credentials away from clients, and the rejection handed them to an audience with usually broader read access. Fixed in 20cd30ada with your exact suggestion: `baseUrl ? baseUrl.origin : publicUrl`.

The operator still gets what they need (`Invalid FOREST_MCP_SERVER_URL "https://mcp.example.com"` — enough to see what was read), and the raw string survives only in the branch where parsing failed, where there is no origin to fall back to and, as you note, userinfo always parses.

Pinned by `does not echo credentials back when rejecting them`, which asserts both halves: the origin is present and `p4ssw0rd` is not.

packages/mcp-server/src/server.ts
Scra3: Fixed in 20cd30ada: `running on port ${port}, advertising ${baseUrl.href}`.

You are right that the loss lands exactly where it hurts. Before the change the line was the true bind by construction; now the two facts differ the moment the option is set, and the one this PR adds is the deployed configuration — where a reverse proxy or a container healthcheck is what someone is debugging, and the bind port is the first thing they look for.

packages/mcp-server/src/server.ts
hercemer42: Claude Opus 5 (claude-opus-5): Fixed. Validating `port` on its own at `:796`, ahead of the URL logic, makes both messages reachable in both configurations — and the message now names `MCP_SERVER_PORT` regardless of whether `FOREST_MCP_SERVER_URL` built the base URL. `listen()` can no longer be the thing that reports a bad port, so the Forest API round trip no longer happens first.

The new `it.each` covers exactly the reachable cases — out of range, fractional, negative. `Number(process.env.MCP_SERVER_PORT) || 3931` already swallows NaN into the default, so there was nothing to test there. Resolving.

One aside, not worth a change: that same `|| 3931` means `MCP_SERVER_PORT=abc` silently becomes 3931 rather than failing, so the new guard reads stricter than it is. Pre-existing, and arguably the friendlier behaviour.

packages/mcp-server/src/server.ts
hercemer42: Claude Opus 5 (claude-opus-5): The credential leak is fixed and tested — `baseUrl.origin` drops userinfo, and `does not echo credentials back when rejecting them` pins both halves (origin shown, secret absent). Thanks.

**Leaving this open for a side effect it introduced**, on an input class that is a plausible operator typo rather than an exotic one. `origin` is the literal string `"null"` for any URL with an opaque scheme, so:

```
FOREST_MCP_SERVER_URL              message echoes
-----------------------------------------------------------
mcp.example.com:8080          →    "null"      ← host:port, no scheme
user:pw@example.com           →    "null"
https://svc:pw@mcp.example.com →   "https://mcp.example.com"   ✓
mcp.example.com               →    "mcp.example.com"           ✓ (canParse false)
https://example.com/p         →    "https://example.com"       ✓
```

`mcp.example.com:8080` is one of the rows in the refusal table at `test/server.test.ts` — it parses with scheme `mcp.example.com:`, so it reaches the origin branch and the operator is told their value was `"null"`. Before the redaction they got their input back. The test passes either way because it only matches `/Invalid FOREST_MCP_SERVER_URL/`.

So the message is now safe but uninformative for exactly the person most likely to hit it: someone who forgot the scheme. Guarding on `origin !== 'null'` alone would restore the echo and reopen the leak for `user:pw@example.com`, so it wants the userinfo stripped from the raw value rather than delegated to `origin` — or, cheaper, keep the origin when it is meaningful and fall back to naming the scheme that was parsed. Your call which; the property to keep is *the operator can see what they set, minus any secret*.

Worth an assertion on the message content in that refusal table too, since the current regex would not have caught this.

packages/mcp-server/src/server.ts
hercemer42: Claude Opus 5 (claude-opus-5): Fixed. `running on port ${port}, advertising ${baseUrl.href}` carries both facts, which is what was needed as soon as they can differ — whoever debugs a reverse proxy or a healthcheck gets the bound port back. Resolving.

packages/mcp-server/src/server.ts
Scra3: Caught, and it was mine — the redaction traded informativeness for safety on exactly the class most likely to hit it. Your table is reproduced as written; `origin` is the literal `"null"` for any opaque scheme, and a forgotten scheme is what parses as one.

Fixed in f8225f86d. Not by guarding on `origin !== 'null'` alone — you are right that it reopens the leak, and worse than you describe: for `user:pw@example.com` the parser does not split the credentials out at all (scheme `user:`, path `pw@example.com`), so `baseUrl.password` is empty and there is nothing to strip by that route.

So the fallback drops everything up to the last `@`, which is never userinfo:

```
mcp.example.com:8080            →  mcp.example.com:8080          (origin is "null", no @, full value)
user:secret@example.com         →  example.com                   (safe and still useful)
https://svc:secret@example.com  →  https://example.com           (origin, scheme kept)
https://example.com/mcp-server  →  https://example.com
```

Every class is now both safe and informative, which is the property you asked to keep.

And you are right that the regex was hiding it: the refusal table now asserts the exact message content per row, `Invalid FOREST_MCP_SERVER_URL "${shown}"`, with a `credentials and no scheme` row added — the one case where both concerns meet. 889 tests green.

On the aside: agreed, `Number(...) || 3931` swallowing `MCP_SERVER_PORT=abc` into the default is pre-existing and friendlier. Left alone deliberately — changing it would fail startup for a value that has always been tolerated, which is not this PR's business.

hercemer42: Claude Opus 5 (claude-opus-5): Fixed, and the test table now pins it. `baseUrl.origin !== 'null' ? origin : publicUrl.slice(lastIndexOf('@') + 1)` keeps both properties — I ran the input classes through it:

```
mcp.example.com:8080            → "mcp.example.com:8080"       ← the regression, restored
user:pw@example.com             → "example.com"
user:p@ss@example.com           → "example.com"                ← @ inside the secret
https://svc:p4ssw0rd@host       → "https://host"
https://u:p@h/path?q#f          → "https://h"
```

No leak in any of them, and `lastIndexOf` rather than `indexOf` is what makes the `@`-inside-the-credential case safe — worth the comment it has.

The added `['credentials and no scheme', 'user:secret@example.com', 'example.com']` row is the one that matters most: it covers exactly the case that made `origin !== 'null'` insufficient on its own, so the next person to simplify this will be caught. Asserting the shown value across the whole refusal table rather than just the prefix closes the hole that let the original issue through. Resolving.

packages/mcp-server/src/server.ts
```diff
@@ -789,15 +789,57 @@ export default class ForestMCPServer {
    */
   async run(): Promise<void> {
     const port = Number(process.env.MCP_SERVER_PORT) || 3931;
```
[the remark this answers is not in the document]
Scra3: Fixed in a2e357d48. The port-0 half is a real bug and I had walked straight past it: `Number('0') || 3931` is `3931`, so `MCP_SERVER_PORT=0` — "bind any free port" — silently became 3931. My own guard accepts `port === 0` (`port < 0` rejects only negatives), so I wrote a check that contemplates a value the line above it made unreachable.

Parsing before defaulting fixes both halves at once:

```ts
const rawPort = process.env.MCP_SERVER_PORT;
const port = rawPort ? Number(rawPort) : 3931;
```

`0` now reaches `listen()` as 0, `abc` reaches the guard as `NaN` and fails at startup, unset and empty still default. Pinned by `binds an ephemeral port on MCP_SERVER_PORT=0 rather than falling back to 3931` (asserts the bound port is neither 0 nor 3931) and a `non-numeric` row in the refusal table.

On `abc`: this makes a previously tolerated value fail at boot. I had declined that on its own when it was raised as an aside, on the grounds that silence was friendlier and pre-existing. The port-0 case changes the calculus — the same expression causes both, and this package's stated convention elsewhere (`FOREST_MCP_FILE_UPLOADS` rejects anything that is not `true`/`false`) is to fail rather than quietly do something else. Failing on a port that was never going to be honoured is consistent with that.

packages/mcp-server/src/server.ts
```diff
@@ -788,16 +788,61 @@ export default class ForestMCPServer {
...
+    const port = rawPort ? Number(rawPort) : 3931;
+
+    if (!Number.isInteger(port) || port < 0 || port > 65535) {
+      throw new Error(
+        `Invalid MCP_SERVER_PORT "${process.env.MCP_SERVER_PORT}": expected an integer between 0 ` +
+          'and 65535.',
+      );
+    }
+
+    const configuredUrl = process.env.FOREST_MCP_SERVER_URL;
```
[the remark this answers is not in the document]
Scra3: Right, and a direct consequence of the port-0 fix an hour earlier: honouring the port made the advertisement wrong, since the url is built before `listen()` picks one. Fixed in 059365a8a with the first of your two options.

`MCP_SERVER_PORT=0` without `FOREST_MCP_SERVER_URL` now fails at startup:

> MCP_SERVER_PORT=0 binds a port chosen by the OS, which cannot be in the url advertised to clients. Set FOREST_MCP_SERVER_URL to the public url they should use.

Deferring construction until after `listen()` was the other option and I did not take it: `buildExpressApp` needs the base url to build the OAuth metadata, and it has to exist before the server can serve, so deferring means creating a bare `http.Server`, listening, reading the port, then attaching the app — real contortion for a configuration that is contradictory anyway. Asking for an unpredictable port *and* expecting us to advertise a url we cannot know are incompatible wishes; port 0 is only sensible when something else already knows the public url, which is exactly when the variable is set.

Two tests: the refusal, and the coherent combination (port 0 + configured url) asserting the bound port is neither 0 nor 3931 while the advertised issuer is the configured host.

hercemer42 (approved):

PR #1835, opened 2026-08-18
fix(datasource-customizer): reject malformed binary filter values instead of crashing

Tonours:
## What

A `Binary` column filtered or written with a malformed value now raises `ValidationError` instead of crashing or converting the wrong bytes.

| Input | Before | After |
| --- | --- | --- |
| `'Anthony'` on a datauri column | `TypeError`, untyped → 500 | `ValidationError` |
| `'Anthony'` on a hex column (every Binary PK/FK) | **empty buffer, wrong results, no error** | `ValidationError` |
| `'data:text/plain,hello'` | decoded to garbage (`85e965`) | `ValidationError` |
| `42` | `TypeError`, untyped → 500 | `ValidationError` |

## Why

`convertScalar` assumed its input was always a well-formed datauri or hex string. It isn't: the value comes from a filter or a record patch, so anything can arrive.

`Buffer.from('Anthony'.split(',')[1], 'base64')` throws a raw `TypeError` — no HTTP status, so it surfaces as a generic 500 and the caller never learns what to send instead.

The hex branch was worse than the crash: `Buffer.from` never throws on bad hex, it stops at the first invalid character. `'Anthony'` yields an **empty buffer**, so a filter on a Binary primary key silently queried for an empty blob and returned wrong results with no error at all. `shouldUseHex` defaults to true for every primary and foreign key, so that was the common path.

## How

- a `Buffer` passes through untouched: it needs no conversion, and `Buffer.from(buf, 'hex')` already returned it verbatim
- any other non-string value rejected up front instead of being cast
- datauri branch delegates to the toolkit's `parseDataUri`, dropping a hand-rolled split that missed its hardening
- hex branch gets `HEX_BYTES` = `/^([0-9a-f]{2})+$/i` — odd lengths included, since `Buffer.from('303','hex')` silently drops a digit
- `replaceValidation` now advertises that same `HEX_BYTES`. It used to advertise `/^[0-9a-f]+$/`, which disagreed with the parser in both directions: `'303'` passed frontend validation then 400'd, and `'0F'` was blocked by the frontend though the backend accepts it
- rejected values are truncated to 32 characters in the message, like `parseDataUri` does — a `ValidationError` message is returned verbatim to the client

Tests assert the message (not just the class) and that the underlying collection was never called.

## Heads-up: writes are stricter too

`create`/`update` go through the same conversion. Three datauri forms that previously reached the database now raise: a raw comma in a `name=` media type, `data:base64,…` with no semicolon, and uppercase `BASE64` (`parseDataUri` is case-sensitive, RFC 2397 is not).

A raw `Buffer` is **not** affected: on a hex column it used to round-trip through `Buffer.from(buf, 'hex')`, and it still does. That is the `Buffer.isBuffer` passthrough above.

These rules come from `parseDataUri`, which already applies them to the S3 plugin's write path. If the uppercase form matters, fix it there for every caller rather than special-casing this decorator.

## Test

```bash
yarn workspace @forestadmin/datasource-customizer test -- test/decorators/binary/collection.test.ts
```

Also verified against a live agent: the same filter went from an opaque 500 to a 400 naming the expected format.

## Known limitation

`convertValueHelper` guards with `if (value)`, so `''`, `0` and `false` still bypass the conversion entirely and reach the connector unconverted. That predates this PR and is unchanged here; `Equal ''` on a Binary column can still produce the 500 class this PR narrows. Left out to keep the diff to the reported crash.

### General

### Security

### DIFF
--- a/packages/datasource-customizer/src/decorators/binary/collection.ts
+++ b/packages/datasource-customizer/src/decorators/binary/collection.ts
@@ -20,7 +20,7 @@ import {
   CollectionDecorator,
   SchemaUtils,
   TypeGetter,
-  ValidationError,
+  parseDataUri,
 } from '@forestadmin/datasource-toolkit';
 import { filetypemime } from 'magic-bytes.js';
 
@@ -252,15 +252,7 @@ export default class BinaryCollectionDecorator extends CollectionDecorator {
 
       if (useHex) return Buffer.from(string, 'hex');
 
-      const payload = string.split(',')[1];
-
-      if (payload === undefined) {
-        throw new ValidationError(
-          `Expected a data uri of the form "data:<mime>;base64,<payload>", received "${string}"`,
-        );
-      }
-
-      return Buffer.from(payload, 'base64');
+      return parseDataUri(string).buffer;
     }
 
     const buffer = value as Buffer;
--- a/packages/datasource-customizer/src/decorators/binary/collection.ts
+++ b/packages/datasource-customizer/src/decorators/binary/collection.ts
@@ -259,6 +259,8 @@ export default class BinaryCollectionDecorator extends CollectionDecorator {
     value: unknown,
   ): Promise<unknown> {
     if (toBackend) {
+      if (Buffer.isBuffer(value)) return value;
+
       if (typeof value !== 'string') {
         throw new ValidationError(
           `Expected a string for a binary field, got ${typeof value}: ${JSON.stringify(value)}`,
--- a/packages/datasource-customizer/src/decorators/binary/collection.ts
+++ b/packages/datasource-customizer/src/decorators/binary/collection.ts
@@ -25,6 +25,8 @@ import {
 } from '@forestadmin/datasource-toolkit';
 import { filetypemime } from 'magic-bytes.js';
 
+const HEX_BYTES = /^([0-9a-f]{2})+$/i;
+
 /**
  * As the transport layer between the forest admin agent and the frontend is JSON-API, binary data
  * is not supported.
@@ -244,7 +246,7 @@ export default class BinaryCollectionDecorator extends CollectionDecorator {
   }
 
   private parseHex(value: string): Buffer {
-    if (!/^([0-9a-f]{2})*$/i.test(value)) {
+    if (!HEX_BYTES.test(value)) {
       throw new ValidationError(
         `Expected an even-length hex string for a binary field, got "${value}"`,
       );
@@ -306,7 +308,7 @@ export default class BinaryCollectionDecorator extends CollectionDecorator {
       const maxLength = schema.validation?.find(v => v.operator === 'ShorterThan')?.value as number;
 
       if (this.shouldUseHex(name)) {
-        validation.push({ operator: 'Match', value: /^[0-9a-f]+$/ });
+        validation.push({ operator: 'Match', value: HEX_BYTES });
         if (minLength) validation.push({ operator: 'LongerThan', value: minLength * 2 + 1 });
         if (maxLength) validation.push({ operator: 'ShorterThan', value: maxLength * 2 - 1 });
       } else {
--- a/packages/datasource-customizer/src/decorators/binary/collection.ts
+++ b/packages/datasource-customizer/src/decorators/binary/collection.ts
@@ -248,7 +248,7 @@ export default class BinaryCollectionDecorator extends CollectionDecorator {
   private parseHex(value: string): Buffer {
     if (!HEX_BYTES.test(value)) {
       throw new ValidationError(
-        `Expected an even-length hex string for a binary field, got "${value}"`,
+        `Expected a hex string of full bytes for a binary field, got "${value.slice(0, 32)}"`,
       );
     }
 
@@ -265,7 +265,9 @@ export default class BinaryCollectionDecorator extends CollectionDecorator {
 
       if (typeof value !== 'string') {
         throw new ValidationError(
-          `Expected a string for a binary field, got ${typeof value}: ${JSON.stringify(value)}`,
+          `Expected a string for a binary field, got ${typeof value}: ${JSON.stringify(
+            value,
+          )?.slice(0, 32)}`,
         );
       }
 
--- a/packages/datasource-customizer/src/decorators/binary/collection.ts
+++ b/packages/datasource-customizer/src/decorators/binary/collection.ts
@@ -20,6 +20,7 @@ import {
   CollectionDecorator,
   SchemaUtils,
   TypeGetter,
+  ValidationError,
   parseDataUri,
 } from '@forestadmin/datasource-toolkit';
 import { filetypemime } from 'magic-bytes.js';
@@ -242,17 +243,31 @@ export default class BinaryCollectionDecorator extends CollectionDecorator {
     return value;
   }
 
+  private parseHex(value: string): Buffer {
+    if (!/^([0-9a-f]{2})*$/i.test(value)) {
+      throw new ValidationError(
+        `Expected an even-length hex string for a binary field, got "${value}"`,
+      );
+    }
+
+    return Buffer.from(value, 'hex');
+  }
+
   private async convertScalar(
     toBackend: boolean,
     useHex: boolean,
     value: unknown,
   ): Promise<unknown> {
     if (toBackend) {
-      const string = value as string;
+      if (typeof value !== 'string') {
+        throw new ValidationError(
+          `Expected a string for a binary field, got ${typeof value}: ${JSON.stringify(value)}`,
+        );
+      }
 
-      if (useHex) return Buffer.from(string, 'hex');
+      if (useHex) return this.parseHex(value);
 
-      return parseDataUri(string).buffer;
+      return parseDataUri(value).buffer;
     }
 
     const buffer = value as Buffer;
--- a/packages/datasource-customizer/src/decorators/binary/collection.ts
+++ b/packages/datasource-customizer/src/decorators/binary/collection.ts
@@ -16,7 +16,12 @@ import type {
   RecordData,
 } from '@forestadmin/datasource-toolkit';
 
-import { CollectionDecorator, SchemaUtils, TypeGetter } from '@forestadmin/datasource-toolkit';
+import {
+  CollectionDecorator,
+  SchemaUtils,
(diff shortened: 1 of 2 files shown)


packages/datasource-customizer/src/decorators/binary/collection.ts
```diff
@@ -237,15 +243,31 @@ export default class BinaryCollectionDecorator extends CollectionDecorator {
     return value;
   }
 
+  private parseHex(value: string): Buffer {
+    if (!/^([0-9a-f]{2})*$/i.test(value)) {
```
nbouliol: The advertised `Match` on hex fields is still `/^[0-9a-f]+$/` (`replaceValidation`, l.305), so `'303'` passes frontend validation and then 400s here, while `'AB'` is the reverse — align that regex to `/^([0-9a-f]{2})+$/i` as well?

nbouliol (commented):
Solid diagnosis, and the silent-empty-buffer case was worth catching. Two things I'd like to settle before approving: the advertised hex `Match` in `replaceValidation` now disagrees with `parseHex` in both directions, and a `Buffer` value on a hex column used to round-trip and now 400s without the Heads-up mentioning it. The rest is smaller. Two comments are anchored on the closest line inside the diff rather than the exact one they discuss (`replaceValidation` l.305 and `convertValueHelper` l.216 are outside the hunks).

packages/datasource-customizer/src/decorators/binary/collection.ts
```diff
@@ -237,15 +243,31 @@ export default class BinaryCollectionDecorator extends CollectionDecorator {
...
+      throw new ValidationError(
+        `Expected an even-length hex string for a binary field, got "${value}"`,
```
nbouliol: Could we slice the value like `parseDataUri` does (32 chars)? A multi-MB filter value lands verbatim in the message and the logs here, and `JSON.stringify` at l.264 dumps every byte of a Buffer. Also "even-length hex string" reads oddly for `'zzzz'`.

packages/datasource-customizer/src/decorators/binary/collection.ts
```diff
@@ -237,15 +243,31 @@ export default class BinaryCollectionDecorator extends CollectionDecorator {
...
+
+    return Buffer.from(value, 'hex');
+  }
+
   private async convertScalar(
     toBackend: boolean,
     useHex: boolean,
     value: unknown,
   ): Promise<unknown> {
     if (toBackend) {
```
nbouliol, line 263: `convertValueHelper`'s `if (value)` short-circuits, so `0`, `false` and `''` still slip past this guard unconverted — worth widening it?

packages/datasource-customizer/src/decorators/binary/collection.ts
```diff
@@ -237,15 +243,31 @@ export default class BinaryCollectionDecorator extends CollectionDecorator {
...
-      const string = value as string;
+      if (typeof value !== 'string') {
```
nbouliol, line 266: `Buffer.from(buf, 'hex')` used to return a copy, so programmatic writes and filters passing raw bytes for a Binary PK worked and now 400 — could `Buffer.isBuffer(value)` pass through?

packages/datasource-customizer/test/decorators/binary/collection.test.ts
```diff
@@ -238,6 +239,66 @@ describe('BinaryCollectionDecorator', () => {
     });
   });
 
+  describe('list filtering a binary column with a value that is not a data uri', () => {
```
nbouliol: What about `it.each` over `[field, value, matcher]` here — the four bodies are the same four lines?

packages/datasource-customizer/test/decorators/binary/collection.test.ts
```diff
@@ -238,6 +239,66 @@ describe('BinaryCollectionDecorator', () => {
...
+    it('should reject the filter instead of crashing on an undefined base64 payload', async () => {
```
nbouliol: This asserts only the class, so any ValidationError passes — match `/must be a data uri/` like its sibling below (same at l.294)?

packages/datasource-customizer/test/decorators/binary/collection.test.ts
```diff
@@ -238,6 +239,66 @@ describe('BinaryCollectionDecorator', () => {
...
+        conditionTree: new ConditionTreeLeaf('id', 'Equal', 'Anthony'),
+      });
+
+      await expect(decoratedBook.list(caller, filter, new Projection('id'))).rejects.toThrow(
+        /even-length hex string/,
+      );
+      expect(books.list).not.toHaveBeenCalled();
+    });
+
+    it('should reject an odd-length hex string, which silently drops its last digit', async () => {
```
nbouliol: The three neighbours also assert `books.list` was never called — worth adding it here?

packages/datasource-customizer/src/decorators/binary/collection.ts
Tonours: Done in a3f2a36 — `replaceValidation` now pushes the same `HEX_BYTES` constant the parser uses, so the two cannot drift again. You were right in both directions: `'303' passed the advertised `Match` then 400d here, and `'0F'/`'ABCD'` were blocked by the frontend though the backend accepts them. I kept `+` rather than `*` for the advertised regex so the empty string stays rejected there as before.

packages/datasource-customizer/src/decorators/binary/collection.ts
Tonours: Done in 8b78e1b — `value.slice(0, 32)` on the hex branch and a sliced `JSON.stringify` on the non-string one, matching `parseDataUri`. Worth noting the message is returned verbatim to the client for a `ValidationError`, so this was a response-size issue too, not only a log one. Also renamed to "hex string of full bytes", which reads correctly for `'zzzz'`. Covered by a test asserting the truncation at 32 chars.

packages/datasource-customizer/src/decorators/binary/collection.ts
Tonours: Verified and left alone, deliberately. `if (value)` predates this PR (identical on main), and it does let `''`, `0` and `false` through unconverted on both branches — so `Equal ''` can still hit the 500 class this PR narrows. It is a real gap, but a wider one than the reported crash, so I documented it as a Known limitation in the PR body rather than widening the diff. Happy to take it in a follow-up.

Side effect worth knowing: it also makes `parseHex`s acceptance of the empty string unreachable, so no empty Buffer is ever produced there.

packages/datasource-customizer/src/decorators/binary/collection.ts
Tonours: Good catch, and it was the worst finding here. Confirmed: `Buffer.from(buf, 'hex')` ignores the encoding and copies the bytes, so a raw `Buffer` on a hex column worked before and my guard broke it — on every Binary PK/FK, i.e. the common path. Fixed in 96ed52d with `Buffer.isBuffer(value)` returning early, plus a test asserting the buffer reaches the child collection unchanged. The datauri branch never worked for a Buffer (`string.split is not a function`), so only hex regressed. The PR bodys Heads-up now says so explicitly.

packages/datasource-customizer/test/decorators/binary/collection.test.ts
Tonours: Done in 8b78e1b — collapsed into one `it.each` over `[field, value, matcher]`. You were right that the precedent exists: `parse-query.test.ts:51` has a near-identical reject-table, and 5 files in this package use it.

packages/datasource-customizer/test/decorators/binary/collection.test.ts
Tonours: Done — both class-only assertions are gone. The data-uri one now matches `/must be a data uri/` via the `it.each` table, and the `create` one asserts `'Expected a string for a binary field, got number: 42'`.

packages/datasource-customizer/test/decorators/binary/collection.test.ts
Tonours: Added — every case in the table now asserts `books.list` was never called, which is the assertion that actually proves the guard fires before the datasource is queried.

packages/datasource-customizer/src/decorators/binary/collection.ts
```diff
@@ -16,9 +16,17 @@ import type {
...
+import {
+  CollectionDecorator,
+  SchemaUtils,
+  TypeGetter,
+  ValidationError,
+  parseDataUri,
+} from '@forestadmin/datasource-toolkit';
 import { filetypemime } from 'magic-bytes.js';
 
+const HEX_BYTES = /^([0-9a-f]{2})+$/i;
```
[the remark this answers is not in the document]
Tonours: Checked this one end-to-end and it does not reproduce — `convertScalar` is never reached with `""`, so `parseHex` never throws on it.

`convertValueHelper` guards the whole conversion with `if (value)` (l.220). `""` is falsy, so an empty hex string short-circuits and is returned as-is, before `columnType === 'Binary'` is even tested. The `parseHex` rejection of `""` is unreachable from create, update and filter alike.

Drove both directions through the decorator to confirm, rather than reading it off the source:

- read: child returns `{ id: Buffer.alloc(0) }` → decorator emits `{"id":""}`, no throw
- write: `create` with `{ id: '' }` → reaches the child collection as `{"id":""}`, no throw

So `+` vs `*` changes nothing for the empty value, and switching to `*` would loosen the `Match` that `replaceValidation` advertises to the frontend, where the empty string is rejected today and was before this PR. That is a real behavior change, in the wrong direction, for a crash that cannot happen.

There is a genuine bug hiding behind this, but it is not the regex: because of that same `if (value)`, `""` reaches the connector as an **empty string rather than an empty Buffer** — unconverted. That is the round-trip gap. It predates this PR (identical on `main`), affects `''`, `0` and `false` on both branches, and is written up as a Known limitation in the PR body. Widening it is a bigger diff than the reported crash, so it stays for a follow-up.

Leaving `HEX_BYTES` as `/^([0-9a-f]{2})+$/i`.

Tonours:
All review threads are answered and the branch is green

nbouliol (approved):

PR #1838, opened 2026-08-19
feat(agent-bff): serve the agent schema contract on GET /agent/v1/context

Tonours:
fixes PRD-944

## What

New route `GET /agent/v1/context`, declared in the served OpenAPI document. Serves the allow-listed agent schema so a trusted UI can build its own requests.

```json
{ "collections": [ { "name": "articles",
    "fields": [ { "field": "id", "type": "Uuid", "isPrimaryKey": true },
                { "field": "status", "type": "Enum", "enums": ["DRAFT","PUBLISHED"] },
                { "field": "thumbnail", "type": "String",
                  "validations": [{ "type": "is like", "value": "/^data:.*;base64,.*/" }] },
                { "field": "orders", "type": "String", "relationship": "HasMany",
                  "reference": "orders.customerId", "inverseOf": "customer" } ],
    "actions": [ { "id": "…", "name": "Ban user", "type": "single", "fields": [ … ] } ] } ],
  "meta": { "schemaRevision": 3, "environmentId": 42 } }
```

Both auth modes reach it — an OAuth session and a BFF API key get the same document, since the schema is not caller-scoped. No timezone required.

## Why

The client needs the schema to know which collections exist, which fields are filterable text, which actions take records. Today it gets that from the Forest SaaS directly, which means a SaaS token in the browser. The BFF already caches this schema for its own data routes.

## How

- reads `SchemaCache` (raw schema), not `ReadModel` (a projection that drops most field metadata)
- `ReadModelStore.getSchemaSnapshot()` returns `{collections, readModel, revision}` in one step, so collections and the read-model filtering them always belong to the same generation
- mounted **before** `createTimezoneMiddleware`, which otherwise rejects any `/agent` request without a timezone

Field types are passed through verbatim. A `Binary` column is advertised as `String` — that is what it is on the wire, bytes travel as a data uri or hex — so `validations` is the only reliable way to tell an encoded field from plain text. `relationship` and `polymorphicTargets` tell a to-one from a to-many, which `reference` alone cannot.

## Scope

No AI relay (`POST /agent/v1/ai/query`) — separate PR. No rendering/project/team identity. Per-field operators stay out (PRD-685).

## Known limitation

`meta.schemaRevision` is a process-local counter, reset on restart: a redeploy shipping a schema change can reproduce a previous value. Matters only if a consumer uses it to invalidate a cache.

## Test

```bash
```

1121 tests. One fixture pins every edge the wire produces: array and composite types, relations with and without `inverseOf`, polymorphic, the three action types, an endpoint-less action, `enums: null`, falsy defaults, malformed validations, dotted and spaced collection names.

### General

### Security


packages/agent-bff/src/context/build-context.ts
```diff
@@ -0,0 +1,146 @@
...
+    .map(entry =>
+      'value' in entry ? { type: entry.type, value: entry.value } : { type: entry.type },
+    );
+}
+
+function toContextField(field: FieldWithWireEnums): ContextField {
+  const serialized: ContextField = { field: field.field, type: field.type };
+
+  if (field.reference) serialized.reference = field.reference;
+  if (field.inverseOf) serialized.inverseOf = field.inverseOf;
```
nbouliol, line 84: `relationship` and `polymorphicReferencedModels` are dropped, so a polymorphic relation serializes identically to a plain `String` column and a consumer can't tell to-one from to-many — `readModel.getRelationTarget()` already returns the `{type, target|targets}` shape here.

packages/agent-bff/src/context/build-context.ts
```diff
@@ -0,0 +1,146 @@
...
+export interface AgentContext {
+  collections: ContextCollection[];
+  meta: ContextMeta;
+}
+
+function toArray<T>(value: T[] | null | undefined): T[] {
+  return Array.isArray(value) ? value : [];
+}
+
+type FieldWithWireEnums = ForestSchemaField & { enums?: string[] };
```
nbouliol, line 64: `ForestSchemaField` declares `enum`, the agent emits `enums` — fixing the type upstream in `forestadmin-client` would drop this intersection and the `as unknown as` cast in the fixture.

packages/agent-bff/src/context/build-context.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+export default function buildContext(
+  collections: ForestSchemaCollection[],
+  readModel: ReadModel,
+  meta: ContextMeta,
+): AgentContext {
+  return {
+    collections: collections
+      .filter(collection => readModel.isCollectionAllowed(collection.name))
```
nbouliol: `readModel` is built from these very collections in `getSchemaSnapshot`, so this filter can never drop one — and the test covering it fabricates a divergence the store can't produce.

packages/agent-bff/src/context/context-routes-middleware.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+async function readSnapshot(store: ReadModelStore): Promise<SchemaSnapshot> {
+  try {
+    return await store.getSchemaSnapshot();
+  } catch (error) {
+    if (error instanceof SchemaUnavailableError) throw schemaUnavailable();
+    throw error;
+  }
+}
```
nbouliol: `resolveReadModel` in `http/agent-route-helpers.ts` is the same try/catch mapping — what about a second export there rather than a private copy?

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -143,6 +143,85 @@ export const CountResponseSchema = z
...
+      'present`, `is longer than`, …) and `value` is passed through unchanged, so its shape ' +
+      'follows the rule: a number for a length rule, a date for a comparison. On `is like` it is ' +
+      'the JavaScript **literal** form of the regular expression — slashes included, and flags ' +
+      'after the closing one (`/^data:.*;base64,.*/`, but also `/^a|b|c$/g`). Parse it as a ' +
+      'literal, splitting on the LAST slash to separate pattern from flags; stripping the outer ' +
+      'characters instead leaves the flags inside the pattern, which then matches nothing. A ' +
+      'rule with no operand carries no `value`.',
+  });
+
+const ContextFieldSchema = z.object({
```
nbouliol, line 186: `ContextField` in `build-context.ts` restates this shape with nothing tying the two together — `z.infer<typeof ContextFieldSchema>` would keep them from drifting.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -143,6 +143,85 @@ export const CountResponseSchema = z
...
+    description:
+      'Everything the agent schema exposes: collection names, typed fields with their relation ' +
+      'markers, and the custom actions that carry an endpoint. Field types are passed through ' +
+      'from the agent wire format, so a type is a string, an array of types, or a composite ' +
+      'object. A `Binary` column is advertised as `String`, because that is what it is on the ' +
+      'wire — the bytes travel as a data uri or as hex — so `type` alone does not tell a text ' +
+      'field from an encoded one: read `validations` for that. `reference` keeps the raw agent ' +
+      'form, the foreign collection and the key joined by a dot — the collection name may itself ' +
+      'contain dots, so drop only the trailing segment to recover it. ' +
+      'The document carries no rendering, environment, project or team identity. It is NOT ' +
```
nbouliol: This denies carrying environment identity two lines before documenting `meta.environmentId` — worth reconciling, it's the only description a consumer reads.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -288,6 +289,27 @@ export function generateOpenApiDocument(version: string, unfolding?: Unfolding):
...
+      200: {
+        description: 'The exposed schema: collections, typed fields, relations and actions',
+        content: { 'application/json': { schema: ContextResponseSchema } },
+      },
+      400: errorRefs.byStatus['400'],
+      401: errorRefs.byStatus['401'],
+      403: errorRefs.byStatus['403'],
+      500: errorRefs.byStatus['500'],
+      501: errorRefs.byStatus['501'],
+      503: errorRefs.byStatus['503'],
```
nbouliol, line 308: 501 only comes from the action-execute mapper, so this route can't return it — and if the 400 is there for the shared bodyParser, 413/415 come from the same place.

nbouliol (commented):
Solid work — the snapshot seam is the right call, and the fixture pins a genuinely wide set of wire edges. One blocking point on the relation metadata, plus a few smaller things.

packages/agent-bff/src/context/build-context.ts
Tonours: Confirmed and fixed in dda999635 — you were right that it was blocking. Reproduced before the fix: a polymorphic relation serialized to `{"field":"poly","type":"String"}`, byte-identical to a plain text column, and a HasMany was indistinguishable from a BelongsTo since both carried only `reference`.

The contract now carries `relationship` (the four `RelationshipType` values) and `polymorphicTargets`. That also repaired my own rationale for `inverseOf`: I added it for the relation route, which only exists for to-many relations — without `relationship` a consumer had to guess which side it was on.

packages/agent-bff/src/context/build-context.ts
Tonours: Right on the facts — `readModel` comes from `getSchemaSnapshot()`, which builds it from the very array it returns, and `ReadModel`'s constructor adds every `collection.name` to its set. So the filter drops nothing in production, and the test at build-context.test.ts:203 fabricates a pair the store cannot produce.

Keeping it deliberately, as a guard rather than a filter: if the two sources ever diverge, serializing a collection the read-model doesn't know would be worse than an over-cautious no-op. I removed the OpenAPI wording that oversold it as curation — the description now says the document is not filtered by anything.

packages/agent-bff/src/context/context-routes-middleware.ts
Tonours: Done in dda999635 — added `resolveSchemaSnapshot` next to `resolveReadModel` in `http/agent-route-helpers.ts` and dropped the private copy. Same try/catch, same SchemaUnavailableError → 503 mapping, one place.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Fixed in f86f1859c. It read "carries no rendering, environment, project or team identity" and then documented `meta.environmentId` four lines down — plainly contradictory. Now: "no rendering, project or team identity, and the only environment datum is `meta.environmentId` below".

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Half right, and I checked both halves.

413/415 added in f86f1859c — you're right, same shared bodyParser as the 400.

501 I'm keeping, because it is reachable on this route: when the read-model bundle is absent (no `FOREST_SERVER_URL`/`FOREST_ENV_SECRET`), `buildContextMiddlewares` returns `[]`, the request falls through to `createAgentStubMiddleware` (cli-core.ts:309) and gets `501 not_implemented`. So the action-execute mapper isn't its only source. The 403 is gone though — the route no longer refuses api keys.

packages/agent-bff/src/context/build-context.ts
Tonours: Agreed on the diagnosis: `ForestSchemaField` declares `enum`, the agent emits `enums` (generator-fields.ts:71, generator-actions.ts:45), and `unfolding.ts:80` already works around the same gap.

Not doing it in this PR though — fixing the declared type in `forestadmin-client` changes a published type for every consumer of that package, which is a wider blast radius than a BFF route deserves to carry. Worth its own ticket; I'll open one. Until then the intersection type is named `FieldWithWireEnums` so the reason is visible at the use site rather than hidden in a cast.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Tempting, and I went back and forth on it. Not taking it: `z.infer<typeof ContextFieldSchema>` would make the serializer depend on the OpenAPI schema, so the documentation layer becomes the source of truth for the runtime shape. I'd rather have the dependency the other way round, or neither.

The drift risk you're pointing at is real though. A conformance test — validate a built context against `ContextResponseSchema` — would catch it without inverting the dependency. Happy to add that here if you'd prefer it over leaving the two declarations side by side.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -288,6 +288,28 @@ export function generateOpenApiDocument(version: string, unfolding?: Unfolding):
...
+      200: {
+        description: 'The exposed schema: collections, typed fields, relations and actions',
+        content: { 'application/json': { schema: ContextResponseSchema } },
+      },
+      400: errorRefs.byStatus['400'],
+      401: errorRefs.byStatus['401'],
+      413: errorRefs.byStatus['413'],
+      415: errorRefs.byStatus['415'],
+      500: errorRefs.byStatus['500'],
+      501: errorRefs.byStatus['501'],
```
nbouliol, line 307: Correcting my earlier comment, which sent this the wrong way: `@koa/bodyparser` only parses POST/PUT/PATCH, so 413/415 can't happen on a GET, and 501 *is* reachable here (no api-key config → `buildContextMiddlewares` returns `[]` → the agent stub answers). The real gap is 403, which this commit made *more* reachable — `forestIdentityNotAllowed` and `originNotAllowed` both fire before this route on the api-key path.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -277,9 +278,8 @@ export function generateOpenApiDocument(version: string, unfolding?: Unfolding):
     type: 'http',
     scheme: 'bearer',
     description:
-      'Mode 1: the BFF session token issued after the OAuth login. It authenticates the caller ' +
-      'but the data and action routes reject it until the BFF mints an agent token from the ' +
-      'OAuth principal, so no operation lists it yet. Use the API key today.',
+      'Mode 1: the BFF session token issued after the OAuth login. Accepted on every agent route, ' +
+      'including the context contract.',
```
nbouliol: The data and action routes still advertise `bffApiKey` only (`SECURITY`, line 29), so "every agent route" isn't true in the document.

packages/agent-bff/src/http/agent-route-helpers.ts
```diff
@@ -25,6 +26,15 @@ export async function resolveReadModel(store: ReadModelStore): Promise<ReadModel
...
 }
 
+export async function resolveSchemaSnapshot(store: ReadModelStore): Promise<SchemaSnapshot> {
+  try {
+    return await store.getSchemaSnapshot();
+  } catch (error) {
+    if (error instanceof SchemaUnavailableError) throw schemaUnavailable();
+    throw error;
+  }
+}
```
nbouliol, line 35: These two are now the same try/catch twice — `resolveReadModel` could be `(await resolveSchemaSnapshot(store)).readModel`.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -143,6 +143,88 @@ export const CountResponseSchema = z
...
+      'after the closing one (`/^data:.*;base64,.*/`, but also `/^a|b|c$/g`). Parse it as a ' +
+      'literal, splitting on the LAST slash to separate pattern from flags; stripping the outer ' +
+      'characters instead leaves the flags inside the pattern, which then matches nothing. A ' +
+      'rule with no operand carries no `value`.',
+  });
+
+const ContextFieldSchema = z.object({
+  field: z.string(),
+  type: ContextFieldTypeSchema,
+  relationship: z.enum(['BelongsTo', 'HasOne', 'HasMany', 'BelongsToMany']).optional(),
```
nbouliol: That makes three declarations of this union — here, `RelationshipType` in `read-model.ts`, and `ForestSchemaField.relationship` in `forestadmin-client` — with no compiler link between them.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -143,6 +143,88 @@ export const CountResponseSchema = z
...
+      'Everything the agent schema exposes: collection names, typed fields with their relation ' +
+      'markers, and the custom actions that carry an endpoint. Field types are passed through ' +
+      'from the agent wire format, so a type is a string, an array of types, or a composite ' +
+      'object. A `Binary` column is advertised as `String`, because that is what it is on the ' +
+      'wire — the bytes travel as a data uri or as hex — so `type` alone does not tell a text ' +
+      'field from an encoded one: read `validations` for that. `reference` keeps the raw agent ' +
+      'form, the foreign collection and the key joined by a dot — the collection name may itself ' +
+      'contain dots, so drop only the trailing segment to recover it. ' +
+      'The document carries no rendering, project or team identity, and the only environment ' +
+      'datum is `meta.environmentId` below. It is served to both auth modes — an OAuth session and a BFF API key get the same document. It is NOT ' +
```
nbouliol: 139 chars where the rest of the file hand-wraps at 100, and prettier can't split a literal.

packages/agent-bff/test/context/context-routes-middleware.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      const fetchSchema = jest.fn().mockResolvedValue(schema);
+      const withKey = makeApp(fetchSchema, 'api-key');
+      const withSession = makeApp(jest.fn().mockResolvedValue(schema));
+
+      const keyResponse = await request(withKey.app.callback()).get(ROUTE);
+      const sessionResponse = await request(withSession.app.callback()).get(ROUTE);
+
+      expect(keyResponse.status).toBe(200);
+      expect(keyResponse.body).toEqual(sessionResponse.body);
+    });
```
nbouliol: The middleware no longer reads `ctx.state.authMode`, so both apps are identical by construction and this `toEqual` can't fail — the thing worth proving is an `X-Forest-Bff-Key` caller reaching the route through the full chain.

nbouliol (commented):
Second pass. Relation metadata and the helper extraction look good.

One thing to flag up front: my previous comment on the declared status codes was wrong in both halves, and you acted on it — details in the thread on `openapi-document.ts`. Net effect is that 403 should come back and 413/415 should go.

On the API-key reversal: the reasoning holds — the document reads no principal, and it is meant to be crossed with `/agent/v1/permissions`. But the ticket still carries `Given le mode API-key, when la route est appelée, then elle est refusée — mode OAuth exigé`, and the widening is real (a rendering-scoped key can now enumerate the whole environment schema). Worth updating PRD-944 rather than leaving the AC contradicted by the PR body.

Still open from the first pass: the allow-list filter in `build-context.ts` is now *provably* dead — your own new `read-model-store.test.ts:32-34` asserts the two can never diverge.

packages/agent-bff/src/context/build-context.ts
Tonours: Taking it — you were right and my "defensive guard" answer did not hold up.

Removed in b1d4c50df. `ReadModel`'s constructor fills its collection set from the array it is handed, and `getSchemaSnapshot()` returns that same array alongside the model built from it, so `isCollectionAllowed` was answering true for every entry by construction. The filter could not drop anything, and keeping dead code as insurance against a divergence the type system already prevents was the wrong trade.

The test that fabricated the impossible pair is gone too — it hand-built a `ReadModel` from a different array than it passed as collections, which the store cannot produce.

`isActionAllowed` stays: it reads `actionEndpoints`, a separate source that genuinely filters. Only the collection-level filter was dead.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Added the conformance test in b1d4c50df, since you did not come back on the offer and the drift risk you raised is real either way.

`build-context.test.ts` now validates a built context against `ContextResponseSchema` with `safeParse`, using the fixture that covers every wire edge. That catches a divergence between the serializer and the document without making the OpenAPI layer the source of truth for the runtime shape — the dependency stays test-only, neither module imports the other.

I checked it can actually fail rather than just pass: typing `meta.schemaRevision` as a string makes it fail with `Invalid input: expected number, received string`, and reverting makes it green again.

Still not taking `z.infer<typeof ContextFieldSchema>` for the reason above, but the two declarations are now pinned to each other by a test instead of by nothing.

packages/agent-bff/src/context/build-context.ts
```diff
@@ -0,0 +1,152 @@
...
+
+function toContextField(field: FieldWithWireEnums): ContextField {
+  const serialized: ContextField = { field: field.field, type: field.type };
+
+  if (field.relationship) serialized.relationship = field.relationship;
+  if (field.reference) serialized.reference = field.reference;
+  if (field.inverseOf) serialized.inverseOf = field.inverseOf;
+
+  const polymorphicTargets = toArray(field.polymorphicReferencedModels);
+  if (polymorphicTargets.length > 0) serialized.polymorphicTargets = [...polymorphicTargets];
```
nbouliol, line 87: `collect-unfolding.ts:157` drops relations whose foreign collection is outside the allow-list so the document doesn't "promise a dead path" — here `reference` and `polymorphicTargets` go out unfiltered, so the contract can name a target absent from `collections[]`.

packages/agent-bff/test/context/build-context.test.ts
```diff
@@ -0,0 +1,298 @@
...
+  // edge, so validating what the serializer emits against what the document promises catches a drift
+  // between the two without either one importing the other's types.
+  describe('when the built context is validated against the published OpenAPI schema', () => {
+    it('should satisfy ContextResponseSchema for every shape the fixture covers', () => {
+      const context = buildContext(schema, readModel, { schemaRevision: 3, environmentId: 42 });
+
+      const result = ContextResponseSchema.safeParse(context);
+
+      expect(result.error?.issues ?? []).toEqual([]);
+      expect(result.success).toBe(true);
```
nbouliol: `z.object()` strips unknown keys rather than rejecting them, so this passes for a field the serializer emits but the document doesn't declare — `expect(ContextResponseSchema.parse(context)).toEqual(context)` catches that direction too, and drops the vacuous first assertion.

packages/agent-bff/test/context/context-routes-middleware.test.ts
```diff
@@ -0,0 +1,178 @@
...
+  const fetcher: SchemaFetcher = { fetchSchema };
+  const schemaCache = new SchemaCache({ fetcher, metrics: makeMetrics() });
+  const store = new ReadModelStore(schemaCache, new CapabilitiesCache({}));
+
+  const app = new Koa();
+  app.use(createErrorMiddleware({ logger: () => {} }));
+  app.use(async (ctx, next) => {
+    ctx.state.authMode = authMode;
+    await next();
+  });
```
nbouliol: Nothing in this chain reads `ctx.state.authMode` any more, so the parameter and this middleware are dead — and the `describe` at line 91 sends no `Authorization` header, so it isn't about an OAuth session either.

nbouliol (commented):
Dropping the filter is right — `buildContext` has one non-test caller, isn't exported from the package index, `getSchemaSnapshot` has no `await` between the collections and the revision read, and the deleted test's invariant is covered directly at `read-model-store.test.ts:30-37`.

Ran locally on this head: 1123 tests / 72 suites pass, `tsc --noEmit` clean, `prettier --check` clean. The only red CI job is the `ai-proxy` LLM integration suite, which is a live OpenAI call unrelated to this PR.

One knock-on from the new conformance test, in the thread below: it also pins `fields: []` for the fields-less fixture collection as contract-valid, which hardens the gap against `jamais de fields: []`.

Still outstanding at the ticket level, not the code: PRD-944 still carries `Given le mode API-key, then elle est refusée — mode OAuth exigé`, and there is now a green test asserting the opposite. Worth updating the AC rather than leaving it contradicted.

packages/agent-bff/test/context/build-context.test.ts
Tonours: Right, and I had verified the wrong direction. Confirmed empirically before changing it: a field carrying `bogusKey` returns `success: true` from `safeParse` and comes back stripped, so the assertion could not see a key the serializer emits and the document does not declare.

Fixed in c917ac38d — now `expect(ContextResponseSchema.parse(context)).toEqual(context)`, and the vacuous `result.error?.issues` assertion is gone. Re-ran the mutation both ways: adding an undeclared key to `ContextField` now fails with `+ "undeclaredKey": "drift"`, and a string `schemaRevision` still fails on the type.

packages/agent-bff/test/context/context-routes-middleware.test.ts
Tonours: Both true — removed in c917ac38d. The `authMode` parameter and the middleware that set it are gone (all four callers used the default), and the `describe` is now `when the route serves the contract`, since it sends no credentials and exercises the serializer and the cache. `makeApp` carries a comment saying the credentialed paths go through `makeEdge`.

packages/agent-bff/src/context/build-context.ts
Tonours: Real gap — the fixture already exhibits it: it references `orders` and `teams` while serving only `users`, `User.address`, `My Coll` and `collectionWithoutFieldsNorActions`. Documented rather than filtered, in c917ac38d.

The two cases are not symmetric. `collect-unfolding` emits **routes**, so documenting `orders` while it is hidden promises an endpoint that 404s — dropping it is right there. `/context` emits a **schema**: `reference: "orders.customerId"` claims the field points at `orders`, which stays true whether or not this document serves it.

Filtering would also undo your round-1 point. Strip `reference` and `polymorphicTargets` and `ownerPolymorphic` goes back to `{"field":"ownerPolymorphic","type":"String","relationship":"BelongsTo"}` — a relation indistinguishable from a text column, which is exactly the bug those two fields were added to fix. I would rather not trade a blocking finding for a smaller one.

So the contract now states it: the description says a target named by `reference` or `polymorphicTargets` is not guaranteed to appear in `collections[]`, and the consumer must cross-check. A test pins the behaviour so it reads as a decision rather than an oversight. If you would rather the document never name an unserved target, that is a contract change worth its own ticket — say so and I will open it.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -143,6 +144,94 @@ export const CountResponseSchema = z
...
+      'contain dots, so drop only the trailing segment to recover it. A target named by ' +
+      '`reference` or `polymorphicTargets` is NOT guaranteed to appear in `collections[]`: the ' +
+      'schema describes a field as the agent declares it, and a relation can point at a ' +
+      'collection this document does not expose. Cross a target against `collections[]` before ' +
+      'following it — unlike the per-collection route documents, nothing here is dropped for ' +
+      'pointing outside the served set. ' +
+      'The document carries no rendering, project or team identity, and the only environment ' +
+      'datum is `meta.environmentId` below. It is served to both auth modes — an OAuth session ' +
+      'and a BFF API key get the same document. It is NOT ' +
+      'filtered by the caller permissions, nor by anything else: it describes the whole exposed ' +
```
nbouliol: Line 212 already says actions are filtered to those carrying an endpoint, so "nor by anything else" contradicts it — `build-context.ts:130` does drop them.

packages/agent-bff/test/context/build-context.test.ts
```diff
@@ -0,0 +1,318 @@
...
+          enums: ['spam', 'abuse'],
+        },
+      ]);
+    });
+  });
+
+  // Deliberate, and the opposite of `collect-unfolding`, which drops a relation whose target is
+  // hidden because documenting a route would promise a 404. Here nothing is a route: dropping the
+  // target would make a relation indistinguishable from a plain text column, which is the very bug
+  // `relationship`/`polymorphicTargets` exist to prevent.
```
nbouliol: `relationship` is a separate key from `reference`/`polymorphicTargets`, so a field stripped of its target would still read `{relationship: 'HasMany'}` — the first sentence is the real reason, the second doesn't hold.

packages/agent-bff/test/context/build-context.test.ts
```diff
@@ -0,0 +1,318 @@
...
+  describe('when a relation points at a collection the document does not serve', () => {
+    it('should still name the target, leaving the cross-check to the consumer', () => {
+      const context = buildContext(schema, readModel, { schemaRevision: 1 });
+      const served = context.collections.map(collection => collection.name);
+
+      expect(served).not.toContain('orders');
+      expect(fieldNamed(context, 'ordersHasManyWithInverseOf')?.reference).toBe(
+        'orders.customerId',
+      );
```
nbouliol, line 264: `users` is in the served set, so this half doesn't show an unserved target yet — `expect(served).not.toContain('teams')` would close it.

packages/agent-bff/test/context/fixtures.ts
```diff
@@ -0,0 +1,122 @@
...
+            { message: 'no type at all' },
+            'garbage',
+            { type: 'contains', value: 'ok' },
+          ],
+        } as unknown as Partial<ForestSchemaField>),
+        field('fieldWithNullValidations', 'String', {
+          validations: null,
+        } as unknown as Partial<ForestSchemaField>),
+        field('ordersHasManyWithInverseOf', 'String', {
+          reference: 'orders.customerId',
```
nbouliol, line 95: The document's "drop only the trailing segment" rule has no fixture behind it — a reference into the served `User.address` would exercise it.

packages/agent-bff/test/context/build-context.test.ts
```diff
@@ -0,0 +1,318 @@
...
+  // edge, so validating what the serializer emits against what the document promises catches a drift
+  // between the two without either one importing the other's types.
+  describe('when the built context is validated against the published OpenAPI schema', () => {
+    // Compared to the input rather than asserted valid: `z.object()` strips unknown keys instead of
+    // rejecting them, so a key the serializer emits and the document does not declare would pass
+    // validation and be silently dropped. Round-tripping catches that direction too.
+    it('should round-trip through ContextResponseSchema for every shape the fixture covers', () => {
+      const context = buildContext(schema, readModel, { schemaRevision: 3, environmentId: 42 });
+
+      expect(ContextResponseSchema.parse(context)).toEqual(context);
```
nbouliol: `toEqual` ignores `undefined`-valued keys, so `toStrictEqual` matches what the title claims.

packages/agent-bff/test/context/context-routes-middleware.test.ts
```diff
@@ -0,0 +1,176 @@
...
+import ReadModelStore from '../../src/read-model/read-model-store';
+import SchemaCache from '../../src/read-model/schema-cache';
+import { makeMetrics } from '../read-model/fixtures';
+
+const ROUTE = '/agent/v1/context';
+const AUTH_SECRET = 'context-secret';
+const RAW_KEY = `fbff_${'a'.repeat(16)}_${'b'.repeat(64)}`;
+
+// No auth middleware: the route reads no principal, so these cases exercise the serializer and the
+// cache. The credentialed paths go through `makeEdge` below, which mounts the real chain.
```
nbouliol: CLAUDE.md asks for no comments in tests — this PR now adds 15 lines across 5 blocks (here, `:53-54`, and `build-context.test.ts:215-218, 304-307, 309-311`); the `describe`/`it` strings carry most of it already.

nbouliol (commented):
All three of the last round's points landed. The round-trip assertion is exactly the fix, and documenting the unserved-target behaviour rather than filtering it is the better call — I'm not reopening that.

Ran locally on this head: 1124 tests / 72 suites pass, `tsc --noEmit` clean, `prettier --check` clean. The only red CI job is still the `ai-proxy` LLM integration suite, a live OpenAI call unrelated to this PR.

What's left is small. Nothing here blocks.

Still outstanding at the ticket level rather than in the code: PRD-944 carries `Given le mode API-key, then elle est refusée — mode OAuth exigé`, and there is a green test asserting the opposite. Worth updating the AC.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Right, and the contradiction was mine — line 212 says the document carries "the custom actions that carry an endpoint", so "nor by anything else" was false. `build-context.ts:130` filters on `readModel.isActionAllowed`, which is `actionEndpoints[collection]?.[action] !== undefined` — an endpoint check, exactly what line 212 announces.

Fixed in 796473d36: the sentence now reads "NOT filtered by the caller permissions: it describes the whole exposed schema minus the endpoint-less actions". The permission claim is what a consumer needs; the endpoint filter is stated once instead of denied once and asserted once.

packages/agent-bff/test/context/build-context.test.ts
Tonours: True — a field keeps `relationship` whether or not its target survives, so "indistinguishable from a plain text column" only held for the round-1 state where neither key existed. The comment overstated it.

Removed in 1720cf4fb, along with the other four blocks (see the CLAUDE.md thread). The `describe`/`it` strings carry the decision: "when a relation points at a collection the document does not serve" / "should still name the target, leaving the cross-check to the consumer".

packages/agent-bff/test/context/build-context.test.ts
Tonours: Added in 1720cf4fb — `expect(served).not.toContain('teams')` now sits next to the `orders` one, so both targets the fixture names are proven absent from `collections[]` before the assertions on `reference` and `polymorphicTargets`.

packages/agent-bff/test/context/fixtures.ts
Tonours: Right — every reference in the fixture pointed at a single-segment name, so the "drop only the trailing segment" rule was documented and unexercised.

Added in 1720cf4fb: `addressBelongsToDottedCollection` carries `reference: 'User.address.city'`, pointing at the served `User.address` collection, with a test pinning that the reference goes out whole. A consumer splitting on the first dot would resolve `User`, which does not exist.

packages/agent-bff/test/context/build-context.test.ts
Tonours: Taken — `toStrictEqual` in 1720cf4fb. `toEqual` treats a key parsed to `undefined` as absent, so an optional key the document declares and the serializer never emits would have round-tripped silently. Re-checked the mutation still fails: an undeclared key on `ContextField` reports `+ "undeclaredKey"`, a string `schemaRevision` still fails on the type.

packages/agent-bff/test/context/context-routes-middleware.test.ts
Tonours: Fair, and CLAUDE.md is unambiguous. All five blocks removed in 1720cf4fb.

Two of them carried something the strings did not, so the code says it instead: `makeApp` is now `makeRouteOnlyApp` and `makeEdge` is `makeFullAgentEdge`, which is what the two comments were there to distinguish. The other three were either restating the `describe` or, in one case, wrong (see the `:218` thread).

packages/agent-bff/src/context/build-context.ts
Tonours: Follow-up on the ticket I promised: PRD-993 — "forestadmin-client: ForestSchemaField declares `enum` but the wire sends `enums`", currently In Review. Once that lands, `FieldWithWireEnums` here and the `as unknown as` casts in the fixture both go away.

Leaving this PR as is unless you want the upstream fix folded in — say so and I will.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Closing the loop: the conformance test landed in b1d4c50df and was tightened in c917ac38d and again in 1720cf4fb — `expect(ContextResponseSchema.parse(context)).toStrictEqual(context)`, which now catches drift in both directions (a key the document declares and the serializer never emits, and one the serializer emits and the document does not declare). Reopen if you would still rather have `z.infer`.

packages/agent-bff/src/context/build-context.ts
Tonours: No answer on the contract-change question, so I am leaving the behaviour as it stands and treating the point as settled by documentation plus a test. The document states the target may be absent from `collections[]`, and 1720cf4fb strengthened the test with `expect(served).not.toContain('teams')` so both unserved targets the fixture names are proven absent.

If you would rather the document never name an unserved target, that is a contract change — reopen this and I will open the ticket.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Correcting myself once more, this time with a measurement instead of a reading. I claimed 501 was reachable here and left it at that; I had not checked what happens *before* the stub.

Booted `runCli` with no `FOREST_ENV_SECRET` and hit the route:
- with `X-Forest-Timezone` → `501 not_implemented`
- without it → `400 missing_timezone`

So both declared statuses are real on this route, but the 400 is not only the shared bodyParser — in degraded mode the timezone middleware answers first, since `/context` is no longer mounted and the request falls through to it. Your original point stands in a way I had missed: the reason I gave for the 400 was the wrong one.

Both are now pinned by an integration test in `test/cli-core.test.ts` ("should fall through to the agent stub when the deployment carries no read-model"), so the declaration is verified rather than argued.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -143,6 +144,104 @@ export const CountResponseSchema = z
       'collection disables count. The pair never disagrees.',
   });
 
+const ContextFieldTypeSchema: z.ZodType = z
+  .lazy(() =>
+    z.union([
+      z.string(),
+      z.array(ContextFieldTypeSchema),
+      z.object({
+        fields: z.array(z.object({ field: z.string(), type: ContextFieldTypeSchema })),
```
nbouliol: `ForestServerColumnType` carries `enums?` on composite sub-fields (`schema/types.ts:19`) and `type` is passed through by reference, so those values ship but the document denies them — the fixture's only composite is a plain `{field, type}`, so the round-trip can't catch it.

packages/agent-bff/src/context/build-context.ts
```diff
@@ -0,0 +1,155 @@
...
+  return meta;
+}
+
+export default function buildContext(
+  collections: ForestSchemaCollection[],
+  readModel: ReadModel,
+  meta: ContextMeta,
+): AgentContext {
+  return {
+    collections: collections.map(collection => toContextCollection(collection, readModel)),
```
nbouliol, line 157: The outermost list is the one left unguarded, so a null collection throws where a null field is dropped — and no fixture puts a non-object in `collection.fields`/`collection.actions`, so the new guards aren't exercised either.

nbouliol (commented):
Good batch. Every point from the last round is closed, and the ticket amendment settles the API-key question I'd been carrying since pass 2 — I'm treating that as resolved.

Verified on this head: `tsc --noEmit` clean, `prettier --check` clean, CI green on `agent-bff`. (My local jest run showed failures, but all of them are the same `supertest` `.send()` resolution error in `timezone-middleware.test.ts`, which this PR never touches — the worktree has no installed `node_modules`. Environmental, ignore it.)

The recursive `ContextFieldType` renders correctly: `components.schemas.ContextFieldType` emits `anyOf: [string, array→$ref self, object{fields:[…$ref self]}]`, and the spec-validity and unfolded tests pass over it.

Two things below. Neither blocks.

One last doc gap with no diff line to hang it on: `packages/agent-bff/README.md` never mentions `/agent/v1/context`, and its Timezone section still states `None → 400 missing_timezone` unconditionally, which is no longer true for this route.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Confirmed empirically before fixing, and it was worse than the document: the BFF's own `FieldType` (`capabilities-cache.ts:7`) omitted `enums` on composite sub-fields too, so my `z.lazy()` was a faithful copy of an already-incomplete type. Parsed a composite carrying `{field: 'country', type: 'Enum', enums: ['FR','BE']}` — it came back as `{field: 'country', type: 'Enum'}`, silently stripped.

Fixed in bba8d23a9:
- `ContextFieldTypeSchema` declares `enums?: string[]` on the sub-field, matching `ForestServerColumnType` (`schema/types.ts:19`)
- `FieldType` widened the same way, so the internal type and the published document now agree
- the fixture's composite carries three sub-fields — a plain string, an `Enum` with `enums`, and a nested array type — so the recursion and the enums are both exercised

Re-ran the mutation to check the test can now fail: undeclaring `enums` on the sub-field makes the round-trip fail. `field-schemas.ts:43` still ignores composite sub-field enums when building JSON Schema, but that is the unfolded document, not this route — separate concern, untouched.

packages/agent-bff/src/context/build-context.ts
Tonours: You are right, and the honest fix was to remove code rather than add a third guard.

Checked reachability for all three lists. `ReadModel` is constructed from the same array `buildContext` receives, and it dereferences an entry before `buildContext` ever runs — `read-model.ts:82` (`collection.name`), `:98` (`field.relationship` via `toRelationTarget`), `:114` (`action.endpoint`). A non-object in `collections`, `collection.fields` or `collection.actions` therefore throws upstream on `main`, independent of this PR. I verified it by adding such a fixture: it fails in `ReadModel`, not here.

`action.fields` is the one exception — `ReadModel` never reads it — and the fixture already puts a `null` there (`fixtures.ts:116`, pinned by `build-context.test.ts:196`).

So in 82f7e0960 the guards are back to that single list, and the `toObjectArray` helper I had introduced is gone — one call site did not earn it. `build-context.ts` is now a pure addition against `main` again. Hardening `ReadModel` is the actual root cause but it is pre-existing and outside this route; I would rather it get its own ticket than ride along here.

packages/agent-bff/src/context/build-context.ts
```diff
@@ -0,0 +1,153 @@
...
+  };
+}
+
+function toContextCollection(
+  collection: ForestSchemaCollection,
+  readModel: ReadModel,
+): ContextCollection {
+  return {
+    name: collection.name,
+    fields: toArray(collection.fields).map(toContextField),
```
nbouliol: A primitive entry survives `ReadModel` (`toRelationTarget` reads `.relationship` off a string and returns null), so it reaches here and serializes as `{}` — no `field`, no `type`, against the published `ContextFieldSchema`. `null` really was dead code; primitives weren't.

packages/agent-bff/src/context/build-context.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  const polymorphicTargets = toArray(field.polymorphicReferencedModels);
+  if (polymorphicTargets.length > 0) serialized.polymorphicTargets = [...polymorphicTargets];
+
+  if (field.isPrimaryKey) serialized.isPrimaryKey = true;
+  if (field.isRequired) serialized.isRequired = true;
+  if (field.isReadOnly) serialized.isReadOnly = true;
+
+  const enums = toArray(field.enums);
+  if (enums.length > 0) serialized.enums = [...enums];
```
nbouliol: `enums: []` is dropped here but kept on action fields at `:107`, so an enum column with no values is indistinguishable from `enums: null` — `build-context.test.ts:178` argues empty lists "carry meaning".

packages/agent-bff/test/context/context-routes-middleware.test.ts
```diff
@@ -0,0 +1,204 @@
...
+    });
+
+    it('should fetch the schema once on a cold cache and never again while it stays warm', async () => {
+      const fetchSchema = jest.fn().mockResolvedValue(schema);
+      const { app } = makeRouteOnlyApp(fetchSchema);
+
+      await request(app.callback()).get(ROUTE);
+      await request(app.callback()).get(ROUTE);
+
+      expect(fetchSchema).toHaveBeenCalledTimes(1);
```
nbouliol: The warm response is never read, so a warm hit returning 500 or an empty `collections` still passes — the TTL test just below captures its second response and asserts on it.

nbouliol (commented):
Both points from last round are properly closed. The composite `enums` fix matches `ForestServerColumnType` and has real fixture coverage (nested array, array-of-composite, sub-field `enums: ['FR','BE']`), and the round-trip test is load-bearing — deleting `enums` from the zod schema does make it fail. CI fully green this time, `tsc` and `prettier` clean.

On the guard removal: you were right about `null` — it throws in `ReadModel` (`read-model.ts:29` for fields, `:105` for actions) before `buildContext` ever runs, so those two guards were dead, and keeping the `action.fields` one is exactly right since `read-model.ts:117` spreads `null` into `{}` without complaint. I said as much before checking primitives, which behave differently — thread below.

Also below: the warm-cache assertion, which I should have raised several rounds ago and didn't.

One AC-wording note, not code: the ticket line says the document is "filtered by nothing", but actions genuinely are filtered. The served OpenAPI already says it correctly ("minus the endpoint-less actions"), so it is the AC sentence that is stale.

packages/agent-bff/src/context/build-context.ts
Tonours: Right, and the guard is back at `build-context.ts:129`. `toRelationTarget` reads `.relationship` off a string and returns null, `buildPrimaryKeys` reads `.isPrimaryKey` and gets undefined, so a primitive walks through `ReadModel` untouched and used to serialize as `{}`. Fixture now carries a bare `'garbage'` entry in `users.fields`, and the round-trip test against `ContextResponseSchema` fails without the filter — so does the new explicit one.

packages/agent-bff/src/context/build-context.ts
Tonours: Aligned on the action-field behaviour: fields now use the same `Array.isArray` check, so `enums: []` survives and `enums: null` still drops. Fixture has `statusWithEmptyEnums`, and the test asserting the old omit-on-empty behaviour is replaced by one pinning the empty list.

packages/agent-bff/test/context/context-routes-middleware.test.ts
Tonours: Fixed: the warm request is captured and asserted — status 200 and body equal to the cold one, so a warm 500 or an emptied `collections` now fails the test rather than passing on the fetch count alone.

Tonours:
Rebased on `main` (force-pushed). Two collisions worth knowing about, since neither is in the review scope:

- `cli-core.ts`: main added the Redoc docs routes while this branch renamed `oauthMiddlewares` to `oauth.middlewares` — kept both.
- `openapi-unfolded.test.ts`: main's two new tag invariants iterate `document.paths` and read `.post` off every entry. `GET /agent/v1/context` is the first get-only path in the document, so both threw `Cannot read properties of undefined`. They now iterate the `paths` constant this file already builds, which excludes `/context` — the invariant is about collection grouping, and an environment-level route carries no collection tag.

The three review points are addressed in `c0d07c1`, `32c688b` and `abd9d43`, each answered in its thread.

On the AC wording: updated PRD-944 — "aucune collection n'est filtrée", with the action filter spelled out explicitly.

Local run on this head: 1179 tests pass. The 10 failures are all missing local deps and reproduce on plain `main` — the Redoc bundle (`docs-routes.test.ts`, one `cli-core.test.ts` case) and `@hey-api/openapi-ts` (`openapi-generated-client.test.ts`). `tsc --noEmit` and `prettier --check` clean.

packages/agent-bff/src/context/build-context.ts
```diff
@@ -0,0 +1,154 @@
...
+  collection: ForestSchemaCollection,
+  readModel: ReadModel,
+): ContextCollection {
+  return {
+    name: collection.name,
+    fields: toArray(collection.fields)
+      .filter(field => typeof field === 'object' && field !== null)
+      .map(toContextField),
+    actions: toArray(collection.actions)
+      .filter(action => readModel.isActionAllowed(collection.name, action.name))
```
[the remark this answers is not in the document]
Tonours: The finding is real, and I reproduced it: with two same-named actions on a collection, `isActionAllowed` is keyed by name only (`read-model.ts:179`), so both pass and `/context` emits two entries — while `collectActions` (`collect-unfolding.ts:170`) iterates the endpoint map's *keys*, so the OpenAPI document declares one. The two surfaces disagree.

The suggested filter closes one half but not the other: with two twins that both carry an endpoint, `Boolean(action.endpoint) && isActionAllowed(...)` still emits both, and the document still declares one. I checked that by running it.

Fixed in `7b7015d` by matching the exact allowed identity instead — the entry whose `id` is the one the read-model kept:

```ts
const allowed = allowedActions[action?.name];

return allowed !== undefined && allowed.id === action.id;
```

That resolves both cases, and it makes `/context` advertise precisely the action that `/actions/{name}/execute` will route to, since every other consumer keys by `(collection, name)` too. Note the `allowed !== undefined` check is load-bearing: a primitive entry in the array would make `allowedActions[undefined]?.id === action?.id` compare `undefined` to `undefined` and pass.

Two fixture pairs and two tests cover it — endpoint-less twin dropped, and both-endpointed resolving to the one the read-model kept.

nbouliol (approved):

PR #365, opened 2022-07-04
feat: allow configuring field validators during agent customization

romain-gilliotte:
### General

- [ ] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org)
- [ ] Test manually the implemented changes
- [ ] Validate the code quality (indentation, syntax, style, simplicity, readability)
- [ ] Ensure that Types have been updated according to your changes (if needed)

### Security

- [ ] Consider the security impact of the changes made

PR #1840, opened 2026-08-20
fix(agent): serve only the columns of collections the caller may read

PMerlet:
## Why

A read is permission-checked on the **root collection only**. Every column a projection, a filter or a sort reaches *through a relation path* is served with no check on the collection it comes from.

A role with `read` on `cards` and nothing at all on `holders` gets this in full:

```
GET /forest/cards
Forest-Projection: id,holder:nationalId,holder:dateOfBirth
```

The header is not even required — `GET /forest/cards` with no `fields[]` returns the same columns, because `ProjectionFactory.all` expands every column of every to-one relation.

The filter is the sharper half. It never returns the column, yet it answers one guess per request:

```
GET /forest/cards?filters={"field":"holder:nationalId","operator":"starts_with","value":"1850"}
```

One row back or zero rows back is one digit. Ten iterations per character recovers a national id in full, from a collection with zero granted permissions, without the value ever appearing in a response.

fixes PRD-900

## The rule

**Check the collection each path *ends* on.** Collections crossed on the way confer and require nothing — reaching one through a relation is a join, not a read — so `account:organization:name` needs `read` on `organizations` alone, and a ManyToMany through-collection stays out of it since it contributes no returned column.

What a denial does depends on **who asked for the field**:

| | |
|---|---|
| Named by the caller (`fields[]`, `Forest-Projection`) | **Refused**, listing every offending path in one message so a client drops them all and retries once |
| Never asked for (the `ProjectionFactory.all` default) | **Dropped from the projection** — refusing would turn an ordinary listing into a 403 |

Filters, sorts, extended searches and a chart's group-by or aggregated field are **always refused**. None has a prunable equivalent: dropping a condition widens the result set, dropping a sort clause silently reorders it, and a grouped-by key is chart output.

The check reads the caller's own query only. Scopes and segments are injected by the agent and may legitimately reference a collection the caller cannot read — a test locks that down.

## One case the rule does not describe

A leaderboard counting a relation has no aggregate field, so no path traces back to the collection being counted. What it exposes is that collection's cardinality, which is what `browse` already governs on `/forest/<collection>/count` and `/relationships/<name>/count` — so that is what gets asserted, on the foreign collection rather than on the through-collection a ManyToMany aggregates.

Found by an adversarial review pass on the first revision of this branch, not by the original implementation.

## Routes covered

`get`, `list`, `csv`, `list-related`, `csv-related`, `count`, `count-related`, the chart routes, and `update`.

`update` is a write, and in scope anyway. `PUT /forest/<collection>/:id` re-lists the row it just wrote with `ProjectionFactory.all` and serializes the result, so a role with `edit` on `cards` and no `read` on `holders` got `holder.nationalId` back from a no-op update — the same disclosure as `GET /forest/cards/:id`, and `edit` almost always comes with it. The projection there is entirely agent-chosen, so it is **redacted, never refused**: a write must not 403 because the row happens to carry a relation the caller cannot read.

The filter-shaped half on the destructive routes — `delete.ts:45` and `dissociate-delete-related.ts:122` accept the caller's `filters` on the same denied paths — is deliberately out of scope and split to PRD-1012. A delete answers `204` whatever it matched, so there is no response oracle and each probe destroys the rows it measures, while refusing would 403 any bulk delete filtered on a belongsTo sub-field, which the filter bar allows and no front change absorbs.

## Sequencing — the front has landed

The front used to ask for these columns on every list and detail view. It stopped in **ForestAdmin/forestadmin#9914**, which prunes projections by `canReadCollection` and renders a denied belongsTo as restricted — **merged into `main` on 2026-08-21** (`d050ef96`). That was the blocker on this PR: merging the agent first would have turned every list or detail view showing a belongsTo on a denied collection into a 403.

One path the front does not cover, by design: chart requests. See the dashboard leaderboard note under *Residual cost*.

## Residual cost

Roles that display a related label today without `read` on the target lose it until an admin grants it. One permission sweep per project, visible and diagnosable rather than silent.

**Dashboard leaderboards that count a relation break for existing roles.** A `Count` leaderboard on `holders` counting `cards`, for a role without `browse` on `cards`, now returns 403 where it returned counts — and the widget still renders, since dashboard chart visibility follows the chart's own collection, so it shows an error rather than disappearing. ForestAdmin/forestadmin#9914 prunes record, list and export projections and does not touch chart requests, so the sequencing note below does not cover this path. It needs the same permission sweep as the rest, on `browse` rather than `read`.

**A projected path that reaches through a collection the caller cannot read serves that collection's id.** The leaf-only rule grants `account:organization:name` on `read` of `organizations` alone, and JSON:API renders the traversed `accounts` record as a resource of its own — an identity in `relationships`, an entry in `included` — so the leaf column cannot be served without it. Its id and nothing more: no column of the traversed collection reaches the payload. Dropping the key instead is not an option, because `withPks` is what supplies the serializer's precondition — a relation with no identity raises `Missing one of expected fields` rather than redacting, a 500 where there was a 200. A OneToOne, or a `foreignKeyTarget` that is not the primary key, is the sharp end: the id is one the root row does not carry. Three cases in `test/security/related-read-permissions.test.ts` pin it, including the payload bound. Raised in review on this branch, where it was diagnosed in a comment and pinned by nothing.

**A `replaceSearch` handler is exempt on a plain search and refused on an extended one.** `getSearchedFields` answers `null` when a handler is installed, and the deciding question is whether the caller can aim. On a plain search they aimed at nothing — the footprint is the one the customer chose, identical for every role, so it is served, the same category as a scope, and refusing would remove search from every customized collection. The extended flag is theirs, though: the same term with it off and then on differs by exactly the rows matched through a relation, a bit per term over collections no check covered, so an unverifiable traversal is refused. Decided in PRD-1024, which applies to node unchanged. **The cost is a 403 where there was a 200** for a customer using `replaceSearch` on a collection where users tick extended search. The residual on the plain half is accepted and now stated in `replaceSearch`'s own JSDoc: a handler searching a column the role cannot read gives the caller the same rows-or-no-rows oracle a filter would, and the agent cannot tell that apart from what the customer intended.

Any client requesting relation fields its role may not read — a customer script, an `agent-client` integration, an MCP tool — starts getting a 403 where it got a 200. That is the fix working: only integrations running under a role that should never have had the data are affected, and the error names the field and the collection so they can be fixed rather than guessed at.

## Not in scope

- `browse` still stands in for a denied `read` when the front resolves a get-one through the list route — tracked in PRD-990.
- On `instantCacheRefresh: false`, each denied check clears and refetches the whole permission cache; denials are routine on the redaction path, so that is one permission fetch per read — tracked in PRD-1002, to be fixed in `forestadmin-client` rather than here.
- Nothing else known. The extended-search sweep no longer derives its own set: `getSearchedFields` walks down the stack and the search decorator answers from `childCollection`, so a field hidden by `.removeField` is still checked, and a replaced search — whose footprint the layer cannot state — is served plain and refused extended, see *Residual cost*.
- Whether a ManyToMany through-collection needs `read` of its own is left open, as PRD-900 states.

## Tests

`test/security/related-read-permissions.test.ts` builds the ticket's schema and role, and each case fails against the previous implementation — verified by reverting the guard, not assumed.

> [!NOTE]
> ### Enforce read permissions on query fields and projections across agent routes
> - Introduces `AuthorizationService.redactProjection` to drop or reject projection fields whose owning related collections the caller cannot read; implicit projections drop fields, explicit field lists return 403
> - Adds `AuthorizationService.assertCanReadQueryFields` to reject filter/sort/search terms that reference fields from unreadable related collections, applied to list, get, count, csv, chart, and related routes
> - Adds `FieldPathUtils.getLeafCollection` to resolve the owning collection of a colon-separated field path, and `SearchCollectionDecorator.getSearchedFields` to surface which fields an extended search will read
> - Updates `CsvGenerator.filterHeader` to keep CSV header labels aligned with redacted columns; `UpdateRoute.handleUpdate` now redacts unreadable fields from the post-update response
> - Risk: callers that previously received data from unreadable related collections via implicit projections or filters now get redacted results or 403s — check `redactProjection` in [authorization.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1840/files#diff-9103e7358427994d7727cf4f652032c5d2d47448022946adda46c493a403c686) and route handlers in [packages/agent/src/routes/access/](https://github.com/ForestAdmin/agent-nodejs/pull/1840/files#diff-9ff1c82c8599c82b6c6abae9421ab17e7978d93a5f0469cfe46ead771c137edf) for regressions in existing API consumers
>
> 
> #### Changes since #1840 opened
>
> - Renamed the `explicit` property to `namedByCaller` in the `RequestedProjection` type and updated all usages throughout `AuthorizationService.redactProjection`, `QueryStringParser.parseProjectionFromHeaderOrQuery`, `UpdateRoute.handleUpdate`, and associated tests [680ec65]
> - Updated an explanatory comment in the `related-read-permissions.test` test suite [aeb3c3c]
> - Extended `AuthorizationService.assertCanReadQueryFields` method to accept optional `consumes` parameter of type `QueryComponent[]` with default value `['filter','sort','search']`, and added exported `QueryComponent` type union and `ALL_QUERY_COMPONENTS` constant [9385592]
> - Updated `ChartRoute.makeChart`, `CountRelatedRoute.handleCountRelated`, and `CountRoute.handleCount` handlers to pass `['filter','search']` as third argument to `assertCanReadQueryFields` [9385592]
> - Added test case verifying count route ignores sort parameter on unreadable related fields [9385592]
> - Added test cases verifying that traversing a OneToOne relation on an unreadable collection correctly projects the relation key and serializes only the key and nested readable fields [e3e61ec]
> - Extended test data source schema with a OneToOne relation from `cards` to a new `contracts` collection and added the `contracts` collection with a ManyToOne relation to `organizations` [e3e61ec]
> - Added `RecordData` type import from `@forestadmin/datasource-toolkit` and `Serializer` import from local services [e3e61ec]
> - Updated test comments clarifying the requirement to include `account:id` in projection for serialization [e3e61ec]
> - Added extended search authorization check in `AuthorizationService.search` that throws `ForbiddenError` when searched fields cannot be determined [f3c2de6]
> - Added test coverage for extended search authorization with unknown field footprints [f3c2de6]
> - Expanded JSDoc documentation in `CollectionCustomizer.replaceSearch` to clarify security semantics of plain versus extended searches [f3c2de6]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2e3e7b5.</sup>
> 
>

### DIFF
--- a/packages/agent/src/routes/access/chart.ts
+++ b/packages/agent/src/routes/access/chart.ts
@@ -1,5 +1,6 @@
 import type {
   Caller,
+  Collection,
   ConditionTreeBranch,
   DateOperation,
   Filter,
@@ -29,6 +30,7 @@ import { DateTime } from 'luxon';
 import { v1 as uuidv1 } from 'uuid';
 
 import ContextFilterFactory from '../../utils/context-filter-factory';
+import FieldPathUtils from '../../utils/field-path';
 import QueryStringParser from '../../utils/query-string';
 import CollectionRoute from '../collection-route';
 
@@ -66,6 +68,8 @@ export default class ChartRoute extends CollectionRoute {
       renderingId,
       chartRequest,
     });
+
+    await this.services.authorization.assertCanReadQueryFields(context, this.collection);
 
     switch (chartRequest.type) {
       case ChartType.Value:
@@ -120,6 +124,11 @@ export default class ChartRoute extends CollectionRoute {
       aggregateFieldName: aggregateField,
     } = <PieChart>context.request.body;
 
+    await this.assertCanReadAggregatedFields(context, this.collection, [
+      ['group a chart by', groupByField],
+      ['aggregate a chart on', aggregateField],
+    ]);
+
     const rows = await this.collection.aggregate(
       QueryStringParser.parseCaller(context),
       await this.getFilter(context),
@@ -143,6 +152,11 @@ export default class ChartRoute extends CollectionRoute {
       groupByFieldName: groupByDateField,
       timeRange,
     } = <LineChart>context.request.body;
+
+    await this.assertCanReadAggregatedFields(context, this.collection, [
+      ['group a chart by', groupByDateField],
+      ['aggregate a chart on', aggregateField],
+    ]);
 
     const filter = await this.getFilter(context);
     const filterOnlyWithValues = filter.override({
@@ -233,9 +247,25 @@ export default class ChartRoute extends CollectionRoute {
     }
 
     if (collection && filter && aggregation) {
-      const rows = await this.dataSource
-        .getCollection(collection)
-        .aggregate(QueryStringParser.parseCaller(context), filter, aggregation, Number(body.limit));
+      const aggregatedCollection = this.dataSource.getCollection(collection);
+
+      await this.assertCanReadAggregatedFields(context, aggregatedCollection, [
+        ['group a leaderboard by', aggregation.groups[0].field],
+        ['aggregate a leaderboard on', aggregation.field],
+      ]);
+
+      // A count exposes the cardinality of the relation, which `/relationships/<name>/count` puts
+      // behind `browse`.
+      if (!aggregation.field) {
+        await this.services.authorization.assertCanBrowse(context, field.foreignCollection);
+      }
+
+      const rows = await aggregatedCollection.aggregate(
+        QueryStringParser.parseCaller(context),
+        filter,
+        aggregation,
+        Number(body.limit),
+      );
 
       return rows.map(row => ({
         key: row.group[aggregation.groups[0].field] as string,
@@ -254,6 +284,10 @@ export default class ChartRoute extends CollectionRoute {
     );
     const aggregation = new Aggregation({ operation: aggregator, field: aggregateField });
 
+    await this.assertCanReadAggregatedFields(context, this.collection, [
+      ['aggregate a chart on', aggregateField],
+    ]);
+
     const rows = await this.collection.aggregate(
       QueryStringParser.parseCaller(context),
       filter,
@@ -261,6 +295,24 @@ export default class ChartRoute extends CollectionRoute {
     );
 
     return rows.length ? (rows[0].value as number) : 0;
+  }
+
+  private async assertCanReadAggregatedFields(
+    context: Context,
+    collection: Collection,
+    fields: Array<[action: string, path: string]>,
+  ): Promise<void> {
+    await this.services.authorization.assertCanReadUsages(
+      context,
+      this.collection.name,
+      fields
+        .filter(([, path]) => path)
+        .map(([action, path]) => ({
+          action,
+          path,
+          collectionName: FieldPathUtils.getLeafCollection(collection, path).name,
+        })),
+    );
   }
 
   private async getFilter(context: Context): Promise<Filter> {
--- a/packages/agent/src/routes/access/chart.ts
+++ b/packages/agent/src/routes/access/chart.ts
@@ -69,7 +69,10 @@ export default class ChartRoute extends CollectionRoute {
       chartRequest,
     });
 
-    await this.services.authorization.assertCanReadQueryFields(context, this.collection);
+    await this.services.authorization.assertCanReadQueryFields(context, this.collection, [
+      'filter',
+      'search',
+    ]);
 
     switch (chartRequest.type) {
       case ChartType.Value:
--- a/packages/agent/src/routes/access/chart.ts
+++ b/packages/agent/src/routes/access/chart.ts
@@ -299,7 +299,7 @@ export default class ChartRoute extends CollectionRoute {
 
   private async assertCanReadAggregatedFields(
     context: Context,
-    collection: Collection,
+    pathCollection: Collection,
     fields: Array<[action: string, path: string]>,
   ): Promise<void> {
     await this.services.authorization.assertCanReadUsages(
@@ -310,7 +310,7 @@ export default class ChartRoute extends CollectionRoute {
         .map(([action, path]) => ({
           action,
           path,
-          collectionName: FieldPathUtils.getLeafCollection(collection, path).name,
+          collectionName: FieldPathUtils.getLeafCollection(pathCollection, path).name,
         })),
     );
   }
--- a/packages/agent/src/routes/modification/update.ts
+++ b/packages/agent/src/routes/modification/update.ts
@@ -40,7 +40,10 @@ export default class UpdateRoute extends CollectionRoute {
     const [updateResult] = await this.collection.list(
       caller,
       new Filter({ conditionTree }),
-      ProjectionFactory.all(this.collection),
+      await this.services.authorization.redactProjection(context, this.collection, {
+        projection: ProjectionFactory.all(this.collection),
+        explicit: false,
+      }),
     );
 
     context.response.body = this.services.serializer.serialize(this.collection, updateResult);
--- a/packages/agent/src/routes/modification/update.ts
+++ b/packages/agent/src/routes/modification/update.ts
@@ -42,7 +42,7 @@ export default class UpdateRoute extends CollectionRoute {
       new Filter({ conditionTree }),
       await this.services.authorization.redactProjection(context, this.collection, {
         projection: ProjectionFactory.all(this.collection),
-        explicit: false,
+        namedByCaller: false,
       }),
     );
 
--- a/packages/agent/src/services/authorization/authorization.ts
+++ b/packages/agent/src/services/authorization/authorization.ts
@@ -1,8 +1,9 @@
+import type { RequestedProjection } from '../../utils/query-string';
 import type { Collection, ConditionTree } from '@forestadmin/datasource-toolkit';
 import type { ForestAdminClient } from '@forestadmin/forestadmin-client';
 import type { Context } from 'koa';
 
-import { UnprocessableError } from '@forestadmin/datasource-toolkit';
+import { ForbiddenError, Projection, UnprocessableError } from '@forestadmin/datasource-toolkit';
 import {
   ChainedSQLQueryError,
   CollectionActionEvent,
@@ -12,6 +13,10 @@ import {
 
 import { HttpCode } from '../../types';
 import ConditionTreeParser from '../../utils/condition-tree-parser';
+import FieldPathUtils from '../../utils/field-path';
+import QueryStringParser from '../../utils/query-string';
+
+type FieldUsage = { action: string; path: string; collectionName: string };
 
 export default class AuthorizationService {
   constructor(private readonly forestAdminClient: ForestAdminClient) {}
@@ -38,6 +43,121 @@ export default class AuthorizationService {
 
   public async assertCanExport(context: Context, collectionName: string) {
     await this.assertCanOnCollection(CollectionActionEvent.Export, context, collectionName);
+  }
+
+  public async canRead(context: Context, collectionName: string): Promise<boolean> {
+    return this.forestAdminClient.permissionService.canOnCollection({
+      userId: context.state.user.id,
+      event: CollectionActionEvent.Read,
+      collectionName,
+    });
+  }
+
+  /**
+   * An unnamed field is redacted rather than refused: `ProjectionFactory.all` expands every column
+   * of every to-one relation when no `fields[]` is sent, so refusing would turn an ordinary
+   * listing into a 403.
+   */
+  public async redactProjection(
+    context: Context,
+    collection: Collection,
+    requested: RequestedProjection,
+  ): Promise<Projection> {
+    const { projection, explicit } = requested;
+    const owners = projection.map(path => FieldPathUtils.getLeafCollection(collection, path).name);
+    const permissions = await this.getReadPermissions(context, collection.name, owners);
+    const isReadable = (index: number) => permissions.get(owners[index]);
+
+    if (explicit) {
+      const denied = projection
+        .map((field, index) => ({ field, collection: owners[index] }))
+        .filter((_, index) => !isReadable(index));
+
+      if (denied.length) {
+        const fields = denied
+          .map(({ field, collection: name }) => `'${field}' from the '${name}' collection`)
+          .join(', ');
+
+        throw new ForbiddenError(`You are not allowed to read ${fields}.`);
+      }
+    }
+
+    return new Projection(...projection.filter((_, index) => isReadable(index)));
+  }
+
+  /**
+   * Refused rather than redacted: dropping a condition widens the result set and dropping a sort
+   * clause silently reorders it, while both leak the value they touch anyway — a `starts_with`
+   * filter answers one guess per request without returning a column of its own.
+   */
+  public async assertCanReadQueryFields(context: Context, collection: Collection): Promise<void> {
+    const usages: FieldUsage[] = [];
+    const push = (action: string, path: string) =>
+      usages.push({
+        action,
+        path,
+        collectionName: FieldPathUtils.getLeafCollection(collection, path).name,
+      });
+
+    QueryStringParser.parseConditionTree(collection, context)?.forEachLeaf(leaf =>
+      push('filter on', leaf.field),
+    );
+
+    for (const { field } of QueryStringParser.parseSort(collection, context)) {
+      push('sort on', field);
+    }
+
+    if (
+      QueryStringParser.parseSearch(collection, context) &&
+      QueryStringParser.parseSearchExtended(context)
+    ) {
+      for (const [name, field] of Object.entries(collection.schema.fields)) {
+        if (field.type === 'ManyToOne' || field.type === 'OneToOne') {
+          usages.push({
+            action: 'run an extended search through',
+            path: name,
+            collectionName: field.foreignCollection,
+          });
+        }
+      }
+    }
+
+    await this.assertCanReadUsages(context, collection.name, usages);
+  }
+
+  public async assertCanReadUsages(
+    context: Context,
+    rootCollectionName: string,
+    usages: FieldUsage[],
+  ): Promise<void> {
+    const permissions = await this.getReadPermissions(
+      context,
+      rootCollectionName,
+      usages.map(usage => usage.collectionName),
+    );
+    const denied = usages.find(usage => !permissions.get(usage.collectionName));
+
+    if (denied) {
+      throw new ForbiddenError(
+        `You cannot ${denied.action} '${denied.path}': you are not allowed to read the ` +
+          `'${denied.collectionName}' collection.`,
+      );
+    }
+  }
+
+  /** The root is skipped: its own route already asserts `browse` on a listing, `read` on a get. */
+  private async getReadPermissions(
+    context: Context,
+    rootCollectionName: string,
+    collectionNames: string[],
+  ): Promise<Map<string, boolean>> {
+    const toCheck = [...new Set(collectionNames)].filter(name => name !== rootCollectionName);
+    const allowed = await Promise.all(toCheck.map(name => this.canRead(context, name)));
+
+    return new Map<string, boolean>([
+      [rootCollectionName, true],
+      ...toCheck.map((name, index): [string, boolean] => [name, allowed[index]]),
+    ]);
   }
 
   private async assertCanOnCollection(
--- a/packages/agent/src/services/authorization/authorization.ts
+++ b/packages/agent/src/services/authorization/authorization.ts
@@ -151,7 +151,7 @@ export default class AuthorizationService {
     }
   }
 
-  /** The root is skipped: its own route already asserts `browse` on a listing, `read` on a get. */
+  /** The root is skipped: `browse` gates a listing, `read` a get, and the signed hash a chart. */
   private async getReadPermissions(
     context: Context,
     rootCollectionName: string,
--- a/packages/agent/src/services/authorization/authorization.ts
+++ b/packages/agent/src/services/authorization/authorization.ts
@@ -133,11 +133,21 @@ export default class AuthorizationService {
 
     if (search) {
       // Asked whatever the extended flag: `relation.column:term` is end-user syntax and reaches a
-      // relation without it. An unknown answer serves the request — the `replaceSearch` exemption.
-      const searched = (collection as CollectionDecorator).getSearchedFields?.(
-        search,
-        QueryStringParser.parseSearchExtended(context),
-      );
+      // relation without it.
+      const extended = QueryStringParser.parseSearchExtended(context);
+      const searched = (collection as CollectionDecorator).getSearchedFields?.(search, extended);
+
+      // An unknown footprint is a `replaceSearch` handler choosing its own fields. On a plain
+      // search the caller aimed at nothing, so it is served — the same category as a scope, and
+      // refusing would remove search from every customized collection. The extended flag is the
+      // caller's own, though: running one term both ways isolates the rows matched through a
+      // relation, a bit per term on collections no check covered. The exemption stops there.
+      if (!searched && extended) {
+        throw new ForbiddenError(
+          `You cannot run an extended search on the '${collection.name}' collection: the fields ` +
+            'it reaches cannot be determined, so they cannot be checked against your permissions.',
+        );
+      }
 
       for (const { path, collection: name } of searched ?? []) {
         usages.push({ action: 'search on', path, collectionName: name });
--- a/packages/agent/src/services/authorization/authorization.ts
+++ b/packages/agent/src/services/authorization/authorization.ts
@@ -67,12 +67,12 @@ export default class AuthorizationService {
     collection: Collection,
     requested: RequestedProjection,
   ): Promise<Projection> {
-    const { projection, explicit } = requested;
+    const { projection, namedByCaller } = requested;
     const owners = projection.map(path => FieldPathUtils.getLeafCollection(collection, path).name);
     const permissions = await this.getReadPermissions(context, collection.name, owners);
     const isReadable = (index: number) => permissions.get(owners[index]);
 
-    if (explicit) {
+    if (namedByCaller) {
       const denied = projection
         .map((field, index) => ({ field, collection: owners[index] }))
         .filter((_, index) => !isReadable(index));
@@ -114,11 +114,8 @@ export default class AuthorizationService {
     const search = QueryStringParser.parseSearch(collection, context);
 
     if (search) {
-      // Asked of the stack rather than derived from the schema: `relation.column:term` is end-user
-      // syntax that needs no extended search, and the fields an extended one reaches are read below
-      // the publication and renaming layers. A `null` answer means the collection cannot say — a
-      // replaced search, where the customer's handler picks the fields and the caller only supplies
-      // the text.
+      // Asked whatever the extended flag: `relation.column:term` is end-user syntax and reaches a
+      // relation without it. An unknown answer serves the request — the `replaceSearch` exemption.
       const searched = (collection as CollectionDecorator).getSearchedFields?.(
         search,
         QueryStringParser.parseSearchExtended(context),
--- a/packages/agent/src/services/authorization/authorization.ts
+++ b/packages/agent/src/services/authorization/authorization.ts
@@ -22,6 +22,10 @@ import QueryStringParser from '../../utils/query-string';
 
 type FieldUsage = { action: string; path: string; collectionName: string };
 
+export type QueryComponent = 'filter' | 'sort' | 'search';
+
+const ALL_QUERY_COMPONENTS: QueryComponent[] = ['filter', 'sort', 'search'];
+
 export default class AuthorizationService {
   constructor(private readonly forestAdminClient: ForestAdminClient) {}
 
@@ -93,8 +97,16 @@ export default class AuthorizationService {
    * Refused rather than redacted: dropping a condition widens the result set and dropping a sort
    * clause silently reorders it, while both leak the value they touch anyway — a `starts_with`
    * filter answers one guess per request without returning a column of its own.
+   *
+   * `consumes` names the query components the calling route applies to its filter. Checking one it
+   * drops would refuse a request the denied field cannot reach: `ContextFilterFactory.build` carries
+   * no sort, which only `buildPaginated` adds.
    */
-  public async assertCanReadQueryFields(context: Context, collection: Collection): Promise<void> {
+  public async assertCanReadQueryFields(
+    context: Context,
+    collection: Collection,
+    consumes: QueryComponent[] = ALL_QUERY_COMPONENTS,
+  ): Promise<void> {
     const usages: FieldUsage[] = [];
     const push = (action: string, path: string) =>
       usages.push({
@@ -103,15 +115,21 @@ export default class AuthorizationService {
         collectionName: FieldPathUtils.getLeafCollection(collection, path).name,
       });
 
-    QueryStringParser.parseConditionTree(collection, context)?.forEachLeaf(leaf =>
-      push('filter on', leaf.field),
-    );
-
-    for (const { field } of QueryStringParser.parseSort(collection, context)) {
-      push('sort on', field);
+    if (consumes.includes('filter')) {
+      QueryStringParser.parseConditionTree(collection, context)?.forEachLeaf(leaf =>
+        push('filter on', leaf.field),
+      );
     }
 
-    const search = QueryStringParser.parseSearch(collection, context);
+    if (consumes.includes('sort')) {
+      for (const { field } of QueryStringParser.parseSort(collection, context)) {
+        push('sort on', field);
+      }
+    }
+
+    const search = consumes.includes('search')
+      ? QueryStringParser.parseSearch(collection, context)
+      : null;
 
     if (search) {
       // Asked whatever the extended flag: `relation.column:term` is end-user syntax and reaches a
--- a/packages/agent/src/services/authorization/authorization.ts
+++ b/packages/agent/src/services/authorization/authorization.ts
@@ -3,6 +3,7 @@ import type { Collection, ConditionTree } from '@forestadmin/datasource-toolkit'
 import type { ForestAdminClient } from '@forestadmin/forestadmin-client';
 import type { Context } from 'koa';
 
+import { getSearchedFieldPaths } from '@forestadmin/datasource-customizer';
 import { ForbiddenError, Projection, UnprocessableError } from '@forestadmin/datasource-toolkit';
 import {
   ChainedSQLQueryError,
@@ -107,17 +108,22 @@ export default class AuthorizationService {
       push('sort on', field);
     }
 
-    if (
-      QueryStringParser.parseSearch(collection, context) &&
-      QueryStringParser.parseSearchExtended(context)
-    ) {
-      for (const [name, field] of Object.entries(collection.schema.fields)) {
-        if (field.type === 'ManyToOne' || field.type === 'OneToOne') {
-          usages.push({
-            action: 'run an extended search through',
-            path: name,
-            collectionName: field.foreignCollection,
-          });
+    const search = QueryStringParser.parseSearch(collection, context);
+
+    if (search) {
+      // `relation.column:term` is end-user search syntax and works without extended search, so the
+      // paths it names are resolved by the search decorator's own resolver rather than guessed at.
+      for (const path of getSearchedFieldPaths(collection, search)) push('search on', path);
+
+      if (QueryStringParser.parseSearchExtended(context)) {
+        for (const [name, field] of Object.entries(collection.schema.fields)) {
+          if (field.type === 'ManyToOne' || field.type === 'OneToOne') {
+            usages.push({
+              action: 'run an extended search through',
+              path: name,
+              collectionName: field.foreignCollection,
+            });
+          }
         }
       }
     }
--- a/packages/agent/src/services/authorization/authorization.ts
+++ b/packages/agent/src/services/authorization/authorization.ts
@@ -1,9 +1,12 @@
 import type { RequestedProjection } from '../../utils/query-string';
-import type { Collection, ConditionTree } from '@forestadmin/datasource-toolkit';
+import type {
+  Collection,
+  CollectionDecorator,
+  ConditionTree,
+} from '@forestadmin/datasource-toolkit';
 import type { ForestAdminClient } from '@forestadmin/forestadmin-client';
 import type { Context } from 'koa';
 
-import { getSearchedFieldPaths } from '@forestadmin/datasource-customizer';
 import { ForbiddenError, Projection, UnprocessableError } from '@forestadmin/datasource-toolkit';
 import {
   ChainedSQLQueryError,
@@ -111,20 +114,18 @@ export default class AuthorizationService {
     const search = QueryStringParser.parseSearch(collection, context);
 
     if (search) {
-      // `relation.column:term` is end-user search syntax and works without extended search, so the
-      // paths it names are resolved by the search decorator's own resolver rather than guessed at.
-      for (const path of getSearchedFieldPaths(collection, search)) push('search on', path);
+      // Asked of the stack rather than derived from the schema: `relation.column:term` is end-user
+      // syntax that needs no extended search, and the fields an extended one reaches are read below
+      // the publication and renaming layers. A `null` answer means the collection cannot say — a
+      // replaced search, where the customer's handler picks the fields and the caller only supplies
+      // the text.
+      const searched = (collection as CollectionDecorator).getSearchedFields?.(
+        search,
+        QueryStringParser.parseSearchExtended(context),
+      );
 
-      if (QueryStringParser.parseSearchExtended(context)) {
-        for (const [name, field] of Object.entries(collection.schema.fields)) {
-          if (field.type === 'ManyToOne' || field.type === 'OneToOne') {
-            usages.push({
-              action: 'run an extended search through',
-              path: name,
-              collectionName: field.foreignCollection,
-            });
-          }
-        }
+      for (const { path, collection: name } of searched ?? []) {
+        usages.push({ action: 'search on', path, collectionName: name });
       }
     }
 
--- a/packages/agent/test/__factories__/authorization/authorization.ts
+++ b/packages/agent/test/__factories__/authorization/authorization.ts
@@ -15,6 +15,12 @@ export class AuthorizationsFactory extends Factory<AuthorizationService> {
       Authorizations.getScope = jest.fn();
       Authorizations.assertCanExecuteChart = jest.fn();
       Authorizations.invalidateScopeCache = jest.fn();
+      Authorizations.canRead = jest.fn().mockResolvedValue(true);
+      Authorizations.assertCanReadQueryFields = jest.fn();
+      Authorizations.assertCanReadUsages = jest.fn();
+      Authorizations.redactProjection = jest
+        .fn()
+        .mockImplementation(async (context, collection, requested) => requested.projection);
     });
   }
 }
--- a/packages/agent/test/security/related-read-permissions.test.ts
+++ b/packages/agent/test/security/related-read-permissions.test.ts
@@ -281,6 +281,22 @@ describe('read permissions on related collections', () => {
   });
 
   describe('sort', () => {
+    // A count applies no sort — `ContextFilterFactory.build` carries none, only `buildPaginated`
+    // does — so refusing one would refuse a request the denied field cannot reach.
+    it('should ignore a sort on the count route, which never applies one', async () => {
+      const dataSource = buildDataSource();
+      const services = buildServices();
+      const aggregate = jest
+        .spyOn(dataSource.getCollection('cards'), 'aggregate')
+        .mockResolvedValue([{ value: 3, group: {} }]);
+
+      await new Count(services, options, dataSource, 'cards').handleCount(
+        buildContext({ query: { sort: '-holder.nationalId' } }),
+      );
+
+      expect(aggregate).toHaveBeenCalled();
+    });
+
     it('should refuse a sort reading a collection the caller cannot read', async () => {
       const dataSource = buildDataSource();
       const services = buildServices();
--- a/packages/agent/test/security/related-read-permissions.test.ts
+++ b/packages/agent/test/security/related-read-permissions.test.ts
@@ -293,6 +293,45 @@ describe('read permissions on related collections', () => {
       ).rejects.toThrow('you are not allowed to read the');
     });
 
+    it('should refuse a search naming a relation column through the dot syntax', async () => {
+      const dataSource = buildDataSource();
+      const services = buildServices();
+      const aggregate = jest.spyOn(dataSource.getCollection('cards'), 'aggregate');
+
+      await expect(
+        new Count(services, options, dataSource, 'cards').handleCount(
+          buildContext({ query: { search: 'holder.nationalId:1850' } }),
+        ),
+      ).rejects.toThrow(
+        "You cannot search on 'holder:nationalId': you are not allowed to read the " +
+          "'holders' collection.",
+      );
+
+      expect(aggregate).not.toHaveBeenCalled();
+    });
+
+    it('should refuse a search reaching a denied collection across a to-many relation', async () => {
+      const dataSource = buildDataSource();
+      const forestAdminClient = factories.forestAdminClient.build();
+      const services = factories.forestAdminHttpDriverServices.build();
+      services.authorization = new AuthorizationService(forestAdminClient);
+      services.serializer.serializeWithSearchMetadata = jest.fn();
+
+      // Searching from `holders`, so `cards` is the denied one here.
+      (forestAdminClient.permissionService.canOnCollection as jest.Mock).mockImplementation(
+        ({ collectionName }) => collectionName !== 'cards',
+      );
+
+      await expect(
+        new List(services, options, dataSource, 'holders').handleList(
+          buildContext({ query: { search: 'cards.panLast4:4242' } }),
+        ),
+      ).rejects.toThrow(
+        "You cannot search on 'cards:panLast4': you are not allowed to read the " +
+          "'cards' collection.",
+      );
+    });
+
     it('should accept a plain search, which never leaves the root collection', async () => {
       const dataSource = buildDataSource();
       const services = buildServices();
--- a/packages/agent/test/security/related-read-permissions.test.ts
+++ b/packages/agent/test/security/related-read-permissions.test.ts
@@ -349,7 +349,10 @@ describe('read permissions on related collections', () => {
     it('should accept a plain search, which never leaves the root collection', async () => {
       const dataSource = buildDataSource();
       const services = buildServices();
-      const list = jest.spyOn(dataSource.getCollection('cards'), 'list').mockResolvedValue([]);
+      const cards = dataSource.getCollection('cards') as CollectionDecorator;
+      const list = jest.spyOn(cards, 'list').mockResolvedValue([]);
+
+      cards.getSearchedFields = () => [{ path: 'panLast4', collection: 'cards' }];
 
       await new List(services, options, dataSource, 'cards').handleList(
         buildContext({ query: { search: 'martin' } }),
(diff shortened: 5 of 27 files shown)


packages/agent/src/services/authorization/authorization.ts
```diff
@@ -40,6 +45,121 @@ export default class AuthorizationService {
...
+
+    QueryStringParser.parseConditionTree(collection, context)?.forEachLeaf(leaf =>
+      push('filter on', leaf.field),
+    );
+
+    for (const { field } of QueryStringParser.parseSort(collection, context)) {
+      push('sort on', field);
+    }
+
+    if (
```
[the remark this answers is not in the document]
PMerlet: The mechanism is real, but this is an exemption rather than a gap, and the proposed fix would reintroduce a bug a test guards against.

`replaceSearch` is server-side code the customer wrote, in the same category as a scope, a segment or a computed field. The caller controls the search *string*, not the fields it maps to — unlike `filters={"field":"holder:nationalId"}`, they cannot aim at a column of their choosing. That is the line the check draws: it reads the caller's own query, and nothing the agent or the customer injects into the filter afterwards.

Authorizing the refined condition tree would cross that line in the wrong direction. By the time `SearchCollectionDecorator.refineFilter` has run, the scope has already been intersected into the same tree, so checking it against the caller's permissions 403s a user out of their own collection whenever a legitimate scope references a related column — a scope on `holder:region`, for instance. `test/security/related-read-permissions.test.ts` covers exactly that case ("should not check the scope, which the agent injects rather than the caller").

The residual risk is real and accepted: a handler that searches a column the role cannot read hands the caller the same rows-or-no-rows oracle a filter would, without them naming the field. The agent cannot distinguish that from what the customer intended, and refusing it would break legitimate `replaceSearch` handlers for the same reason refusing a scope would. Recorded as a deliberate exemption in PRD-900 so the Ruby implementation does not resolve it the other way.

packages/agent/src/services/authorization/authorization.ts
```diff
@@ -40,6 +45,121 @@ export default class AuthorizationService {
...
+      QueryStringParser.parseSearch(collection, context) &&
+      QueryStringParser.parseSearchExtended(context)
```
hercemer42: Claude Opus 5 (claude-opus-5): **Must fix** — the filter oracle PRD-900 read 3 describes stays open through `search`, which this guard only inspects when `searchExtended` is also set.

`GET /forest/cards/count?search=holder.nationalId:1850&timezone=UTC` returns a count for a role with `browse` on `cards` and zero permissions on `holders`. No `filters`, no `sort`, no `searchExtended` — so `assertCanReadQueryFields` pushes no usages at all and passes. Extend the value one character at a time and you read `holders.nationalId` out of a collection the caller has no `read` on; `AND` in the same string narrows it to one holder. Same on `/forest/cards`, `.csv`, the related lists and the Value/Objective charts.

`relation.childProperty:term` is documented end-user syntax (`/product/execute/browse` § Advanced search syntax) and is deliberately independent of extended search:

- `search/collection.ts:84` — `extractSpecifiedFields(parsedQuery)` runs unconditionally; `options.extended` gates only `defaultFields` at `:88`.
- `search/custom-parser/fields-query-walker.ts:9` — the name is turned into a path (`.` → `:`).
- `search/collection.ts:139-164` — `lenientGetSchema` resolves it fuzzily (`normalizeName`) through `ManyToOne`, `OneToOne` **and** `OneToMany`, at any depth.
- `search/custom-parser/condition-tree-query-walker.ts:131-155` — emits a leaf on that path alone; `build-string-field-filter.ts:9-13` prefers `IContains`, i.e. a substring oracle rather than equality.

Live on the standard SQL agent: `datasource-sequelize/.../model-to-collection-schema-converter.ts:156` sets `searchable: false`, so `SearchCollectionDecorator.refineFilter` takes the "implement search ourselves" branch — the ANTLR path is the normal one, not a fallback.

`test/security/related-read-permissions.test.ts:296` — *"should accept a plain search, which never leaves the root collection"* — pins the premise with `search: 'martin'`, which contains no `:` and never reaches the property-matching branch.

The check has to reuse the decorator's own resolution (`extractSpecifiedFields` + a `lenientGetSchema` equivalent) and assert `read` on each resolved leaf regardless of `searchExtended`. A name-equality allowlist built from the top-level schema will not match, because that resolution is fuzzy and runs below the rename decorator.

packages/agent/src/services/authorization/authorization.ts
```diff
@@ -40,6 +45,121 @@ export default class AuthorizationService {
...
+    ) {
+      for (const [name, field] of Object.entries(collection.schema.fields)) {
+        if (field.type === 'ManyToOne' || field.type === 'OneToOne') {
```
hercemer42: Claude Opus 5 (claude-opus-5): **Should fix** — this sweep enumerates the wrong set, in both directions.

**Under-inclusive, and it is a leak.** It reads `collection.schema.fields` at the top of the decorator stack — above `PublicationDataSourceDecorator` and `RenameFieldCollectionDecorator` (`decorators-stack.ts:71-72`) — while `SearchCollectionDecorator` sits below both (`:49`) and enumerates `this.childCollection` (`search/collection.ts:88,121-137`). A customer who calls `.removeField('holder')` on `cards` while `holders` stays published hides the relation from this loop but not from the search decorator, so `searchExtended=1` still emits `holder:<col>` conditions and no `read` check ever fires. It is also to-one only, while `lenientGetSchema` traverses `OneToMany` as well.

**Over-inclusive, on a collection the PR deliberately exempts.** `refineFilter` hands the string to `this.replacer` and never calls `getFields` when a `replaceSearch` is installed (`search/collection.ts:56-61`), so a customer whose custom search only ever touches `panLast4` now loses extended search entirely for any role missing `read` on any to-one target. Fail-closed, so not a leak — but it contradicts the ticket's `replaceSearch` exemption in the strict direction, and the route cannot currently tell a customized collection from a default one (`schema.searchable` is true either way).

packages/agent/src/services/authorization/authorization.ts
```diff
@@ -40,6 +45,121 @@ export default class AuthorizationService {
...
+  }
+
+  /** The root is skipped: its own route already asserts `browse` on a listing, `read` on a get. */
+  private async getReadPermissions(
+    context: Context,
+    rootCollectionName: string,
+    collectionNames: string[],
+  ): Promise<Map<string, boolean>> {
+    const toCheck = [...new Set(collectionNames)].filter(name => name !== rootCollectionName);
+    const allowed = await Promise.all(toCheck.map(name => this.canRead(context, name)));
```
hercemer42, line 187: Claude Opus 5 (claude-opus-5): **Should fix** — on agents running `instantCacheRefresh: false`, every one of these denials clears the whole environment permission cache and refetches it from the Forest API.

`canRead` → `PermissionServiceWithCache.canOnCollection` → `ActionPermissionService.can`, which passes `allowRefetch: !options.instantCacheRefresh` (`action-permission.ts:33`); on a denial `hasPermissionOrRefetch` calls `invalidateCache()` → `permissionsCache.clear()` and refetches before answering (`:49-57`, `:134-138`). The default is safe — `options-validator.ts:42` defaults `instantCacheRefresh` to `true`, so `allowRefetch` is false — but `instantCacheRefresh: false` is the documented opt-out for deployments that cannot hold the SSE connection, and it is also the only configuration where `permissionsCacheDurationInSeconds` is honoured (`options-validator.ts:66`).

Before this PR a denied `can()` was a rare event: someone attempting an action they lack. This PR makes denial the steady state — PRD-900 expects roles to be missing `read` on related collections until an admin sweeps — so one unreadable relation in a layout turns into a cache wipe plus a full permission refetch on *every* list, get, CSV and chart request, process-wide for all concurrent users. Only observable as `Invalidating roles permissions cache..` at Debug.

A read lookup that does not refetch on deny, or memoising the permission map per request, removes it.

packages/agent/src/services/authorization/authorization.ts
```diff
@@ -40,6 +45,121 @@ export default class AuthorizationService {
...
+
+    if (denied) {
+      throw new ForbiddenError(
+        `You cannot ${denied.action} '${denied.path}': you are not allowed to read the ` +
+          `'${denied.collectionName}' collection.`,
+      );
+    }
+  }
+
+  /** The root is skipped: its own route already asserts `browse` on a listing, `read` on a get. */
```
hercemer42: Claude Opus 5 (claude-opus-5): **Preferential** — the justification does not hold for the chart route, which is one of the callers.

`chart.ts:50-51` asserts only `assertCanExecuteChart`; nothing asserts `browse` or `read` on `this.collection`. So for charts the unconditional `[rootCollectionName, true]` entry rests on the chart-hash check alone, not on a collection-level assert. Charts were never collection-permission-gated, so this is not a regression — but the comment tells the next reader (and whoever ports this to agent-ruby) that an assert exists where it does not.

Compounding it slightly: `assertCanReadAggregatedFields` (`chart.ts:300-316`) passes `this.collection.name` as the root while resolving paths against `aggregatedCollection`. That is correct for both leaderboard shapes today, but two unrelated collections arriving as arguments to the same call is worth a word in the comment.

packages/agent/src/routes/access/chart.ts
```diff
@@ -233,9 +247,25 @@ export default class ChartRoute extends CollectionRoute {
...
+
+      await this.assertCanReadAggregatedFields(context, aggregatedCollection, [
+        ['group a leaderboard by', aggregation.groups[0].field],
+        ['aggregate a leaderboard on', aggregation.field],
+      ]);
+
+      // A count exposes the cardinality of the relation, which `/relationships/<name>/count` puts
+      // behind `browse`.
+      if (!aggregation.field) {
+        await this.services.authorization.assertCanBrowse(context, field.foreignCollection);
```
hercemer42, line 263: Claude Opus 5 (claude-opus-5): **Should fix** — this breaks dashboard leaderboards for existing roles, and neither the sequencing note nor the front PR covers that path.

A `Count` leaderboard on `holders` counting `cards`, for a role without `browse` on `cards`, now returns 403 where it returned counts. The chart is still rendered — dashboard chart visibility follows the chart's own collection (`/product/manage/dashboards` § Access control: *"Users only see the charts they have access to based on their role's collection permissions"*) — so the widget shows an error rather than disappearing.

ForestAdmin/forestadmin#9914 prunes record, list and export projections; it does not touch chart requests, so "do not merge before the front" does not protect this, and "Residual cost" names only related labels in list and detail views. The assertion itself is right and matches `count-related.ts:20`; what is missing is that dashboards are in the blast radius with no front-side change to absorb it. Worth naming in the PR body and the ticket before this ships.

hercemer42 (commented):
## Validator review — PR #1840

**Spec (PRD-900):** conforms on every route the ticket's *What shipped* section lists, and the three documented divergences (leaf-only checking, refuse-named / redact-default, no redaction announcement) are implemented as written — with one gap: the ticket's own **read 3** oracle stays reachable through `search`, not `filters`, because the guard only inspects search when `searchExtended` is also set. See the must-fix on `authorization.ts:112`. The scope/segment exemption is implemented and tested as specified.

---

Claude Opus 5 (claude-opus-5): **Should fix**

**Applies to:** `packages/agent/src/routes/modification/update.ts:40-46` (not in this diff)

`PUT /forest/<collection>/:id` re-lists the record with `ProjectionFactory.all(this.collection)` and serializes the result straight into the response, so a role with `edit` on `cards` and no `read` on `holders` gets `holder.nationalId` back from a no-op update. It is the same disclosure this PR closes on `GET /forest/cards/:id`, one HTTP verb away, and a role with `edit` almost always has it.

`modification/delete.ts:43-45` and `dissociate-delete-related.ts:119-122` accept `filters` on the same denied related paths — destructive, so a poor oracle, but the guard is absent there too. `create.ts:41` serializes only caller-supplied PKs, so it does not leak.

Routing `update.ts` through `redactProjection` with `explicit: false` is a two-line change and closes the read-shaped half. The filter-shaped half on the modification routes is a follow-up worth naming in the ticket's *Not in scope*, since "serve only the columns of collections the caller may read" reads as a claim about the whole agent.



---

Claude Opus 5 (claude-opus-5): **Should fix**

No ADR records the decisions this diff embeds, and `agent-nodejs` has no `docs/adr` directory at all — the org-wide corpus is four records, none touching permissions (control query verified, so this is an absence rather than a failed search). Three calls here are hard to reverse, surprising without context, and a real trade-off:

- explicitly named denied paths are refused while the implicit `ProjectionFactory.all` expansion is silently redacted — one class of request 403s, another is quietly narrowed;
- `browse` stands in for `read` on a `Count` leaderboard;
- rendering scopes, segments and a customized `replaceSearch` are exempt.

PRD-900 carries the rationale and the comments at `authorization.ts:55-60,88-92,148` carry part of it, but this is a cross-SDK contract — the ticket says agent-ruby should be built from it, and agent-python/php have no record either. Worth running `/adr` in this repo so the next implementer inherits the reasoning rather than the behaviour.

packages/agent/test/__factories__/authorization/authorization.ts
```diff
@@ -15,6 +15,12 @@ export class AuthorizationsFactory extends Factory<AuthorizationService> {
       Authorizations.getScope = jest.fn();
       Authorizations.assertCanExecuteChart = jest.fn();
       Authorizations.invalidateScopeCache = jest.fn();
+      Authorizations.canRead = jest.fn().mockResolvedValue(true);
+      Authorizations.assertCanReadQueryFields = jest.fn();
```
hercemer42, line 19: Claude Opus 5 (claude-opus-5): **Should fix** — these stubs mean the three `*-related` guards and two of the four chart shapes are asserted by nothing.

`assertCanReadQueryFields` is stubbed to a no-op and `redactProjection` to an identity passthrough, so `test/routes/access/list-related.test.ts`, `csv-related.test.ts` and `count-related.test.ts` exercise the absent guard and stay green whether `list-related.ts:21`, `csv-related.ts:26` and `count-related.ts:23` call it or not. The new security suite imports only `Chart`, `Count`, `Csv`, `Get` and `List` — no related route — and its chart cases cover Pie group-by and Leaderboard `Count` only, leaving `makeLineChart` (`chart.ts:156`) and `computeValue` (`chart.ts:287`, reached from Value and Objective) unpinned.

Those three routes are also the only sites passing `this.foreignCollection` rather than `this.collection`, so passing the wrong one there reproduces the exact cross-collection read this PR closes, with a fully green suite.

`skills/conventions/testing.md#Cover error and edge paths, not only the happy path`

packages/agent/test/security/related-read-permissions.test.ts
```diff
@@ -0,0 +1,388 @@
...
+    it('should accept a filter reading a collection the caller can read', async () => {
+      const dataSource = buildDataSource();
+      const services = buildServices(['holders']);
+      const list = jest.spyOn(dataSource.getCollection('cards'), 'list').mockResolvedValue([]);
+
+      await new List(services, options, dataSource, 'cards').handleList(
+        buildContext({ query: { filters: oracleFilter } }),
+      );
+
+      expect(list).toHaveBeenCalled();
```
hercemer42: Claude Opus 5 (claude-opus-5): **Violates conventions** — bare called-ness on all four allow-side cases (`:248`, `:264`, `:305`, `:317`) proves the request was not refused, not that the clause survived.

These are the only guards on the permissive half of `assertCanReadQueryFields`. If a later change to the guard or to `ContextFilterFactory` dropped the `holder:nationalId` condition, the injected scope, or `search`/`searchExtended` from the filter handed to `Collection.list`, all four still pass — and silently narrowing a permitted query is the failure mode this PR chose 403 over. Assert the filter instead: the condition tree at `:248`, the injected scope leaf at `:264`, `search: 'martin'` at `:305`, `search: 'martin', searchExtended: true` at `:317`.

The `not.toHaveBeenCalled()` assertions at `:128`, `:222`, `:236` are fine — a call that must not happen has no arguments to assert.

`skills/conventions/testing.md#Assert behavior with args and outputs`

packages/agent/test/security/related-read-permissions.test.ts
```diff
@@ -0,0 +1,388 @@
...
+    it('should traverse a collection it cannot read to reach a column it can', async () => {
+      const dataSource = buildDataSource();
+      const services = buildServices(['organizations']);
+      const list = jest.spyOn(dataSource.getCollection('cards'), 'list').mockResolvedValue([]);
+
+      await new List(services, options, dataSource, 'cards').handleList(
+        buildContext({}, { 'forest-projection': 'id,account:organization:name' }),
+      );
+
+      // The `account:` primary keys `withPks` re-adds are already on the row as `cards.accountId`.
```
hercemer42: Claude Opus 5 (claude-opus-5): **Preferential** — the justification is `ManyToOne`-only, and the test is the place someone will read it as the rule.

For `card.account` as a `ManyToOne` it is exactly right: `account:id` is `cards.accountId`, already on the row. For a `OneToOne` intermediate the key lives on the foreign side, so `withPks` (`projection/index.ts:56-64`) re-adds a primary key that nothing the caller may read carries — and `Serializer.buildRelationshipsConfiguration` (`serializer.ts:110-119`) emits it as `relationships.<relation>.data.id`, telling the caller that a record exists in a collection they were denied, and which one.

Narrow, and it follows from the leaf-only rule the ticket chose deliberately — so the ask is just to scope the comment to the relation type it is true for, since agent-ruby will be built from this reasoning.

packages/agent/test/services/authorization/authorization.test.ts
```diff
@@ -433,4 +434,109 @@ describe('AuthorizationService', () => {
       expect(forestAdminClient.markScopesAsUpdated).toHaveBeenCalledWith(42);
     });
   });
+
+  describe('keepReadableProjection', () => {
```
hercemer42: Claude Opus 5 (claude-opus-5): **Violates conventions** — this block names a method that does not exist: every assertion inside it exercises `AuthorizationService.redactProjection`, and no `keepReadableProjection` symbol appears anywhere in `packages/agent/src`.

Anyone grepping for the suite that pins the redact-vs-refuse policy — the load-bearing decision in this PR — finds nothing. Rename to `redactProjection`.

`skills/conventions/testing.md#Test name states the exact behavior it asserts`

packages/agent/src/services/authorization/authorization.ts
PMerlet: Confirmed and fixed — the hole was real, and I had dismissed it once on a bad check.

My first pass ran `holder:nationalId:1850` through the ANTLR grammar, which rejects it, and I concluded there was nothing here. The documented form is the dot syntax, and `FieldsQueryWalker` normalises `.` to `:` before the fields are resolved — so the oracle was reachable exactly as you wrote it, with no `filters`, no `sort` and no `searchExtended`.

The guard no longer derives anything at the route. `CollectionDecorator.getSearchedFields(search, extended)` returns the paths a search will actually reach and the collection each one ends on; `assertCanReadQueryFields` turns those into read checks. a88dbfed, then 78944c53 which moved the resolution itself down into the search decorator.

Pinned by `should refuse whatever the stack says the search will reach` and `should ask the stack with the extended flag the caller sent`.

packages/agent/src/services/authorization/authorization.ts
PMerlet: Same root cause as the oracle above, and closed by the same change — the route stopped enumerating anything, so both directions disappear rather than getting patched.

- **Under-inclusive.** `SearchCollectionDecorator.getSearchedFields` answers against `this.childCollection`, which is what the search actually reads, so a field hidden by publication or renaming above it is still accounted for. Resolution goes through `lenientGetSchema`, so `OneToMany` is traversed too.
- **Over-inclusive.** It returns `null` when a replacer is installed, and the caller must read `null` as "unknown", never as "none". The exemption argued in the Macroscope thread stands, but it is now expressed by the layer that owns the decision instead of being approximated one level up.

Four tests in `datasource-customizer/test/decorators/search/collections.test.ts`, plus `should serve the request when the stack cannot say what a search reaches` on the agent side.

packages/agent/src/services/authorization/authorization.ts
PMerlet: Accurate, including the part that matters most: the mechanism predates this PR, but this PR is what turns denial from a rare event into the steady state, so the cost is fairly attributed here.

Deferred to [PRD-1002](https://linear.app/forestadmin/issue/PRD-1002) rather than fixed on this branch, deliberately:

- the fix belongs in `forestadmin-client` — `hasPermissionOrRefetch` wiping the cache on *every* deny is wrong for every caller, not just this guard, and it ships on its own release cycle;
- memoising per request here would only cover the routes this PR touches and would leave the same storm on `assertCanBrowse`, `assertCanEdit` and the rest;
- it is a performance regression on an opt-out configuration, not a correctness or security one, so it should not gate the security fix.

Already in: `getReadPermissions` dedupes collection names within a call, so a request costs one lookup per distinct collection rather than one per path.

packages/agent/src/services/authorization/authorization.ts
PMerlet: Both halves fixed.

The comment no longer claims an assert that does not exist. It now reads: *"The root is skipped: `browse` gates a listing, `read` a get, and the signed hash a chart."* — naming the actual guarantee on each route, including that the chart's rests on the hash rather than on a collection permission.

For the second half I renamed the parameter instead of adding a comment: `assertCanReadAggregatedFields(context, pathCollection, fields)`. The two collections reaching the same call are now distinguishable by name — `pathCollection` is what paths resolve against, `this.collection.name` is the permission root — so the leaderboard call site passing `aggregatedCollection` reads as intentional. 46659eb7.

packages/agent/src/routes/access/chart.ts
PMerlet: Agreed, and now named in both places you asked for.

**PR body** — a paragraph under "Residual cost": a `Count` leaderboard on `holders` counting `cards` now 403s for a role without `browse` on `cards`, the widget still renders and shows an error rather than disappearing, and ForestAdmin/forestadmin#9914 does not touch chart requests, so the front-first sequencing does not cover this path.

**PRD-900** — the same paragraph under "Newly settled: a leaderboard that counts", so agent-ruby inherits it rather than rediscovering it.

The assertion itself stays as you read it. Worth flagging separately to whoever runs the sweep: for this path it is `browse`, not `read`.

packages/agent/test/__factories__/authorization/authorization.ts
PMerlet: Fixed in d90b26ce — by covering the routes rather than by un-stubbing the factory, which would have rewritten a large number of unrelated route tests for no gain here.

`related-read-permissions.test.ts` now imports `ListRelated`, `CsvRelated` and `CountRelated` and drives them through a real `AuthorizationService`, so passing `this.collection` where `this.foreignCollection` is required now fails the suite — which is the substitution those three routes are uniquely exposed to.

Chart coverage extended to the two unpinned shapes, `makeLineChart` and `computeValue` (reached from Value and Objective), alongside the Pie group-by and Leaderboard `Count` cases already there.

packages/agent/test/security/related-read-permissions.test.ts
PMerlet: Fixed. The four allow-side cases now assert the clause survived, not that the request escaped refusal:

- filter: `list.mock.calls[0][1].conditionTree` matches the `holder:nationalId` leaf;
- scope: the injected leaf is asserted in the tree handed to `list`;
- search: `{ search: 'martin', searchExtended: true }` on the filter, and `getSearchedFields` asserted to have been called with `('martin', true)`.

Agreed on `not.toHaveBeenCalled()` for the refusal cases — a call that must not happen has no arguments to assert. Left as they were.

packages/agent/test/security/related-read-permissions.test.ts
PMerlet: Scoped, as asked. The comment now reads:

> `withPks` re-adds `account:id`, which a ManyToOne already carries on the row as `cards.accountId`. A OneToOne intermediate would expose a key the row does not carry.

The reasoning agent-ruby inherits now states the relation type it holds for, and names the OneToOne case as something the leaf-only rule leaves open rather than letting it read as covered.

packages/agent/test/services/authorization/authorization.test.ts
PMerlet: Renamed to `redactProjection` in 46659eb7. 26 tests in that file, all green.

PMerlet:
Replying to the two findings in the [validator review](https://github.com/ForestAdmin/agent-nodejs/pull/1840#pullrequestreview-4984300713) body, which have no inline thread.

## `update.ts:40-46` — fixed

Verified independently before fixing, and your map of the three sites is exact: `ProjectionFactory.all` appears in exactly one route (`update.ts:43`), `create.ts:41` serializes only caller-supplied data, and the two filter-shaped sites are `delete.ts:45` and `dissociate-delete-related.ts:122`.

Fixed in bc3925a4 as suggested — routed through `redactProjection` with `explicit: false`. Redaction rather than refusal is load-bearing here: the projection is entirely agent-chosen, so a `PUT` must not 403 because the row it wrote happens to carry a relation the caller cannot read. Two tests, one per side of the branch: the redacted case, and a readable relation surviving so the fix cannot be satisfied by dropping relations wholesale.

You were right that the PR title reads as a claim about the whole agent. The "Routes covered" section now names `update` and says why a write is in it.

**The filter-shaped half is deferred, not overlooked** — [PRD-1012](https://linear.app/forestadmin/issue/PRD-1012), and named in PRD-900's *Not in scope*. Two reasons, and the first weakens your "poor oracle" reading further: a delete answers `204` whatever it matched, so there is no response channel at all — the caller learns only by observing which rows vanished, and each probe destroys the rows it is measuring. Second, `assertCanReadQueryFields` is a one-line addition to both sites, but it would 403 any bulk delete filtered on a belongsTo sub-field, which the filter bar lets a user build and which #9914 does not prune. That is the same breaking-change class as the leaderboard, on a destructive path, and it needs its own sequencing decision rather than riding on this one.

## No ADR — declined, deliberately

Not adding one. PRD-900 carries all three decisions with their counter-arguments, and its *What shipped* section is explicitly marked as authoritative over the earlier decision for whoever ports this — that is the artefact agent-ruby will be built from, and it is linked from this PR.

The gap you name is real: Linear is less discoverable from a repo than `docs/adr` would be. But `agent-nodejs` has no ADR directory, so this would create the convention rather than follow it, and a corpus of one that nobody else maintains is worse than a ticket people already read. Worth raising as its own decision for the SDK repos rather than settling it inside a security fix that is waiting on a front PR to land.

## qlty structure findings

- `lenientGetSchema` complexity 16 / nesting level 4 — **fixed** in bc3925a4. The real branch is on `suffix`, so it is now taken once at the top instead of inside the loop. `continue` guards were the first attempt and ESLint rejects them (`no-continue`), which is why it reads this way. A named `SEARCHABLE_THROUGH` constant with a type predicate replaces the inline three-way type disjunction.
- `makeChart` many returns (5) — **left as is.** `git show origin/main` confirms it already had 5 returns; it is flagged only because this PR added a line to the function. It is a `switch` dispatching one return per chart type, which is the clearest form available.

packages/agent/test/security/related-read-permissions.test.ts
```diff
@@ -0,0 +1,593 @@
...
+      await new List(services, options, dataSource, 'cards').handleList(
+        buildContext({ query: { search: 'martin' } }),
+      );
+
+      expect(list.mock.calls[0][1]).toMatchObject({ search: 'martin', searchExtended: false });
+    });
+
+    it('should accept an extended search once every to-one relation is readable', async () => {
+      const dataSource = buildDataSource();
+      const services = buildServices(['holders', 'accounts']);
```
hercemer42, line 514: Claude Opus 5 (claude-opus-5): **Should fix** — this grant is decorative: the test passes with no permissions at all, so neither of the two allow-side search cases exercises the guard.

The collection factory does not define `getSearchedFields` — the four refusal cases assign it inline (`:303`, `:322`, `:339`) precisely because it is absent by default. So on this test and on `:349`, `assertCanReadQueryFields` reaches `(collection as CollectionDecorator).getSearchedFields?.(...)`, gets `undefined`, and `?? []` pushes zero usages. Swap `buildServices(['holders', 'accounts'])` for `buildServices()` and it still passes.

The name claims the permissive half of the search check — *once every to-one relation is readable* — and nothing in the test would notice if that half stopped working. Stubbing `getSearchedFields` to the paths an extended search really reaches, as the refusal cases do, makes both tests assert what they are named for. Same shape as the bare-called-ness finding these replaced, one layer further in.

`skills/conventions/testing.md#Assert behavior with args and outputs`

hercemer42 (commented):
## Validator re-review — PR #1840

Follow-up to run `1840-20260820144736-hercemer42`, re-verified against `bc3925a4`. All nine inline findings from that review are addressed: seven fixed, one deferred to PRD-1002 with reasoning I accept, and the `update.ts` disclosure fixed in `bc3925a4` — the exact-projection tests on both sides of the branch are the right way to pin it. The `lenientGetSchema` refactor is semantically identical to what it replaced. CI is green on this head.

**Spec (PRD-900):** the code conforms. The ticket does not — see the finding below; it is a documentation gap with a code consequence for the SDK ports, not a defect in this diff.

---

Claude Opus 5 (claude-opus-5): **Should fix**

**Applies to:** PRD-900

The ticket does not record the search fix. No mention of the `relation.column:term` dot syntax, of `getSearchedFields`, or of the oracle that motivated `a88dbfed` / `78944c53` — and *What shipped* still describes the implementation those commits replaced:

> Filters, sorts, extended search, and a chart's group-by or aggregated field are **always refused**

> **A customized** `replaceSearch` **is exempt, deliberately.** Its condition tree is built inside `SearchCollectionDecorator.refineFilter`, long after the route has run its check, so a handler searching `holder:nationalId` is never verified.

Both sentences reason from a route that derives the check from the schema. The route no longer derives anything: it asks the stack, and the exemption is now the decorator returning `null` because it owns the decision — which is a better rule, stated nowhere.

This matters because of the argument for not adding an ADR, which I accept on its merits: PRD-900 carries the decisions, and it is what agent-ruby will be built from. It now carries three of four, and the missing one is the largest change in the follow-up. Built from the ticket as it stands, the Ruby port lands on the original design — enumerate to-one relations at the route, gate on `searchExtended`, never look at the search string — which is the hole this PR closed. The dot syntax is documented end-user behaviour (`/product/execute/browse` § Advanced search syntax), so the port would ship it reachable.

The other three follow-ups are recorded well: `update` in *Routes covered*, the dashboard leaderboard under *Newly settled*, the delete filters split to PRD-1012 in *Not in scope*. This is the one gap, and it is the one that carries a vulnerability into the next SDK.

packages/datasource-customizer/src/decorators/search/field-paths.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export function getLeafCollectionName(collection: Collection, path: string): string {
+  const index = path.indexOf(':');
+
+  if (index === -1) return collection.name;
+
+  const relation = collection.schema.fields[path.substring(0, index)];
+
+  if (!relation || relation.type === 'Column') return collection.name;
```
hercemer42: Claude Opus 5 (claude-opus-5): **Preferential** — this default fails open, in a helper whose whole job is to decide which collection a permission check applies to.

An unresolvable prefix yields the root collection name, and `getReadPermissions` pins the root to `true` unconditionally — so "I cannot resolve this path" becomes "this path is allowed". Not reachable today: every path reaching here comes from `lenientGetSchema`, which returns canonical schema keys, or from `getFields`, which builds them from `Object.entries(schema.fields)`. So this is about the shape rather than a live hole.

Worth aligning with the agent's own resolver, which took the opposite default: `FieldPathUtils.getLeafCollection` lets `SchemaUtils.getRelation` throw. Throwing, or returning `null` and having `getSearchedFields` drop the path, keeps a future caller from inheriting a silent allow.

packages/agent/src/routes/modification/update.ts
```diff
@@ -40,7 +40,10 @@ export default class UpdateRoute extends CollectionRoute {
     const [updateResult] = await this.collection.list(
       caller,
       new Filter({ conditionTree }),
-      ProjectionFactory.all(this.collection),
+      await this.services.authorization.redactProjection(context, this.collection, {
```
hercemer42, line 43: Claude Opus 5 (claude-opus-5): **Preferential** — the fix is right; it just widens PRD-1002 beyond what that ticket currently describes.

A `PUT` now issues one `canRead` per distinct related collection, so on an `instantCacheRefresh: false` deployment every update also clears the environment permission cache and refetches it — PRD-1002 scopes the symptom to "every list request", and its title says a *denied permission check*, which a write now performs routinely too.

No change needed here. One sentence in PRD-1002 naming `update` alongside the read routes keeps whoever picks it up from sizing the blast radius from reads alone.

packages/agent/test/security/related-read-permissions.test.ts
PMerlet: Confirmed empirically before fixing — I swapped `buildServices(['holders', 'accounts'])` for `buildServices()` and the test still passed. The collection factory defines no `getSearchedFields` at all, so `getSearchedFields?.(...)` returned `undefined` and `?? []` pushed zero usages, exactly as you described.

Both allow-side cases now stub it to the paths a search really reaches, as the refusal cases do:

- the extended one to `panLast4` / `holder:fullName` / `account:iban`, so the grant it names is the thing being exercised;
- the plain-search one to `panLast4`, which had the same hole — its name claims "never leaves the root collection", and nothing was checking that the root is what the stack reported.

Re-falsified after the fix: dropping the grant now fails with `ForbiddenError: You cannot search on 'holder:fullName': you are not allowed to read the 'holders' collection.` The test bites.

packages/datasource-customizer/src/decorators/search/field-paths.ts
PMerlet: Fixed rather than left as a shape note. A fail-open default in the one helper whose job is to decide which collection a permission check applies to is worth closing even when nothing reaches it today — the next caller inherits the default, not the reachability argument.

`getLeafCollectionName` now hands the prefix to `SchemaUtils.getRelation`, which throws `MissingRelationError`. That is the same resolver and the same default as `FieldPathUtils.getLeafCollection` on the agent side, so the two halves of the check can no longer disagree about what an unresolvable path means.

I did not take the `null`-and-drop option: `getSearchedFields` dropping the path pushes no usage, so `assertCanReadQueryFields` checks nothing and serves the request — the same silent allow, one layer further up. Throwing is the only variant that actually fails closed.

Pinned by four tests in a new `field-paths.test.ts`, including the case you named — a `Column` prefix carrying a suffix (`panLast4:nationalId`), which is the branch that used to answer `cards`.

packages/agent/src/routes/modification/update.ts
PMerlet: Agreed, and no code change here.

PRD-1002 now carries a paragraph naming `update` next to the read routes: the route re-lists the row it wrote with `ProjectionFactory.all`, so a write performs the same one-`canRead`-per-related-collection sweep and carries the same cache wipe. It ends by listing the routes to size the blast radius from — `get`, `list`, `csv`, the related and count routes, the chart routes and `update` — so whoever picks it up does not read "every list request" as the whole surface.

Left the title alone: the mechanism is still a denied permission check, it is just no longer reads-only that performs one.

.releaserc.js
```diff
@@ -31,7 +31,38 @@ module.exports = {
...
+              context,
+            );
+
+            if (!transformed) return transformed;
+
+            // The preset relabels every note "BREAKING CHANGES"; keep each title so an upgrade
+            // note does not announce itself as a breaking change on a patch release.
+            transformed.notes = commit.notes.map(note => ({
+              ...note,
+              title: note.title === 'BREAKING CHANGE' ? 'BREAKING CHANGES' : `${note.title}S`,
```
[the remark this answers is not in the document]
PMerlet: Moot — this code is no longer in the branch. The release-notes change was reverted in full and `.releaserc.js` is now identical to `main`: the release policy of a public repo should not change for one case. The mitigation moved to explicit front-side errors ([PRD-1014](https://linear.app/forestadmin/issue/PRD-1014)) plus direct communication with the large self-hosted clients.

For the record, since it may matter if this is ever revisited: the premise did not reproduce when I ran it. I fed the real squash body — the config's `parserOpts` and `writerOpts` with the angular preset — through `conventional-commits-parser` and `conventional-changelog-writer`, and the output was a correct `### UPGRADE NOTES` section, not `BREAKING CHANGESS`. So either the preset's `transform` does not mutate `commit.notes` in place at the installed version, or it does so after my callback reads them. Worth re-checking empirically rather than from the code shape if anyone picks this up again.

packages/agent/test/security/related-read-permissions.test.ts
hercemer42: Verified at `aeb3c3c2`. Both cases now stub the paths a search really reaches, and the grant is load-bearing: the extended one names `holder:fullName` and `account:iban`, so `buildServices(['holders', 'accounts'])` is what keeps it green, and the plain-search one asserts the root-only claim in its name instead of assuming it. Resolving.

packages/datasource-customizer/src/decorators/search/field-paths.ts
hercemer42: Verified at `aeb3c3c2`. `SchemaUtils.getRelation` is the right landing: same resolver and same default as `FieldPathUtils.getLeafCollection`, so the two halves of the check can no longer disagree about an unresolvable path. Your rejection of null-and-drop is correct and I should have said so — a dropped path pushes no usage, so `assertCanReadQueryFields` checks nothing and serves the request, which is the same silent allow one layer up. I checked no live path can now throw: every path reaching the helper comes from `lenientGetSchema` (canonical keys, relation-traversed) or `getFields` (built from schema entries), so the four new tests exercise a contract rather than a reachable 500. Resolving.

packages/agent/src/routes/modification/update.ts
hercemer42: Verified — PRD-1002 now carries the *Writes too, not only reads* paragraph with the route list, which is what this was asking for. Agreed on leaving the title alone. Resolving.

hercemer42:
Three inline threads from the re-review are verified at `aeb3c3c2` and resolved. The `getLeafCollectionName` fix went further than the finding asked — you closed it instead of accepting the reachability argument, and the `withPks` comment now names a case I had not raised (a `foreignKeyTarget` that is not the primary key, where `account:id` genuinely carries something the row does not). Both correct.

Pushing back on the one body finding, which is a **Should fix** and still open.

**PRD-900 does not record the search fix.** Re-checked just now: no mention of the `relation.column:term` dot syntax, of `getSearchedFields`, or of the oracle anywhere in the description. The ticket was edited at `08:45Z`, after the re-review posted, so this is not a lag — and both passages that describe the deleted implementation are still there verbatim:

> Filters, sorts, extended search, and a chart's group-by or aggregated field are **always refused**

> **A customized** `replaceSearch` **is exempt, deliberately.** Its condition tree is built inside `SearchCollectionDecorator.refineFilter`, long after the route has run its check, so a handler searching `holder:nationalId` is never verified.

The second one is the sharper half, and it is not only stale wording — the *reason* is now wrong. The exemption no longer exists because the route's check runs too early; it exists because `SearchCollectionDecorator.getSearchedFields` returns `null`, deliberately, as the layer that owns the decision. A port that implements the recorded reason builds the check at the route, sees that a replaced search is unverifiable there, and concludes it has parity — while the dot-syntax path it never looked at stays open. That path is documented end-user syntax (`/product/execute/browse` § Advanced search syntax), so it ships reachable.

This is the same ticket you offered in place of an ADR, and I accepted that on its merits — a corpus of one is worse than a ticket people read. The trade only holds while the ticket is complete. It now carries `update`, the dashboard leaderboard and the delete split, and not the largest change of the three rounds.

Concretely, what agent-ruby needs from *What shipped*: the route asks the collection what a search will reach rather than deriving it from the schema; the answer is resolved below the publication and renaming layers; `relation.column:term` reaches a relation with no `searchExtended`; to-many hops resolve too; and an unknown answer serves the request, which is where the `replaceSearch` exemption now lives. Happy to be told this belongs in an ADR after all, or in the Ruby ticket instead — but not nowhere.

hercemer42 (approved):

packages/agent/src/services/authorization/authorization.ts
```diff
@@ -40,6 +53,135 @@ export default class AuthorizationService {
...
+      collectionName,
+    });
+  }
+
+  /**
+   * An unnamed field is redacted rather than refused: `ProjectionFactory.all` expands every column
+   * of every to-one relation when no `fields[]` is sent, so refusing would turn an ordinary
+   * listing into a 403.
+   */
+  public async redactProjection(
```
matthv, line 69: Every caller of `redactProjection` (`list.ts:33`, `get.ts:36`, `list-related.ts:43`) runs `.withPks(collection)` on its result. `Projection.withPks` recurses through `this.relations` and pushes the foreign collection's PK for every relation still present in the *already redacted* projection — so a key gets re-added without going through the leaf-collection check this method just did.

The test at `related-read-permissions.test.ts:183-187` already names the residual in a comment: harmless for a ManyToOne targeting the primary key (the row already carries the FK as e.g. `accountId`), but for a OneToOne or a `foreignKeyTarget` that isn't the primary key, this adds a value the row doesn't otherwise carry and the caller has no independent way to read.

Worth flagging: that comment documents the gap but nothing pins it — there's no OneToOne or non-PK `foreignKeyTarget` fixture in this suite, so the case is diagnosed, not tested. Same gap exists identically on the ruby port (#365).

Minimal fix: do the `withPks` expansion here, inside `redactProjection`, so the keys it adds go through `isReadable`/the same leaf check as everything else, instead of being appended by each route after the check has already run.

matthv (changes requested):
Requesting changes for the permission bypass flagged inline on `authorization.ts:69`: every route calling `redactProjection` (`list.ts`, `get.ts`, `list-related.ts`) runs `.withPks(collection)` on the result, which re-adds a foreign-collection primary key for every relation still present in the projection — without that key going through the leaf-collection read check `redactProjection` just ran. Harmless for a ManyToOne targeting the primary key, but exposes a value the caller has no read permission on for a OneToOne or a non-PK `foreignKeyTarget`. The gap is already named in a comment on the test at `related-read-permissions.test.ts:183-187`, but nothing pins or fixes it. See the inline comment for the trace and a suggested fix (do the `withPks` expansion inside `redactProjection` so it goes through the same check).

packages/agent/src/services/authorization/authorization.ts
PMerlet: The mechanism is exact, and your reading of the ManyToOne/OneToOne split is right. The minimal fix isn't available though: `withPks` is the serializer's precondition, not an afterthought. `IdUtils.packId` (`id.ts:23`), reached from `beforeSerialize` (`serializer.ts:81`), throws on a relation object with no identity — so moving the expansion inside `redactProjection` turns `GET /forest/cards` with `Forest-Projection: id,account:organization:name` into `Error: Missing one of expected fields: 'id'`, a 500 where it used to be a 200. Verified end-to-end through `ListRoute.handleList` with the real `Serializer`, not reasoned about.

And the leak reduces entirely to that one case. A relation only survives the redaction if a path through it survives; either the leaf is its own collection (readable, so its key is too) or it is traversed to a deeper readable leaf. Traversal is the only shape where the key is denied — and it is the shape where dropping it breaks serialization, because JSON:API renders the traversed relation as a resource with an identity: `relationships.account.data.id` plus an `accounts` entry in `included` are what carry `organization.name` to the caller at all.

So the real choice is *serve the key* or *refuse the traversal* — not *drop it*. Refusing would mean requiring `read` on every collection a projected path materializes rather than on the leaf, which contradicts the rule PRD-900 states, so the key stays and the residual is accepted: an id, and only an id.

Your second point is the one that landed. It was diagnosed in a comment and pinned by nothing, and the reason it slipped is that the suite stubs `services.serializer`. Three tests now, on a OneToOne fixture whose key the root row does not carry: the projection keeps `contract:id`; the payload exposes that identity and no column of `contracts` (`reference`, `cardId` and `organizationId` are asserted absent); and the `list` mock reprojects the row, so dropping the key from the projection fails the test with the exact 500 above rather than passing quietly.

The ruby port (#365) inherits the same reasoning, including the reason the minimal fix is unavailable there.

packages/agent/src/services/authorization/authorization.ts
```diff
@@ -40,6 +53,135 @@ export default class AuthorizationService {
...
+
+    if (search) {
+      // Asked whatever the extended flag: `relation.column:term` is end-user syntax and reaches a
+      // relation without it. An unknown answer serves the request — the `replaceSearch` exemption.
+      const searched = (collection as CollectionDecorator).getSearchedFields?.(
+        search,
+        QueryStringParser.parseSearchExtended(context),
+      );
+
+      for (const { path, collection: name } of searched ?? []) {
```
matthv, line 152: `searched ?? []` turns "unknown" into "nothing to check". `getSearchedFields()` returns `null` on purpose when a `replaceSearch(...)` handler is installed (`collection.ts:127-128`: *"the customer's handler chooses the fields, and the caller only supplies the text"*) — but here that `null` is read the same as an empty result, so no permission check runs on the search at all once a `replaceSearch` is in place.

Concrete case: a common-enough setup —
```ts
cardsCollection.replaceSearch((value) => ({
  field: 'holder:nationalId',
  operator: 'StartsWith',
  value,
}));
```
Any caller with just `browse` on `cards` (zero permission on `holders`) can send `?search=1`, `?search=18`, `?search=185`... and read from whether rows come back — the exact one-guess-per-request oracle this PR closes for filters (see the comment a few lines up), reopened here.

I see it's deliberate and pinned by `related-read-permissions.test.ts:351-365` ("should serve the request when the stack cannot say what a search reaches"), so this may be an accepted trade-off (trust the customer's own handler) rather than an oversight — but it's not written down anywhere visible (PR description, SDK doc for `replaceSearch`), so a future reader comparing this PR's claims against the diff would reasonably assume `search` is fully covered. Worth at least a line in the PR description or the `replaceSearch` doc making the trade-off explicit, so it's a decision on record rather than something a security review has to rediscover.

PMerlet: Your trace is right on all three links, and the ask lands — but the decision moved further than the doc line, and it moved this morning.

PRD-1024, settled 2026-08-24 at 10:20: **a plain search is served, an extended search is refused**, and it says in as many words that node has only the `replaceSearch` cause and the rule applies there unchanged. PRD-900 §8 carries it now ("the exemption stops at the extended flag"). So this was no longer a trade-off to write down — it was a gap between the code and a rule decided a few hours before you read it, and `related-read-permissions.test.ts:351-365` was pinning exactly the case that is now refused.

The deciding question is not whether the fields are known but **whether the caller can aim**, which is the principle the rest of the change already runs on: refuse what the caller named, prune what they did not. On a plain search they aim at nothing — the footprint is the one the customer chose, identical for every role, same category as a scope, and refusing it would remove search from every collection carrying a handler. The `searchExtended` flag is theirs, though, and that is what your oracle actually rides on: the same term with the flag off and then on differs by exactly the rows that matched through a relation, one bit per term over collections no check ever covered.

Fixed in 88d1674b, plus the doc you asked for in f3c2de6a:

- `assertCanReadQueryFields` refuses an unknown footprint when the search is extended, naming the collection, and keeps serving the plain half.
- A collection that does not answer at all is read the same way as one answering `null` — silence is not an empty footprint. Its own test, because `!searched` → `searched === null` is a mutation nothing else catches.
- The old "cannot say" test becomes the plain-search half; the extended half has its own.
- `replaceSearch`'s JSDoc now states the exemption, the oracle it leaves open on a plain search, and that the extended mode is refused — where a customer installing a handler will read it, rather than only in the ticket.
- PR description updated: the *Not in scope* line no longer reads as "search is fully covered", and *Residual cost* carries the split and its price.

Worth naming the cost, since it widens this PR's blast radius: a customer using `replaceSearch` on a collection where users tick extended search now gets a 403 where they got a 200. That is PRD-1024's decided cost, not a side effect. And PRD-1024 is titled `[agent-ruby]` with a ruby implementation section, so there is no node task for what just shipped here — I will get the ticket scoped to both rather than leave the node half undocumented.

matthv (approved):
LGTM

PR #1841, opened 2026-08-21
chore(claude): register the forest plugin marketplace

hercemer42:
Installing the Forest Claude Code toolkit in this repo currently takes two commands, the first of which nobody can guess:

```
/plugin marketplace add ForestAdmin/claudine
/plugin install forest@forest
```

This commits the marketplace source so only the second one is needed.

## What the entry does, and does not do

`.claude/settings.json` registers `ForestAdmin/claudine` as a marketplace named `forest`, and nothing else. It does not enable the plugin.

`extraKnownMarketplaces` is a declaration, not a fetch: a session never clones from it. Measured in a scratch project against an unreachable marketplace, and against a reachable one not yet on disk — in both cases the session started normally, wrote nothing to `installed_plugins.json` or `known_marketplaces.json`, cloned nothing, and printed no error even under `--debug`. So a developer who has not run `/plugin install forest@forest` has no plugin fetched, loaded or run for them, and someone without access to the private repo sees nothing at all rather than a failure.

The pre-enable key `"forest@forest": true` is deliberately absent, and not for the reason one might assume. Measured in the same harness, on a machine where the marketplace is *already* cloned, that key installs the plugin at project scope on session start with nobody running `/plugin install` — recorded in the ledger as `"scope": "project"` against the project path. That is what business rules 1 and 2 rule out: the entry must install nothing, and installing stays an explicit act by the developer. It has no bearing on people without access, who are unaffected either way.

Observed in print mode (`claude -p`); an interactive session may surface warnings print mode does not.

Access to the private `ForestAdmin/claudine` comes from the developer's own git credentials. There is no field for it here. Someone without access gets a resolution failure when they run `/plugin install` themselves — not before, since nothing here reaches for the marketplace unprompted.

Names come from the marketplace manifest rather than a guess: `ForestAdmin/claudine` → `.claude-plugin/marketplace.json` declares marketplace `"name": "forest"` with one plugin `"name": "forest"`, hence `forest@forest`.

## Scope

`.claude/settings.json` only, +6/-0, purely additive. The two `anthropics` marketplaces, the six already-enabled plugins and the `permissions.deny` block are byte-identical to `main` — `enabledPlugins` is no longer touched at all.

Refs: [PRD-713](https://linear.app/forestadmin/issue/PRD-713/install-the-toolkit-with-one-command-in-the-repos-we-work-in)

> [!NOTE]
> ### Register the `forest` plugin source and enable `forest@forest` skill
> Adds a new GitHub-backed source named `forest` pointing to `ForestAdmin/claudine` in the Claude settings. Enables the `forest@forest` skill for use.
>
> 
> #### Changes since #1841 opened
>
> - Removed `forest@forest` plugin registration from Claude settings configuration [86d446f]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2bc9ea1.</sup>
> 
>

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -25,8 +25,7 @@
     "feature-dev@claude-code-plugins": true,
     "pr-review-toolkit@claude-code-plugins": true,
     "document-skills@anthropic-agent-skills": true,
-    "example-skills@anthropic-agent-skills": true,
-    "forest@forest": true
+    "example-skills@anthropic-agent-skills": true
   },
   "permissions": {
     "deny": [
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -11,6 +11,12 @@
         "source": "github",
         "repo": "anthropics/skills"
       }
+    },
+    "forest": {
+      "source": {
+        "source": "github",
+        "repo": "ForestAdmin/claudine"
+      }
     }
   },
   "enabledPlugins": {
@@ -19,7 +25,8 @@
     "feature-dev@claude-code-plugins": true,
     "pr-review-toolkit@claude-code-plugins": true,
     "document-skills@anthropic-agent-skills": true,
-    "example-skills@anthropic-agent-skills": true
+    "example-skills@anthropic-agent-skills": true,
+    "forest@forest": true
   },
   "permissions": {
     "deny": [


.claude/settings.json
```diff
@@ -19,7 +25,8 @@
     "feature-dev@claude-code-plugins": true,
     "pr-review-toolkit@claude-code-plugins": true,
     "document-skills@anthropic-agent-skills": true,
-    "example-skills@anthropic-agent-skills": true
+    "example-skills@anthropic-agent-skills": true,
+    "forest@forest": true
```
Tonours: Claude Opus 5 (claude-opus-5[1m]): **Should fix** — this is the one of the four repositories that is public, so pre-enabling a plugin from a private marketplace lands on people who cannot resolve it.

Who observes what: `ForestAdmin/agent-nodejs` is public and `ForestAdmin/claudine` is private (`gh api repos/… --jq .private` → `false` / `true`). Every external contributor, fork, or credential-less sandbox that opens a Claude Code session in a clone of this repo carries a tracked config that enables `forest@forest` from a marketplace their credentials cannot fetch. By the PR description's own account ("someone without access gets a resolution failure"), that is a resolution failure per session, for the population least able to act on it. Forest developers with `gh` auth see it work.

The marketplace entry alone is inert for them — nothing resolves until someone runs `/plugin install`. It is `"forest@forest": true` that makes it happen unprompted. Note this repo's existing six pre-enabled plugins all come from public `anthropics` marketplaces, so this is the first entry that cannot resolve for an outsider.

Two one-line options: drop `"forest@forest": true` and keep the marketplace entry — which is the part the ticket actually asks for, and matches PRD-713's implementation proposal, where the pre-enable key was deliberately left out — or keep it and confirm what an unauthenticated session in this repo actually prints, since that is the half nobody has observed yet.

Tonours (commented):
**Spec (PRD-713)**: conforms — `agent-nodejs` is one of the four repositories AC 1 names, and the entry makes `forest@forest` resolvable without a manual marketplace source. The "purely additive" claim checks out exactly: a recursive structural diff of the parsed file between the merge-base and this SHA yields only `ADDED /extraKnownMarketplaces/forest` and `ADDED /enabledPlugins/forest@forest` — the two `anthropics` marketplaces, the six existing plugins and `permissions.deny` are unchanged, in place and in order. `forest@forest` is a real pair: `ForestAdmin/claudine`'s `.claude-plugin/marketplace.json` declares marketplace `forest` with one plugin `forest`.

AC 5 (confirmed by somebody other than the author, on a machine where the plugin is not installed) is still outstanding; AC 4 (the toolkit README naming the four repositories) belongs to `claudine`.

One finding inline.

.claude/settings.json
hercemer42: Agreed — took the first option, dropped `"forest@forest": true` in `86d446f775c1bb6683bd4e46ba256da6c0a8f37c`. `enabledPlugins` is now byte-identical to `main` and the PR is `extraKnownMarketplaces` only, +6/-0.

Confirmed both halves of your premise independently: `gh api repos/ForestAdmin/agent-nodejs --jq .private` → `false`, `ForestAdmin/claudine` → `true`.

The part that settles it beyond the public/private asymmetry is that PRD-713's own implementation proposal already ruled on this key, and the PR contradicted it:

> The pre-enable key is deliberately left out: it is not needed for the install to resolve, and what it does for somebody who has not installed yet is unverified — it can be added as a one-line change once somebody has watched it happen.

So this wasn't a judgement call between your two options — option two would have been re-opening a decision the ticket had already made, on the strength of an observation nobody had made. It also collided with the ticket's last acceptance criterion, "nothing shipped claims behaviour nobody has observed", which is the same rule you're applying here.

Two things your finding surfaces that go wider than this PR, both flagged to @brianfox rather than fixed here:

- **`ForestAdmin/agent-ruby` #367 has the identical defect.** It is also public (`private=false`) and its `.claude/settings.json` carries the same `"forest@forest": true`. Same fix, not applied yet because it is outside the two PRs I was asked to touch.
- **`forestadmin` #9921 and `forestadmin-server` #8457 also carry the key.** Both repos are private, so your stated harm doesn't reach them — but they still contradict the ticket's proposal, so they're arguably wrong for the weaker reason.

Description updated: it no longer says the entry "marks the `forest` plugin enabled", and now states why a public repo deliberately stops at the marketplace entry.

hercemer42: Correction to my own reply above — I repeated "a resolution failure per session" from your finding without testing it. I've now tested it, and it does not reproduce. Nothing happens to a credential-less session at all.

Method: scratch project whose `extraKnownMarketplaces` points at `ForestAdmin/claudine-no-access-probe`. GitHub returns 404 for a private repo you cannot see and for one that does not exist alike, so that is a faithful stand-in for no access. Ran a session in each shape, captured stdout/stderr, and diffed `installed_plugins.json` / `known_marketplaces.json` / the marketplace clone dir before and after.

| project settings | marketplace | what happened |
|---|---|---|
| entry only | unreachable | session normal, zero writes, no error |
| entry + `"forest@forest": true` | unreachable | identical — silent, even under `--debug` |
| entry + switch | reachable but not yet on disk | no clone, no install |
| entry + switch | **already on disk** | **auto-installed at project scope** |

The mechanism is that `extraKnownMarketplaces` is lazy — it is a declaration, and a session never clones from it. So for somebody without access there is nothing to fail: no fetch is attempted, no error is printed, no prompt appears. The last row is where the real behaviour is, and it only reaches people who already have the marketplace cloned.

So the harm you named does not land, but your recommendation was right, for a reason that is stronger and points the other way — at Forest developers rather than outsiders:

`"forest@forest": true` silently installs the plugin at project scope on session start, with nobody running `/plugin install`. Ledger entry from the run:

```
"scope": "project",
"installedAt": "2026-09-03T09:15:04.093Z",
"projectPath": "…/B"
```

That is exactly what PRD-713's business rules 1 and 2 forbid — "installs nothing, loads nothing, and runs nothing on anyone's behalf", and "installing stays an explicit act by the developer". So the key had to go regardless; the public/private asymmetry just was not the reason.

It also means this PR's description was wrong before I touched it — "It installs nothing: a developer who has not run `/plugin install forest@forest` has no plugin fetched" is false for anyone with the marketplace already cloned, which is every Forest developer who has installed the toolkit once. The "verified in a scratch repo" note behind that sentence tested a machine where the plugin was **absent**, which is the one case that does not exercise the install path. Fixing the description now, and dropping the credential-less-resolution-failure claim I added to it earlier today.

Caveat on all of the above: observed in print mode (`claude -p`). An interactive session could surface a warning print mode does not; I could not drive the TUI to check.

hercemer42:
Closing — the premise doesn't hold up.

A user-scope `/plugin install forest@forest` makes the toolkit available in **every** repo, not just the four. So a committed per-repo marketplace entry only helps somebody who has not installed yet, in one of four specific repos, and becomes dead weight the moment they install once anywhere. Verified the plain way: `ForestAdmin/claudine` itself commits no `.claude/settings.json`, and a session opened there still loads all seven forest skills, off a `scope=user` install.

Two further findings from measuring the mechanism rather than reasoning about it:

- `extraKnownMarketplaces` is lazy. A session never clones from it, so it changes nothing until somebody types `/plugin install` — naming `forest@forest`, which is exactly as unguessable as the repo name the entry was meant to spare them. Someone without access to the private repo sees nothing at all, rather than the per-session resolution failure this PR's description claimed.
- The `"forest@forest": true` pre-enable key installs at **project** scope, silently, at session start, with nobody running `/plugin install`. That is the narrower install shape, obtained without asking.

So the friction this targeted — `/plugin marketplace add ForestAdmin/claudine` — was not the real barrier. Not knowing the toolkit exists is, and no committed file fixes that, as PRD-713 concedes in its own value section. The demo does: two commands, once per person, at user scope, covering every repo forever.

PRD-713 goes back to that. Nothing here is worth the review and maintenance it would carry.

PR #1842, opened 2026-08-21
fix(agent): stop the audit store's queries from adopting a host's CLS transaction

bexchauveto:
## Summary

- Found while enabling `auditTrail` on a CLS-using host: `Sequelize.useCLS(ns)` sets `Sequelize._cls`, a class-level static shared by every `Sequelize` instance in the same loaded copy of the package — including the audit store's own, otherwise-unrelated connection. Any query issued without an explicit `transaction` reads that static (`options.transaction === void 0 && Sequelize._cls`) to decide whether to adopt an ambient transaction. A host using CLS for its own request-scoped transactions therefore had every audit-store query silently run against the **host's** connection, inside the **host's** transaction, instead of the audit store's own pool.
- Verified directly against Sequelize's own source (`Sequelize.prototype.query()`) and reproduced with a fake CLS namespace causing a live `insertPending`/`confirm` call to reach for a bogus hijacked connection instead of its own.
- Consequences: if the audit and host databases differ, the write 404s inside the host's transaction and aborts it (an audited business write breaks even though `critical: false`); if they're the same database, a `pending` row rolls back with the host's transaction, defeating the "a pending row is evidence a write was attempted" guarantee the write protocol is built on.
- Fix: every query the store issues now passes `transaction: null` explicitly, opting out of the CLS lookup unconditionally — `insertPending`, `insertPendingBatch`, `confirm`, `listByRecord`, `countByRecord`, `listDistinctUsers`, `listByCorrelation(s)`, and `authenticate`. Also threads a migration's own transaction through `assertOwnsTable`'s `describeTable` probe, which previously passed no transaction option at all (same class of gap, narrower window).
- A second finding from the same report (pending rows storing `previousValues` unredacted) was investigated and does **not** reproduce: `buildRecord` in `instrument.ts` already applies `redactValues` unconditionally before returning, including for the Before-Update/Before-Delete pending inserts — verified against both current source and the published `1.96.0` dist bundle.

## Test plan

- [ ] `yarn workspace @forestadmin/agent test` — 1439 passing (was 1438; one new regression test added)
- [ ] `yarn workspace @forestadmin/agent lint` — 0 errors
- [ ] New test (`sql-store.test.ts`) sets a fake `Sequelize._cls` namespace with a transaction whose `.connection` getter is spied, then confirms a pending audit row and asserts the getter was never called — verified this test fails without the fix and passes with it.
- [ ] Manually configure an agent with `auditTrail` alongside a host that calls `Sequelize.useCLS(...)` for its own transactions, and confirm an audited write inside a host transaction no longer touches the host's connection.

> [!NOTE]
> ### Stop audit store queries from adopting host CLS transactions
> - Pass `{ transaction: null }` to every Sequelize model operation in [sql-store.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1842/files#diff-7690e57f0c6b0cb4de141c8ccd19e1dc4f197da2d4e1a6135565f246c9e57c0e) (`create`, `bulkCreate`, `update`, `findAll`, `count`, and `authenticate`) so the store never adopts an ambient CLS transaction set by the host process.
> - Forward `context.transaction` through `assertOwnsTable` → `columnNames` in [migrations.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1842/files#diff-49e098e6ade9bb88bf50facbfdf95712bdbd8efdc47ecc96b5c1eb31c230fd49) so the migration ownership/idempotency check runs inside the migration's own transaction instead of outside it.
> - Add a regression test in [sql-store.test.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1842/files#diff-bc1e72a908f58bce866969a059eff4f8a9afd9f3584f6301a49049db403916b8) that injects a fake CLS namespace via `Reflect.set(Sequelize, "_cls", ...)` and asserts `store.confirm` does not touch the hijacked transaction's connection getter.
> - Behavioral Change: audit store queries that previously ran inside a host CLS transaction now always execute on their own connection with no transaction. Migration ownership checks now run transactionally when `context.transaction` is provided.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 245fe35.</sup>
> 
>


arnaud-moncel (approved):
LGTM

PR #1843, opened 2026-08-21
feat(agent-bff): expose full-text search on the list and count endpoints

nbouliol:
Fixes PRD-961

## What

A BFF client can now run the agent's native full-text search by sending `search` and `searchExtended` on `POST /agent/v1/{collection}/list` and `/count`, instead of rebuilding a per-field `Or` tree of `IContains`/`Equal` that cannot cover relations or non-text columns.

- `search?: string` and `searchExtended?: boolean` on `ListRequestBody` and `CountRequestBody`, validated in both parsers against the shared request schemas.
- Both builders emit them under the wire names the agent reads (`search`, `searchExtended`) via a shared `applySearch`.
- OpenAPI: new `Search` and `SearchExtended` components on `ListRequest` and `CountRequest`.

## Decisions

The ticket's open question was settled before implementing (see the [Linear comment](https://linear.app/forestadmin/issue/PRD-961/expose-full-text-search-on-the-bff-list-and-count-endpoints) for the full reasoning):

- **Non-searchable collections: pass-through.** The agent answers `400 validation_error` with `Collection is not searchable`. No `searchable` flag added to `/_internal/capabilities`, no agent change. A fail-closed variant would read the *published* Forest schema, which can be stale relative to the running agent, and would reject searches the agent would accept.
- **Blank search is treated as absent.** The parser validates the type; the builder decides emission via `trim()`. This also smooths over an agent-side inconsistency: `parseSearch` guards on a truthy value, so `''` passes while `'   '` raises on a non-searchable collection. A cleared search box must not depend on how many spaces it holds.
- **`searchExtended` is a strict boolean.** `'true'`, `1`, `'0'` are rejected with `invalid_request`, consistent with `page.limit` requiring a real integer. The agent's string coercion exists because it reads query params; the BFF is a JSON contract.
- **`searchExtended` alone is ignored, not rejected**, and never emitted — so a search-less body produces the exact query it does today.

## Relations gained search, and the ticket's non-goal is void

`parseRelationListRequest` delegates to `parseListRequest`, and the relation handlers call the same builders. Adding search to the type and the builder exposes it on `/relations/{relation}/list` and `/count` automatically — and it works agent-side, since both relation routes build their filter through `ContextFilterFactory` on the foreign collection. Rather than add code to forbid something the agent already supports correctly, this PR accepts it: the relation OpenAPI schemas and tests are updated alongside the top-level ones.

## A search query reaches relation fields on its own

Review surfaced something sharper, since verified against a real agent. The `search` value is a **query**, not a plain term: `FieldsQueryWalker` turns `.` into `:` (`fields-query-walker.ts:9`), `lenientGetSchema` resolves `prefix:suffix` through relations, and `extractSpecifiedFields` runs **outside** the `extended` branch (`search/collection.ts:84`). So `search: 'relation.column:value'` crosses a relation with no `searchExtended` at all, and `assertNoRelationFieldPaths` never sees it — `collectListFieldPaths` only reads projection, filter and sort.

Measured with `authors` removed from the BFF read-model, `books` exposed:

| Path to the hidden collection's columns | Result |
| --- | --- |
| `POST /agent/v1/authors/list` | 404 `unknown_collection` |
| `POST /agent/v1/books/relations/author/list` | 404 `unknown_relation` |
| `filter: { field: 'author:name', operator: 'IContains' }` | 422 `relation_field_not_supported` |
| `search: 'author.name:asimov'` | **200, the two matching books** |

`search` is the only one of the four doors open. The hidden collection's data is not returned — the projection stays on the searched collection — but a caller can test predicates against hidden columns and infer their values from which records come back.

This PR **documents and pins** that behaviour rather than blocking it: the `Search` schema now describes the query syntax, its escape from `relation_field_not_supported`, and that it can filter on a collection the BFF does not expose; two integration tests hold the contrast in place. Blocking it means running the agent's own search-query parse BFF-side — a new runtime dependency on `datasource-customizer` and ANTLR — and rests on a question this PR should not answer alone: is the BFF exposure allow-list a security boundary, or a convenience that trims what a client sees? Tracked in PRD-1037.

## Known gap, deliberately out of scope

The agent returns `meta.decorators` on list — per record, which attributes matched. The BFF loses it twice: JSON:API deserialization drops the root `meta`, and `mapListResponse` only builds `{ data, meta: { countStatus } }`. So search filters correctly but the client cannot tell *why* a record matched.

Carrying it through means exposing a deserializer from `agent-client` (`jsonapi-serializer` is not an `agent-bff` dependency), changing the `AgentDataClient` interface and the `ListResponse` contract, plus a cross-package bump. And the data is low quality: decorators are computed by naive substring match over already-serialized attributes, so a native-search datasource (the Zendesk case) or a `replaceSearch` yields wrong or empty decorators, and a `searchExtended` match in a relation yields none.

PRD-962 should confirm whether the Zendesk data tab needs highlighting for parity with the pre-BFF behaviour.

## Tests

`76 suites, 1268 tests` green from `packages/agent-bff`, which is how CI runs jest.

- **Unit** (`agent-query.test.ts`): parsers accept/reject/absent, builders' exact outgoing parameter names.
- **Route-level** (`data-routes-middleware.test.ts`): list, count, relation list and relation count, asserting the exact query reaching the data client — including that a search-less body is byte-identical to today.
- **Integration against a real in-process agent** (`search-agent.integration.test.ts`, run by its own `bff-integration-tests` job rather than the unit one, following the `llm.integration` precedent): AC #3 ("the count reflects the searched rows") is not provable against a stub, so this boots a real agent with `createTestableAgent` over an ad hoc in-memory fixture and asserts observed behaviour:
  - count without search → 3, with `search: 'foundation'` → 1
  - `search: 'asimov'` alone → 0 records; with `searchExtended: true` → the 2 books by that author, matched through the relation
  - `search: 'author.name:asimov'` crosses the relation with no `searchExtended`, while the same path in a filter draws 422
  - search intersects the filter rather than replacing it
  - a collection with `.disableSearch()` is rejected `400 validation_error` on both list and count, rather than listed in full

The fixture is ad hoc rather than `datasource-dummy`, for two measured reasons recorded in its docblock: dummy's collections never call `enableCount()` and the customizer only exposes `disableCount()`, so count answers `deactivated` — the one assertion the harness exists to make; and `emulateFieldOperator` does not reach the search decorator's operator choice, so the suite would exercise case-sensitive `Contains` instead of the `IContains` production takes (`'Foundation'` matches, `'foundation'` does not).

New devDependencies on `agent-bff`: `@forestadmin/agent`, `@forestadmin/agent-testing`, pinned at the current workspace versions — an exact pin that falls behind stops resolving from the workspace and silently pulls the published package instead.

## Also in this branch

`refactor(agent-bff): declare the request body shape once` (separate commit). `ListRequestSchema` published a shape that `parseListRequest` re-checked by hand. The rules now live once in `src/data/request-schemas.ts`; the parsers validate against them and the OpenAPI schemas decorate them with names and descriptions. `filter` keeps its own path, since its operators are checked against the collection capabilities and a node readable as both leaf and branch has no schema equivalent. The generated document is byte-identical, every previously rejected input is still rejected with the same status, and the messages gained the failing path (`sort.0.direction`, `page.limit`, `projection.0`).

## Unrelated finding

`yarn test` at the repo root is broken on `main`, independently of this PR: `packages/forest-cloud/test/commands/__mocks__/form-data.ts` is a manual mock for a node module, which jest applies project-wide when run from the root. Superagent then evaluates `x instanceof FormData` against an ESM namespace object and 94 supertest assertions fail. CI is green because it runs jest per package. Worth a separate fix.

https://claude.ai/code/session_01SktMf1BRjkFKR3gY25ngJt

> [!NOTE]
> Fixes PRD-961
>
> ## What
>
> A BFF client can now run the agent's native full-text search by sending `search` and `searchExtended` on `POST /agent/v1/{collection}/list` and `/count`, instead of rebuilding a per-field `Or` tree of `IContains`/`Equal` that cannot cover relations or non-text columns.
>
> - `search?: string` and `searchExtended?: boolean` on `ListRequestBody` and `CountRequestBody`, validated in both parsers against the shared request schemas.
> - Both builders emit them under the wire names the agent reads (`search`, `searchExtended`) via a shared `applySearch`.
> - OpenAPI: new `Search` and `SearchExtended` components on `ListRequest` and `CountRequest`.
>
> ## Decisions
>
> The ticket's open question was settled before implementing (see the [Linear comment](https://linear.app/forestadmin/issue/PRD-961/expose-full-text-search-on-the-bff-list-and-count-endpoints) for the full reasoning):
>
> - **Non-searchable collections: pass-through.** The agent answers `400 validation_error` with `Collection is not searchable`. No `searchable` flag added to `/_internal/capabilities`, no agent change. A fail-closed variant would read the *published* Forest schema, which can be stale relative to the running agent, and would reject searches the agent would accept.
> - **Blank search is treated as absent.** The parser validates the type; the builder decides emission via `trim()`. This also smooths over an agent-side inconsistency: `parseSearch` guards on a truthy value, so `''` passes while `'   '` raises on a non-searchable collection. A cleared search box must not depend on how many spaces it holds.
> - **`searchExtended` is a strict boolean.** `'true'`, `1`, `'0'` are rejected with `invalid_request`, consistent with `page.limit` requiring a real integer. The agent's string coercion exists because it reads query params; the BFF is a JSON contract.
> - **`searchExtended` alone is ignored, not rejected**, and never emitted — so a search-less body produces the exact query it does today.
>
> ## Relations gained search, and the ticket's non-goal is void
>
> `parseRelationListRequest` delegates to `parseListRequest`, and the relation handlers call the same builders. Adding search to the type and the builder exposes it on `/relations/{relation}/list` and `/count` automatically — and it works agent-side, since both relation routes build their filter through `ContextFilterFactory` on the foreign collection. Rather than add code to forbid something the agent already supports correctly, this PR accepts it: the relation OpenAPI schemas and tests are updated alongside the top-level ones.
>
> ## A search query reaches relation fields on its own
>
> Review surfaced something sharper, since verified against a real agent. The `search` value is a **query**, not a plain term: `FieldsQueryWalker` turns `.` into `:` (`fields-query-walker.ts:9`), `lenientGetSchema` resolves `prefix:suffix` through relations, and `extractSpecifiedFields` runs **outside** the `extended` branch (`search/collection.ts:84`). So `search: 'relation.column:value'` crosses a relation with no `searchExtended` at all, and `assertNoRelationFieldPaths` never sees it — `collectListFieldPaths` only reads projection, filter and sort.
>
> Measured with `authors` removed from the BFF read-model, `books` exposed:
>
> | Path to the hidden collection's columns | Result |
> | --- | --- |
> | `POST /agent/v1/authors/list` | 404 `unknown_collection` |
> | `POST /agent/v1/books/relations/author/list` | 404 `unknown_relation` |
> | `filter: { field: 'author:name', operator: 'IContains' }` | 422 `relation_field_not_supported` |
> | `search: 'author.name:asimov'` | **200, the two matching books** |
>
> `search` is the only one of the four doors open. The hidden collection's data is not returned — the projection stays on the searched collection — but a caller can test predicates against hidden columns and infer their values from which records come back.
>
> This PR **documents and pins** that behaviour rather than blocking it: the `Search` schema now describes the query syntax, its escape from `relation_field_not_supported`, and that it can filter on a collection the BFF does not expose; two integration tests hold the contrast in place. Blocking it means running the agent's own search-query parse BFF-side — a new runtime dependency on `datasource-customizer` and ANTLR — and rests on a question this PR should not answer alone: is the BFF exposure allow-list a security boundary, or a convenience that trims what a client sees? Tracked in PRD-1037.
>
> ## Known gap, deliberately out of scope
>
> The agent returns `meta.decorators` on list — per record, which attributes matched. The BFF loses it twice: JSON:API deserialization drops the root `meta`, and `mapListResponse` only builds `{ data, meta: { countStatus } }`. So search filters correctly but the client cannot tell *why* a record matched.
>
> Carrying it through means exposing a deserializer from `agent-client` (`jsonapi-serializer` is not an `agent-bff` dependency), changing the `AgentDataClient` interface and the `ListResponse` contract, plus a cross-package bump. And the data is low quality: decorators are computed by naive substring match over already-serialized attributes, so a native-search datasource (the Zendesk case) or a `replaceSearch` yields wrong or empty decorators, and a `searchExtended` match in a relation yields none.
>
> PRD-962 should confirm whether the Zendesk data tab needs highlighting for parity with the pre-BFF behaviour.
>
> ## Tests
>
> `76 suites, 1268 tests` green from `packages/agent-bff`, which is how CI runs jest.
>
> - **Unit** (`agent-query.test.ts`): parsers accept/reject/absent, builders' exact outgoing parameter names.
> - **Route-level** (`data-routes-middleware.test.ts`): list, count, relation list and relation count, asserting the exact query reaching the data client — including that a search-less body is byte-identical to today.
> - **Integration against a real in-process agent** (`search-agent.integration.test.ts`, run by its own `bff-integration-tests` job rather than the unit one, following the `llm.integration` precedent): AC #3 ("the count reflects the searched rows") is not provable against a stub, so this boots a real agent with `createTestableAgent` over an ad hoc in-memory fixture and asserts observed behaviour:
>   - count without search → 3, with `search: 'foundation'` → 1
>   - `search: 'asimov'` alone → 0 records; with `searchExtended: true` → the 2 books by that author, matched through the relation
>   - `search: 'author.name:asimov'` crosses the relation with no `searchExtended`, while the same path in a filter draws 422
>   - search intersects the filter rather than replacing it
>   - a collection with `.disableSearch()` is rejected `400 validation_error` on both list and count, rather than listed in full
>
> The fixture is ad hoc rather than `datasource-dummy`, for two measured reasons recorded in its docblock: dummy's collections never call `enableCount()` and the customizer only exposes `disableCount()`, so count answers `deactivated` — the one assertion the harness exists to make; and `emulateFieldOperator` does not reach the search decorator's operator choice, so the suite would exercise case-sensitive `Contains` instead of the `IContains` production takes (`'Foundation'` matches, `'foundation'` does not).
>
> New devDependencies on `agent-bff`: `@forestadmin/agent`, `@forestadmin/agent-testing`, pinned at the current workspace versions — an exact pin that falls behind stops resolving from the workspace and silently pulls the published package instead.
>
> ## Also in this branch
>
> `refactor(agent-bff): declare the request body shape once` (separate commit). `ListRequestSchema` published a shape that `parseListRequest` re-checked by hand. The rules now live once in `src/data/request-schemas.ts`; the parsers validate against them and the OpenAPI schemas decorate them with names and descriptions. `filter` keeps its own path, since its operators are checked against the collection capabilities and a node readable as both leaf and branch has no schema equivalent. The generated document is byte-identical, every previously rejected input is still rejected with the same status, and the messages gained the failing path (`sort.0.direction`, `page.limit`, `projection.0`).
>
> ## Unrelated finding
>
> `yarn test` at the repo root is broken on `main`, independently of this PR: `packages/forest-cloud/test/commands/__mocks__/form-data.ts` is a manual mock for a node module, which jest applies project-wide when run from the root. Superagent then evaluates `x instanceof FormData` against an ESM namespace object and 94 supertest assertions fail. CI is green because it runs jest per package. Worth a separate fix.
>
>
> https://claude.ai/code/session_01SktMf1BRjkFKR3gY25ngJt
> 
> #### Changes since #1843 opened
>
> - Added `bff-integration-tests` as a required dependency to the `Release packages` and `Publish API Reference` jobs in the build workflow [6b63560]
> 
>


packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -61,22 +61,48 @@ export const TimezoneSchema = z.string().openapi('Timezone', {
     'missing_timezone.',
 });
 
+export const SearchSchema = z.string().openapi('Search', {
+  description:
```
Tonours, line 67: The relation-field warning is on the wrong schema: `search: "author.name:asimov"` reaches an `authors` column with no `searchExtended` at all, while `filter: { field: "author:name" }` on the same path is refused with 422 relation_field_not_supported. Move that paragraph into `Search`, or say here that plain `search` already crosses relations through its `field.subfield:value` syntax.

Mechanism: `datasource-customizer/src/decorators/search/collection.ts:84` calls `extractSpecifiedFields` outside the `extended` branch, and `lenientGetSchema` (:139-164) resolves `prefix:suffix` through ManyToOne/OneToOne/OneToMany. `assertNoRelationFieldPaths` never sees it: `collectListFieldPaths`/`collectCountFieldPaths` only read projection, filter and sort.

Worth deciding beyond the doc: the BFF exposure allow-list is separate from agent permissions (`data-routes-middleware.ts:289-293` says a browsable collection may still be hidden here), so a relation-scoped search can read columns of a collection that list hides, without going through `/relations/`. The hole predates this PR agent-side, but this is the first time it is reachable through the BFF.

packages/agent-bff/src/data/agent-query.ts
```diff
@@ -55,11 +59,29 @@ function assertNoNodeReadableAsBothLeafAndBranch(node: unknown, depth = 0): void
 // Validate the untyped request body before it reaches the query builders, so malformed shapes
 // (e.g. `projection` or `sort` as a string) surface as 400 invalid_request rather than a 500 from
 // an array method blowing up downstream.
+/**
```
Tonours, line 62: `assertValidSearch` slots between the `parseListRequest` comment and the function it documents, so that comment now reads as this helper's. Move the new function above the three `// Validate the untyped request body...` lines.

packages/agent-bff/src/data/agent-query.ts
```diff
@@ -138,6 +162,24 @@ function serializePage(page: BffPage): Record<string, number> {
...
+ *
+ * A blank search is dropped rather than forwarded. The agent's search decorator already treats it
+ * as absent, but `parseSearch` guards on a truthy value, so a whitespace-only search would raise
+ * "Collection is not searchable" on a non-searchable collection while an empty one would not — a
+ * cleared search box must not depend on how many spaces it holds.
+ *
+ * `searchExtended` only ships alongside a real search: on its own it changes nothing agent-side,
+ * and emitting it would alter the outgoing query of every search-less request.
+ */
+function applySearch(query: AgentQuery, body: CountRequestBody): void {
```
Tonours: `applySearch(query, body: CountRequestBody)` is also called from `buildListAgentQuery` with a `ListRequestBody`; it works by structural typing but the signature reads as count-only. Narrow it to `Pick<CountRequestBody, 'search' | 'searchExtended'>`, the two fields it actually touches.

packages/agent-bff/test/openapi/openapi-document.test.ts
```diff
@@ -377,6 +377,48 @@ describe('generateOpenApiDocument', () => {
...
+    expect(propertiesOf(name).searchExtended).toEqual({
+      $ref: '#/components/schemas/SearchExtended',
+    });
+  });
+
+  it.each([
+    ['RelationListRequest', 'ListRequest'],
+    ['RelationCountRequest', 'CountRequest'],
+  ])('should let %s inherit the search inputs from %s', (relation, base) => {
+    expect(refsOf(relation)).toContain(`#/components/schemas/${base}`);
```
Tonours: These two pass unchanged with the `src/` hunks reverted: `RelationListRequestSchema = ListRequestSchema.extend(...)` predates this PR, so `refsOf` only re-asserts zod's inheritance. What about asserting the resolved `Search`/`SearchExtended` refs are reachable from the relation schemas instead?

packages/agent-bff/test/data/agent-query.test.ts
```diff
@@ -113,6 +191,43 @@ describe('parseListRequest', () => {
       expect.objectContaining({ type: 'invalid_request', status: 400 }),
     );
   });
+
```
Tonours, line 198: `toBe(body)` cannot fail: `parseListRequest` returns the same reference whatever it validates, so this only proves nothing threw. Assert on the built query instead (`buildListAgentQuery(...).search`), or drop the identity check for `expect(() => parseListRequest(body)).not.toThrow()`, which at least says what is being tested.

Tonours (commented):
Reviewed against a real agent booted in a throwaway worktree, so the findings below are probed rather than read off the diff.

The integration test never covers `searchExtended: false`, which is the one round-trip where the JSON to query-string boundary bites: superagent serializes it to `"false"` and `agent/src/utils/query-string.ts:115` converts it back through `extended !== 'false'`. I probed it and it works, but it is the only coercion in this PR that a transport refactor could break silently, and the unit test stops before the wire. A case in the integration test would pin it.

Two things I checked and found fine, so nobody re-runs them: `query.search` going out untrimmed is harmless (the agent trims twice, `search/collection.ts:42` and `parse-query.ts:22`), and search skipping `hasCapabilityConstrainedInput` is by design since it names no field the capabilities can invalidate.

The integration test is the good part of this PR: a real agent with the real search decorator, and `disableSearch()` asserted on a collection whose label carries the same term as a book title, so the rejection cannot pass by accident. Exact-object assertions throughout instead of `objectContaining`.

One blocking comment inline on the OpenAPI description, four suggestions.

packages/agent-bff/src/openapi/schemas.ts
nbouliol: Verified, and it is worse than the doc placement. I reproduced the mechanism you describe: `FieldsQueryWalker.enterPropertyMatching` does `.replace(/\./g, ':')` (`fields-query-walker.ts:9`), so `author.name:asimov` becomes the path `author:name`, which `lenientGetSchema` resolves through the ManyToOne — and `extractSpecifiedFields` is indeed called outside the `extended` branch, so no `searchExtended` is needed.

Against a real agent with `authors` removed from the BFF read-model, `books` exposed:

| Path to the hidden collection's columns | Result |
| --- | --- |
| `POST /agent/v1/authors/list` | 404 `unknown_collection` |
| `POST /agent/v1/books/relations/author/list` | 404 `unknown_relation` |
| `filter: { field: 'author:name', operator: 'IContains' }` | 422 `relation_field_not_supported` |
| `search: 'author.name:asimov'` | **200, `["Foundation","I, Robot"]`** |

Search is the only one of the four doors open. The hidden data is not returned (projection stays on `books`), but it is a filtering oracle: with `IContains` you can extract a hidden string column character by character.

Doc fixed as you asked, on `Search` rather than `SearchExtended`, and stated as the mechanism rather than as a side effect of the flag: the value is a query, `column:value` narrows to one column, `relation.column:value` narrows to a related collection's column with no `searchExtended`, it escapes the 422 the same path draws in `filter`/`sort`/`projection`, and the agent resolves the relation against its own schema so a query can filter on a collection this BFF does not expose. `SearchExtended` now points at `Search` for that instead of claiming it.

Two integration tests pin the contrast so it cannot drift silently: the search crossing the relation, and the same path in a filter drawing 422.

On "worth deciding beyond the doc" — agreed, and I did not want to decide it inside this PR. Blocking it needs the relation paths extracted from the query string, which means running the agent's own `parseQuery` + `extractSpecifiedFields` BFF-side: a runtime dependency on `datasource-customizer` and ANTLR, plus keeping the BFF's reading of that grammar in step with the agent's. A regex over the query string is the tempting shortcut and the likely source of both false rejections and missed paths.

It also rests on a question I do not think this PR should answer alone: is the BFF exposure allow-list a security boundary, or a convenience that trims what a client sees? `data-routes-middleware.ts:289-290` reads like the former, and every other route honours it, which is what makes this inconsistent rather than merely undocumented.

Filed as PRD-1037 with the evidence, both options (authorize the relation target against the read-model, vs reject any relation path) and that open question. If you read the allow-list as a real boundary, say so and I will treat PRD-1037 as a prerequisite to this merge rather than a follow-up.

packages/agent-bff/src/data/agent-query.ts
nbouliol: Right, and I had not noticed. The `// Validate the untyped request body...` lines documented `parseListRequest` and my helper landed between the two, so they now read as its docblock. `assertValidSearch` moved above them.

packages/agent-bff/src/data/agent-query.ts
nbouliol: Agreed, it works only by structural typing and the signature lies about it. Narrowed to `Pick<CountRequestBody, 'search' | 'searchExtended'>`, the two fields it reads.

packages/agent-bff/test/openapi/openapi-document.test.ts
nbouliol: Correct — they asserted zod's `.extend()`, not my change. Replaced with a `resolvedPropertiesOf` that walks `allOf` into the components the refs point at, so the relation bodies are checked on the properties a client actually resolves.

Checked the fix the way you checked the flaw: with `src/openapi/schemas.ts` reverted to its pre-PR version, all 8 OpenAPI tests fail, including the two relation ones. They could not fail before.

packages/agent-bff/test/data/agent-query.test.ts
nbouliol: Right, `parseListRequest` returns `body as ListRequestBody` — the same reference whatever it validates, so the assertion was unfalsifiable. I took neither of your two options: `toMatchObject({ search, searchExtended })` can actually fail, since it also proves the parser passes the fields through instead of stripping them. The values reaching the wire stay asserted in the builder describe.

Left the pre-existing `toBe(body)` on line 174 alone — same weakness, but not mine to change in this PR.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -88,7 +118,7 @@ export const RelationListRequestSchema = ListRequestSchema.extend({
   parentId: ParentIdSchema,
 }).openapi('RelationListRequest', {
   description:
-    'Filter, sort and projection apply to the FOREIGN collection; the parent only resolves ' +
+    'Filter, sort, projection and search apply to the FOREIGN collection; the parent only resolves ' +
```
Tonours, line 123: PRD-961's non-goal is explicit: no search on any other BFF endpoint. This publishes it on relation list and count, with tests pinning it. `const { search, searchExtended, ...rest } = parseListRequest(body)` in `parseRelationListRequest` withholds it; otherwise let's amend the non-goal on the ticket before merge.

packages/agent-bff/test/data/fixtures/in-memory-collection.ts
```diff
@@ -0,0 +1,87 @@
...
+  'In',
+  'Present',
+]);
+
+/**
+ * A read-only in-memory collection: enough to let the agent's own search decorator build a real
+ * condition tree and apply it. `searchable` stays false, as on any datasource with no native
+ * search, which is the path the decorator implements itself.
+ */
+export default class InMemoryCollection extends BaseCollection {
```
Tonours, line 39: This re-implements `datasource-dummy/src/collections/base.ts` with a different operator set. Both reasons for the fork are curable from the test: search sits above `lateOpEmulate` (`decorators-stack.ts:45`), so `emulateFieldOperator('title', 'IContains')` reaches the search decorator, and `createDummyDataSource()` does declare relations. `createDummyDataSource()` + `emulateFieldOperator` + `disableSearch()` on one collection drops these 139 lines and both fixture files.

packages/agent-bff/package.json
```diff
@@ -41,6 +41,8 @@
     "zod": "4.3.6"
   },
   "devDependencies": {
+    "@forestadmin/agent": "1.96.0",
```
Tonours: These pins miss the workspace versions (1.97.0 / 1.2.0), so yarn resolves them from npm instead of linking locally and the integration test boots a published agent rather than `packages/agent`. Pin `1.97.0` / `1.2.0` like `forest-cloud/package.json:6`, and commit the `yarn.lock` update: CI only hides it because it runs plain `yarn` with no `--frozen-lockfile`.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -61,22 +61,52 @@ export const TimezoneSchema = z.string().openapi('Timezone', {
...
+    "The agent's native full-text search, applied on top of `filter` rather than instead of it. " +
+    'An empty or whitespace-only value is treated as absent, so clearing a search box is not an ' +
+    'error. Searching a collection whose search is disabled is not rejected here: the agent ' +
+    'answers 400 validation_error with "Collection is not searchable". The response does not say ' +
+    'which field matched. The value is a query, not a plain term: `column:value` narrows the ' +
+    'search to one column, and `relation.column:value` narrows it to a column of a related ' +
+    'collection — so a search reaches relation fields on its own, with no `searchExtended`, and ' +
+    'escapes the 422 relation_field_not_supported that the same path draws in `filter`, `sort` or ' +
+    '`projection`. The agent resolves a relation named in a query against its own schema, so a ' +
+    'query can filter on a column of a collection this BFF does not expose.',
```
Tonours: `search: 'author.name:asimov'` filters on a collection the read-model excludes, and `collectListFieldPaths` never inspects the search value so `assertNoRelationFieldPaths` cannot fire, while the same path in `filter` returns 422. Shipping that sentence as public contract locks in an open authorization gap; can we get the epic owner's call on PRD-1037 first and hold the line until then?

packages/agent-bff/src/data/agent-query.ts
```diff
@@ -52,6 +56,22 @@ function assertNoNodeReadableAsBothLeafAndBranch(node: unknown, depth = 0): void
   }
 }
 
+/**
+ * The agent reads `search`/`searchExtended` from query params, so it coerces both from strings.
+ * The BFF is a JSON contract: a real boolean is required here, like `page.limit` requires a real
+ * integer. Only the type is checked — whether a blank search is worth sending is the builder's
+ * call, so a cleared search box parses the same way as an absent one.
+ */
+function assertValidSearch(search: unknown, searchExtended: unknown): void {
```
Tonours: `ListRequestSchema` in `src/openapi/schemas.ts` already declares this shape, so the hand-rolled check is a second source of truth. Parsing just the flat inputs (`projection`, `sort`, `page`, `search`, `searchExtended`) against a runtime zod schema and leaving `filter` on the current path removes ~40 lines here plus the `return body as ListRequestBody` casts at `:121` and `:135`. A full zod parse doesn't work: operator validation is owned by `capabilities-validator.ts:88-90` and `assertNoNodeReadableAsBothLeafAndBranch` has no schema equivalent.

packages/agent-bff/src/data/agent-query.ts
```diff
@@ -138,6 +162,27 @@ function serializePage(page: BffPage): Record<string, number> {
...
+ * cleared search box must not depend on how many spaces it holds.
+ *
+ * `searchExtended` only ships alongside a real search: on its own it changes nothing agent-side,
+ * and emitting it would alter the outgoing query of every search-less request.
+ */
+function applySearch(
+  query: AgentQuery,
+  body: Pick<CountRequestBody, 'search' | 'searchExtended'>,
+): void {
+  if (!body.search?.trim()) return;
```
Tonours: The guard trims but `:182` forwards the untrimmed value, so the trim decision lives in the builder while the type check lives in the parser. `search?.trim() || undefined` in both parsers collapses `applySearch` to two unconditional lines and removes this branch.

packages/agent-bff/test/data/search-agent.integration.test.ts
```diff
@@ -0,0 +1,281 @@
+import type { Logger } from '../../src/ports/logger-port';
+import type { SchemaFetcher } from '../../src/read-model/forest-schema-client';
+import type { TestableAgent } from '@forestadmin/agent-testing';
+import type { ForestSchemaCollection } from '@forestadmin/forestadmin-client';
+
+import { createTestableAgent } from '@forestadmin/agent-testing';
```
Tonours, line 6: This boots a real agent inside the unit jest project (`jest.config.ts` matches `test/**/*.test.ts`), so `yarn test` now opens HTTP with a 60s timeout, and it makes `@forestadmin/agent` a devDep of a package the agent otherwise sits above. Renaming to `*.integration.test.ts` matches `ai-proxy/test/llm.integration.test.ts`, which `build.yml:111` already excludes from the unit job.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -61,22 +61,52 @@ export const TimezoneSchema = z.string().openapi('Timezone', {
     'missing_timezone.',
 });
 
+export const SearchSchema = z.string().openapi('Search', {
```
Tonours: Ten lines of description against two or three for every sibling (`PageSchema`, `TimezoneSchema`). Keeping the two client-actionable sentences (blank value treated as absent, non-searchable answered by the agent) and moving the query-syntax detail to the endpoint docs reads better in Redoc.

packages/agent-bff/test/openapi/openapi-document.test.ts
```diff
@@ -377,6 +377,75 @@ describe('generateOpenApiDocument', () => {
...
+    'should resolve the search inputs on %s, which shares the parsers and builders',
+    name => {
+      expect(resolvedPropertiesOf(name).search).toEqual({ $ref: '#/components/schemas/Search' });
+      expect(resolvedPropertiesOf(name).searchExtended).toEqual({
+        $ref: '#/components/schemas/SearchExtended',
+      });
+    },
+  );
+
+  it('should say a blank search is treated as absent rather than rejected', () => {
```
Tonours: These six assert substrings of a description, so rewording the doc reds the build with no behaviour change. The `$ref` tests above already pin the contract; what about dropping the prose ones?

packages/agent-bff/test/data/fixtures/in-memory-collection.ts
nbouliol: Both your reasons were right and mine were wrong — `createDummyDataSource()` does add `addManyToOneRelation('author', 'persons', { foreignKey: 'authorId' })`, I had only looked at the raw `books.ts` class; and `search` is at `decorators-stack.ts:49` above `lateOpEmulate` at `:45`, so the emulated operator is in reach. I tried it, and it still does not work, for two measured reasons.

`datasource-dummy` collections never call `enableCount()`, and the customizer only exposes `disableCount()` — there is no supported way to enable it. So with dummy:

```
POST /agent/v1/books/count {}                       -> {"count":null,"countStatus":"deactivated"}
POST /agent/v1/books/count {"search":"foundation"}   -> {"count":null,"countStatus":"deactivated"}
```

AC #3 ("the count reflects the searched rows, not the whole collection") is the one assertion this harness exists to make, and it becomes unassertable.

And `emulateFieldOperator` does not reach the search decorator's operator choice, whatever the stack order suggests:

```
search: 'foundation'  -> []                    // lowercase
search: 'Foundation'  -> [{"id":"2","title":"Foundation"}]   // exact case
```

So the suite would silently exercise case-sensitive `Contains` instead of the `IContains` the decorator prefers in production. I did not chase why the emulation does not land, since the count blocker already settles it.

Kept the fixture, with both measurements recorded in its docblock so the next reader does not have to redo them. If you want dummy to serve this, the prerequisite is `enableCount()` in `BaseDummyCollection` — which changes what `_example` does, so not from here.

packages/agent-bff/test/data/search-agent.integration.test.ts
nbouliol: Agreed on the substance, and I had missed the precedent — thanks. Done: renamed to `search-agent.integration.test.ts` and given its own job, modelled on `llm-integration-tests` but without `continue-on-error`, since it reaches no third party and must gate.

One correction on the mechanics, because the shortcut is dangerous. `:111` ignores `llm.integration` specifically, so the rename alone excludes nothing — and widening the pattern to `integration` would silently stop running hundreds of genuine unit tests:

```
datasource-mongoose/test/integration/**        (20 files)
datasource-replica/test/integrations/**        (10 files)
ai-proxy/test/integrations/**                  (18 files)
datasource-customizer/test/decorators/write/integration/**
datasource-mongo, datasource-sql, datasource-sequelize, agent-client/test/integration/**
agent/test/agent-integration.test.ts, workflow-executor/test/integration/**
```

Most packages keep ordinary unit tests under `test/integration*/`. So the ignore list now names each suite that has its own job — `'llm.integration|search-agent.integration'` — with a comment saying why it is not a broad match. Verified: the unit job selects 75 files, the integration job selects exactly one.

packages/agent-bff/package.json
nbouliol: Bumped to `1.97.3` / `1.2.3`. The mechanism you describe is exactly right and the pins were rotting, but the conclusion did not hold at the commit you read: the branch was based on `6124b4197`, where the workspace was at exactly `1.96.0` / `1.1.79`, so yarn linked them and the suite did boot `packages/agent`. Checked rather than assumed —

```
node_modules/@forestadmin/agent -> ../../packages/agent     # symlink
packages/agent-bff/node_modules/@forestadmin                # does not exist
grep '@forestadmin/agent@' yarn.lock                        # no match
```

The real problem is the one your comment implies: an exact pin on a workspace sibling that releases often goes stale on the next rebase and then silently resolves from npm. `main` has since released past both, which is what you were looking at.

On the lockfile, though: nothing to commit. A satisfied pin resolves to the workspace link, which has no `yarn.lock` entry — after the bump, `yarn install` leaves the lockfile untouched and no nested copy appears. The convention is the exact pin (`forest-cloud/package.json:6`, `agent`'s own devDeps on `datasource-sql: 1.17.14`), kept in step by `multi-semantic-release --deps.bump=override` at `build.yml:248`.

packages/agent-bff/src/data/agent-query.ts
nbouliol: Done, in its own commit (`43354ae0f`) so it stays separable from the search work.

The rules now live once in `src/data/request-schemas.ts`; the parsers check bodies against them, and `openapi/schemas.ts` decorates the same schemas with names and descriptions. `filter` keeps its own path for the two reasons you named.

One wrinkle worth recording: sharing them needs a single extended zod instance. A schema built from a plain `zod` import has no `.openapi()` — `extendZodWithOpenApi` does not reach it — so the extension moved to `src/zod.ts` and `openapi/zod-openapi.ts` re-exports it, which also keeps the request schemas from depending on the openapi folder.

The `body as` casts stayed. `safeParse` is used for validation only and the parsers still return the caller's body by reference: `parseRelationListRequest` spreads it and a zod-rebuilt object would strip `filter`, `timezone` and `parentId`.

Behaviour checked rather than assumed:

- the generated document is **byte-identical** before and after (46504 bytes, empty diff), so the published contract did not move
- all 11 inputs the hand-rolled checks rejected are still rejected, same status
- the messages gained the failing path, which the old ones never named: `sort.0.direction: Invalid option: expected one of "asc"|"desc"`, `page.limit: Too small: expected number to be >0`, `projection.0: Invalid input: expected string, received number`. Pinned by a new test.

37 lines shorter net.

packages/agent-bff/src/data/agent-query.ts
nbouliol: Fixed: `applySearch` now sends the trimmed value instead of trimming only to decide.

Kept the trim in the builder rather than the parsers, though. `parseListRequest` returns the caller's body by reference — `parseRelationListRequest` spreads it, and a pre-existing test asserts the identity — so normalising there means either mutating the request body or rebuilding it, and rebuilding strips `filter`, `timezone` and `parentId`. So the parsers stay pure validation and the emission owns the trim, which is documented on the function.

packages/agent-bff/src/openapi/schemas.ts
nbouliol: Amended the ticket rather than withholding it — the decision to accept relation search was taken deliberately before implementing, it just never made it out of a Linear comment and into the description. PRD-961's Non-goals now strikes that line and records why: relation list and count share the parsers (`parseRelationListRequest` delegates to `parseListRequest`) and the builders, so search reaches them whether we want it or not, and it works agent-side since both relation routes build their filter through `ContextFilterFactory` on the foreign collection. Withholding it would mean writing code to forbid something the agent already supports correctly.

.github/workflows/build.yml
```diff
@@ -147,6 +150,35 @@ jobs:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
 
+  # Boots a real agent on a local HTTP port, so it stays out of the unit job. Unlike the LLM suite
+  # it reaches no third party and is deterministic, hence no continue-on-error: it must gate.
+  bff-integration-tests:
```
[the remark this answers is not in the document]
nbouliol: Correct, and it caught me contradicting my own comment. Fixed in 6b63560dd: `bff-integration-tests` added to both `release` and `publish-api-reference`.

Worth naming the regression precisely, since it is worse than a missing gate on a new job. The suite used to run inside the `test` matrix, and `test` gates both through `send-coverage`. Moving it into its own job therefore *removed* a gate that already existed, while the comment I wrote on that job claimed the opposite. `llm-integration-tests` is absent from those `needs` lists on purpose — it has `continue-on-error: true` because it calls real LLM APIs — so copying its shape is what dropped the gating.

Verified the chain parses and resolves: both jobs now need `[build-api-reference, send-coverage, lint, bff-integration-tests]`, and the job still carries no `continue-on-error`.

Tonours (approved):
LGTM.

PR #1845, opened 2026-08-24
feat(agent-bff): proxy the ai query on post /agent/v1/ai/query

Tonours:
fixes PRD-1033

## What

`POST /agent/v1/ai/query` relays the body unchanged to `/api/ai-proxy/ai-query`, authenticated with the OAuth session the BFF already holds. Second half of PRD-944; `GET /agent/v1/context` shipped in #1838.

## Why

After the OAuth switch the Zendesk app has no SaaS token left — it sends its `bff_access` to `api.forestadmin.com`, which rejects a JWT it never issued. The provider keys live on the SaaS, so this one route is a relay rather than a contract.

## How

- Outgoing headers are a **closed allow-list** built server-side: `Authorization` (session `saasAccessToken` via `ensureFreshServerAccess`), `Content-Type`, `forest-environment-id`, `forest-rendering-id`. Nothing incoming is copied — the SaaS prefers a session cookie over the Bearer (`make-fetch-token.ts:26-30`), and `x-mcp-oauth-tokens` is injected as `Authorization` toward an MCP server.
- Upstream path is fixed. Never a dynamic `:route`.
- **Session required.** An API key gets `403 oauth_required`: it produces no `sid`, so there is no `saasAccessToken` to forward. A third-party app authenticating via OAuth will work with no code change.
- **1 MB body on this route only.** A scoped parser mounted before the global one — `@koa/bodyparser` skips when `ctx.request.body` is set. Data routes keep 16 KB, `/oauth/token` keeps parsing. The global 16 KB breaks AI prompts past ~30 collections (measured).
- **No upstream body reaches the iframe unless it is JSON and below 500.** A WAF answers HTML on `403`/`429`; status is preserved, body replaced. Timeout → `504`, bounded by `BFF_AI_TIMEOUT_MS` (default 120s) — nothing else on the chain bounds an AI generation today.
- `ai-name` is a constant, not an env var: it selects the SaaS billing line and model, so a caller must not choose it.
- **The OpenAPI document publishes this path only when OAuth is configured**, since that is the only case where the route is mounted. The `forest-bff openapi` export therefore omits it: it cannot know the runtime configuration.

## Scope and safety

`ALLOWED_HEADERS` is deliberately **unchanged**. The `forest-*` headers the app sends today are exactly what the BFF now derives server-side; letting them through the preflight would mean proving we ignore them. The app drops them instead.

No body validation (the SaaS validates, and its schema is deliberately loose), no rate limiting, no session store change.

## How to test

```
yarn workspace @forestadmin/agent-bff test   # 1250 tests
```

Two tests exist only to catch a mount-order inversion, and both were verified to fail when the lines are swapped: "900 KB passes" (a `2 MB → 413` test does **not** catch it — both orders give 413), and the `runCli` test reaching the route without `X-Forest-Timezone`.

## Known limitation

Not reachable from the Zendesk app until its own switch lands: `fetchAIProxy` still targets the SaaS and still sends `forest-*` headers, which this preflight rejects. Separate PR, same ticket.

Behind the BFF every user exits on one IP, so the SaaS per-IP anti-brute-force becomes a shared quota. Degrades cleanly (the `429` is relayed), worth watching in production.

### General

### Security


packages/agent-bff/test/ai/ai-routes-middleware.test.ts
```diff
@@ -0,0 +1,338 @@
...
+  authMode = 'oauth',
+  renderingId = '42',
+  environmentId = 7,
+}: AppOptions = {}) {
+  const sessionStore = store ?? makeSessionStore({ saasAccessToken: freshAccessToken() });
+  const logger = jest.fn();
+  const app = new Koa();
+
+  app.use(createErrorMiddleware({ logger: () => undefined }));
+  app.use(bodyParser({ jsonLimit: '1mb', enableTypes: ['json'] }));
```
nbouliol: Importing `AI_BODY_LIMIT` here instead of the literal would make the fixture track the real limit.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -95,6 +95,132 @@ describe('runCli', () => {
...
+
+    it('should reject a body above 1mb', async () => {
+      const server = await runCli(OAUTH_ENV, noopLogger);
+
+      try {
+        const response = await request(server.callback)
+          .post('/agent/v1/ai/query')
+          .set('Authorization', `Bearer ${sessionToken()}`)
+          .set('Content-Type', 'application/json')
+          .set('Content-Length', String(2 * 1024 * 1024))
```
nbouliol, line 213: This declares a length it never sends, so `raw-body` short-circuits on the header and the byte-counting path is never exercised; with the pass case at 900 KB, neither test sits at the boundary — a real ~1.05 MB body would cover both sides.

packages/agent-bff/test/ai/ai-proxy-client.test.ts
```diff
@@ -0,0 +1,196 @@
...
+    beforeEach(() => {
+      global.fetch = jest
+        .fn()
+        .mockResolvedValue(jsonFetchResponse(200, { choices: [] })) as unknown as typeof fetch;
+    });
+
+    it('should target the fixed ai-query path with the ai-name of this deployment', async () => {
+      await makeClient().query(params());
+
+      expect(firstCall()[0]).toBe(`${FOREST_SERVER_URL}/api/ai-proxy/ai-query?ai-name=zendesk`);
```
nbouliol, line 49: What about a `POST /agent/v1/ai/query?ai-name=other` case asserting the outgoing URL still carries `zendesk`? Both `ctx.path` comparisons would break silently on a refactor to `ctx.url`.

packages/agent-bff/src/ai/ai-routes-middleware.ts
```diff
@@ -0,0 +1,148 @@
...
+  upstreamTimeout,
+  upstreamUnreachable,
+} from '../http/bff-local-errors';
+import { OAuthRequestError } from '../oauth/oauth-error';
+import ensureFreshServerAccess from '../oauth/session-lifecycle';
+
+export const AI_QUERY_ROUTE = '/agent/v1/ai/query';
+
+const JSON_CONTENT_TYPE = 'application/json';
+const MAX_CAUSE_DEPTH = 3;
```
nbouliol, line 24: No test walks a cause chain deeper than 3, so nothing proves the recursion actually stops.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -249,6 +249,20 @@ export const ContextResponseSchema = z
       'deployment runs without the OAuth configuration.',
   });
 
+export const AiQueryRequestSchema = z
```
nbouliol: Nothing validates against this, yet the served contract now carries a BFF-authored shape nobody will keep in sync with `routeArgsSchema` — `z.unknown()`, as used for the 200, would match the passthrough decision.

packages/agent-bff/test/config/env-config.test.ts
```diff
@@ -235,24 +236,63 @@ describe('parseConfig', () => {
...
     ])('should throw ConfigurationError for a %s value', (_label, value) => {
       expect(() => parseConfig({ ...VALID_ENV, BFF_AGENT_TIMEOUT_MS: value })).toThrow(
         ConfigurationError,
       );
     });
   });
 
+  describe('BFF_AI_TIMEOUT_MS', () => {
+    it('should default to 2 minutes when unset, since an AI generation is slow', () => {
+      expect(parseConfig({ ...VALID_ENV }).aiTimeoutMs).toBe(120_000);
```
nbouliol: `DEFAULT_AI_TIMEOUT_MS` is already imported and used three cases below — using it here too keeps the default in one place.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -186,6 +312,71 @@ describe('runCli', () => {
       }
     });
 
+    it('should not mount the AI query route without the OAuth configuration, since it has no session', async () => {
+      const token = jsonwebtoken.sign(
```
nbouliol: A `sessionToken()` helper already exists in the sibling describe — reusing it drops three copies of this block.

packages/agent-bff/README.md
```diff
@@ -121,9 +125,35 @@ normalized away, no trailing slash, no wildcard, no subdomain matching):
...
+
+`POST /agent/v1/ai/query` relays its body to the Forest AI proxy under the OAuth session the BFF
+holds, and relays the answer back:
+
+- **Session only.** An API key is refused with `403 oauth_required`: only the OAuth flow yields the
+  Forest access token this route forwards. The environment and rendering are derived server-side, so
+  sending `forest-*` headers has no effect.
+- **Body.** `1mb` instead of the `16kb` of every other route, and `application/json` only — any other
+  content type arrives as an empty body, is relayed upstream as `{}`, and the Forest server rejects
+  the query. Over the limit → `413`.
```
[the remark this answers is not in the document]
Tonours: Rejecting this one, with a measurement.

`ctx.request.body` is not `undefined` for a non-JSON content type. `@koa/bodyparser` returns `{}` when no body type matches (`dist/body-parser.mjs:93`, `if (!bodyType) return {}`) and then assigns it (`:130`, `"parsed" in response ? response.parsed : {}`). So the scoped json-only parser leaves `{}`, and `JSON.stringify({})` is `'{}'` — `fetch` sends a two-byte body, not no body.

Measured against the real selector from `cli-core.ts`, three non-JSON requests:

```
form             -> ctx.request.body = {} , fetch sends "{}"
text/plain       -> ctx.request.body = {} , fetch sends "{}"
no content-type  -> ctx.request.body = {} , fetch sends "{}"
```

The sentence as written is accurate; the suggested patch would make it wrong. Keeping it.

For what it is worth, this same `if (!bodyType) return {}` line has now produced two opposite wrong conclusions on this PR — an earlier review claimed the generic parser was parsing the form body at its 56 kb limit. It is worth checking before reasoning about what reaches the upstream.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -186,12 +361,63 @@ describe('runCli', () => {
...
+
+      try {
+        const outcome = await oversizedBodyOutcome(
+          server.callback,
+          '/agent/v1/ai/query',
+          token,
+          900_000,
+        );
+
+        expect(outcome).toMatch(/^(413|aborted:)/);
```
nbouliol: This was a hard `toBe(413)` before — the 16 kb parser rejects 900 kB long before the client finishes writing, so de-flaking the 1.05 MB sibling cost a strict assertion on a test that was not flaky.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -95,6 +123,157 @@ describe('runCli', () => {
...
+
+      try {
+        const outcome = await oversizedBodyOutcome(
+          server.callback,
+          '/agent/v1/ai/query',
+          sessionToken(),
+          1_050_000,
+        );
+
+        expect(outcome).toMatch(/^(413|aborted:)/);
```
nbouliol: This passes on a real 413 and on any thrown error alike, since the helper's `catch` maps every exception to `aborted:<code ?? 'unknown'>` — and it is the only test at the true byte-counting boundary. Could we assert the abort code we actually expect?

packages/agent-bff/test/openapi/openapi-spec-validity.test.ts
```diff
@@ -14,8 +14,12 @@ const REDOCLY_BIN = path.join(
 );
 
 describe.each([
-  ['generic', () => generateOpenApiDocument('1.0.0')],
-  ['unfolded', () => generateOpenApiDocument('1.0.0', unfoldingFixture())],
+  ['generic', () => generateOpenApiDocument('1.0.0', { hasAiQueryRoute: true })],
```
nbouliol, line 17: Both rows now pin `hasAiQueryRoute: true`, so redocly never lints the document a non-OAuth deployment serves — the branch the conditional registration exists to protect. A third row with `false` would cover it.

packages/agent-bff/src/cli-core.ts
```diff
@@ -56,6 +59,24 @@ function agentScoped(middleware: Middleware): Middleware {
...
 
+function createBodyParser(hasAiQueryRoute: boolean): Middleware {
+  const parseBody = bodyParser({ jsonLimit: BODY_LIMIT });
+
+  if (!hasAiQueryRoute) return parseBody;
+
+  const parseAiBody = bodyParser({ jsonLimit: AI_BODY_LIMIT, enableTypes: ['json'] });
+
+  return async function selectedBodyParser(ctx, next) {
+    if (ctx.path === AI_QUERY_ROUTE) {
```
nbouliol, line 70: Selected by path alone, this runs ahead of `createAuthModeMiddleware`, so an unauthenticated caller gets 1 MB buffered and JSON-parsed — 64x the pre-auth surface of every other route. Worth a deliberate note even if rate limiting is a non-goal.

packages/agent-bff/test/ai/ai-routes-middleware.test.ts
```diff
@@ -0,0 +1,382 @@
...
+        .set('Authorization', 'Bearer incoming-bff-access-token')
+        .send({ messages: [] });
+
+      expect(query).toHaveBeenCalledWith(
+        expect.objectContaining({ saasAccessToken, renderingId: 42, environmentId: 7 }),
+      );
+    });
+
+    it('should keep the deployment ai-name on the outgoing url whatever the caller asks for', async () => {
+      global.fetch = jest.fn().mockResolvedValue({
```
nbouliol: This file has no `afterEach`, unlike `ai-proxy-client.test.ts` and `cli-core.test.ts`, so the stub leaks into every later test here.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -95,6 +123,157 @@ describe('runCli', () => {
...
+          .send(`messages=${'x'.repeat(100_000)}`);
+
+        expect(response.status).not.toBe(413);
+        expect(response.status).toBe(401);
+      } finally {
+        await server.stop();
+      }
+    });
+
+    it('should reject a body above 1mb on its declared length, without reading it', async () => {
```
nbouliol: Nothing here shows the body went unread — 15 bytes are sent behind a 2 MB `Content-Length`, so the test cannot tell. Drop the clause, or assert the route was never reached?

packages/agent-bff/test/ai/ai-routes-middleware.test.ts
```diff
@@ -0,0 +1,382 @@
...
+
+    if (authMode === 'oauth') {
+      ctx.state.principal = { sid: 'session-id', rendering_id: renderingId };
+    }
+
+    await next();
+  });
+  app.use(
+    createAiRoutesMiddleware({
+      client: client ?? ({ query } as unknown as AiProxyClient),
```
nbouliol: Two competing seams here: when `client` is passed, the `query` mock `makeApp` returns is dead and records nothing, silently.

packages/agent-bff/test/cli-core.test.ts
Tonours: It was flaky too — the premise was wrong on my side as well, since I only saw it break on the 1.05 MB sibling. Probed the 900 kB one 6x locally: 5 runs returned `413`, one returned `aborted:EPIPE`. Same race, the server answers 413 and destroys the socket while supertest is still writing; 900 kB is just small enough to usually flush first.

So I did not restore `toBe(413)` — it would reintroduce the flake. Both tests now assert against a closed set instead of a prefix regex:

```ts
const OVERSIZED_BODY_OUTCOMES = ['413', 'aborted:EPIPE', 'aborted:ECONNRESET'];
expect(OVERSIZED_BODY_OUTCOMES).toContain(outcome);
```

`ECONNRESET` is there for Linux CI. Any other exception — a TypeError, an assertion inside the helper — now fails.

packages/agent-bff/test/cli-core.test.ts
Tonours: Done, same closed set as the sibling: `[413, aborted:EPIPE, aborted:ECONNRESET]` instead of `/^(413|aborted:)/`. An unexpected throw no longer passes.

packages/agent-bff/test/openapi/openapi-spec-validity.test.ts
Tonours: Added a third row, `ai-less`, with `hasAiQueryRoute: false`. Redocly lints it clean (suite green, 3 documents).

packages/agent-bff/src/cli-core.ts
Tonours: Agreed, and it is worse than 64x in intent: the parser is picked on `ctx.path` alone, so the buffering happens for a caller with no credentials at all. Documented it in the README next to the body limit rather than in code — bounding it is a reverse-proxy / rate-limiting concern, not something the BFF can fix without moving the parser behind auth (which would mean parsing twice or a custom raw-body read).

packages/agent-bff/test/ai/ai-routes-middleware.test.ts
Tonours: Fixed: `const originalFetch = global.fetch` captured at the top of the describe, restored in `afterEach`. Note `jest.restoreAllMocks()` would not have helped here either — `global.fetch = jest.fn()` is an assignment, not a spy, so the two other files leak the same way; this one is now the only one that actually restores it.

packages/agent-bff/test/cli-core.test.ts
Tonours: Renamed to `before the route is reached` — the 413 already proves that, since the mounted route answers 401 `session_expired` on this env. Dropped the unproven claim rather than adding an assertion the 15-byte body cannot support.

packages/agent-bff/test/ai/ai-routes-middleware.test.ts
Tonours: Made them mutually exclusive at the type level:

```ts
type AppOptions = AppContext &
  ({ query?: jest.Mock; client?: never } | { client: AiProxyClient; query?: never });
```

`makeApp({ query, client })` is now a compile error, so the dead mock cannot be read by mistake.

packages/agent-bff/README.md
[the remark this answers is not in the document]
Tonours: Not reproducible — the README is correct and the finding rests on a wrong premise about `@koa/bodyparser`.

`parseBody` returns `{}` when no enabled type matches the content-type, and the middleware then assigns `ctx.request.body = "parsed" in response ? response.parsed : {}`. With `enableTypes: ['json']`, a form request on the AI path has `bodyType === null`, so `ctx.request.body` is `{}`, not `undefined`.

Probed it against the real scoped parser rather than reading the source only:

```
seen: {}
JSON.stringify(seen): "{}"
```

So `AiProxyClient.query` sends the literal `{}` upstream, exactly as documented. No change.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -95,6 +125,157 @@ describe('runCli', () => {
...
+
+    it('should keep parsing /oauth/token, which the scoped parser must not shadow', async () => {
+      const server = await runCli(OAUTH_ENV, noopLogger);
+
+      try {
+        const response = await request(server.callback)
+          .post('/oauth/token')
+          .send({ grant_type: 'refresh_token' });
+
+        expect(response.body.error_description).toContain('refresh_token');
```
nbouliol: Both branches satisfy this: parsed gives `invalid_request` "Missing required parameter: refresh_token", shadowed gives `unsupported_grant_type` "Only authorization_code and refresh_token are supported" — the word is in both. `expect(response.body.error).toBe('invalid_request')` separates them.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -95,6 +125,157 @@ describe('runCli', () => {
...
   });
 
+  describe('when the AI query route is mounted', () => {
+    const OAUTH_ENV = {
+      ...VALID_ENV,
+      BFF_TOKEN_ENCRYPTION_KEY: Buffer.alloc(32).toString('base64'),
+    } satisfies NodeJS.ProcessEnv;
+
+    beforeEach(() => {
+      global.fetch = jest.fn().mockResolvedValue({
```
nbouliol: This 6-line stub is byte-identical at `:93`, `cli-ai-wiring.test.ts:23` and `openapi-routes.test.ts:128` — three of the four added here. One `stubEnvironmentIdFetch()` would carry them all.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -95,6 +125,157 @@ describe('runCli', () => {
...
+        ok: true,
+        status: 200,
+        statusText: 'ok',
+        json: async () => ({ data: { id: '42' } }),
+      }) as unknown as typeof fetch;
+    });
+
+    afterEach(() => {
+      jest.restoreAllMocks();
```
nbouliol: `restoreAllMocks` only undoes `jest.spyOn`, so the direct `global.fetch` assignment survives it. Same at `openapi-routes.test.ts:124` and `cli-ai-wiring.test.ts:32` — the `originalFetch` pattern just added at `ai-routes-middleware.test.ts:92` is the one that works.

packages/agent-bff/test/ai/ai-routes-middleware.test.ts
```diff
@@ -0,0 +1,389 @@
...
+      sessionStore,
+      serverClient,
+      environmentId,
+      logger,
+    }),
+  );
+
+  return { app, query, sessionStore, logger };
```
nbouliol: The union now forbids passing both, but `makeApp({ client })` still hands back a defaulted `jest.fn()` that can never fire — an `expect(query).not.toHaveBeenCalled()` would pass vacuously.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -95,6 +125,157 @@ describe('runCli', () => {
...
+
+      try {
+        const outcome = await oversizedBodyOutcome(
+          server.callback,
+          '/agent/v1/ai/query',
+          sessionToken(),
+          1_050_000,
+        );
+
+        expect(OVERSIZED_BODY_OUTCOMES).toContain(outcome);
```
nbouliol: The race is real, so the closed list is a fair narrowing — but this still passes on two outcomes where the caller never sees a 413, and it is the only test carrying the "above the limit → 413" half of that AC for the AI path. A regression that reset the socket instead of answering 413 would stay green.

Tonours: You were right that the closed list was a workaround, and the workaround was unnecessary — the flake was on the client side, not the server side. `supertest` rejects its promise when the write fails, so the 413 the server had already sent was thrown away.

Dropped the outcome list and the helper. A raw `http.request` that swallows write errors and waits for the response gets a deterministic 413:

```ts
outgoing.on('error', () => undefined);
outgoing.setTimeout(10_000, () => reject(new Error('the server never answered')));
```

No `Content-Length`, so Node sends it chunked and `raw-body` has to count. `expect(status).toBe(413)` — 8/8 on the probe, then 6 consecutive runs of the whole file green. A regression that reset the socket instead of answering 413 now fails on the 10 s timeout instead of passing.

One ceiling worth naming: the helper only works when the overshoot is small (1_050_000 against a 1 MB limit, ~2 kB left to write). I tried it on the no-route test too — 900 kB against the 16 kb limit — and it hangs: ~884 kB still has to go into a socket nobody is reading. That test is back on a declared `Content-Length: 900000` with a 15-byte body, which is all it needs to prove the 1 MB parser is not mounted there. Named the helper `countedOversizedBodyStatus` so the distinction is in the call site.

packages/agent-bff/test/ai/ai-routes-middleware.test.ts
Tonours: Right, the union only stopped the illegal call, not the vacuous one. Split the harness instead:

```ts
function buildApp(client: AiProxyClient, context: AppContext) { … }

function makeApp({ query = jest.fn(), ...context }: AppContext & { query?: jest.Mock } = {}) {
  return { ...buildApp({ query } as unknown as AiProxyClient, context), query };
}

function makeAppWithClient(client: AiProxyClient, context: AppContext = {}) {
  return buildApp(client, context);
}
```

`makeAppWithClient` returns no `query` at all, so `expect(query).not.toHaveBeenCalled()` does not compile there rather than passing vacuously. The union is gone.

packages/agent-bff/test/cli-core.test.ts
Tonours: Extracted to `test/helpers/environment-id-fetch.ts`:

```ts
export function stubEnvironmentIdFetch(): jest.Mock
export function restoreFetch(original: typeof global.fetch): void
```

Used at all four sites — `cli-core.test.ts:93` and `:135`, `cli-ai-wiring.test.ts:23`, `openapi-routes.test.ts:128`. The one that asserted on the stub now holds the returned mock (`expect(fetchEnvironmentId).toHaveBeenCalledTimes(1)`) instead of reading `global.fetch` back.

packages/agent-bff/test/cli-core.test.ts
Tonours: Confirmed and fixed at all three. Every `jest.restoreAllMocks()` that only existed to undo a `global.fetch` assignment is replaced by the `originalFetch` capture plus `restoreFetch(originalFetch)` — `cli-core.test.ts` (hoisted to the top-level describe, so it covers the deliberate `mockRejectedValue` override too), `cli-ai-wiring.test.ts`, and the OAuth describe in `openapi-routes.test.ts`. The remaining `restoreAllMocks` calls in `openapi-routes.test.ts` guard real `jest.spyOn` mocks and stay.

packages/agent-bff/test/cli-core.test.ts
Tonours: Fixed — `expect(response.body.error).toBe('invalid_request')`. As you said, `error_description` carries the word in both branches, so the assertion could not tell a parsed body from a shadowed one.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -95,6 +143,148 @@ describe('runCli', () => {
...
+
+    it('should keep parsing /oauth/token, which the scoped parser must not shadow', async () => {
+      const server = await runCli(OAUTH_ENV, noopLogger);
+
+      try {
+        const response = await request(server.callback)
+          .post('/oauth/token')
+          .send({ grant_type: 'refresh_token' });
+
+        expect(response.body.error).toBe('invalid_request');
```
nbouliol: My earlier comment here was wrong. `handleToken` calls `parseBody` first (`oauth-routes.ts:476`), so a shadowed parser throws `invalid_request` / "Missing request body", never `unsupported_grant_type` — the original `toContain('refresh_token')` did discriminate, and this replacement does not. `expect(response.body.error_description).toBe('Missing required parameter: refresh_token')` pins the two branches apart.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -17,7 +21,61 @@ const VALID_ENV = {
...
+          method: 'POST',
+          headers: { 'Content-Type': 'application/json', Authorization: `Bearer ${token}` },
+        },
+        response => {
+          response.resume();
+          resolve(response.statusCode as number);
+        },
+      );
+
+      outgoing.on('error', () => undefined);
```
nbouliol: `outgoing.on('error', reject)` is strictly better — after `resolve`, a later `reject` is a no-op. As written, an EPIPE while writing 1 MB settles the promise never: the handler discards it and Node clears the socket timer on destroy, so it surfaces as an unrelated jest timeout.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -17,7 +21,61 @@ const VALID_ENV = {
...
+      outgoing.setTimeout(10_000, () => reject(new Error('the server never answered')));
```
nbouliol: No `testTimeout` is configured anywhere, so jest's 5 s default kills the test first and this diagnostic is unreachable — a named constant under 5 s, or a per-test timeout.

packages/agent-bff/test/helpers/environment-id-fetch.ts
```diff
@@ -0,0 +1,16 @@
...
+    statusText: 'ok',
+    json: async () => ({ data: { id: '42' } }),
+  });
+
+  global.fetch = stub as unknown as typeof fetch;
+
+  return stub;
+}
+
+export function restoreFetch(original: typeof global.fetch): void {
```
nbouliol: Having `stubEnvironmentIdFetch()` capture the previous value and return `{ fetchEnvironmentId, restore }` would drop the `const originalFetch = global.fetch` line all four callers hoist, and this function with it.

packages/agent-bff/test/ai/ai-routes-middleware.test.ts
```diff
@@ -0,0 +1,387 @@
...
+
+function jsonResponse(status: number, body: unknown): AiProxyResponse {
+  return { status, body, isJson: true };
+}
+
+describe('createAiRoutesMiddleware', () => {
+  const originalFetch = global.fetch;
+
+  afterEach(() => {
+    global.fetch = originalFetch;
```
nbouliol: This bypasses the `restoreFetch` helper the same commit introduced.

packages/agent-bff/test/cli-core.test.ts
Tonours: Confirmed against the code and against a mutation, and you are right on both halves — my replacement made it worse.

`handleToken` calls `parseBody` at `oauth-routes.ts:475`, and `parseBody` throws `invalidRequest('Missing request body')` at `:219`. So both branches carry `error: invalid_request`, and the assertion I had just added discriminated nothing.

Now `expect(response.body.error_description).toBe('Missing required parameter: refresh_token')`. Mutated `createBodyParser(...)` to `agentScoped(createBodyParser(...))` to confirm it actually separates them:

```
✕ should keep parsing /oauth/token, which the scoped parser must not shadow
  Expected: "Missing required parameter: refresh_token"
  Received: "Missing request body"
```

Which also confirms the old `error` assertion would have stayed green through that mutation.

packages/agent-bff/test/cli-core.test.ts
Tonours: Fixed — `outgoing.on('error', reject)`. You are right that a settled promise makes the later `reject` a no-op, so swallowing bought nothing and cost the diagnostic.

Chasing this surfaced a second defect in the same helper: with the error path now reachable, the `finally` hung. `server.close()` waits for open connections, so a request that never completed kept the helper pending until jest killed it 2 minutes later. Added `server.closeAllConnections()` before the close.

packages/agent-bff/test/cli-core.test.ts
Tonours: Right — `jest.config.js` sets no `testTimeout`, so the 5 s default won. Named it `ANSWER_TIMEOUT_MS = 3_000`.

Verified the diagnostic is now reachable rather than assuming it: ran the helper against a server that never answers, and it rejects with its own message at 3032 ms instead of surfacing as a jest timeout.

packages/agent-bff/test/helpers/environment-id-fetch.ts
Tonours: Done, and taken one step further so the callers keep nothing at all. The helper file is now `test/helpers/fetch-stub.ts`:

```ts
export function stubFetch(response: unknown): jest.Mock
export function stubEnvironmentIdFetch(): jest.Mock
export function restoreFetchAfterEach(): void
```

`restoreFetchAfterEach()` captures the previous value and registers the `afterEach` itself, so each caller is one line in the describe body instead of the three it used to hoist — `const originalFetch`, the `afterEach`, and the restore call. `restoreFetch` is gone.

I went with this over the `{ fetchEnvironmentId, restore }` shape because two of the four sites stub inside a `beforeEach`, which would have meant threading `restore` through a describe-scoped `let` to reach the `afterEach`.

packages/agent-bff/test/ai/ai-routes-middleware.test.ts
Tonours: Fixed. That describe now calls `restoreFetchAfterEach()` like the other three, and the one test that stubbed `global.fetch` directly uses `stubFetch(...)` and asserts on the returned mock (`upstream.mock.calls[0][0]`) instead of reading `global.fetch` back. No file assigns `global.fetch` by hand any more.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -341,6 +353,27 @@ describe('a client generated from the emitted OpenAPI document', () => {
...
       expect(receivedKeys).toEqual([API_KEY]);
     });
+
+    it('should send the key alone when the caller reads the scheme it is answering', async () => {
+      await sdk.countRecordsUsers({ body: {} });
+
+      expect(receivedAuthorizations).toEqual(['']);
+    });
+
+    it('should send both credentials when the caller answers every scheme with the same token, which the BFF rejects with 400 ambiguous_credentials', async () => {
```
nbouliol: Nothing here checks a status, and `buildApp()` mounts no auth chain (`:172` says so) — the "rejects with 400 ambiguous_credentials" half of the name would survive any change to that rejection.

packages/agent-bff/src/docs/docs-samples.ts
```diff
@@ -107,27 +107,27 @@ const SAMPLES_SCRIPT = `
...
+                apiKey = { name: scheme.name, prefix: '', secret: true };
               } else if (scheme.type === 'http' && scheme.scheme === 'bearer') {
-                header = { name: 'Authorization', prefix: 'Bearer ', secret: true };
+                bearer = { name: 'Authorization', prefix: 'Bearer ', secret: true };
               }
             });
           });
 
-          return header || { name: KEY_HEADER, prefix: '', secret: true };
+          return apiKey || bearer || { name: KEY_HEADER, prefix: '', secret: true };
```
nbouliol: For the session-only AI route this falls through to the bearer branch and emits `Authorization: Bearer $BFF_KEY` — the API key the page unlocks with, which that route rejects. `test/docs/docs-page.test.ts:375` pins it as intended; a distinct `$BFF_SESSION` placeholder on the bearer branch would keep the sample copy-pastable.

packages/agent-bff/test/ai/ai-proxy-client.test.ts
```diff
@@ -0,0 +1,196 @@
...
+  };
+}
+
+function firstCall() {
+  return (global.fetch as jest.Mock).mock.calls[0];
+}
+
+describe('AiProxyClient', () => {
+  afterEach(() => {
+    jest.restoreAllMocks();
```
nbouliol, line 38: This is the same `restoreAllMocks`-does-not-undo-`global.fetch` gap the other four suites just fixed, and this file installs nine stubs (`:40` through `:184`), leaving the last one for whatever runs next. `restoreFetchAfterEach()` and `stubFetch` now exist.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
Tonours: Right on both halves — nothing asserts a status, and `buildApp()` resolves credentials by hand at `:163` with a comment saying auth is out of scope.

Dropped the clause: the test is now `should send both credentials when the caller answers every scheme with the same token`, which is exactly what it verifies. Asserting the 400 here would mean mounting the auth chain in a fixture that deliberately does not have one, and the rejection itself is already covered at `auth-mode-middleware.test.ts:56` and `error-contract.test.ts:71`.

packages/agent-bff/src/docs/docs-samples.ts
Tonours: Good catch, and it was worse than cosmetic: the emitted command could not work. The AI route answers `403 oauth_required` to an API key, and the sample handed the reader `Authorization: Bearer $BFF_KEY` — the unlock secret.

The env variable now travels on the header rather than being a module-level constant, so each scheme names its own:

```js
apiKey  = { name: scheme.name, prefix: ,        variable: KEY_VARIABLE };     // BFF_KEY
bearer  = { name: 'Authorization', prefix: 'Bearer ', variable: SESSION_VARIABLE }; // BFF_SESSION
```

`header.secret` is gone — the three generators (cURL, node, ruby) read `header.variable`. Updated `docs-page.test.ts:377` and the ruby assertion, and added one that pins the negative across all three languages:

```ts
["'cURL'", 'JavaScript', 'Ruby'].forEach(language => {
  const source = page.sourceOf(AI_QUERY, language);
  expect(source).toContain('BFF_SESSION');
  expect(source).not.toContain('BFF_KEY');
});
```

Mutated `SESSION_VARIABLE` back to `KEY_VARIABLE` to check it actually bites — three tests fail, including the new one. The `toContain('BFF_SESSION')` half is there on purpose: `sourceOf` returns `''` for a missing sample, so a bare negative would pass vacuously.

packages/agent-bff/test/ai/ai-proxy-client.test.ts
Tonours: Fixed — that suite now calls `restoreFetchAfterEach()` and all nine stubs go through the helpers. Added `rejectFetch(error)` alongside `stubFetch(response)` since four of the nine are `mockRejectedValue`. No file in the package assigns `global.fetch` by hand any more.

Kept the `jest.restoreAllMocks()` in that file: it also guards a real `jest.spyOn(AbortSignal, 'timeout')` at `:183`, which is the case `restoreAllMocks` is actually for.

packages/agent-bff/test/docs/docs-page.test.ts
```diff
@@ -361,11 +372,31 @@ describe('the code samples the docs page injects', () => {
...
+    expect(page.sourceOf(AI_QUERY, 'cURL')).not.toContain('X-Forest-Bff-Key');
+  });
+
+  it('should not sample the unlock key on the session-only route, which answers 403 to it', async () => {
+    const page = await render(API_KEY_SPEC);
+
+    ['cURL', 'JavaScript', 'Ruby'].forEach(language => {
+      const source = page.sourceOf(AI_QUERY, language);
+
+      expect(source).toContain('BFF_SESSION');
```
nbouliol, line 388: `cURL` is pinned exactly at `:378` and Ruby at `:431`, but nothing pins the JavaScript one — `:405` covers `RELATION`, whose scheme has no prefix. So the JS prefix branch (`docs-samples.ts:232`) rests on this `toContain`, which passes whether it emits `"Bearer " + process.env.BFF_SESSION` or just `process.env.BFF_SESSION`.

packages/agent-bff/test/helpers/fetch-stub.ts
```diff
@@ -0,0 +1,32 @@
+export function stubFetch(response: unknown): jest.Mock {
+  const stub = jest.fn().mockResolvedValue(response);
+
+  global.fetch = stub as unknown as typeof fetch;
+
+  return stub;
+}
+
+export function rejectFetch(error: unknown): jest.Mock {
```
nbouliol: Identical to `stubFetch` above but for the mock method — an `installFetch(stub)` both delegate to would remove the repeat.

packages/agent-bff/src/docs/docs-samples.ts
```diff
@@ -261,8 +265,8 @@ const SAMPLES_SCRIPT = `
           ];
 
           headers.forEach(function (header) {
-            var read = "ENV.fetch('" + KEY_VARIABLE + "')";
-            var value = header.secret
+            var read = "ENV.fetch('" + header.variable + "')";
```
nbouliol, line 274: `read` is built from `header.variable` before the ternary that tests it, so a variable-less header (`X-Forest-Timezone`, `Content-Type`) makes it `ENV.fetch('undefined')`. Discarded today, but the two sibling generators (`:202`, `:231`) build the env expression inside the truthy branch.

packages/agent-bff/test/docs/docs-page.test.ts
Tonours: Correct — the JS prefix branch was the one language with no exact assertion, and my `toContain(BFF_SESSION)` passed either way. Added a test that pins the whole line:

```ts
expect(page.sourceOf(AI_QUERY, 'JavaScript')).toContain(
  '    "Authorization": "Bearer " + process.env.BFF_SESSION,',
);
```

Checked it bites: dropping the prefix from `docs-samples.ts:231` so the branch emits `process.env.BFF_SESSION` bare makes it the only failing test in the file. All three languages now have an exact assertion on that header — cURL at `:378`, JS at `:401`, Ruby at `:439`.

packages/agent-bff/test/helpers/fetch-stub.ts
Tonours: Done:

```ts
function installFetch(stub: jest.Mock): jest.Mock {
  global.fetch = stub as unknown as typeof fetch;

  return stub;
}

export function stubFetch(response: unknown): jest.Mock {
  return installFetch(jest.fn().mockResolvedValue(response));
}

export function rejectFetch(error: unknown): jest.Mock {
  return installFetch(jest.fn().mockRejectedValue(error));
}
```

packages/agent-bff/src/docs/docs-samples.ts
Tonours: Right, and it is latent only by accident — nothing stops a future header from carrying a prefix without a variable. Moved it behind a guard clause, which also matches the two siblings:

```js
headers.forEach(function (header) {
  if (!header.variable) {
    lines.push('request[' + rubyQuoted(header.name) + '] = ' + rubyQuoted(header.value));

    return;
  }

  var read = "ENV.fetch('" + header.variable + "')";
  var value = header.prefix ? '"' + header.prefix + '#{' + read + '}"' : read;

  lines.push('request[' + rubyQuoted(header.name) + '] = ' + value);
});
```

No new test: the old expression was computed and discarded, so no output changed, and making it observable would need a header shape the document cannot produce. The generated Ruby is already pinned exactly at `:437` (key) and `:439` (bearer).

nbouliol (approved):
Seven rounds, and the relay itself has been sound since the first pass — the work went into the tests. The validation matrix now discriminates on every item: status and body pinned together, both sides of the 1 MB bound covered on both the declared-length and counted-bytes paths, and an /oauth/token guard that actually fails when the scoped parser shadows it. Closed header allow-list, fixed upstream path and the API-key refusal all match the settled arbitrages.

One correction on my side: my round-3 comment on the /oauth/token assertion was wrong — parseBody runs before the grant_type check, so the original assertion did discriminate. Thanks for the rework it cost you, and for the exact fix afterwards.

PR #1848, opened 2026-08-25
feat(agent-bff): publish a docker image to ghcr

nbouliol:
Ships `@forestadmin/agent-bff` as a Docker image on GHCR, mirroring the pipeline the workflow-executor already uses.

## What lands

| File | Role |
|---|---|
| `packages/agent-bff/Dockerfile` | 3 stages: builder (lerna, scoped) → isolated prod deps → runtime |
| `packages/agent-bff/docker/build-deps-manifest.js` | merges the external deps of the 5 workspace packages into one manifest |
| `packages/agent-bff/docker/deps/yarn.lock` | pins every transitive version of that manifest |
| `packages/agent-bff/docker/check-deps-closure.js` | fails when the hardcoded package list or the Dockerfile `COPY`s drift from the real `@forestadmin` closure |
| `packages/agent-bff/docker/smoke-test.sh` | CLI surface, module graph, Redoc bundle, boot, `/health`, `/docs` |
| `packages/agent-bff/docker/scan-gate.js` | renders the Trivy library findings into the job summary |
| `packages/agent-bff/docker-compose.yml` | single-instance setup |
| `.github/workflows/docker-publish-bff.yml` | validates on PR, builds amd64 + arm64 and publishes a multi-arch manifest |
| `.github/workflows/build.yml` | dispatches the publish after a `@forestadmin/agent-bff@x.y.z` release |

Tags follow the npm package: `:latest`, `:1`, `:1.20`, and the immutable `:1.20.2`. The mutable tags only move for the latest stable version, so a prerelease or a rebuild of an older version never regresses `:latest`.

## Gating

Nothing is pushed before the image is smoke-tested and scanned locally, so a broken or vulnerable image is never pullable by digest.

Trivy gates by **origin**, not severity: OS packages block (they can only be fixed here, by bumping the base image), while library findings are report-only — the BFF's npm dependencies already ship to npm consumers, so blocking the image would desync GHCR from npm without removing the vuln.

## Verified locally (arm64)

- build OK, image 271 MB
- smoke test passes end to end
- Trivy OS: 0 vulnerabilities. Trivy libraries: findings only in the npm CLI the base image bundles, none in the BFF's own dependencies
- `yarn install --frozen-lockfile` is reproducible against the committed isolated lockfile; `yarn audit` is clean

## Deliberate differences from the workflow-executor pipeline

- **No `docker-entrypoint.sh`.** The BFF CLI takes positional subcommands (`openapi`), which the executor's "is the first argument a flag?" wrapper would mistake for a command to `exec`. The Dockerfile uses an exec-form `ENTRYPOINT` instead.
- **The smoke test boots without `BFF_TOKEN_ENCRYPTION_KEY`.** Fully configured, `runCli` fetches the environment id from `FOREST_SERVER_URL` and dies on an unreachable host. Without that one key the whole middleware chain is still built and nothing touches the network — `/health` answers `degraded`, which is the point: it answers.
- **`scan-gate.js` is report-only.** The image has no Docker-only npm dependency to gate yet.
- **No `notify-server` job.** This publishes to GHCR and stops there.

## Follow-up

OpenTelemetry ships in a stacked PR on top of this one.

> [!NOTE]
> ### Add Docker image build and publishing pipeline for `agent-bff`
> - Adds a production multi-stage `Dockerfile` that builds the BFF and its workspace deps, isolates runtime deps, and runs as non-root `node` with a healthcheck on port 3450
> - Adds a `docker-compose.yml` using the published `latest` image with configurable host/container ports and host-gateway mapping
> - Adds a dispatch-triggered `docker-publish-bff` workflow that validates version input, builds amd64/arm64 images with Trivy scans, and publishes a multi-arch manifest with rolling `major`, `minor`, and `latest` tags only for the highest stable version
> - Adds a snapshot/compare step in the release workflow that picks the highest newly released `agent-bff` tag and dispatches the publish workflow
> - Adds a dependency-closure checker, a deterministic external-deps manifest generator with a dedicated lockfile, a smoke-test script, a library-vulnerability report script, and Dockerfile healthcheck port tests
> - Risk: mutable `latest`/`major`/`minor` tags move only for the highest stable version at merge time; the `check-deps-closure.js` checker fails the build when the hardcoded workspace package list or Dockerfile `COPY` instructions diverge from the transitive `@forestadmin` closure
>
> 
> #### Changes since #1848 opened
>
> - Scoped GitHub Actions workflow permissions to job level [e6ce046]
> - Added non-blocking Trivy OS vulnerability scans for unfixed issues in the Docker publishing workflow [61ee388]
> - Added documentation explaining GHCR package visibility behavior [0e3d9a6]
> - Added manifest digest validation to the `docker-publish-bff` workflow [879a480]
> - Modified `docker-publish-bff` workflow publish step to fail when `imagetools inspect` returns errors other than explicit not-found cases [cae987f]
> - Reworked concurrency configuration in `docker-publish-bff` workflow to use version-scoped groups for `workflow_dispatch` events and apply `cancel-in-progress` only to `pull_request` events [cae987f]
> - Expanded `pull_request` trigger paths in `docker-publish-bff` workflow to include `src/**` for `agent-client`, `agent-toolkit`, `datasource-toolkit`, and `forestadmin-client` packages [cae987f]
> - Pinned all GitHub Actions in the `docker-publish-bff` workflow to specific commit SHAs instead of floating version tags [a93f0bf]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 74a653b.</sup>
> 
>


.github/workflows/docker-publish-bff.yml
```diff
@@ -0,0 +1,278 @@
...
+      is_latest: ${{ steps.version.outputs.is_latest }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Need all tags to determine whether this is the latest stable version.
+          fetch-depth: 0
+      - name: Parse version from input
+        id: version
+        run: |
+          VERSION="${{ github.event.inputs.version }}"
```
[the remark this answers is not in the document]
nbouliol: Fixed in 204c4390. The input now reaches the shell through `env` and is validated as a semver before anything uses it; the build job's `ref:` takes the validated output instead of the raw input. Worth noting the same pattern exists in `docker-publish.yml` (workflow-executor), which this was copied from — that one should get the same treatment separately.

packages/agent-bff/README.md
```diff
@@ -31,6 +31,43 @@ dropped collection reachable.
...
+docker compose up
+```
+
+The `docker-compose.yml` at the root of this package starts a single BFF instance. See
+`.env.example` for the full list of environment variables and their descriptions.
+
+Or run the image directly:
+
+```bash
+docker run -d \
```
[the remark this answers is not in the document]
nbouliol: Fixed in 204c4390 (and in the APM example on the stacked PR). Added `--add-host host.docker.internal:host-gateway` and spelled out in the note why Docker Desktop does not need it and Docker Engine for Linux does.

.github/workflows/docker-publish-bff.yml
```diff
@@ -0,0 +1,278 @@
...
+        required: true
+
+permissions:
+  contents: read
+  packages: write
+
+concurrency:
+  # Include the dispatch version so two manual runs for different versions from the
+  # same branch don't cancel each other mid-release.
+  group: docker-publish-bff-${{ github.ref }}-${{ github.event_name == 'workflow_dispatch' && github.event.inputs.version || '' }}
```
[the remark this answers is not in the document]
nbouliol: Fixed in 204c4390, but not the way suggested. A single shared group with `cancel-in-progress: true` cancels whichever run is already going, which is wrong when the older dispatch happens to start second — it would cancel the newer publish. Instead the merge job now re-decides latest-stable against the tags as they are at that moment, so the outcome no longer depends on which run finishes first.

packages/agent-bff/docker/check-deps-closure.js
```diff
@@ -0,0 +1,73 @@
...
+    fs.readFileSync(path.join(PACKAGES_DIR, dir, 'package.json'), 'utf8'),
+  );
+  for (const dep of Object.keys(dependencies)) {
+    if (dep.startsWith('@forestadmin/')) queue.push(dep);
+  }
+}
+
+const actual = [...closure].sort();
+const declared = [...WORKSPACE_PACKAGES].sort();
+const dockerfile = fs.readFileSync(DOCKERFILE, 'utf8');
```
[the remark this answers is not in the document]
nbouliol: Fixed in 204c4390. The check now matches only lines that start with an active `COPY`, and requires the `package.json` alongside the `dist` — that was the real hole, since without a manifest Node has no `main` to resolve the package by. Verified by commenting out one COPY line: the guard fails as expected.

packages/agent-bff/docker/check-deps-closure.js
```diff
@@ -0,0 +1,73 @@
...
+for (const dir of fs.readdirSync(PACKAGES_DIR)) {
+  const manifestPath = path.join(PACKAGES_DIR, dir, 'package.json');
+  if (!fs.existsSync(manifestPath)) continue;
+  const { name } = JSON.parse(fs.readFileSync(manifestPath, 'utf8'));
+  if (name) nameToDir[name] = dir;
+}
+
+// BFS the transitive @forestadmin closure (including the root itself).
+const closure = new Set();
+const queue = [ROOT_PACKAGE];
```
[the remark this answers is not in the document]
nbouliol: Fixed in 204c4390. An `@forestadmin` dependency with no package under `packages/` now exits 1 instead of being skipped — it would fall through every mechanism the image has (the walk cannot reach it, `build-deps-manifest` skips the whole `@forestadmin/` prefix, and there is no dist to COPY), so it can only be a hard failure.

packages/agent-bff/README.md
```diff
@@ -31,6 +31,46 @@ dropped collection reachable.
 
 ## Usage
 
+### Docker (recommended)
+
+```bash
+cp .env.example .env   # then fill in the secrets
```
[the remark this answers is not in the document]
nbouliol: Fixed in a402d50f. Documented right under the `cp` rather than inside it: the template is also what the non-containerised "Without Docker" section copies, where `localhost` is the correct value — so the line to change belongs in the Docker section, not in the template.

packages/agent-bff/Dockerfile
```diff
@@ -0,0 +1,101 @@
...
+WORKDIR /app
+
+# Install from manifests + lockfile only (before the source COPY) so editing
+# source doesn't bust the install layer — keeps the install cached on
+# source-only changes (CI cache-warm + local rebuilds). --parents preserves the
+# packages/*/ layout that yarn workspaces needs.
+# --frozen-lockfile guarantees the resolved tree matches yarn.lock exactly.
+# --ignore-scripts skips husky, native (node-gyp) builds, and binary downloads
+# irrelevant here (sqlite3 is dev-only and never loaded at runtime).
+COPY --parents package.json yarn.lock packages/*/package.json packages/workflow-executor/example/package.json ./
```
Tonours: The BFF Dockerfile naming another package's nested workspace is a trap for the next person who adds one: the build fails on an opaque yarn error, far from the cause.

`packages/*/package.json packages/*/*/package.json` covers both without naming anyone.

packages/agent-bff/Dockerfile
```diff
@@ -0,0 +1,101 @@
...
+# OCI image metadata (https://github.com/opencontainers/image-spec).
+LABEL org.opencontainers.image.title="forestadmin-agent-bff" \
+      org.opencontainers.image.description="Backend-for-frontend for Forest Admin agents." \
+      org.opencontainers.image.source="https://github.com/ForestAdmin/agent-nodejs" \
+      org.opencontainers.image.licenses="GPL-3.0"
+
+# Exec form, no wrapper script: the CLI takes positional subcommands (`openapi`),
+# so a "is the first argument a flag?" wrapper would misread them as a command to
+# exec. Use `--entrypoint` to run anything else in the image.
+ENTRYPOINT ["node", "/app/packages/agent-bff/dist/cli.js"]
```
Tonours, line 112: node is PID 1 here and nothing registers a SIGTERM handler. I grepped the 5 closure packages, and `BFFHttpServer.stop()` (`bff-http-server.ts:90`) is never wired.

The kernel drops the default disposition for PID 1, so `docker stop` waits the full 10s grace then SIGKILLs, dropping in-flight requests every time.

Two options: `init: true` in the compose file for the quick win, or wire `stop()` on SIGTERM/SIGINT for a real graceful shutdown. The second one is what helps anyone running this behind k8s.

.github/workflows/docker-publish-bff.yml
```diff
@@ -0,0 +1,326 @@
...
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: ".nvmrc"
+
+      - name: Check @forestadmin dependency closure
+        run: node packages/agent-bff/docker/check-deps-closure.js
```
Tonours, line 52: This only runs in the PR `validate` job. The release `build` job never runs it, and it isn't wired into lint or test either.

A dep change that reaches `main` through a path outside the `paths:` filter hits the release build unchecked, and the failure mode is a container dying on `Cannot find module`. Cheap to add the same step to `build`.

.github/workflows/docker-publish-bff.yml
```diff
@@ -0,0 +1,326 @@
...
+          LATEST_STABLE=$(git tag --list "${PREFIX}*" \
+            | sed "s|^${PREFIX}||" \
+            | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' \
+            | sort -V | tail -n1)
+          case "$VERSION" in
+            *-*) IS_LATEST=false ;;                                  # prerelease
+            "$LATEST_STABLE") IS_LATEST=true ;;
+            *) IS_LATEST=false ;;
+          esac
+          echo "is_latest=$IS_LATEST" >> $GITHUB_OUTPUT
```
Tonours: This decision is fully subsumed by the re-check in `merge` (:283-293). `LATEST_STABLE` there is grepped to stable-only, so a prerelease can never match it, and neither can a rebuild of an older version. The only consumer of this output is the `if [ "$IS_LATEST" != "true" ]` short-circuit at :279.

You can drop the `case` block and the `is_latest` output, keep the merge check. -17 lines and one less "decided twice, second one wins" to hold in your head.

.github/workflows/docker-publish-bff.yml
```diff
@@ -0,0 +1,326 @@
...
+          severity: CRITICAL,HIGH
+          ignore-unfixed: true
+          exit-code: '0'
+          format: json
+          output: trivy-libs.json
+      - name: Report npm deps
+        run: node packages/agent-bff/docker/scan-gate.js trivy-libs.json
+
+      # All gates passed (smoke + OS scan). Publish by digest (cache hit).
+      - name: Push by digest
```
Tonours, line 222: The gated image and the pushed image are two separate builds. This one only carries `cache-from`, so nothing asserts the digest you scanned is the digest you push.

In practice the buildx container keeps its local cache for the whole job and will reuse the layers, so I don't think this bites today. But the guarantee in the description ("nothing is pushed before the image is smoke-tested") rests on a cache hit rather than on anything checked. `docker tag` + `docker push` on the loaded image makes it exact, or at minimum add `cache-to` here.

packages/agent-bff/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+# Minimal single-instance setup for the agent BFF.
+#
+# 1. Copy .env.example to .env and fill in your secrets.
```
Tonours, line 3: The `a402d50` fix landed in the README prose but not in the file this line tells people to copy: `.env.example:5` still has `AGENT_URL=http://localhost:3351`, which points at the container itself.

So the documented quickstart is broken for anyone following it without reading the note further down. A commented `host.docker.internal` variant in `.env.example` closes it.

packages/agent-bff/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+    # list of knobs). Unset optional vars fall back to the image's built-in
+    # defaults — no whitelist to keep in sync.
+    env_file:
+      - .env
+    # Lets AGENT_URL reach host services via host.docker.internal on Linux too
+    # (resolves natively on Docker Desktop, no-op there).
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    ports:
+      - "${HTTP_PORT:-3450}:${HTTP_PORT:-3450}"
```
Tonours: `HTTP_PORT=0` breaks this. Verified:

```
$ HTTP_PORT=0 docker compose config
service ports services.agent-bff.ports.[0] is missing a target port
```

And `env-config.ts:59` documents 0 as valid ("must be an integer between 0 and MAX_PORT"). Edge case, but it's a value you tell users they can set. Decoupling host port from container port would also let people remap without touching app config.

packages/agent-bff/docker/smoke-test.sh
```diff
@@ -0,0 +1,87 @@
...
+fi
+if ! grep -q '"status":"degraded"' /tmp/bff-health.json; then
+  echo "::error::/health body is not the degraded payload"
+  cat /tmp/bff-health.json
+  exit 1
+fi
+
+# The docs page and its bundle are public routes; a 200 proves the asset is served,
+# not merely present on disk.
+bundle=$(curl -s -o /dev/null -w '%{http_code}' "http://127.0.0.1:$PORT/docs/redoc.standalone.js")
```
Tonours: Two small gaps in an otherwise solid test:

- Only the bundle is fetched, never `/docs` itself. That's the shell page a user actually opens (`DOCS_PATH`, `docs-routes.ts:9`), so a broken `renderDocsPage` or `enabled: false` slips through.
- The 200 path of `/health` is never exercised, only the 503. The image's own HEALTHCHECK (`Dockerfile:90`) demands a 200, so if `hasAllRequired` ever regressed to always-false, CI stays green and every published container reports unhealthy forever. A second boot with `BFF_TOKEN_ENCRYPTION_KEY` set would cover both the 200 and the happy config path.

packages/agent-bff/docker/build-deps-manifest.js
```diff
@@ -0,0 +1,120 @@
...
+        `Conflicting ranges for "${name}" not arbitrated by agent-bff: ${detail}. ` +
+          'Align the version in the source packages.',
+      );
+    }
+    console.warn(
+      `[build-deps] "${name}": conflicting ranges (${ranges.join(', ')}); ` +
+        `using the BFF's pin "${bffRange}".`,
+    );
+    deps[name] = bffRange;
+  }
```
Tonours, line 61: I ran this on the real tree: 13 deps, zero warnings, zero throws. The conflict arbitration never fires today, and `yarn install --frozen-lockfile` is already the arbiter since an unresolvable set fails the install.

That's ~30 lines and two concepts for a case that doesn't exist. A merge + sort gets you there, and an inline `if (existing && existing !== range) throw` keeps the guard if you want to keep one.

packages/agent-bff/docker/scan-gate.js
```diff
@@ -0,0 +1,48 @@
...
+  lines.push('| Package | Severity | ID | Installed | Fixed |', '|---|---|---|---|---|');
+  for (const f of findings) {
+    lines.push(`| ${f.pkg} | ${f.severity} | ${f.id} | ${f.installed} | ${f.fixed} |`);
+  }
+}
+const summary = lines.join('\n');
+console.log(summary);
+if (process.env.GITHUB_STEP_SUMMARY) fs.appendFileSync(process.env.GITHUB_STEP_SUMMARY, `${summary}\n`);
+
+console.log('\nNo blocking library vulnerabilities — npm-shared deps are report-only.');
```
Tonours: Small one: the only `process.exit(1)` left is the usage error at :21, so this gates nothing. The workflow-executor version still has a real one at :54 for its OTel deps.

The header comment explains why, which is fair, but the name plus a hardcoded "No blocking library vulnerabilities" reads like a gate to whoever greps for it in six months. `report-library-scan.js` would be honest. Trivy's `format: template` into `$GITHUB_STEP_SUMMARY` would also let you delete the file.

Tonours (commented):
Nice pipeline, the gating story is well thought out and the smoke test has real teeth. I mutation-tested it: dropping `agent-toolkit` from `WORKSPACE_PACKAGES` and removing a runtime `COPY` both fail as intended.

Two things worth deciding before merge, the rest is in line comments.

**1. The hardening landed only in this copy.** Commit 204c4390 fixes three real problems, and the workflow-executor set still has all three:

- `.github/workflows/docker-publish.yml:113` interpolates `${{ github.event.inputs.version }}` straight into `run:`, so a crafted dispatch input runs shell with that job's `packages:write` token. It needs repo write to trigger, but it's still a write -> token escalation.
- `packages/workflow-executor/docker/check-deps-closure.js:61` still uses `dockerfile.includes(...)`, so a commented-out `COPY` passes the check, and it never verifies the `package.json` half.
- Same file, line 37: `if (!dir || closure.has(dir)) continue` silently skips an `@forestadmin` dep living outside the monorepo.

Can you backport those here? Otherwise the fixes stay stranded in the BFF copy.

**2. The clone is going to keep costing.** `docker-publish-bff.yml` is 246 code lines, 133 of them differ from `docker-publish.yml`, and the job graph is identical (`validate` -> `extract-version` -> `build` matrix -> `merge`). Point 1 is what that duplication already cost. A `workflow_call` parameterized on package / image / dockerfile / smoke test would collapse both into one reusable plus two ~20 line callers, same for the two dispatch blocks in `build.yml` where a matrix over `{package, workflow}` covers it.

Not asking you to do that in this PR. But the third image will make it three copies.

packages/agent-bff/Dockerfile
nbouliol: Confirmed and fixed in 9eaa394d. I reproduced it before touching anything:

    no handler        -> docker stop -> exit 137 (SIGKILL)
    handler + exit()  -> exit 0

Went with wiring `stop()` rather than `init: true`, as you suggested — the second option is the one that helps behind k8s. SIGTERM/SIGINT now close the server, bound the wait for in-flight requests at 10s (`closeAllConnections` past the deadline, or `close` waits on a request that never completes), then exit. A second signal gives up on the wait and exits 1.

Two things that fell out of it and are worth flagging:

- the exit sets `process.exitCode` and lets the loop drain instead of calling `process.exit`. Under Docker stdout is a pipe, so writes to it are async and exiting outright dropped both shutdown log lines — the one place an operator looks. An unref'd 1s timer still forces the exit if some other handle holds the loop.
- this also invalidated the span-flush handler on the stacked OTel PR: it re-raised the signal expecting the default disposition to terminate, which at PID 1 does not exist. It survived the re-raise and got SIGKILLed anyway, so the flush it existed to protect could be cut. Fixed there — it flushes and lets this handler own the exit. Verified: exit 0 with OTel armed too.

`stop_grace_period: 15s` in the compose file, since `docker stop` defaults to exactly the 10s deadline.

packages/agent-bff/Dockerfile
nbouliol: Fixed in 9eaa394d — `packages/*/package.json packages/*/*/package.json`. Agreed on the trap; the failure mode you describe is exactly the one worth avoiding.

.github/workflows/docker-publish-bff.yml
nbouliol: Fixed in 9eaa394d — the closure check now runs in `build` too, right after the checkout of the version tag. Cheap, and the failure mode you name is invisible to every other step in that job.

.github/workflows/docker-publish-bff.yml
nbouliol: Agreed, dropped in 9eaa394d — the `case` block, the `is_latest` output, and the short-circuit that consumed it. Your reasoning holds: `LATEST_STABLE` is grepped to stable-only, so neither a prerelease nor a rebuild of an older version can match it, and the merge check alone gives the same answer.

Went one further: `extract-version` only needed a checkout for those tags, so it has none now — splitting a version string needs no repository.

.github/workflows/docker-publish-bff.yml
nbouliol: You are right that the description overclaimed, and I fixed that in 9eaa394d rather than the structure — the comment now says it is a second build, that buildx reuses the layers the gating build wrote to the cache, and that this is a cache hit rather than a checked identity. What the gate actually buys is that a build failing smoke or scan never reaches the registry at all, which is true either way.

I did not go with `docker tag` + `docker push`: the docker driver cannot load a push-by-digest result, and push-by-digest is what the multi-arch manifest is assembled from, so that route also costs the SBOM attestation.

The exact version I considered is comparing `imageid` (the config digest) between the gating step and the push step and failing on a mismatch — that turns the assumption into a checked invariant for ~5 lines. I left it out because `sbom: true` is only on the push step and I am not certain enough it cannot perturb the config digest; a false positive there blocks a release. Happy to add it if you would rather have the check and eat that risk.

packages/agent-bff/docker-compose.yml
nbouliol: Right, my earlier fix stopped at the prose. Fixed in 9eaa394d: `.env.example` now carries the commented `host.docker.internal` variant next to the `localhost` line, with the reason. Left `localhost` as the active value since the same template is what the non-containerised section copies, where it is correct.

packages/agent-bff/docker-compose.yml
nbouliol: Fixed in 9eaa394d by decoupling the two sides, as you suggested: `${BFF_HOST_PORT:-3450}:${HTTP_PORT:-3450}`. Remapping the published port no longer touches app config.

`HTTP_PORT=0` still cannot work here — an ephemeral port the app picks at boot has no target port to publish, whatever the mapping looks like. It stays valid for a local run, so rather than reject it I documented the limit in both `.env.example` and the compose file.

packages/agent-bff/docker/smoke-test.sh
nbouliol: Both fixed in 9eaa394d, and the second one was the better catch — you are right that a regression making `hasAllRequired` always false would leave every published container permanently unhealthy with CI green.

- `/docs` is now fetched alongside the bundle.
- The 200 path gets its own boot. That needed the token encryption key, which is what makes the BFF fetch its environment id from FOREST_SERVER_URL at boot — the reason the first boot deliberately omits it. So the second one runs against a stub serving `/liana/environment`, reached over host-gateway. It stays within curl, openssl and python3, no extra image to pull.

Then `/health` must answer 200 with `"status":"ok"` or the test fails.

packages/agent-bff/docker/build-deps-manifest.js
nbouliol: Agreed — replaced in 9eaa394d with the merge plus an inline throw on a genuine disagreement, which is your suggestion. ~30 lines down to ~15, and the "BFF's pin wins" policy is gone rather than sitting there unexercised. Verified the generated manifest is byte-identical to what the old code produced.

packages/agent-bff/docker/scan-gate.js
nbouliol: Fair, and the last line was the worst part of it — changed in 9eaa394d to "Every library finding above is report-only, this step gates nothing today".

Kept the name, though, because the report-only state is transient: the stacked PR (#1849) restores a real `process.exit(1)` on the Docker-only @opentelemetry packages, exactly like the workflow-executor version. Renaming here would mean renaming back one commit later. Same reason the Trivy-template route does not work — #1849 needs the gate logic, not just a summary.

If you would rather the two PRs not be read as one unit, say so and I will rename in this one and back in the other.

packages/agent-bff/src/shutdown.ts
```diff
@@ -0,0 +1,90 @@
...
+      return;
+    }
+
+    stopping = true;
+    logger?.('Info', 'Shutting down', { signal, graceMs });
+
+    server
+      .stop(graceMs)
+      .then(() => {
+        logger?.('Info', 'Forest BFF stopped');
```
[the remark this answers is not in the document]
nbouliol: Confirmed and fixed in 8f675661 — the escalation set exit 1, then the pending stop() resolved, logged a clean stop and exited 0 over it. It now returns early on both the resolve and the reject path, so an interrupted shutdown never reports success. Two tests cover it.

packages/agent-bff/Dockerfile
```diff
@@ -0,0 +1,104 @@
...
+COPY --from=builder /app/packages/agent-bff/package.json ./packages/agent-bff/package.json
+
+USER node
+
+# HTTP server (GET /health, the /agent edge, GET /docs).
+# Override with the HTTP_PORT env var.
+EXPOSE 3450
+
+HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
+  CMD node -e "require('http').get('http://localhost:'+(process.env.HTTP_PORT||3450)+'/health',r=>process.exit(r.statusCode===200?0:1)).on('error',()=>process.exit(1))"
```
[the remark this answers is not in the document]
nbouliol: Real, and I documented it rather than papering over it (8f675661).

The healthcheck cannot discover an ephemeral port: nothing outside the process learns which one the OS handed out. Neither can the operator — an unpublishable port makes the container unreachable whatever the healthcheck does, so a healthcheck that somehow probed correctly would only report a container nobody can reach. Making it exit 0 regardless would be worse: healthy without checking anything.

Rejecting it in the app is not right either — 0 is legitimate outside Docker and the tests rely on it. So HTTP_PORT=0 is now called out as unusable in the image, in the Dockerfile next to EXPOSE, in the README config table, and in .env.example and the compose file where the port mapping already could not express it.

packages/agent-bff/Dockerfile
```diff
@@ -0,0 +1,107 @@
...
+
+# HTTP server (GET /health, the /agent edge, GET /docs).
+# Override with the HTTP_PORT env var — but only with a real port. `0` asks the OS
+# for an ephemeral one, which is useful for a local run and unusable here: nothing
+# outside the process learns which port it got, so it can be neither published nor
+# probed, and the healthcheck below would report the container unhealthy forever.
+EXPOSE 3450
+
+HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
+  CMD node -e "require('http').get('http://localhost:'+(process.env.HTTP_PORT||3450)+'/health',r=>process.exit(r.statusCode===200?0:1)).on('error',()=>process.exit(1))"
```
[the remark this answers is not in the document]
nbouliol: Real, fixed in 9f777501 — reproduced in the container first: the old expression throws ERR_INVALID_URL on `http://localhost: 8080 /health` while the server is happily listening on 8080.

Took the trim but placed it before the default rather than after. Your suggestion trims after the `||`, so HTTP_PORT='   ' becomes an empty string and the URL falls back to port 80, where parsePort returns 3450 — a mismatch in the other direction.

Rather than restate the expression in a test, the test lifts it out of the Dockerfile and evaluates it against the real parseConfig across unset / empty / whitespace / plain / padded / zero. Confirmed it fails on the old expression before I changed it.

packages/agent-bff/Dockerfile
Tonours: Glob widened, no package named anymore. Good.

packages/agent-bff/Dockerfile
Tonours: `shutdown.ts` handles it properly, and `stop_grace_period: 15s` beats the 10s grace plus the 1s force exit. Nice.

.github/workflows/docker-publish-bff.yml
Tonours: Same step in `build` now, with its own setup-node. Good.

.github/workflows/docker-publish-bff.yml
Tonours: Gone, and you dropped the checkout it was the only reason for. Even better.

.github/workflows/docker-publish-bff.yml
Tonours: Fair. The docker driver cannot load a push-by-digest output, so a single build is not on the table, and the comment now says what the gate actually buys instead of overselling it. Good enough for me.

packages/agent-bff/docker-compose.yml
Tonours: `.env.example` carries the commented `host.docker.internal` line now. Good.

packages/agent-bff/docker-compose.yml
Tonours: `BFF_HOST_PORT` splits the two knobs, and the 0 case is documented in the compose file, the Dockerfile and `.env.example`. Good.

packages/agent-bff/docker/smoke-test.sh
Tonours: Both covered, and the fully-configured second boot with a stub goes further than I asked. Nice.

packages/agent-bff/docker/build-deps-manifest.js
Tonours: That is exactly it. Good.

packages/agent-bff/docker/scan-gate.js
Tonours: Log line is honest now. The filename still reads like a gate to anyone grepping for it, but that is cosmetic, not worth another round.

Tonours:
Went back over the branch, all ten threads resolved. The shutdown path, the second smoke boot against a stub, and dropping `is_latest` from `extract-version` all go further than what I asked. CI green on 32 checks.

One thing from my review body is still open, and it's the one that matters: the workflow-executor backport. `docker-publish.yml:113` still does `VERSION="${{ github.event.inputs.version }}"` inside `run:`, and `workflow-executor/docker/check-deps-closure.js` still has both weaknesses (`dockerfile.includes` at :61, the silent `!dir` skip at :37). Fine by me if that's a separate PR, just don't let it drop.

Two small ones from the new code left as line comments, neither a blocker.

packages/agent-bff/README.md
```diff
@@ -31,6 +31,55 @@ dropped collection reachable.
...
+set `AGENT_URL=http://host.docker.internal:3351`. Left at `localhost`, it resolves to the BFF
+container itself and every agent call fails (see the note below).
+
+The `docker-compose.yml` at the root of this package starts a single BFF instance. See
+`.env.example` for the full list of environment variables and their descriptions.
+
+Or run the image directly:
+
+```bash
+docker run -d \
```
Tonours, line 51: `docker stop` defaults to a 10s timeout, and the new shutdown budget is `DEFAULT_GRACE_MS` 10s plus `FORCE_EXIT_MS` 1s. Under in-flight load that SIGKILLs right as the grace expires, which is the case the graceful shutdown was built for.

The compose file covers it with `stop_grace_period: 15s`. This example doesn't, so `--stop-timeout 15` here would keep the two consistent. Not a blocker.

packages/agent-bff/docker/smoke-test.sh
```diff
@@ -0,0 +1,173 @@
...
+STUB_DIR=$(mktemp -d)
+mkdir -p "$STUB_DIR/liana"
+printf '{"data":{"id":1}}' > "$STUB_DIR/liana/environment"
+# Bound to every interface, not just loopback: host-gateway resolves to the docker
+# bridge address on Linux, so a stub listening only on 127.0.0.1 is unreachable from
+# the container. It serves one static file for the length of this script.
+#
+# --directory rather than a `cd` subshell: `$!` must be python's own pid, or the
+# kill below reaps the subshell and leaves the server holding the port.
+python3 -m http.server "$STUB_PORT" --bind 0.0.0.0 --directory "$STUB_DIR" >/dev/null 2>&1 &
```
Tonours, line 112: The stub brings in python3 as a new dependency of the smoke test. Fine on the runners, and I'd rather have the stub than skip the `/health` 200 path.

Just worth a line in `docker/README.md` next to the other prerequisites, so whoever runs this locally on a slim box knows before they get a bare `command not found`.

packages/agent-bff/README.md
nbouliol: Right, the budget is 11s against a 10s default — fixed in 3f00878a. Added `--stop-timeout 15` to the docker run example so it matches the compose file, and rewrote the paragraph to say where the 11 comes from (10s deadline + 1s exit fallback) rather than leaving the reader to add it up. Also noted that k8s' default terminationGracePeriodSeconds of 30 already covers it.

packages/agent-bff/docker/smoke-test.sh
nbouliol: Fair, documented in 3f00878a. Added a "Running the smoke test" section to docker/README.md naming all four prerequisites (docker, curl, openssl, python3), why python3 is there, and that it is the one likely missing on a slim box.

packages/agent-bff/README.md
Tonours: `--stop-timeout 15` is in the example now, and spelling out the 11s budget plus the k8s 30s default goes further than what I asked. Good.

packages/agent-bff/docker/smoke-test.sh
Tonours: Prerequisites section covers it, and naming python3 as the one likely to be missing is the useful part. Good.

Tonours (approved):
LGTM. Everything I raised is fixed, and rather than take the claims on trust I built and ran the image locally on arm64.

Built from `9f7775014`. The only commit since is docs-only, so the image is unchanged.

| Check | Result |
|---|---|
| `docker build`, arm64 | exit 0, 271 MB, matches the number in the description |
| Image metadata | `USER node`, exec-form entrypoint, OCI labels present |
| `node_modules` | 36 MB, exactly the 4 closure packages, nothing else hoisted in |
| `smoke-test.sh` | exit 0, both boots |
| `docker stop` (SIGTERM) | stops in 1s, exit code 0, `Shutting down` then `Forest BFF stopped` in the logs |
| Container HEALTHCHECK | reaches `healthy` with a full config, `/health` 200 `{"status":"ok"}` |
| `HTTP_PORT` handling | blank and whitespace-only fall back to 3450, `" 8080 "` binds 8080, probe and server agree |
| `BFF_HOST_PORT=9999` | published 9999 / target 3450, the two knobs are genuinely separate |
| `HTTP_PORT=0` on compose | still rejected, exactly as the comment says |
| `openapi` | OpenAPI 3.1.0, 7 paths |
| Trivy OS | 0 vulnerabilities (debian 12.14) |
| Trivy libraries | 9 findings, all under `usr/local/lib/node_modules/npm/`, none in `/app/node_modules` |

The library findings back the origin argument: `tar`, `sigstore`, `ip-address`, `picomatch` and `brace-expansion` all come from the npm CLI the base image bundles. One is a CRITICAL (`CVE-2026-59873`, tar gzip bomb), not reachable from the BFF runtime, and only fixable by bumping the pinned base digest.

Two things I did not verify: amd64, since I only built arm64, and I ran Trivy from a container rather than through `trivy-action`, so the pipeline's exact invocation is untested.

Still worth doing, in a separate PR if you prefer: the workflow-executor backport. `docker-publish.yml:113`, and `workflow-executor/docker/check-deps-closure.js:37,61`.

packages/agent-bff/Dockerfile
```diff
@@ -0,0 +1,120 @@
...
+COPY --from=builder /app/packages/datasource-toolkit/dist   ./node_modules/@forestadmin/datasource-toolkit/dist
+COPY --from=builder /app/packages/datasource-toolkit/package.json ./node_modules/@forestadmin/datasource-toolkit/package.json
+COPY --from=builder /app/packages/forestadmin-client/dist   ./node_modules/@forestadmin/forestadmin-client/dist
+COPY --from=builder /app/packages/forestadmin-client/package.json ./node_modules/@forestadmin/forestadmin-client/package.json
+
+# BFF entry point. dist/docs carries the Redoc bundle the docs page serves.
+COPY --from=builder /app/packages/agent-bff/dist         ./packages/agent-bff/dist
+COPY --from=builder /app/packages/agent-bff/package.json ./packages/agent-bff/package.json
+
+USER node
```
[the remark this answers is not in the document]
nbouliol: Confirmed and fixed in a7b35723. Reproduced in the image before touching it:

    $ docker run --rm bff:before openapi --output
    Cannot write /app/openapi.json: EACCES: permission denied, open '/app/openapi.json'

Took the suggestion as written, with one thing worth stating: the chown is deliberately **not** recursive. It grants the one thing the CLI needs — creating a file in WORKDIR — and leaves node_modules and dist root-owned, so the process still cannot rewrite the code it runs. Checked after the fix:

    drwxr-xr-x 1 node node /app
    drwxr-xr-x 1 root root /app/node_modules
    touch: cannot touch '/app/node_modules/x': Permission denied

Also added a smoke-test step for it. The existing coverage used the stdout form (`openapi > file`), which writes nothing inside the container and so could never have caught this — verified the new step fails on the pre-fix image and passes on the fixed one.

.github/workflows/docker-publish-bff.yml
```diff
@@ -0,0 +1,327 @@
...
+      # --frozen-lockfile and wouldn't catch an inconsistency).
+      - 'yarn.lock'
+      - '.github/workflows/docker-publish-bff.yml'
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version tag to publish (e.g. 1.20.2)'
+        required: true
+
+permissions:
```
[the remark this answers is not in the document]
nbouliol: Fixed in e6ce046b. Checked the three jobs before applying it:

- `validate` (the `pull_request` one) builds with `push: false` and has no `docker/login-action` step — it never touches the registry, so it needed none of it.
- `build` logs in and pushes by digest.
- `merge` logs in and assembles the multi-arch manifest.

So `contents: read` is now the workflow floor and `packages: write` is granted per job to those two. `extract-version` gets neither beyond the floor — it has no checkout and only splits a version string.

Worth stating why this is not theoretical here: `validate` runs the PR's own Dockerfile, its `docker/*.js` and its `smoke-test.sh`, so the token was reachable from code the PR controls. Fork PRs get a read-only token regardless, but these branches live in the repo, so they did get the writable one.

Same shape exists in `docker-publish.yml` (workflow-executor), which this was copied from — not touching it here.

.github/workflows/docker-publish-bff.yml
```diff
@@ -0,0 +1,327 @@
...
+
+      # Gate by origin (see the validate job): OS packages block; the BFF's npm
+      # deps are report-only.
+      - name: Scan OS packages (blocking)
+        uses: aquasecurity/trivy-action@v0.36.0
+        with:
+          image-ref: agent-bff:${{ matrix.arch }}
+          vuln-type: os
+          severity: CRITICAL,HIGH
+          ignore-unfixed: true
```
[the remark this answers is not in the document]
nbouliol: Half right, so half applied — I measured the image before answering.

Trivy on the built image, OS packages, CRITICAL+HIGH:

- with `ignore-unfixed: true` (today's gate): **0**
- with `ignore-unfixed: false` (the suggestion): **56**

So every HIGH/CRITICAL OS finding on this base is unfixed — `perl-base` 8, the `util-linux` family 5 each (CVE-2026-53613), `zlib1g` CVE-2023-45853, `ncurses`, `libsystemd0`... Flipping the flag would not surface a fixable CVE that is being hidden; it would fail 100% of builds on vulnerabilities the pipeline cannot act on. The `apt-get upgrade` in the Dockerfile already pulls every patch Debian has published, which is exactly why the fixed count is zero. The only way to publish again would be a `.trivyignore` with 56 entries, i.e. the same suppression with more maintenance.

What *was* right: those 56 were invisible. `ignore-unfixed` dropped them before anyone could read them, so a newly-disclosed unfixed CVE showed up nowhere. Added a second OS scan, `ignore-unfixed: false` / `exit-code: 0`, in both the PR job and the publish job — the findings are printed, the gate still blocks only on what a rebuild can fix. 61ee388

If we do want a hard ceiling on unfixed OS CVEs later, the lever is the base image (distroless/chainguard), not the flag.

.github/workflows/docker-publish-bff.yml
```diff
@@ -0,0 +1,362 @@
...
+      # identity. What the gate really buys is that a build which fails smoke or scan
+      # never reaches the registry at all.
+      - name: Push by digest
+        id: build
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: packages/agent-bff/Dockerfile
+          platforms: ${{ matrix.platform }}
+          outputs: type=image,name=ghcr.io/forestadmin/agent-bff,push-by-digest=true,name-canonical=true,push=true
```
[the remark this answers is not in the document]
nbouliol: Right, and documented — but it cannot be fixed in the workflow.

Visibility is a property of the GHCR *package*, not of the push. `packages: write` on `GITHUB_TOKEN` grants write to the registry; changing a package to public needs admin on the package, which that token does not have and cannot grant itself. Adding a `gh api` call here would just fail, or force a PAT with `admin:packages` into the repo secrets to work around a one-time click.

And it is one-time: the next 200 pushes inherit the visibility of the package the first one created. `ghcr.io/forestadmin/workflow-executor` already went through exactly this — verified it answers an anonymous token today (`HTTP 200` on `/v2/forestadmin/workflow-executor/tags/list`), while `agent-bff` does not, because the package does not exist yet.

So: documented on both sides. 0e3d9a6

- README, under the tag list: the package is public, that is set by hand once, and until then pulls need `docker login ghcr.io -u <user> -p <token-with-read:packages>`.
- A comment at the `Push by digest` step, so whoever ships the first image knows the flip is theirs to make.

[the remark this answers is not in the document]
nbouliol: It is there — `0e3d9a6`, pushed after that round. Current `feat/agent-bff-docker-image`:

- `packages/agent-bff/README.md` L82-86, right under the tag list: the package is public, that visibility is set once by hand, and until it is, pulls need `docker login ghcr.io -u <user> -p <token-with-read:packages>`.
- `.github/workflows/docker-publish-bff.yml` L256, above `Push by digest`: the first push creates the package private, `packages: write` grants the push and not the admin, so the flip is a manual one-time step.

Nothing to prepare, thanks.

.github/workflows/docker-publish-bff.yml
```diff
@@ -0,0 +1,367 @@
...
+          FULL: ${{ needs.extract-version.outputs.full }}
+          MINOR: ${{ needs.extract-version.outputs.minor }}
+          MAJOR: ${{ needs.extract-version.outputs.major }}
+          IS_LATEST: ${{ steps.recheck.outputs.is_latest }}
+        run: |
+          # Always publish the immutable full-version tag. The mutable tags
+          # (minor, major, latest) only move when this is the latest stable
+          # version, so neither a prerelease nor a rebuild of an older version
+          # ever overwrites the current stable :latest / :1 images.
+          TAGS="-t $IMAGE:$FULL"
```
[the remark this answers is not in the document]
nbouliol: Fair, and fixed — this one is a real hole. `879a480d`

The README calls `:1.20.2` immutable and the merge job did not hold that up: it always passed `-t $IMAGE:$FULL`, so re-dispatching a version that already shipped repointed its tag at whatever that run built, under a digest someone may have pinned.

The merge step now resolves the tag first and compares its children against the digests this run pushed, minus the SBOM's attestation manifests (no `.platform.architecture`, so they are excluded from the comparison):

- **identical** → a replay: a merge job that died, or a run whose only purpose is to move the mutable tags. Goes through, `imagetools` writes back the same list.
- **different** → fails, printing both sides. The answer to a rebuilt version is a new version, not a silently different `1.20.2`.
- **absent** → first publish, nothing to check.

Verified the comparison against a fixture manifest list, including the attestation entry: identical set → SAME, one digest swapped → DIFFERENT.

.github/workflows/docker-publish-bff.yml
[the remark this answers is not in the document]
nbouliol: Holding here, and stating it plainly so the thread closes on a decision rather than a loop.

An approval policy on unfixed OS findings means a human ticks a box, on every release, against the same 56 entries — and ticks it yes, because there is no other answer available: no patch exists, and the rebuild already pulls everything Debian has published. A gate whose only possible outcome is approval is not a gate, it is a step that trains people to click through security prompts.

What would actually move the number is the base image — distroless or Chainguard, where the unfixed count is near zero because the surface is not there. That is a change to the image, worth its own PR, not a flag on this one.

So: the blocking scan stays on what a rebuild can fix, the report makes the rest visible, and the ceiling on unfixed CVEs is a base-image decision. Resolving.

.github/workflows/docker-publish-bff.yml
```diff
@@ -0,0 +1,388 @@
...
+  pull_request:
+    # Validate on any change to the BFF or to a dependency package's manifest — a
+    # dep bump in one of the 4 source packages can drift the image's dedicated
+    # lockfile (see packages/agent-bff/docker/).
+    paths:
+      - 'packages/agent-bff/**'
+      - 'packages/agent-client/package.json'
+      - 'packages/agent-toolkit/package.json'
+      - 'packages/datasource-toolkit/package.json'
+      - 'packages/forestadmin-client/package.json'
```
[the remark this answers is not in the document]
nbouliol: Right, and fixed. `cae987f4`

The 4 dependency packages ship as `dist` inside the image, so their `src` is image code — a change to `agent-client/src` could break startup with nothing in this repo to catch it before release.

Took `src/**` rather than `**` for each: their `test/`, README and tooling changes do not reach the image, and paying a two-arch Docker build on every PR that touches a test in `datasource-toolkit` would be a tax with nothing to show for it. Their `package.json` stays listed for the separate reason it was there — a dep bump drifts the image's dedicated lockfile.

.github/workflows/docker-publish-bff.yml
```diff
@@ -0,0 +1,412 @@
...
+          cache-from: type=gha,scope=bff-${{ matrix.arch }}
+          cache-to: type=gha,mode=max,scope=bff-${{ matrix.arch }}
+
+      - name: Smoke test (CLI + module graph + boot)
+        run: sh packages/agent-bff/docker/smoke-test.sh agent-bff:${{ matrix.arch }}
+
+      # Gate by origin (see the validate job): OS packages block; the BFF's npm
+      # deps are report-only, and unfixed OS findings are reported, not blocked.
+      - name: Scan OS packages (blocking)
+        uses: aquasecurity/trivy-action@v0.36.0
```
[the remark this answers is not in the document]
nbouliol: Agreed, and done — all 21 `uses:` in the file, not just the two privileged jobs. `a93f0bf9`

This is not a theoretical tag rewrite: tj-actions/changed-files had every one of its tags repointed in March 2025 at a commit that dumped runner memory into the build log. `build` and `merge` hold `packages: write` and a live registry login, so the same move here runs attacker code under a token that can publish to GHCR.

Each one is pinned to the commit the tag it already named resolves to, with the release in a trailing comment — no version changes in this commit:

```yaml
uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6.19.2
```

`validate` is pinned too. It holds only `contents: read`, but it builds and runs code the PR controls, and leaving one job on tags would just invite the question of why.

One thing worth being explicit about, since it is the real cost: a pin does not receive the action's own fixes, and this repo has no Dependabot or Renovate config for `github-actions` — the other 15 workflow files are all on tags (one is on `@master`). These move deliberately from now on, like the Dockerfile's base digest. Pinning the rest of the repo, or wiring up the updater, is worth its own PR.

PR #1849, opened 2026-08-25
feat(agent-bff): bundle opentelemetry apm in the docker image

nbouliol:
> Stacked on #1848 — review that one first, this PR's diff is only the OpenTelemetry layer.

Bundles OpenTelemetry APM into the `agent-bff` Docker image, the way the workflow-executor image already does.

## How it works

`src/tracing.ts` holds the setup; `src/tracing-preload.ts` is the two-line `--require` entry point the Dockerfile's `ENTRYPOINT` loads before `cli.js`, so the auto-instrumentation patches land before the app imports anything. Keeping them apart is what lets the setup be imported by a test without arming an SDK.

Tracing is **off until `OTEL_EXPORTER_OTLP_ENDPOINT` is set**, so a deployment that never opted into APM pays nothing but the require. Configuration is entirely standard OTel environment variables (`OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`, `OTEL_SDK_DISABLED`).

The packages are pinned in the image's isolated dependency tree only, never shipped to npm consumers — hence the dynamic require behind a seam, and a warning rather than a crash when they are absent.

## Two things worth a look

**The signal handler re-raises.** The BFF installs no signal handler of its own, and registering a listener *suppresses* Node's default termination — arming one to flush spans without re-raising would leave `docker stop` waiting out its timeout for a SIGKILL. So the handler flushes, then re-raises the same signal; `once` has already removed the listener, so the second delivery finds none and terminates. A rejected flush re-raises anyway, so a dead collector cannot hang the process.

**`scan-gate.js` blocks again.** `@opentelemetry/*` exists only in this image and can only be fixed here, so a fixable CRITICAL/HIGH in one of them now fails the publish. Everything else stays report-only. `**/@opentelemetry/propagator-jaeger` is pinned to 2.9.0 (CVE-2026-59892) — `auto-instrumentations-node` still asks for 2.8.0.

## Verified locally (arm64)

- image rebuilt, smoke test passes and now asserts `OpenTelemetry tracing enabled` in the logs — the packages exist only in the image, so nothing else would prove they are loadable
- `yarn audit` on the regenerated isolated lockfile: 0 vulnerabilities across 335 packages
- 11 new tests on the setup (both `load` and signal paths through injected seams); the full agent-bff suite is green at 1217 tests
- lint clean

> [!NOTE]
> ### Bundle OpenTelemetry APM in `agent-bff` Docker image
> - Adds `tracing.ts`, `tracing-preload.ts`, and `tracing-handle.ts` to initialize OpenTelemetry NodeSDK with auto-instrumentation when trace export environment variables are set; tracing stays inactive otherwise.
> - Changes the Docker `ENTRYPOINT` to require `tracing-preload.js` before `cli.js`, and updates [build-deps-manifest.js](https://github.com/ForestAdmin/agent-nodejs/pull/1849/files#diff-c8a6558db0f08393b6685be3c45ae9cfecefc81f4257fc2db9e29452324d1b5a) to pin `sdk-node`, `auto-instrumentations-node`, and the OTLP HTTP exporter into the isolated Docker dependency manifest.
> - `initTracing` honors `OTEL_SERVICE_NAME` / `OTEL_RESOURCE_ATTRIBUTES` for service naming, respects `OTEL_SDK_DISABLED`, redacts credentials from logged endpoints, and defaults metrics/logs exporters to none.
> - Updates [scan-gate.js](https://github.com/ForestAdmin/agent-nodejs/pull/1849/files#diff-cda3d321ea4e5bdcc3d31b500f5848b2b4b25fba7d010eead69b10c7eb3f028e) to fail the build on any `@opentelemetry/` library vulnerability finding, while other library findings remain report-only.
> - Risk: the npm CLI binary does not load the preload, so tracing only works inside the Docker image; the scan gate now exits non-zero on OpenTelemetry vulnerabilities that were previously report-only.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized acc0d68.</sup>
> 
>

### DIFF
--- a/packages/agent-bff/Dockerfile
+++ b/packages/agent-bff/Dockerfile
@@ -109,4 +109,8 @@ LABEL org.opencontainers.image.title="forestadmin-agent-bff" \
 # Exec form, no wrapper script: the CLI takes positional subcommands (`openapi`),
 # so a "is the first argument a flag?" wrapper would misread them as a command to
 # exec. Use `--entrypoint` to run anything else in the image.
-ENTRYPOINT ["node", "/app/packages/agent-bff/dist/cli.js"]
+#
+# tracing-preload runs before cli.js so the OTel auto-instrumentation patches land
+# before the app imports anything. It is inert until OTEL_EXPORTER_OTLP_ENDPOINT is
+# set, so this costs an install that never opted into APM nothing but the require.
+ENTRYPOINT ["node", "--require", "/app/packages/agent-bff/dist/tracing-preload.js", "/app/packages/agent-bff/dist/cli.js"]
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -90,8 +90,10 @@ The Docker image ships with [OpenTelemetry](https://opentelemetry.io/) APM built
 any OTLP-compatible backend (Datadog, Grafana Tempo, Jaeger, Honeycomb, etc.). It is **off by
 default** and turns on as soon as you point it at an OTLP receiver — no code changes or extra
 installs required. Tracing is set up before the app starts (auto-instrumentation for HTTP and the
-outbound calls to the agent and the Forest SaaS), and buffered spans are flushed on `SIGTERM` /
-`SIGINT`, alongside the graceful shutdown described above.
+outbound calls to the agent and the Forest SaaS). The graceful shutdown described above waits for
+the buffered spans to be exported before it exits, but gives that its own 2 second deadline rather
+than the 10 seconds in-flight requests get: an unreachable collector costs you the last spans, never
+the ability to stop. Worst case it adds ~3 seconds to a shutdown.
 
 Configure it entirely through the standard OTel environment variables:
 
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -84,6 +84,41 @@ so under load it would SIGKILL exactly when the shutdown is doing its job. Hence
 above and `stop_grace_period: 15s` in the Compose file; on Kubernetes the default
 `terminationGracePeriodSeconds` of 30 already covers it.
 
+### Observability (OpenTelemetry)
+
+The Docker image ships with [OpenTelemetry](https://opentelemetry.io/) APM built in, and works with
+any OTLP-compatible backend (Datadog, Grafana Tempo, Jaeger, Honeycomb, etc.). It is **off by
+default** and turns on as soon as you point it at an OTLP receiver — no code changes or extra
+installs required. Tracing is set up before the app starts (auto-instrumentation for HTTP and the
+outbound calls to the agent and the Forest SaaS), and buffered spans are flushed on `SIGTERM` /
+`SIGINT` before the process exits.
+
+Configure it entirely through the standard OTel environment variables:
+
+| Variable | Description |
+| --- | --- |
+| `OTEL_EXPORTER_OTLP_ENDPOINT` | OTLP receiver URL (e.g. `http://collector:4318`). **Tracing stays disabled until this is set.** |
+| `OTEL_SERVICE_NAME` | Service name reported in traces. Default: `forestadmin-agent-bff`. |
+| `OTEL_RESOURCE_ATTRIBUTES` | Extra resource attributes, e.g. `deployment.environment=production`. |
+| `OTEL_SDK_DISABLED` | Set to `true` to force-disable tracing even when an endpoint is configured. |
+
+```bash
+docker run -d \
+  -p 3450:3450 \
+  -e FOREST_AUTH_SECRET="..." \
+  -e FOREST_ENV_SECRET="..." \
+  -e FOREST_SERVER_URL="https://api.forestadmin.com" \
+  -e FOREST_APP_URL="https://app.forestadmin.com" \
+  -e AGENT_URL="http://host.docker.internal:3351" \
+  -e BFF_TOKEN_ENCRYPTION_KEY="$(openssl rand -base64 32)" \
+  -e OTEL_EXPORTER_OTLP_ENDPOINT="http://collector:4318" \
+  ghcr.io/forestadmin/agent-bff:latest
+```
+
+> **Note:** OpenTelemetry is bundled only in the Docker image. It is not shipped with the npm
+> package — with an endpoint set but the packages absent, the BFF logs a warning and starts
+> untraced rather than failing.
+
 ### Without Docker
 
 Packaged / production — run the bin:
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -100,7 +100,7 @@ Configure it entirely through the standard OTel environment variables:
 | `OTEL_EXPORTER_OTLP_ENDPOINT` | OTLP receiver URL (e.g. `http://collector:4318`). **Tracing stays disabled until this is set.** |
 | `OTEL_SERVICE_NAME` | Service name reported in traces. Default: `forestadmin-agent-bff`. |
 | `OTEL_RESOURCE_ATTRIBUTES` | Extra resource attributes, e.g. `deployment.environment=production`. |
-| `OTEL_SDK_DISABLED` | Set to `true` to force-disable tracing even when an endpoint is configured. |
+| `OTEL_SDK_DISABLED` | Set to `true` (case-insensitive) to force-disable tracing even when an endpoint is configured. |
 
 ```bash
 docker run -d \
@@ -116,9 +116,10 @@ docker run -d \
   ghcr.io/forestadmin/agent-bff:latest
 ```
 
-> **Note:** OpenTelemetry is bundled only in the Docker image. It is not shipped with the npm
-> package — with an endpoint set but the packages absent, the BFF logs a warning and starts
-> untraced rather than failing.
+> **Note:** These variables only do anything in the Docker image. The image's entry point loads
+> the tracing preload before the CLI; the npm `forest-bff` bin runs the CLI on its own, and the
+> OpenTelemetry packages are not npm dependencies — so outside Docker an `OTEL_*` variable is
+> read by nothing and the process starts untraced, silently.
 
 ### Without Docker
 
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -103,11 +103,12 @@ Configure it entirely through the standard OTel environment variables:
 | `OTEL_SERVICE_NAME` | Service name reported in traces. Default: `forestadmin-agent-bff`. |
 | `OTEL_RESOURCE_ATTRIBUTES` | Extra resource attributes, e.g. `deployment.environment=production`. |
 | `OTEL_SDK_DISABLED` | Set to `true` (case-insensitive) to force-disable tracing even when an endpoint is configured. |
-| `OTEL_TRACES_EXPORTER` | Set to `none` to stop exporting while keeping instrumentation on, so trace context still propagates to the agent and the Forest SaaS. Any other value exports over OTLP. |
+| `OTEL_TRACES_EXPORTER` | Unset (or `otlp`) exports over OTLP to the endpoint above. Any other value — `none`, `console`, a list — is left to the SDK to configure from the environment, and only the OTLP exporter ships in the image. Instrumentation stays on either way, so trace context keeps propagating to the agent and the Forest SaaS. |
 
 ```bash
 docker run -d \
   -p 3450:3450 \
+  --stop-timeout 15 \
   --add-host host.docker.internal:host-gateway \
   -e FOREST_AUTH_SECRET="..." \
   -e FOREST_ENV_SECRET="..." \
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -104,7 +104,9 @@ Configure it entirely through the standard OTel environment variables:
 | `OTEL_SERVICE_NAME` | Service name reported in traces. Falls back to `service.name` in `OTEL_RESOURCE_ATTRIBUTES`, then to `forestadmin-agent-bff`. |
 | `OTEL_RESOURCE_ATTRIBUTES` | Extra resource attributes, e.g. `deployment.environment=production`. A `service.name` here is honoured when `OTEL_SERVICE_NAME` is unset. |
 | `OTEL_SDK_DISABLED` | Set to `true` (case-insensitive) to force-disable tracing whatever else is configured. |
-| `OTEL_TRACES_EXPORTER` | Unset (or `otlp`) exports over OTLP to the endpoint above. Any other value — `none`, `console`, a list — is left to the SDK to configure from the environment; only the OTLP and console exporters ship in the image. Setting it alone turns tracing on without an OTLP endpoint, which is what makes `console` usable for debugging. Instrumentation stays on either way, so trace context keeps propagating to the agent and the Forest SaaS. |
+| `OTEL_TRACES_EXPORTER` | Which exporter the SDK builds: `otlp` (the default), `console`, `zipkin`, `none`, or a list. Setting it alone turns tracing on without an OTLP endpoint, which is what makes `console` usable for debugging. `none` keeps instrumentation running with nothing exported, so trace context still propagates to the agent and the Forest SaaS. |
+| `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` | Per-signal endpoint, taking precedence over the generic one above. Setting either turns tracing on. |
+| `OTEL_EXPORTER_OTLP_PROTOCOL` | `http/protobuf` (the default), `http/json` or `grpc`, with `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` for traces alone. |
 
 ```bash
 docker run -d \
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -89,7 +89,8 @@ above and `stop_grace_period: 15s` in the Compose file; on Kubernetes the defaul
 The Docker image ships with [OpenTelemetry](https://opentelemetry.io/) APM built in, and works with
 any OTLP-compatible backend (Datadog, Grafana Tempo, Jaeger, Honeycomb, etc.). It is **off by
 default** and turns on as soon as you point it at an OTLP receiver — no code changes or extra
-installs required. Tracing is set up before the app starts (auto-instrumentation for HTTP and the
+installs required. A setup that cannot start logs a warning and runs untraced rather than taking
+the process down with it. Tracing is set up before the app starts (auto-instrumentation for HTTP and the
 outbound calls to the agent and the Forest SaaS). The graceful shutdown described above waits for
 the buffered spans to be exported before it exits, but gives that its own 2 second deadline rather
 than the 10 seconds in-flight requests get: an unreachable collector costs you the last spans, never
@@ -99,11 +100,11 @@ Configure it entirely through the standard OTel environment variables:
 
 | Variable | Description |
 | --- | --- |
-| `OTEL_EXPORTER_OTLP_ENDPOINT` | OTLP receiver URL (e.g. `http://collector:4318`). **Tracing stays disabled until this is set.** |
-| `OTEL_SERVICE_NAME` | Service name reported in traces. Default: `forestadmin-agent-bff`. |
-| `OTEL_RESOURCE_ATTRIBUTES` | Extra resource attributes, e.g. `deployment.environment=production`. |
-| `OTEL_SDK_DISABLED` | Set to `true` (case-insensitive) to force-disable tracing even when an endpoint is configured. |
-| `OTEL_TRACES_EXPORTER` | Unset (or `otlp`) exports over OTLP to the endpoint above. Any other value — `none`, `console`, a list — is left to the SDK to configure from the environment, and only the OTLP exporter ships in the image. Instrumentation stays on either way, so trace context keeps propagating to the agent and the Forest SaaS. |
+| `OTEL_EXPORTER_OTLP_ENDPOINT` | OTLP receiver URL (e.g. `http://collector:4318`). **Tracing stays off until this or `OTEL_TRACES_EXPORTER` is set.** |
+| `OTEL_SERVICE_NAME` | Service name reported in traces. Falls back to `service.name` in `OTEL_RESOURCE_ATTRIBUTES`, then to `forestadmin-agent-bff`. |
+| `OTEL_RESOURCE_ATTRIBUTES` | Extra resource attributes, e.g. `deployment.environment=production`. A `service.name` here is honoured when `OTEL_SERVICE_NAME` is unset. |
+| `OTEL_SDK_DISABLED` | Set to `true` (case-insensitive) to force-disable tracing whatever else is configured. |
+| `OTEL_TRACES_EXPORTER` | Unset (or `otlp`) exports over OTLP to the endpoint above. Any other value — `none`, `console`, a list — is left to the SDK to configure from the environment; only the OTLP and console exporters ship in the image. Setting it alone turns tracing on without an OTLP endpoint, which is what makes `console` usable for debugging. Instrumentation stays on either way, so trace context keeps propagating to the agent and the Forest SaaS. |
 
 ```bash
 docker run -d \
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -105,6 +105,7 @@ Configure it entirely through the standard OTel environment variables:
 ```bash
 docker run -d \
   -p 3450:3450 \
+  --add-host host.docker.internal:host-gateway \
   -e FOREST_AUTH_SECRET="..." \
   -e FOREST_ENV_SECRET="..." \
   -e FOREST_SERVER_URL="https://api.forestadmin.com" \
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -103,6 +103,7 @@ Configure it entirely through the standard OTel environment variables:
 | `OTEL_SERVICE_NAME` | Service name reported in traces. Default: `forestadmin-agent-bff`. |
 | `OTEL_RESOURCE_ATTRIBUTES` | Extra resource attributes, e.g. `deployment.environment=production`. |
 | `OTEL_SDK_DISABLED` | Set to `true` (case-insensitive) to force-disable tracing even when an endpoint is configured. |
+| `OTEL_TRACES_EXPORTER` | Set to `none` to stop exporting while keeping instrumentation on, so trace context still propagates to the agent and the Forest SaaS. Any other value exports over OTLP. |
 
 ```bash
 docker run -d \
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -91,7 +91,7 @@ any OTLP-compatible backend (Datadog, Grafana Tempo, Jaeger, Honeycomb, etc.). I
 default** and turns on as soon as you point it at an OTLP receiver — no code changes or extra
 installs required. Tracing is set up before the app starts (auto-instrumentation for HTTP and the
 outbound calls to the agent and the Forest SaaS), and buffered spans are flushed on `SIGTERM` /
-`SIGINT` before the process exits.
+`SIGINT`, alongside the graceful shutdown described above.
 
 Configure it entirely through the standard OTel environment variables:
 
--- a/packages/agent-bff/docker/build-deps-manifest.js
+++ b/packages/agent-bff/docker/build-deps-manifest.js
@@ -24,7 +24,11 @@ const WORKSPACE_PACKAGES = [
   'forestadmin-client',
 ];
 
-// Pinned to exact versions — OTel ships only in the Docker image.
+// Pinned to exact versions — OTel ships only in the Docker image, so nothing else bumps them.
+// A fixable CRITICAL/HIGH in this tree blocks every publish of this image until someone raises a
+// pin here and refreshes deps/yarn.lock (see the propagator-jaeger resolution below for the shape
+// of that fix). Renovate does not watch this file — it is a plain object, not a manifest — so an
+// unrelated release will stall on it unless someone is looking.
 const OTEL_DEPENDENCIES = {
   '@opentelemetry/sdk-node': '0.219.0',
   '@opentelemetry/auto-instrumentations-node': '0.77.0',
@@ -114,4 +118,4 @@ if (require.main === module) {
   generate(packagesDir, outFile);
 }
 
-module.exports = { WORKSPACE_PACKAGES, OTEL_DEPENDENCIES, generate };
+module.exports = { WORKSPACE_PACKAGES, generate };
--- a/packages/agent-bff/docker/build-deps-manifest.js
+++ b/packages/agent-bff/docker/build-deps-manifest.js
@@ -1,7 +1,8 @@
 // Generates the package.json for the Docker image's isolated runtime deps.
 //
 // It merges the external (non-@forestadmin) runtime dependencies of the BFF and
-// its 4 workspace dependencies into a single manifest.
+// its 4 workspace dependencies into a single manifest, plus the OpenTelemetry
+// packages used for APM (Docker-only; not shipped to npm consumers of the CLI).
 //
 // The output is deterministic (sorted keys). A committed yarn.lock sits next to
 // the generated manifest; the Docker build regenerates the manifest and runs
@@ -23,6 +24,13 @@ const WORKSPACE_PACKAGES = [
   'forestadmin-client',
 ];
 
+// Pinned to exact versions — OTel ships only in the Docker image.
+const OTEL_DEPENDENCIES = {
+  '@opentelemetry/sdk-node': '0.219.0',
+  '@opentelemetry/auto-instrumentations-node': '0.77.0',
+  '@opentelemetry/exporter-trace-otlp-http': '0.219.0',
+};
+
 // Security pins for transitive deps whose parents never ship a patched range.
 // They mirror the monorepo root's `resolutions` for the packages that actually
 // appear in this closure — the isolated install does not inherit the root ones.
@@ -31,6 +39,8 @@ const RESOLUTIONS = {
   '**/qs': '>=6.15.2',
   // jsonapi-serializer pins lodash ^4.17.x.
   '**/lodash': '^4.18.0',
+  // auto-instrumentations-node pins propagator-jaeger 2.8.0 (CVE-2026-59892, fixed in 2.9.0).
+  '**/@opentelemetry/propagator-jaeger': '2.9.0',
 };
 
 function generate(packagesDir, outFile) {
@@ -59,6 +69,8 @@ function generate(packagesDir, outFile) {
       declaredBy[name] = pkg;
     }
   }
+
+  Object.assign(deps, OTEL_DEPENDENCIES);
 
   const sorted = Object.fromEntries(Object.keys(deps).sort().map(key => [key, deps[key]]));
 
@@ -102,4 +114,4 @@ if (require.main === module) {
   generate(packagesDir, outFile);
 }
 
-module.exports = { WORKSPACE_PACKAGES, generate };
+module.exports = { WORKSPACE_PACKAGES, OTEL_DEPENDENCIES, generate };
--- a/packages/agent-bff/docker/build-deps-manifest.js
+++ b/packages/agent-bff/docker/build-deps-manifest.js
@@ -25,6 +25,8 @@ const WORKSPACE_PACKAGES = [
 ];
 
 // Pinned to exact versions — OTel ships only in the Docker image, so nothing else bumps them.
+// sdk-node drags in every OTLP exporter plus the Zipkin one, which is why the image can honour
+// OTEL_TRACES_EXPORTER and OTEL_EXPORTER_OTLP_PROTOCOL without naming an exporter here.
 // A fixable CRITICAL/HIGH in this tree blocks every publish of this image until someone raises a
 // pin here and refreshes deps/yarn.lock (see the propagator-jaeger resolution below for the shape
 // of that fix). Renovate does not watch this file — it is a plain object, not a manifest — so an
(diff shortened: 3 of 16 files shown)


packages/agent-bff/src/tracing.ts
```diff
@@ -0,0 +1,127 @@
...
+    load = loadOtelModules,
+    onSignal = (signal, handler) => {
+      process.once(signal, handler);
+    },
+    raise = signal => {
+      process.kill(process.pid, signal);
+    },
+  } = options;
+
+  if (!env.OTEL_EXPORTER_OTLP_ENDPOINT || env.OTEL_SDK_DISABLED === 'true') return undefined;
```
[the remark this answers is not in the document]
nbouliol: Fixed in 86d36a1c — good catch, the OTel spec defines its boolean variables as case-insensitive and this one is the kill switch, so reading TRUE as "not disabled" was the worst possible direction to fail in. Normalized with a trim as well, and covered by a test over TRUE / True / ' true '.

While in there: a blank OTEL_EXPORTER_OTLP_ENDPOINT now counts as unset too. `env_file` hands an empty string through for a line left blank in a .env, which would have armed the SDK against the OTel default endpoint.

packages/agent-bff/README.md
```diff
@@ -69,6 +69,42 @@ docker run --rm ghcr.io/forestadmin/agent-bff:latest openapi > openapi.json
...
+  -e FOREST_ENV_SECRET="..." \
+  -e FOREST_SERVER_URL="https://api.forestadmin.com" \
+  -e FOREST_APP_URL="https://app.forestadmin.com" \
+  -e AGENT_URL="http://host.docker.internal:3351" \
+  -e BFF_TOKEN_ENCRYPTION_KEY="$(openssl rand -base64 32)" \
+  -e OTEL_EXPORTER_OTLP_ENDPOINT="http://collector:4318" \
+  ghcr.io/forestadmin/agent-bff:latest
+```
+
+> **Note:** OpenTelemetry is bundled only in the Docker image. It is not shipped with the npm
```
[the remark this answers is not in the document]
nbouliol: Right, the note was wrong — nothing loads the preload outside the image, so there is no warning either. Corrected in 86d36a1c to say the OTEL_* variables are read by nothing outside Docker and the process starts untraced silently.

Not wiring tracing into the npm entrypoint on purpose: the packages are not npm dependencies, so every npm user who already has OTEL_EXPORTER_OTLP_ENDPOINT set in their environment would get a boot warning they can do nothing about.

packages/agent-bff/src/tracing.ts
```diff
@@ -0,0 +1,143 @@
...
+ * process exits on its own. A rejected flush is swallowed for the same reason: it must not become
+ * an unhandled rejection that outlives the shutdown it is running beside.
+ */
+function armFlush(sdk: OtelSdk, onSignal: NonNullable<TracingOptions['onSignal']>): void {
+  const signals: NodeJS.Signals[] = ['SIGTERM', 'SIGINT'];
+
+  for (const signal of signals) {
+    onSignal(signal, () => {
+      void sdk.shutdown().catch(() => undefined);
+    });
```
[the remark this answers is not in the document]
nbouliol: Right, and the fix in 20670ef7 goes further than re-emitting: the flush no longer registers a signal handler at all.

Re-emitting would not have closed the window you found. If the signal lands before the CLI arms armShutdown, the re-raise finds no listener either — and at PID 1 there is no default disposition to fall back on, so the process still would not terminate. I verified that exact behaviour earlier on this stack: a re-raising handler at PID 1 gets SIGKILLed, exit 137.

So the preload now parks the SDK on a shared handle and arms nothing. armShutdown is the only termination handler, and it flushes through that handle. A signal arriving before it is armed hits no listener at all, which at PID 1 is the state that was already there before tracing existed.

packages/agent-bff/src/tracing.ts
```diff
@@ -0,0 +1,143 @@
...
+ * So this only flushes. Once both the flush and that shutdown settle the loop empties and the
+ * process exits on its own. A rejected flush is swallowed for the same reason: it must not become
+ * an unhandled rejection that outlives the shutdown it is running beside.
+ */
+function armFlush(sdk: OtelSdk, onSignal: NonNullable<TracingOptions['onSignal']>): void {
+  const signals: NodeJS.Signals[] = ['SIGTERM', 'SIGINT'];
+
+  for (const signal of signals) {
+    onSignal(signal, () => {
+      void sdk.shutdown().catch(() => undefined);
```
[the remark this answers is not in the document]
nbouliol: Fixed in 20670ef7, the same change as the comment above — armShutdown now awaits the flush alongside server.stop(), so the exit cannot cut an export in flight.

One deliberate difference from 'coordinate it with the shutdown': the flush gets its own 2s deadline, not the 10s in-flight requests get. Awaiting it under graceMs made a dead collector hold the process for the full grace period, and I measured that turning into exit 137 — the orchestrator's own patience starts at the same 10 seconds, so we would have traded a few lost spans for a lost shutdown. Worst case now adds ~3s (2s flush deadline + the 1s force-exit fallback), verified exit 0 with an unreachable collector.

packages/agent-bff/src/tracing.ts
```diff
@@ -0,0 +1,115 @@
...
+  const serviceName = env.OTEL_SERVICE_NAME || DEFAULT_SERVICE_NAME;
+  const sdk = new NodeSDK({
+    serviceName,
+    traceExporter: new OTLPTraceExporter(),
+    instrumentations: [getNodeAutoInstrumentations()],
+  });
+
+  sdk.start();
+
+  logger('Info', 'OpenTelemetry tracing enabled', { serviceName, endpoint });
```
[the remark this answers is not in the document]
nbouliol: Good catch, fixed in 0e9a5d33. Basic auth in the endpoint is a real way to reach a collector, and this package is careful not to echo a secret anywhere else.

Stripped the userinfo rather than reducing to `origin`: the path matters here, since OTLP endpoints commonly carry `/v1/traces` and the line is there to say where spans go. An endpoint that does not parse is dropped from the log entirely instead of passed through — it cannot be redacted, so it cannot be shown, and the SDK will fail on it on its own soon enough.

Verified in the container: the line reads `https://127.0.0.1:4318/v1/traces` and `docker logs` contains zero occurrences of the secret.

packages/agent-bff/src/shutdown.ts
```diff
@@ -75,8 +109,7 @@ export default function armShutdown(options: ShutdownOptions): void {
     stopping = true;
     logger?.('Info', 'Shutting down', { signal, graceMs });
 
-    server
-      .stop(graceMs)
+    Promise.all([server.stop(graceMs), flush ? bounded(flush(), flushMs) : undefined])
       .then(() => {
```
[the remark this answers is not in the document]
nbouliol: Right, fixed in be4c5a9e. Promise.all is fail-fast, so a rejected stop() reached exit(1) with the export still in flight and the 1s force-exit fallback cut it — losing exactly the spans that explain the failed shutdown.

Used allSettled and read the server's result out of it rather than rethrowing, which keeps the two outcomes in one place instead of a then/catch pair. Covered by a test that releases the flush after the stop has already rejected and asserts nothing exits before it.

packages/agent-bff/src/tracing.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    return undefined;
+  }
+
+  const { NodeSDK, getNodeAutoInstrumentations, OTLPTraceExporter } = modules;
+  const serviceName = env.OTEL_SERVICE_NAME || DEFAULT_SERVICE_NAME;
+  const sdk = new NodeSDK({
+    serviceName,
+    traceExporter: new OTLPTraceExporter(),
+    instrumentations: [getNodeAutoInstrumentations()],
```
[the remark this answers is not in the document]
nbouliol: Correct, fixed in be4c5a9e — took the suggestion as written. Supplying traceExporter put NodeSDK on its manual path where it never reads OTEL_TRACES_EXPORTER, so `none` still exported.

Omitting it on `none` keeps the instrumentations armed, which is the spec-faithful reading: only the export stops, trace context still propagates to the agent and the Forest SaaS. The log line says `exporter: none` instead of naming an endpoint it will not export to. Tested on none / NONE / ' none ' and on any other value.

One honest limitation: this leans on NodeSDK's "no exporter supplied → read the environment" branch, so it is coupled to the pinned SDK version. The tests assert our call shape, not the SDK's behaviour — an upstream change to that branch would not be caught here.

packages/agent-bff/Dockerfile
```diff
@@ -109,4 +109,8 @@ LABEL org.opencontainers.image.title="forestadmin-agent-bff" \
 # Exec form, no wrapper script: the CLI takes positional subcommands (`openapi`),
 # so a "is the first argument a flag?" wrapper would misread them as a command to
 # exec. Use `--entrypoint` to run anything else in the image.
-ENTRYPOINT ["node", "/app/packages/agent-bff/dist/cli.js"]
+#
+# tracing-preload runs before cli.js so the OTel auto-instrumentation patches land
+# before the app imports anything. It is inert until OTEL_EXPORTER_OTLP_ENDPOINT is
+# set, so this costs an install that never opted into APM nothing but the require.
+ENTRYPOINT ["node", "--require", "/app/packages/agent-bff/dist/tracing-preload.js", "/app/packages/agent-bff/dist/cli.js"]
```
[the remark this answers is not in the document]
nbouliol: You are right about the behaviour — `openapi` does drop its spans — but I implemented the fix, measured it, and backed it out.

Calling shutdown on the SDK forces an export attempt, and its retries hold the event loop long after the flush stops being awaited. My bound stops the waiting; it does not release the handles. Measured in the image against an unreachable collector:

    without the flush   1.3s
    with the flush      9.0s

Capping that would need a hard process.exit on this path, and this is the command whose entire output is the OpenAPI document on stdout — under Docker that is a pipe, so exiting outright risks truncating the document it was asked for. Trading a reliable export for a few spans from a one-shot CLI run is the wrong way round.

So it stays unflushed, with the reasoning and the numbers in a comment in cli.ts rather than left as an oversight. Happy to reinstate it if you would rather have the spans — say so and I will.

packages/agent-bff/src/tracing.ts
```diff
@@ -0,0 +1,149 @@
...
+    ...(exportsTraces ? { traceExporter: new OTLPTraceExporter() } : {}),
+    instrumentations: [getNodeAutoInstrumentations()],
+  });
+
+  sdk.start();
+
+  logger('Info', 'OpenTelemetry tracing enabled', {
+    serviceName,
+    // Naming the endpoint it will not export to would read as a promise it is not keeping.
+    ...(exportsTraces ? { endpoint: redactEndpoint(endpoint) } : { exporter: 'none' }),
```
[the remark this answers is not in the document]
nbouliol: Right — same bug as the last one, one step further. Fixed in d5a1f197, took the suggestion as written.

We now supply an exporter only when nothing else was requested: unset, or otlp said out loud. Everything else goes to the SDK. Your framing of why is the one that convinced me: exporting to OTLP because we did not recognise the value sends telemetry somewhere the operator never asked for, which is worse than not exporting at all.

Verified in the image, since only the OTLP exporter ships there: console and zipkin both log `exporter: <value>` and export nothing, rather than quietly reaching the OTLP endpoint. Tested on none / NONE / ' none ' / console / zipkin / 'otlp,console', and on otlp / OTLP / ' otlp ' still exporting.

Tonours (commented):
Good layer, and the design is better than the description says (more on that below). Splitting `tracing.ts` (pure setup) from `tracing-preload.ts` (the `--require`) from `tracing-handle.ts` (the one shared bit of state) is the right split, and wiring the flush into `armShutdown` instead of letting the preload arm its own handler is what makes the exit wait for the export rather than race it. 19 tests on `shutdown.ts` covering the flush deadline, a failing `stop()` with a flush still in flight, and the interrupted paths.

I built the image and exercised it rather than take the claims on trust. Built from `d5a1f197b`, arm64.

| Check | Result |
|---|---|
| `docker build` | exit 0, **337 MB**, ENTRYPOINT `node --require .../tracing-preload.js .../cli.js` |
| OTel packages shipped | 79 under `node_modules/@opentelemetry`, `node_modules` 36 MB -> 152 MB |
| `smoke-test.sh` | exit 0 |
| Tracing off, no endpoint | no tracing line, `docker stop` in 0s |
| Tracing on, collector unreachable | `OpenTelemetry tracing enabled`, `docker stop` in 3s, exit 0 |
| Flush deadline | `Shutting down` -> `Forest BFF stopped` = **2.006s**, exactly `flushMs` |
| `OTEL_SDK_DISABLED=TRUE` | disabled, the uppercase fix holds |
| Credential redaction | `https://user:s3cret@collector.example:4318` logged as `https://collector.example:4318/`, zero occurrences of the secret |

So the `bounded()` deadline does exactly what it promises: a dead collector costs 2 seconds, not the 10 the in-flight requests get, and the process still ends on 0. The "~3 seconds" in the README is the real number.

Two things on the description itself, since neither is a line in the diff:

**The "signal handler re-raises" section is stale.** It describes `once` plus a re-raise of the same signal, and none of that is in the final code: `4a087d940` and `6a8bf5cae` moved the flush into `armShutdown` through `tracing-handle`, and `tracing.ts:13-14` now says arming a handler there would be the bug. Worth rewriting before merge, it's the part reviewers will read, and it points them at a mechanism that no longer exists.

**The size cost isn't mentioned.** 271 MB -> 337 MB, +24%, and `node_modules` more than quadruples, for a layer that is off by default. Nothing to fix, `auto-instrumentations-node` pulls the 79 packages and that's the price. But #1848 put "image 271 MB" in its description, so this one saying nothing means someone finds the number in production instead.

Line comments below. Nothing blocking, one is worth three lines of guard before merge.

packages/agent-bff/src/tracing.ts
```diff
@@ -0,0 +1,157 @@
...
+export type ModuleLoader = (id: string) => Record<string, unknown>;
+
+export function loadOtelModules(load: ModuleLoader = require): OtelModules | undefined {
+  try {
+    return {
+      NodeSDK: load(OTEL_MODULE_IDS.sdk).NodeSDK,
+      getNodeAutoInstrumentations: load(OTEL_MODULE_IDS.instrumentations)
+        .getNodeAutoInstrumentations,
+      OTLPTraceExporter: load(OTEL_MODULE_IDS.exporter).OTLPTraceExporter,
+    } as OtelModules;
```
Tonours: The `try/catch` above only covers resolution, so this returns a **truthy** object whose members can be `undefined`. A package that loads but no longer exports `NodeSDK` (a rename on a version bump) skips the `if (!modules)` warning branch and reaches `new NodeSDK(...)` at :140. Reproduced:

```
modules truthy (so no warning branch): true | NodeSDK = undefined
TypeError: NodeSDK is not a constructor
```

And this lives in a `--require`, so node aborts **before the entry point**. I confirmed that separately: `loadPreloadModules` in the stack, `cli.js` never runs. So a dead container, for an APM layer that's optional and best-effort, and it contradicts what :24 promises ("a warning rather than a crash").

The smoke test would catch it before publish, since the boot fails, so it can't reach the registry. Still worth closing: checking the three are functions before returning turns a dead container into a logged warning. There's no test for a loaded-but-incomplete module either, and `tracing-preload.test.ts` never exercises `initTracing` throwing.

packages/agent-bff/src/tracing.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export default function initTracing(options: TracingOptions = {}): OtelSdk | undefined {
+  const { env = process.env, logger = createConsoleLogger(), load = loadOtelModules } = options;
+
+  const endpoint = env.OTEL_EXPORTER_OTLP_ENDPOINT?.trim();
+
+  // A blank endpoint counts as unset, not as "export to the OTel default": `env_file` hands an
+  // empty string through for a variable left blank in a `.env`, and arming the SDK against
+  // localhost:4318 is not what leaving the line empty asks for.
+  if (!endpoint || isDisabled(env.OTEL_SDK_DISABLED)) return undefined;
```
Tonours: This gates on the endpoint before anything else, so `OTEL_TRACES_EXPORTER=console` needs an OTLP endpoint set to do anything, which reads as nonsense for `console`. The README table presents it as usable on its own.

Either gate on `endpoint || OTEL_TRACES_EXPORTER`, or say it in the table. Low, but someone will try `console` first when debugging.

packages/agent-bff/src/tracing.ts
```diff
@@ -0,0 +1,157 @@
...
+  // which configures it from the environment. Sending spans to the OTLP endpoint because we did not
+  // recognise the value would be worse than not sending them: it is telemetry going somewhere the
+  // operator did not ask for, and the SDK says so out loud when it cannot honour the request.
+  //
+  // Instrumentation stays on in every case, so trace context keeps propagating to the agent and the
+  // Forest SaaS whatever the export does.
+  const requestedExporter = env.OTEL_TRACES_EXPORTER?.trim().toLowerCase();
+  const exportsTraces = !requestedExporter || requestedExporter === 'otlp';
+  const sdk = new NodeSDK({
+    serviceName,
```
Tonours: Two small ones on this block.

`serviceName` is always passed, so it overrides `OTEL_RESOURCE_ATTRIBUTES=service.name=...`, which the README documents as a supported knob. Someone setting only that gets `forestadmin-agent-bff` silently. Passing it only when `OTEL_SERVICE_NAME` is set lets the SDK resolve the rest by the spec's own precedence.

And `new OTLPTraceExporter()` reads the real `process.env`, not `options.env`. No effect in production since they're the same object, but it means the injected-env tests never cover where the exporter actually points: `tracing.test.ts:116` asserts it was constructed, not its target. The seam stops half way.

packages/agent-bff/docker/scan-gate.js
```diff
@@ -45,4 +44,15 @@ const summary = lines.join('\n');
 console.log(summary);
 if (process.env.GITHUB_STEP_SUMMARY) fs.appendFileSync(process.env.GITHUB_STEP_SUMMARY, `${summary}\n`);
 
-console.log('\nEvery library finding above is report-only — this step gates nothing today.');
+// Block only on Docker-only OTel packages — everything else is report-only.
+const blocking = findings.filter(f => f.pkg.startsWith('@opentelemetry/'));
```
Tonours, line 48: The prefix here and `OTEL_DEPENDENCIES` in `build-deps-manifest.js` are two sources of truth for "what is Docker-only". Importing the list keeps them together, and it would also stop this from blocking on an `@opentelemetry/*` package that later became a real BFF npm dependency, which is exactly the GHCR/npm desync the origin rule exists to prevent.

Separately: this now means a fixable CVE in any transitive OTel package blocks **every** release of this image until someone bumps a pin and regenerates the isolated lockfile. That's the trade-off you're choosing on purpose, and the `propagator-jaeger` resolution shows the workflow. But the base image digest carries a "Bump via Renovate/Dependabot" note and these pins carry nothing, so a Renovate rule on `OTEL_DEPENDENCIES` (or a named owner) would keep it from stalling an unrelated release.

packages/agent-bff/docker/build-deps-manifest.js
```diff
@@ -102,4 +114,4 @@ if (require.main === module) {
   generate(packagesDir, outFile);
 }
 
-module.exports = { WORKSPACE_PACKAGES, generate };
+module.exports = { WORKSPACE_PACKAGES, OTEL_DEPENDENCIES, generate };
```
Tonours: `OTEL_DEPENDENCIES` is exported and nothing imports it. Either wire it into `scan-gate.js` (see the comment there) or drop it from the export.

packages/agent-bff/docker/smoke-test.sh
```diff
@@ -37,12 +38,17 @@ grep -q '"openapi"' /tmp/bff-openapi.json
...
+# OTEL_EXPORTER_OTLP_ENDPOINT is set so the SDK actually initialises: the packages
+# exist only in this image, so nothing else would prove they are loadable. The
+# receiver is unreachable on purpose — exporting is asynchronous and best-effort.
 CONTAINER=$(docker run -d -p "127.0.0.1:$PORT:3450" \
   -e FOREST_AUTH_SECRET=smoke-test \
   -e FOREST_ENV_SECRET="$(openssl rand -hex 32)" \
   -e FOREST_SERVER_URL=http://127.0.0.1:1 \
   -e FOREST_APP_URL=http://127.0.0.1:1 \
   -e AGENT_URL=http://127.0.0.1:1 \
+  -e OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:4318 \
```
Tonours, line 51: Tracing is armed on this boot only. The fully-configured second boot has no `OTEL_EXPORTER_OTLP_ENDPOINT`, so the flush-on-shutdown path never runs in the image, and that's the riskiest new code here.

I ran it by hand and it's fine: `docker stop` on this container with a dead collector takes 3s and exits 0, with 2.006s between `Shutting down` and `Forest BFF stopped`. Timing that here would pin the "~3 seconds" the README promises, and would catch a `bounded()` regression that no unit test can see from inside the process.

packages/agent-bff/src/tracing.ts
nbouliol: Confirmed and fixed in 296a6c27 — I reproduced your case before touching it:

    modules truthy (so no warning branch): true | NodeSDK = undefined
    TypeError: m.NodeSDK is not a constructor

You are right that it contradicts what the header promises, and worse in a --require than anywhere else. loadOtelModules now checks the three are callable before returning, so an incomplete package takes the warning branch like a missing one.

Went one further on your last point: the preload wraps the whole call, so nothing thrown there can abort the process — not just this case, but whatever a future SDK version decides to throw from start(). Tests for both: one per package loading without the export we read, and one asserting the preload logs and continues when initTracing throws.

packages/agent-bff/src/tracing.ts
nbouliol: Fixed in 296a6c27 — took the first option rather than documenting the limitation, since console is genuinely useful on its own and the console exporter ships with sdk-node, so it works in the image as-is.

Either variable now turns tracing on. Verified: OTEL_TRACES_EXPORTER=console with no OTLP endpoint logs `exporter: console` and arms the SDK, where before it did nothing at all. A blank value still counts as unset either way. README table updated to say so instead of presenting the endpoint as the only switch.

packages/agent-bff/src/tracing.ts
nbouliol: First one fixed in 296a6c27. serviceName is now passed only when the environment names no service, so OTEL_SERVICE_NAME and a service.name resource attribute both reach the SDK and it applies the spec's precedence; our default fills in only when neither is set. Verified in the image: OTEL_RESOURCE_ATTRIBUTES=service.name=bff-from-attrs now reports bff-from-attrs where it used to report the default.

Second one I did not fix, because the honest fix is worse. Passing a url would mean re-deriving it from OTEL_EXPORTER_OTLP_ENDPOINT ourselves — appending /v1/traces, honouring OTEL_EXPORTER_OTLP_TRACES_ENDPOINT — which is the SDK's job and a good way to get it subtly wrong. So the seam stops where you say it does, and I documented that boundary on the option itself: env steers our decisions, not the SDK's, and a test injecting it cannot assert where spans go.

packages/agent-bff/docker/scan-gate.js
nbouliol: Pushing back on the first half. Importing OTEL_DEPENDENCIES would narrow the gate to the three direct pins, and the finding that actually showed up is propagator-jaeger — transitive, not in that list. So the exact-match version would have let through the one CVE this gate has caught so far, to prevent an over-block that has not happened: no BFF npm dependency brings @opentelemetry today.

The prefix is an approximation, and I would rather it err toward blocking. If a BFF dependency ever does pull OTel in, that is the moment to make the rule precise — and it would be visible, because the gate would start failing on a package nobody here pinned.

Second half taken, in 296a6c27. You are right that the pins carried nothing while the base image digest says how it gets bumped. OTEL_DEPENDENCIES now records that a fixable CVE in that tree blocks every publish until someone raises a pin and refreshes the lockfile, and that Renovate does not watch this file — it is a plain object, not a manifest — so it stalls silently unless someone is looking. A Renovate rule would be better than a comment; that needs a decision about where it lives, so I left it as a flag rather than guessing.

Also dropped the unused export, per your other comment.

packages/agent-bff/docker/build-deps-manifest.js
nbouliol: Dropped from the export in 296a6c27 — see the scan-gate thread for why I did not wire it in instead.

packages/agent-bff/docker/smoke-test.sh
nbouliol: Good catch, and done in 296a6c27 — you are right that it was the riskiest new code with no coverage in the image at all.

The fully-configured boot now arms tracing against a collector that is not listening, and the test stops it with -t 30 (so it measures the BFF, not whatever docker stop defaults to locally), then asserts exit 0, that 'Forest BFF stopped' appears, and that the whole thing takes at most 8s. It prints the elapsed time either way.

Reproduces your numbers exactly: exit 0 in 3s, 2.005s between 'Shutting down' and 'Forest BFF stopped'.

The 8s ceiling is set against the two regressions I actually measured on this branch rather than a round number: when the flush shared graceMs it took the full 10s and came back 137, and unbounded it ran ~9s on the OTLP retry budget. Both trip it.

packages/agent-bff/src/tracing.ts
```diff
@@ -0,0 +1,198 @@
...
+  // which configures it from the environment. Sending spans to the OTLP endpoint because we did not
+  // recognise the value would be worse than not sending them: it is telemetry going somewhere the
+  // operator did not ask for, and the SDK says so out loud when it cannot honour the request.
+  //
+  // Instrumentation stays on in every case, so trace context keeps propagating to the agent and the
+  // Forest SaaS whatever the export does.
+  const exportsTraces = !requestedExporter || requestedExporter === 'otlp';
+  const sdk = new NodeSDK({
+    ...(named ? {} : { serviceName: DEFAULT_SERVICE_NAME }),
+    ...(exportsTraces ? { traceExporter: new OTLPTraceExporter() } : {}),
```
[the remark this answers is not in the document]
nbouliol: Both correct, and chasing them one at a time was the actual mistake. Fixed in 5a07686b by removing the cause: we no longer pass traceExporter at all.

OTEL_TRACES_EXPORTER, then OTEL_EXPORTER_OTLP_PROTOCOL, then the per-signal endpoint — three rounds of the same root cause, which is that supplying an exporter puts NodeSDK on its manual path where it stops reading the environment. Special-casing one variable per round was never going to converge.

What unlocked it: sdk-node already depends on every OTLP exporter and on the Zipkin one. Verified inside the image:

    present exporter-trace-otlp-http
    present exporter-trace-otlp-proto
    present exporter-trace-otlp-grpc
    present exporter-zipkin

So handing the choice to the SDK costs no new dependency and makes the whole standard surface work — exporter, protocol, per-signal endpoint and per-signal protocol. Our decisions are now only whether to arm at all, and the service name when nothing else supplies one. Activation counts OTEL_EXPORTER_OTLP_TRACES_ENDPOINT too, and the log line reports what was asked for rather than a destination we no longer pick.

Verified in the image: the per-signal endpoint alone now arms and reports itself; protocol=grpc no longer silently builds an HTTP exporter.

Note this also changes the default wire format from http/json to the spec's http/protobuf, since the SDK now picks it. Nothing is published yet, so there is no migration.

packages/agent-bff/docker/scan-gate.js
nbouliol: Correction to something I told you two rounds ago, on the OTEL_TRACES_EXPORTER thread: I said "only the OTLP exporter ships in the image", so console and zipkin would export nothing. That was wrong, and I only found out while chasing the protocol comment.

sdk-node depends on every OTLP exporter and on the Zipkin one. Checked inside the image:

    present exporter-trace-otlp-http
    present exporter-trace-otlp-proto
    present exporter-trace-otlp-grpc
    present exporter-zipkin

The log-line behaviour I reported was real — I ran it — but my explanation for it was not, and the README repeated the same claim. Both fixed in 5a07686b.

It does not change the argument on this thread: the prefix rule still errs toward blocking, and the transitive propagator-jaeger finding is still the one it caught.

packages/agent-bff/src/tracing.ts
Tonours: Closed twice over: `loadOtelModules` now checks every export is a function, and the preload catches whatever is left and starts untraced with a warning. Good.

packages/agent-bff/src/tracing.ts
Tonours: Gate is `(!endpoint && !requestedExporter)` now, and picking up `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` on the way is a bonus. Good.

packages/agent-bff/src/tracing.ts
Tonours: Both halves. `serviceNameFromEnv` follows the spec precedence, and dropping `traceExporter` entirely removes the seam problem rather than patching it. Better than what I suggested.

packages/agent-bff/docker/build-deps-manifest.js
Tonours: Gone from the export. Good.

packages/agent-bff/docker/smoke-test.sh
Tonours: Goes further than I asked: `docker stop -t 30`, exit 0 asserted with 137 named, `Forest BFF stopped` checked, and an 8s ceiling. That last one is the regression guard `bounded()` needed.

packages/agent-bff/docker/scan-gate.js
Tonours: Taking the pushback. Narrowing the gate to the three direct pins would have let `propagator-jaeger` through, which is the only CVE it has caught — I was optimising against an over-block that has not happened, and no BFF npm dependency brings `@opentelemetry` today. Erring toward blocking is the right default here, and the day a real dependency pulls OTel in, the gate fails on a package nobody pinned, so it surfaces on its own.

Second half is in `build-deps-manifest.js:30-33`. A Renovate rule is still better than a comment, but where it lives is a separate decision and the flag is enough to keep it from stalling a release silently.

Closing this.

Tonours (approved):
Approving. I went back through all 16 findings and checked each one against the code at `50d3f1d`, not against the replies.

What I verified:

- Signal handling — `tracing.ts` arms nothing, the SDK is parked on `tracing-handle` and `armShutdown` is the only terminating handler. Pinned by `tracing.test.ts:239`.
- Credentials in the logs — `redactEndpoint` (`tracing.ts:121`), applied at the log line. Pinned, user:password and token-as-user.
- `OTEL_SDK_DISABLED` — trim + lowercase, pinned over `TRUE` / `True` / `' true '`.
- `Promise.all` on the shutdown path — now `allSettled` with the server result read out of it. Pinned by `shutdown.test.ts:118`.
- Module that resolves without the export we read — every export checked callable, plus the preload catching whatever is left. Pinned per package.
- Exporter, protocol and per-signal endpoint — root cause removed rather than patched per variable: no `traceExporter` at all, the SDK reads the environment. Pinned.
- The flush path in the image — `smoke-test.sh:160-214` arms OTLP against a dead collector on the second boot and asserts exit 0 with a bounded elapsed. Confirmed it runs in CI on this head (`docker-publish-bff` run 32994136258, success), so this and the scan gate are actually exercised.

Two things I did not ask you to change: `openapi` not flushing is the right call at 9s versus 1.3s, and the `options.env` seam stopping short of where spans go is the honest boundary — both are documented where someone will hit them.

One nit, not blocking: the fix for the `OTEL_TRACES_EXPORTER` gate left three comments behind saying tracing only arms on `OTEL_EXPORTER_OTLP_ENDPOINT` — `Dockerfile:114`, `tracing.ts:10`, and the variable list at `tracing.ts:15-20` which omits `OTEL_TRACES_EXPORTER` and `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`. The README table is correct; these are not.

Note for merge order: this stacks on `feat/agent-bff-docker-image`, so the parent goes first.

packages/agent-bff/src/tracing.ts
```diff
@@ -0,0 +1,195 @@
...
+ * that token should end up — this package does not echo a secret anywhere else either.
+ *
+ * An endpoint that does not parse is dropped from the log entirely rather than passed through: it
+ * cannot be redacted, so it cannot be shown. The SDK will fail on it soon enough on its own.
+ */
+export function redactEndpoint(endpoint: string): string | undefined {
+  try {
+    const url = new URL(endpoint);
+
+    if (!url.username && !url.password) return endpoint;
```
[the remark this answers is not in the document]
nbouliol: Right, fixed — the query goes too. `2f906f9`

Stripping only the userinfo was half a redaction: a collector addressed as `?api_key=...` wrote its key straight to the container logs, which is exactly what this function exists to prevent. Telling a credential parameter apart from a harmless one means knowing every vendor's spelling, so the whole query is dropped unread — the host and the path are what that log line is for.

One difference from your diff: the early return keeps its place, widened to `!url.username && !url.password && !url.search`. Returning `url.toString()` for every endpoint would also normalise the ones with nothing to hide (`http://collector:4318` → `http://collector:4318/`), and the log should echo what the operator set. Tests cover a bare query key, and a query alongside userinfo.

packages/agent-bff/src/tracing.ts
```diff
@@ -0,0 +1,195 @@
...
+  // No `traceExporter`, deliberately. Passing one puts NodeSDK on its manual-configuration path,
+  // where it stops reading the environment — which is how OTEL_TRACES_EXPORTER, then
+  // OTEL_EXPORTER_OTLP_PROTOCOL, then the per-signal endpoint each turned out to be silently
+  // ignored, one review round after another. They were three symptoms of doing the SDK's job.
+  //
+  // sdk-node depends on every OTLP exporter and on the Zipkin one, so all of them are in the image
+  // already; leaving the choice to the SDK costs nothing and makes the whole standard surface work,
+  // protocol and per-signal overrides included. Our only decisions are whether to arm at all, and
+  // the service name when nothing else supplies one.
+  const sdk = new NodeSDK({
```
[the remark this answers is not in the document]
nbouliol: Confirmed against the pinned SDK, and fixed. `2f906f9`

Unpacked `@opentelemetry/sdk-node@0.219.0` rather than take it on trust — `getMetricReadersFromEnv` and `configureLoggerProviderFromEnv` both do the same thing:

```js
if (enabledExporters.length === 0) {
  diag.debug('OTEL_METRICS_EXPORTER is empty. Using default otlp exporter.');
  enabledExporters.push('otlp');
}
```

So unset means otlp, and arming traces started a metric reader and a log processor aimed at `http://localhost:4318`. Under `OTEL_TRACES_EXPORTER=console`, which asks for no collector at all, that is a POST a minute to a port nobody named; where something else on the host does listen there, it is telemetry that install never opted into.

`initTracing` now fills both in with `none` when the environment left them blank, right before constructing the SDK. Opting a signal back in still works through its own standard variable, since only a blank one is filled. Documented in the README's OTel table and in the module header; tests assert both the default and that an explicit `otlp` survives, and that a run which never arms leaves the environment untouched.

Note for the record: passing `metricReaders: []` / `logRecordProcessors: []` would *not* have been equivalent — an empty array is truthy, so the SDK would still build a provider for each, just without an exporter. The `none` path skips construction outright.

.github/workflows/docker-publish-bff.yml
```diff
@@ -352,6 +352,27 @@ jobs:
...
         run: |
+          # :$FULL is advertised as immutable, so a re-dispatch of a version that
+          # already shipped must not repoint it at the images this run just built.
+          # Compare what the tag resolves to against those digests — the attestation
+          # manifests the SBOM adds carry no architecture, so they are not part of the
+          # comparison. Identical means this is a replay (recovering a merge job that
+          # died, or moving the mutable tags): let it through, imagetools rewrites the
+          # same list. Different means the version was rebuilt, and the answer is a new
+          # version, not a silently different 1.20.2.
+          if EXISTING=$(docker buildx imagetools inspect "$IMAGE:$FULL" --raw 2>/dev/null); then
```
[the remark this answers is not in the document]
nbouliol: Correct, and it defeated the check I had just added. Fixed in `cae987f4`.

The group was `docker-publish-bff-${{ github.ref }}-<version>`, so two dispatches for the same version from different refs sat in different groups: both resolve the tag as free, both publish, the later `imagetools create` wins. The tag check is only worth anything if same-version publishes are serialised.

A publish is now keyed by the version alone:

```yaml
group: docker-publish-bff-${{ github.event_name == 'workflow_dispatch' && format('publish-{0}', github.event.inputs.version) || github.ref }}
cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

Queued, not cancelled — cancelling the first run would leave the second checking against a tag the first was midway through writing, which is the same race with extra steps. Waiting means the second run's inspect sees exactly what the first published, and stops there. Different versions still publish concurrently, and the merge job still re-decides latest-stable against the tags as they are then, so out-of-order finishes cannot regress the mutable tags.

PRs keep the old behaviour, grouped by ref and cancelling: a superseded validation of an older push is waste, not a release.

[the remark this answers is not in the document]
nbouliol: Yes — `2>/dev/null` made every failure look like an absent tag, which is the one reading that lets the publish proceed. Fixed in `cae987f4`.

Probed the two shapes against ghcr rather than guess at the matching:

```
$ docker buildx imagetools inspect ghcr.io/forestadmin/workflow-executor:no-such-tag-xyz --raw
ERROR: ghcr.io/forestadmin/workflow-executor:no-such-tag-xyz: not found

$ docker buildx imagetools inspect ghcr.io/forestadmin/agent-bff:1.0.0 --raw
ERROR: failed to authorize: failed to fetch anonymous token: ... 403 Forbidden
```

stderr is kept now, and only an explicit not-found (`not found`, `manifest unknown`, `name unknown`) means the tag is free. Anything else — a blip, an expired token, a registry 5xx — prints the error and stops the publish, rather than handing this run a tag it never resolved.

Ran the four branches against a stubbed inspect: same digests → no-op, different digests → blocked, not-found → publish, authorize failure → blocked.

PR #1850, opened 2026-08-25
fix(agent-bff): advertise mode 1 on the data and action routes

Tonours:
Stacked on #1845 (`feat/prd-1033-ai-query-proxy`), which merges first. Only the last commit belongs to this PR; everything else in the diff comes from the parent.

fixes PRD-936

## What

`SECURITY` in `openapi-document.ts` now lists `bffSession` alongside `bffApiKey`, so the six data and action paths advertise both auth modes. The `bffSession` scheme description no longer says those routes take the API key only; it states the real constraint instead — the session must carry a usable rendering, or the request is rejected with 401.

## Why

Mode 1 has reached data and action routes since #1800 (PRD-866): `auth-mode-middleware.ts` mints an agent token from the OAuth principal into `ctx.state.agentToken`, and `requireAgentToken` reads it. `security` is the field a code generator reads, so every generated client kept sending an API key on data routes and Mode 1 looked unsupported.

Two lines of production code, plus the fallout of advertising a second scheme: the docs page samples and the generated client both picked the first scheme in the list and switched to the bearer, so a reader who unlocked with an API key got a snippet sending a credential they do not hold, and a client configured with one blanket `auth` callback sent BOTH headers and got 400 `ambiguous_credentials`. `docs-samples.ts` now prefers the API key wherever an operation accepts it, and the generated-client test pins the scheme-discriminating callback with a regression test on the `Authorization` header. No auth behaviour changes.

## How to test

```
cd packages/agent-bff && npx jest test/openapi
node dist/cli.js openapi | jq -r '.paths | to_entries[] | .key as $p | .value | to_entries[] | "\($p) -> \([.value.security[]? | keys[]] | join(", "))"'
```

Every data and action path prints `bffSession, bffApiKey`.

---

### General

### Security

### DIFF
--- a/packages/agent-bff/src/docs/docs-samples.ts
+++ b/packages/agent-bff/src/docs/docs-samples.ts
@@ -110,21 +110,22 @@ const SAMPLES_SCRIPT = `
         function authHeader(spec, operation) {
           var schemes = (spec.components || {}).securitySchemes || {};
           var requirements = operation.security || spec.security || [];
-          var header = null;
+          var apiKey = null;
+          var bearer = null;
 
           requirements.forEach(function (requirement) {
             Object.keys(requirement).forEach(function (name) {
               var scheme = schemes[name] || {};
 
               if (scheme.type === 'apiKey' && scheme.in === 'header') {
-                header = { name: scheme.name, prefix: '', secret: true };
-              } else if (!header && scheme.type === 'http' && scheme.scheme === 'bearer') {
-                header = { name: 'Authorization', prefix: 'Bearer ', secret: true };
+                apiKey = { name: scheme.name, prefix: '', secret: true };
+              } else if (scheme.type === 'http' && scheme.scheme === 'bearer') {
+                bearer = { name: 'Authorization', prefix: 'Bearer ', secret: true };
               }
             });
           });
 
-          return header || { name: KEY_HEADER, prefix: '', secret: true };
+          return apiKey || bearer || { name: KEY_HEADER, prefix: '', secret: true };
         }
 
         /**
--- a/packages/agent-bff/src/docs/docs-samples.ts
+++ b/packages/agent-bff/src/docs/docs-samples.ts
@@ -107,7 +107,6 @@ const SAMPLES_SCRIPT = `
           return body;
         }
 
-        // The scheme the operation names, so a document that renames or re-types it stays right.
         function authHeader(spec, operation) {
           var schemes = (spec.components || {}).securitySchemes || {};
           var requirements = operation.security || spec.security || [];
@@ -115,13 +114,11 @@ const SAMPLES_SCRIPT = `
 
           requirements.forEach(function (requirement) {
             Object.keys(requirement).forEach(function (name) {
-              if (header) return;
-
               var scheme = schemes[name] || {};
 
               if (scheme.type === 'apiKey' && scheme.in === 'header') {
                 header = { name: scheme.name, prefix: '', secret: true };
-              } else if (scheme.type === 'http' && scheme.scheme === 'bearer') {
+              } else if (!header && scheme.type === 'http' && scheme.scheme === 'bearer') {
                 header = { name: 'Authorization', prefix: 'Bearer ', secret: true };
               }
             });
--- a/packages/agent-bff/src/docs/docs-samples.ts
+++ b/packages/agent-bff/src/docs/docs-samples.ts
@@ -107,6 +107,8 @@ const SAMPLES_SCRIPT = `
           return body;
         }
 
+        // The page only ever unlocks with the API key, so it wins wherever an operation takes it;
+        // a session-only operation still samples the bearer it names.
         function authHeader(spec, operation) {
           var schemes = (spec.components || {}).securitySchemes || {};
           var requirements = operation.security || spec.security || [];
--- a/packages/agent-bff/src/openapi/openapi-document.ts
+++ b/packages/agent-bff/src/openapi/openapi-document.ts
@@ -409,7 +409,7 @@ export function generateOpenApiDocument(
     path: `${ROUTE_PREFIX}/context`,
     operationId: 'getContext',
     summary: 'Read the exposed schema contract',
-    security: [{ [SESSION_SCHEME]: [] }, { [API_KEY_SCHEME]: [] }],
+    security: SECURITY,
     request: {},
     responses: {
       200: {
--- a/packages/agent-bff/src/openapi/openapi-document.ts
+++ b/packages/agent-bff/src/openapi/openapi-document.ts
@@ -27,7 +27,7 @@ export const ROUTE_PREFIX = '/agent/v1';
 const SESSION_SCHEME = 'bffSession';
 const API_KEY_SCHEME = 'bffApiKey';
 
-const SECURITY = [{ [API_KEY_SCHEME]: [] }];
+const SECURITY = [{ [SESSION_SCHEME]: [] }, { [API_KEY_SCHEME]: [] }];
 
 const ERROR_STATUSES: Record<string, string> = {
   400: 'Malformed body, a malformed URL-encoded path segment, an invalid filter operator, a filter nested too deep, ambiguous credentials, an unsupported page, a missing or invalid timezone, an unknown submitted action field, or a rejected action form (type action_error)',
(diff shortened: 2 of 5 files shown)


packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -27,7 +27,7 @@ export const ROUTE_PREFIX = '/agent/v1';
 const SESSION_SCHEME = 'bffSession';
 const API_KEY_SCHEME = 'bffApiKey';
 
-const SECURITY = [{ [API_KEY_SCHEME]: [] }];
+const SECURITY = [{ [SESSION_SCHEME]: [] }, { [API_KEY_SCHEME]: [] }];
```
nbouliol, line 30: Now that `SECURITY` lists both schemes, the inline literal on the context route (line 412) is character-for-character the same — could it just be `security: SECURITY`?

packages/agent-bff/src/docs/docs-samples.ts
```diff
@@ -107,27 +107,25 @@ const SAMPLES_SCRIPT = `
...
+                apiKey = { name: scheme.name, prefix: '', secret: true };
               } else if (scheme.type === 'http' && scheme.scheme === 'bearer') {
-                header = { name: 'Authorization', prefix: 'Bearer ', secret: true };
+                bearer = { name: 'Authorization', prefix: 'Bearer ', secret: true };
               }
             });
           });
 
-          return header || { name: KEY_HEADER, prefix: '', secret: true };
+          return apiKey || bearer || { name: KEY_HEADER, prefix: '', secret: true };
```
nbouliol, line 130: The precedence is right (the docs page only ever unlocks with the key, `docs-page.ts:114`) but the deleted comment was the only place saying why — worth a line here?

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -341,6 +345,12 @@ describe('a client generated from the emitted OpenAPI document', () => {
 
       expect(receivedKeys).toEqual([API_KEY]);
     });
+
+    it('should send no Authorization header beside the key, which the BFF rejects as ambiguous', async () => {
+      await sdk.countRecordsUsers({ body: {} });
+
+      expect(receivedAuthorizations).toEqual(['']);
```
nbouliol, line 360: This is guaranteed by the harness callback at line 271 rather than by the document — the case worth pinning is the blanket `auth: () => key`, which now sends `Authorization: Bearer <apiKey>` and gets 401.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -265,7 +267,8 @@ describe('a client generated from the emitted OpenAPI document', () => {
 
     client.setConfig({
       baseUrl: `http://127.0.0.1:${port}`,
-      auth: () => API_KEY,
+      auth: (scheme: { name?: string }) =>
+        scheme.name === 'X-Forest-Bff-Key' ? API_KEY : undefined,
```
nbouliol: `BFF_KEY_HEADER` is exported from `src/api-key/api-key-middleware`, like the `TIMEZONE_HEADER` this file already imports.

packages/agent-bff/test/docs/docs-page.test.ts
```diff
@@ -131,7 +131,7 @@ const API_KEY_SPEC = {
   paths: {
     '/agent/v1/My%20Coll/list': {
       post: {
-        security: [{ bffApiKey: [] }],
+        security: [{ bffSession: [] }, { bffApiKey: [] }],
```
nbouliol, line 134: The ACTION fixture at line 158 is still session-only, which no real action route is after this change — repointing it at `/ai/query` would keep the bearer branch honest and the name at line 364 accurate.

nbouliol (commented):
Reviewed the four commits against PRD-936: the `SECURITY` fix and the description rewrite both land, and the docs-samples precedence flip is the right call (the docs page only ever unlocks with the API key, so preferring it keeps the samples runnable, and Mode-1-only operations still get the bearer). Nothing blocking. Five notes below, mostly around the tests and one now-redundant literal.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Done — `security: SECURITY` on the context route. Both routes accept both modes for the same reason (the auth chain resolves either credential), so the copy bought nothing.

packages/agent-bff/src/docs/docs-samples.ts
Tonours: Added, right above `authHeader`: the page only ever unlocks with the key, so it wins wherever an operation takes it, and a session-only operation still samples the bearer it names.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
Tonours: Right, that one only proved my own harness callback. Kept it as the good-path case and added the blanket `auth: () => key` one next to it: it asserts the client sends `X-Forest-Bff-Key` AND `Authorization: Bearer <apiKey>`, which only happens because the document now lists two schemes — revert `SECURITY` and it fails. One correction: the BFF answers 400 `ambiguous_credentials`, not 401 — `resolveAuthMode` throws before either credential is verified (`auth-mode.ts:36`, `bff-http-error.ts:69`). The test name says 400.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
Tonours: Done, imported from `src/api-key/api-key-middleware` and used in the recording middleware too.

packages/agent-bff/test/docs/docs-page.test.ts
Tonours: Done. The ACTION fixture now carries both schemes like the real action routes, and a new `/agent/v1/ai/query` entry carries the session alone — the bearer assertions (cURL and Ruby) moved onto it, and the both-schemes assertion now covers LIST and ACTION.

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -341,6 +353,27 @@ describe('a client generated from the emitted OpenAPI document', () => {
...
       expect(receivedKeys).toEqual([API_KEY]);
     });
+
+    it('should send the key alone when the caller reads the scheme it is answering', async () => {
+      await sdk.countRecordsUsers({ body: {} });
+
+      expect(receivedAuthorizations).toEqual(['']);
+    });
+
+    it('should send both credentials when the caller answers every scheme with the same token, which the BFF rejects with 400 ambiguous_credentials', async () => {
```
nbouliol, line 363: The harness mounts no auth middleware (`ctx.state.agentToken` is set directly at line 174), so nothing here reaches `ambiguousCredentials` — this is the claim `b7fffdd4d` trimmed off the test above; cut the 400 clause, or assert the status?

packages/agent-bff/test/openapi/openapi-generated-client.test.ts
```diff
@@ -197,6 +205,7 @@ function documentedOperators(document: {
 
 describe('a client generated from the emitted OpenAPI document', () => {
   let codegenOutput: string;
+  let configure: (auth: AuthCallback) => unknown;
```
nbouliol, line 208: `configure` doesn't say what it configures — only `auth` varies (baseUrl and headers are fixed at 276-279), so `configureAuth` would read truer.

nbouliol (commented):
Second pass on `285c7e872`. All five notes from the first review are cleanly addressed: `security: SECURITY` on the context route, the why-comment back above `authHeader`, `BFF_KEY_HEADER` imported, the blanket-auth test replacing the circular assertion, and the ACTION fixture now carrying both schemes with a session-only `/ai/query` fixture beside it. Test isolation on the new test is sound — the `try/finally` restores `SCHEME_AWARE_AUTH` and `beforeEach` clears the captures, so ordering does not matter. Two notes left, both on the new test.

nbouliol (approved):
The two production lines are right and the knock-on fallout is handled properly: preferring the API key in the docs samples keeps every unlockable sample runnable, session-only operations still get the bearer, and the scheme-discriminating client callback is now pinned alongside the blanket-callback case. The only note left is the 400 clause in the new test's name, which the harness cannot exercise — worth trimming, not worth holding the PR.

PR #1851, opened 2026-08-25
feat(agent-client): allow per-request timezone override on action form hooks

Tonours:
fixes PRD-768

## What

`Collection.action()` takes an optional `timezone` in its action context and threads it into the `/hooks/load` and `/hooks/change` query. The BFF passes `ctx.state.timezone` on the action form and execute routes.

The BFF middleware also loads the action once before dispatching on the verb, and passes its handler and `loadAction` arguments as objects.

## Why

Action form hooks always ran in `Europe/Paris`: `HttpRequester` hardcodes that default and `FieldFormStates` sent no query. A load or change hook computing `now` or `today` returned the wrong date for every user outside that timezone, and the BFF could not forward the timezone it already resolves per request.

No timezone means no query at all, so existing callers keep the `Europe/Paris` default unchanged.

Out of scope, per the ticket: `Action.execute()` (still on the default), the list/count path (already BFF-controlled), and the two other `Collection.action()` consumers, `mcp-server` and `workflow-executor`, which have no request-resolved timezone to forward. No route contract moves: the action operations already declare the `X-Forest-Timezone` header and a `timezone` body field, and the timezone middleware runs before the action routes.

## How to test

```bash
yarn workspace @forestadmin/agent-client test
```

### General

### Security


packages/agent-client/src/action-fields/field-form-states.ts
```diff
@@ -30,6 +31,7 @@ export default class FieldFormStates {
     hooks?: ForestSchemaAction['hooks'],
     fallbackFields?: ForestSchemaAction['fields'],
     fallbackLayout?: ForestSchemaAction['layout'],
+    timezone?: string,
```
nbouliol: Internal class, so no consumer breaks — what about an options object here? The new tests already pay for it with `undefined, undefined, undefined, timezone`.

packages/agent-bff/src/action/agent-action-client.ts
```diff
@@ -29,8 +29,15 @@ export interface Action extends ActionForm {
   execute(): Promise<unknown>;
 }
 
+export interface LoadActionParams {
+  collection: string;
+  action: string;
```
nbouliol: `action` holds the action name while `Action` in this module is the loaded object — `actionName` would read clearer.

packages/agent-bff/src/action/agent-action-client.ts
```diff
@@ -29,8 +29,15 @@ export interface Action extends ActionForm {
...
+  recordIds: string[];
+  timezone: string;
```
nbouliol, line 36: `timezone` is required here, but the only caller feeds it `ctx.state.timezone as string` — optional would make the type honest.

packages/agent-bff/src/action/agent-action-client.ts
```diff
@@ -42,8 +49,8 @@ export interface AgentActionClientOptions {
 
 // The raw layout must be read AFTER tryToSetFields: a change hook rebuilds fields+layout in place.
 // agent-client's `Action.getLayout()` only returns an `ActionLayoutRoot` wrapper whose element array
-// lives in a protected field. The rollback contract forbids agent-client changes, so we read it
-// through a cast rather than adding a public accessor. `extract-raw-layout.test.ts` builds a real
+// lives in a protected field. PRD-673's rollback contract forbade agent-client changes, so we read
```
nbouliol: We usually keep ticket ids out of code comments — "the PRD-673 rollback contract" could just be "the rollback contract".

packages/agent-bff/src/action/agent-action-client.ts
```diff
@@ -70,7 +77,7 @@ export default function createAgentActionClient({
   });
 
   return {
-    loadAction: (collection, action, recordIds) =>
-      client.collection(collection).action(action, { recordIds }),
+    loadAction: ({ collection, action, recordIds, timezone }) =>
+      client.collection(collection).action(action, { recordIds, timezone }),
```
nbouliol: The execute POST itself still ships the `Europe/Paris` default, so a `now` default rendered in the form and the value the agent sees at execute time can disagree within one request. Follow-up ticket?

packages/agent-client/test/action-fields/field-form-states.test.ts
```diff
@@ -275,6 +275,91 @@ describe('FieldFormStates', () => {
...
+        }),
+      );
+    });
+
+    it('should send no query in the load hook when no timezone is provided', async () => {
+      httpRequester.query.mockResolvedValue({ fields: [], layout: [] });
+
+      await withTimezone().loadInitialState();
+
+      expect(httpRequester.query.mock.calls[0][0].query).toBeUndefined();
```
nbouliol: `calls[0][0]` relies on call ordering rather than the path — `toHaveBeenCalledWith(expect.objectContaining({ path: '.../hooks/load', query: undefined }))` would name what it checks. Same for `calls[1][0]` below.

nbouliol (commented):
Threading looks right: `HttpRequester` merges `{ timezone: 'Europe/Paris', ...query }` so a caller-supplied timezone wins, and `buildTimezoneQuery` returns `undefined` rather than `{ timezone: undefined }`, so absent-timezone callers keep the default. Both hook paths and both BFF verbs are covered by tests. A few suggestions inline.

packages/agent-client/src/action-fields/field-form-states.ts
Tonours: Done in 84b5541: `FieldFormStatesOptions`, 17 call sites converted, and the `undefined, undefined, undefined, timezone` chain is gone. It also killed the qlty `count = 9` nit the previous push introduced.

packages/agent-bff/src/action/agent-action-client.ts
Tonours: Renamed in f6e104f.

packages/agent-bff/src/action/agent-action-client.ts
Tonours: Dropped in 11c2a83.

packages/agent-client/test/action-fields/field-form-states.test.ts
Tonours: Done in 768de7f: both now assert `toHaveBeenCalledWith(expect.objectContaining({ path, query: undefined }))`.

packages/agent-bff/src/action/agent-action-client.ts
Tonours: Pushing back on this one. `resolveTimezone` returns `string` or throws (`timezone.ts:31-38`: `missingTimezone()` when no source resolves, `invalidTimezone()` on a bad IANA name), and `createTimezoneMiddleware` sits at `cli-core.ts:300`, ahead of the action routes in the same chain. So on this path `ctx.state.timezone` is always a validated string, and `timezone: string` is the honest type.

The cast is Koa: `ctx.state` is untyped, so every consumer asserts. Same line already exists on main at `data-routes-middleware.ts:344` and `agent-stub.ts:7`.

Making it optional would let a caller pass nothing and silently fall back to `Europe/Paris`, which is the bug this PR fixes. Happy to type `ctx.state` properly instead if you want the cast gone, but that is its own change across the three sites.

nbouliol (approved):
Timezone threading holds up: the query merge order lets a caller override the default, and `buildTimezoneQuery` returning `undefined` rather than `{ timezone: undefined }` is what keeps existing callers on `Europe/Paris`. Both hooks and both BFF verbs are covered. The options-object migration lands cleanly — all nine params still map 1:1.

PR #1855, opened 2026-08-28
fix(agent-bff): reject unknown keys on list and count bodies

Tonours:
## What

List and count bodies reject an undeclared top-level key with 400 `invalid_request`. Same for a stray key inside a `sort` clause, `page`, or a filter node.

| Body | Before | After |
| --- | --- | --- |
| `{"filters": {...}}` | 200, all rows | 400 `Unrecognized key: "filters"` |
| `{"projections": ["id"]}` | 200, all fields | 400 |
| `{"sort": [{"field": "a", "direciton": "desc"}]}` | 200, sorted ascending | 400 |
| `{"parentId": "7"}` on a plain list | 200, ignored | 400 |
| `{"projection": [...]}` on a **count** | 200, ignored | 400 |
| `{"filter": ..., "timezone": ..., "parentId": ...}` | 200 | 200, unchanged |

fixes PRD-1098

## Why

The bodies were parsed with plain zod objects, which strip unknown keys. `filters` instead of `filter` was dropped silently, so the query ran unfiltered and answered 200: a typo returned the wrong rows with a success status. The TS client protects nobody using curl, plain JS or an LLM-built body, and that is this BFF's audience.

## How

Every request object is `z.strictObject`. `filter` and `parentId` are declared but not re-validated: each already has its own validator, and a second rule would only disagree with the first at the margins.

`timezone` **is** typed, `z.string().regex(/\S/).optional()`. It cannot disagree with `resolveTimezone`, which is strictly stricter — it needs a non-empty valid IANA name and rejects the rest with its own error. What the type adds is a 400 for a non-string, and that case mattered: `bodyTimezone` returns `undefined` for anything that is not a string, so `timezone: null` used to be dropped and the request resolved from the header or the deployment default. A blank used to take the same silent path from the other side — `resolveTimezone` skips empty candidates rather than rejecting them — so the regex refuses one instead. A date filter (`Today`, `PreviousXDays`) then ran in the wrong timezone, in a 200, with no trace.

The published components are derived from those same schemas, with a `satisfies OverridesOf<...>` clause that refuses a decoration for a key the runtime does not accept. The document cannot advertise a shape the parser rejects. Body types are `z.infer` of the same schemas.

## The document closes too, structurally

Every request body in both document forms carries `additionalProperties: false`, including the per-collection `SortClause_${key}` and every leaf and branch of the filter tree — a generated client now refuses `direciton` or `valu` before the request leaves.

Getting there meant dropping the `allOf` composition on relation bodies: `additionalProperties: false` on one branch forbids the key the other adds, and `oas31` has no `unevaluatedProperties` to lift that. Relation bodies now spread the foreign collection's properties into one flat closed object with `parentId`, which is what `schemas.ts` already did with `RelationListFlatInputs.extend(...)`.

## Breaking

A stray top-level key now draws a 400 where it used to give a silently different result set. The full surface, on all four routes (list, count, relation list, relation count):

- any undeclared top-level key;
- an undeclared key inside a `sort` element or inside `page`;
- a non-string or blank `timezone` (`null` and `""` included), which used to be silently dropped or skipped;
- `parentId` on a plain list or count, where a parent id means nothing;
- `projection`, `sort` or `page` on a count or a relation count, where paging and projection mean nothing.

The last two are the likely ones in practice: a client reusing one body builder across route families used to have the extra keys stripped.

## Observability

Every body the parser rejects logs a `Warn` carrying the reason — the offending key name, never a submitted value, which a test asserts. That covers every throw site, not only the zod one: a non-object body, a non-object `filter`, a stray key anywhere inside the filter tree, and a bad `parentId`, which on relation routes is parsed before the schema. Without it this contract change would ship blind: `BffHttpError` is serialized and returned before `error-middleware` reaches its logger, and there is no access log, so nobody would see a client that has been sending `filters` for six months.

When several rules fail at once, the message names the unrecognized key rather than whichever issue zod happened to list first — `{"filters": {}, "page": {"limit": 0}}` used to answer `page.limit: Too small`, sending the caller to fix the wrong thing.

## Known limitations

One sibling of the same bug stays open, worth a ticket:

- `values` misspelled on an action body: the key is read with a cast, not parsed, so `valeus` executes the action with no submitted values at all.

PRD-1099 — a non-JSON content type makes `@koa/bodyparser` set `ctx.request.body = {}`, so a `fetch` POST with no `Content-Type` header loses the whole body and answers 200 unfiltered. That one is the same symptom as this ticket one layer up, and it is fixed in #1856.

Stale claim spotted, not rewritten here: the `ConditionTree` description says relation list and count forward the filter without checks, while both checks run.

## How to test

```
yarn workspace @forestadmin/agent-bff test && yarn workspace @forestadmin/agent-bff lint
```

By hand, POST a list body with `filters` instead of `filter`: 400 naming the key, where it used to return every row.

### General

### Security


packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -254,17 +257,21 @@ function fieldRefs(deps: Deps, plan: Pick<CollectionPlan, 'key' | 'collection'>)
   };
 }
 
+const CLOSED_BODY_PROSE_ONLY = `${CLOSED_BODY_NOTE} This schema does not say so structurally yet: treat it as closed.`;
```
nbouliol: `Page` comes out closed here via the now-strict `PageInput`, but `SortClause_${key}` is hand-built in `fieldRefs` without `additionalProperties: false` — so the unfolded doc still invites the `direciton` typo this PR turns into a 400. What about closing it there too?

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -349,7 +358,7 @@ function registerRelationRequests(
   const description =
     `Filter, sort and projection apply to ${quoted(foreign.collection.name)}, the foreign ` +
```
nbouliol: `search` now reaches relation bodies through the `allOf`, so this could read "Filter, sort, projection and search" like its twin in `schemas.ts`.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -349,7 +358,7 @@ function registerRelationRequests(
...
     `collection of ${quoted(plan.collection.name)}.${quoted(relation.name)}; the parent only ` +
-    'resolves which records are related.';
+    `resolves which records are related. ${CLOSED_BODY_PROSE_ONLY}`;
```
nbouliol: Here the relation body stays an `allOf` and can only be closed in prose, while `schemas.ts` solved it by flattening `RelationListFlatInputs.extend(...)` — could the same flattening apply here and drop the caveat?

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -85,48 +88,62 @@ export const SearchExtendedSchema = SearchExtendedInput.openapi('SearchExtended'
...
+  'of `filter` is an error rather than a silently unfiltered result. Inside the filter tree the ' +
+  'check stops there: an unknown key on a condition node still reaches the agent, so a leaf ' +
+  'carrying `valu` runs with no value rather than being rejected.';
+
+const countOverrides = {
+  filter: ConditionTreeSchema.optional(),
+  search: SearchSchema.optional(),
+  searchExtended: SearchExtendedSchema.optional(),
+  timezone: TimezoneSchema.optional(),
+} satisfies Partial<Record<keyof typeof CountFlatInputs.shape, z.ZodType>>;
```
nbouliol: This guards key names but not types — `timezone` is published as a string while the runtime tolerates `null`. Also repeated four times; a small `overrides<S>()` helper would carry it.

packages/agent-bff/test/data/agent-query.test.ts
```diff
@@ -395,6 +404,66 @@ describe('parseRelationCountRequest', () => {
   });
 });
 
+describe('unknown keys', () => {
```
nbouliol, line 466: Strictness also makes `parseCountRequest({ projection: [...] })` a 400 — a valid key on the wrong route, arguably the likeliest break after `parentId`. Worth a case here, and a line in the Breaking section?

nbouliol (commented):
No blocker: the closure holds end-to-end on list, count and both relation parsers, and `filter`/`timezone`/`parentId` still pass. A few remarks inline, mostly about the unfolded document staying more open than the generic one.

nbouliol (approved):
Closure verified end-to-end on list, count and both relation parsers, and the generic document is derived from the same schemas rather than restated. Remaining remarks are about the unfolded document staying more open than the runtime, none blocking.

packages/agent-bff/src/openapi/unfolded-paths.ts
Tonours: Closed. `SortClause_${key}` gets `additionalProperties: false` like the pooled `SortClause` does through the strict `PageInput`/`SortClauseInput`. Test: "should close a per-collection sort clause, so a misspelled direction is not silently kept".

packages/agent-bff/src/openapi/unfolded-paths.ts
Tonours: Done — "Filter, sort, projection and search apply to …", matching its twin.

packages/agent-bff/src/openapi/unfolded-paths.ts
Tonours: Yes, and it drops the caveat entirely. The relation body now spreads the foreign collection's properties instead of composing them with `allOf`, exactly like `RelationListFlatInputs.extend(...)`:

```ts
const body = (properties: BodyProperties): SchemaObject => ({
  type: 'object',
  description,
  properties: { ...properties, parentId },
  required: ['parentId'],
  additionalProperties: false,
});
```

`registerRequests` was split so the property maps are built once and carried on the plan, which is what lets the relation builder reach the foreign collection's properties rather than only its `$ref`.

Consequence: `ListRequest_${key}` and `CountRequest_${key}` are `additionalProperties: false` too — they were prose-only *because* the relation bodies composed them. `CLOSED_BODY_PROSE_ONLY` is gone, every unfolded body now says the same thing structurally and in prose. The "per-collection document stays structurally open" limitation comes off the PR description.

Tests: "should close a relation request structurally, not only in prose", "should close the per-collection list and count bodies structurally", and the relation assertions moved from `allOf[1].properties` to `properties`.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Helper added — `type OverridesOf<S extends { shape: object }> = Partial<Record<keyof S['shape'], z.ZodType>>`, used by all four. A type alias rather than a function: `satisfies` keeps the literal type that `.extend()` needs, a wrapper function would have to thread a second generic to preserve it.

On the type half: the guard cannot check types here, by construction. The keys it guards are the `validatedElsewhere` ones, declared `z.unknown().optional()` in `request-schemas.ts` precisely because `filter`, `timezone` and `parentId` have their own validators. Every override is assignable to `unknown`, so a type-level check passes whatever the override says. Tightening the runtime side to make the check bite is the thing the PR deliberately does not do — a second rule for `timezone` would disagree with `resolveTimezone` at the margins.

`timezone: null` specifically is not a case the document gets wrong. `bodyTimezone` (`timezone-middleware.ts:17`) returns `undefined` for anything that is not a string, so `null` is **ignored**, not accepted — the request then resolves from the header or the deployment default, or answers 400 `missing_timezone` if there is neither. Publishing `type: string` says "send a string or omit the key", which is the honest instruction; `string | null` would advertise `null` as meaningful. The parser test that pins the tolerance is "should leave a null timezone to the middleware".

packages/agent-bff/test/data/agent-query.test.ts
Tonours: Added, and widened past `projection` — `sort` and `page` break the same way:

```ts
it('should reject a list-only key on a count body, where paging and projection mean nothing', () => {
  expect(() => parseCountRequest({ projection: ['id'] })).toThrow(REJECTED);
  expect(() => parseCountRequest({ sort: [{ field: 'id' }] })).toThrow(REJECTED);
  expect(() => parseCountRequest({ page: { limit: 10, offset: 0 } })).toThrow(REJECTED);
});
```

It is the likelier break of the two, agreed: a client reusing one body builder across list and count sends all three, and before this PR they were stripped. The Breaking section names it now.

Tonours:
Second push, from the review passes.

**`timezone` is typed now.** The PR said re-typing it would turn a tolerated `null` into a 400, and that reasoning was wrong: `z.string()` cannot disagree with `resolveTimezone`, which is strictly stricter — it needs a non-empty valid IANA name and rejects everything else with its own error. All the type adds is a 400 for a non-string, and that case is not benign. `bodyTimezone` (`timezone-middleware.ts:18`) returns `undefined` for anything that is not a string, so `timezone: null` was dropped and the request resolved from the header or the deployment default. A `Today` or `PreviousXDays` filter then ran in the wrong timezone — rows around midnight in or out of the result — in a 200, with no trace. That is the same class of bug as `filters`, which is what this PR is about. The test that pinned the old behaviour was named "should leave a null timezone to the middleware"; the middleware never handled it, it discarded it.

**Rejections are logged.** `BffHttpError` is serialized and returned before `error-middleware` reaches its logger, and there is no access log — so this contract change would have shipped with no way to see it working. Every rejected body now logs a `Warn` with the reason. The reason is a key name, never a submitted value, and a test asserts that.

**The message names the right key.** `assertFlatInputs` took `issues[0]`, and zod does not order issues by relevance: `{"filters": {}, "page": {"limit": 0}}` answered `page.limit: Too small`, so the caller fixed their page, replayed, and got a second 400 for the typo that actually changed their results. It now prefers the `unrecognized_keys` issue.

**One published claim corrected.** `CLOSED_BODY_NOTE` said a leaf carrying `valu` "runs with no value". Optimistic — the leaf is built with `value: undefined`, `TypeGetter` reads it as `null`, `FieldValidator` accepts it, and `query-converter` emits `{ [Op.eq]: undefined }`. What comes back is datasource-dependent. The note now says the contract does not define it.

**Not doing here, and why:**

- **Non-JSON content type.** `@koa/bodyparser` sets `ctx.request.body = {}` when the content type is neither JSON nor urlencoded, so a `fetch` POST with no `Content-Type` header loses filter, projection, sort, page and timezone and answers 200 on the whole collection — this ticket's symptom, one layer up, and it bypasses the strict parse entirely. It is PRD-1099 and it is fixed in #1856. Worth knowing the two interact: `curl -d` sends urlencoded, so the body becomes one long unknown key and *this* PR catches it; only the `fetch` path needs #1856.
- **Strict filter leaves** and **`values` on an action body** — named in Known limitations, each worth its own ticket. The action one is partly closed by #1860, which rejects a missing required field at execute.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -85,48 +88,65 @@ export const SearchExtendedSchema = SearchExtendedInput.openapi('SearchExtended'
...
 });
 
-export const RelationListRequestSchema = ListRequestSchema.extend({
-  parentId: ParentIdSchema,
-}).openapi('RelationListRequest', {
+export const CLOSED_BODY_NOTE =
+  'The runtime rejects an undeclared TOP-LEVEL key with 400 invalid_request, so `filters` instead ' +
+  'of `filter` is an error rather than a silently unfiltered result. Inside the filter tree the ' +
+  'check stops there: an unknown key on a condition node still reaches the agent, so a leaf ' +
+  'carrying `valu` instead of `value` is not rejected here and what the agent then does with it ' +
```
nbouliol: `assertNoStrayKey` now rejects `valu` on a leaf with a 400, so this note says the opposite of what the runtime does — and it ships on all four bodies in both document forms. Worth rewriting alongside the Known limitations bullet that still lists the leaf typo as open?

packages/agent-bff/test/openapi/openapi-unfolded.test.ts
```diff
@@ -217,57 +230,109 @@ describe('the unfolded document', () => {
     expect(branch.not.properties.field.type).toBe('string');
   });
 
-  it('should not forbid an unknown extra key on a filter node, which the runtime strips', () => {
+  it('should not forbid an unknown extra key on a filter node, which the runtime forwards', () => {
```
nbouliol: This locks in the pre-`259dc95` contract: the runtime no longer forwards the extra key, it 400s. `FilterLeaf_*` and `Filter_*` still publish no `additionalProperties: false`, so a generated client sends `valu` and gets rejected.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -341,25 +370,24 @@ function registerRelationRequests(
...
-    properties: { parentId },
-    required: ['parentId'],
-  };
+  const { degraded } = foreign.collection.fields;
+  const foreignNote = degraded ? ` ${DEGRADED_NOTE[degraded]}` : '';
   const description =
-    `Filter, sort and projection apply to ${quoted(foreign.collection.name)}, the foreign ` +
-    `collection of ${quoted(plan.collection.name)}.${quoted(relation.name)}; the parent only ` +
-    'resolves which records are related.';
+    `Filter, sort, projection and search apply to ${quoted(foreign.collection.name)}, the ` +
```
nbouliol: One `description` serves both bodies, so `RelationCountRequest_*` advertises sort and projection while `body(foreign.properties.count)` now closes without them — a client following the prose gets a 400. `schemas.ts` keeps the bare note on the count for that reason.

packages/agent-bff/src/data/request-schemas.ts
```diff
@@ -29,17 +20,26 @@ export const TimezoneInput = z.string();
...
+const validatedElsewhere = z.unknown().optional();
+
+export const ListFlatInputs = z.strictObject({
+  filter: validatedElsewhere,
   projection: ProjectionInput.optional(),
   sort: z.array(SortClauseInput).optional(),
   page: PageInput.optional(),
   search: SearchInput.optional(),
   searchExtended: SearchExtendedInput.optional(),
+  timezone: TimezoneInput.optional(),
```
nbouliol, line 44: `z.string()` still admits `""`, and `resolveTimezone` skips empty candidates rather than rejecting them, so `{"timezone": ""}` falls back to the header or the default — the same silent wrong-timezone path the How section says this closes. `.regex(/\S/)` like `ParentIdInput`?

packages/agent-bff/src/data/agent-query.ts
```diff
@@ -53,53 +59,59 @@ function assertNoNodeReadableAsBothLeafAndBranch(node: unknown, depth = 0): void
...
   if (result.success) return;
 
-  const [issue] = result.error.issues;
+  const { issues } = result.error;
+  const issue = issues.find(candidate => candidate.code === 'unrecognized_keys') ?? issues[0];
   const path = issue.path.join('.');
+  const reason = path ? `${path}: ${issue.message}` : issue.message;
 
-  throw invalidRequest(path ? `${path}: ${issue.message}` : issue.message);
+  logger('Warn', 'Request body rejected', { reason });
```
nbouliol: Only the zod path logs, so the filter-tree strays, a non-object body, `filter must be an object` and every `parseParentId` failure still reject silently — and on relation routes `parseParentId` runs before the schema. "Every rejected body logs a Warn" needs either the other throw sites or a narrower claim.

nbouliol (commented):
The five points from the previous round are fixed and verified — `SortClause_${key}` closed including the degraded branch, relation bodies flattened out of the `allOf` while keeping the foreign refs and `required: ['parentId']`, `OverridesOf`, the count break tested and documented, and the relation prose mentioning search.

The filter-tree strictness in `259dc95` is the one that landed without the document following it: the published note, the `FilterLeaf_*` schema, one test's premise and the Known limitations bullet all still describe the old behaviour. Details inline.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -59,8 +62,9 @@ export const PageSchema = PageInput.openapi('Page', {
 export const TimezoneSchema = TimezoneInput.openapi('Timezone', {
   description:
     'Used when the X-Forest-Timezone header is absent. The header wins when both are sent. A ' +
-    'deployment with no configured default rejects a request carrying neither with 400 ' +
-    'missing_timezone.',
+    'blank or whitespace-only string is rejected with 400 rather than silently resolving from ' +
```
nbouliol: These two sentences disagree: the timezone middleware runs before the data routes, so with `X-Forest-Timezone: Europe/Paris` and `{"timezone": ""}` the header does resolve and the body still 400s. Worth saying the header wins only when the body key is absent?

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -85,48 +89,64 @@ export const SearchExtendedSchema = SearchExtendedInput.openapi('SearchExtended'
...
+  description:
+    'Filter, sort, projection and search apply to the FOREIGN collection; the parent only ' +
+    `resolves which records are related. ${CLOSED_BODY_NOTE}`,
+});
+
+export const RelationCountRequestSchema = RelationCountFlatInputs.extend(
+  relationCountOverrides,
+).openapi('RelationCountRequest', { description: CLOSED_BODY_NOTE });
 
 export const ActionRequestSchema = z
```
nbouliol, line 160: The `Timezone` component is reused here, but action bodies are never zod-parsed — `action-routes-middleware.ts:191` reads them by cast — so `{"timezone": ""}` still resolves from the header or the default on `/actions/*`, which is what its description now says cannot happen.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -31,7 +34,7 @@ const ConditionTreeSchema: z.ZodType = z
   .lazy(() =>
     z.union([
```
nbouliol, line 35: `agent-query.ts:82` keeps `filter: {}` allowed on purpose, but it matches neither union member, so a generated client refuses a body the BFF accepts. An empty-object member, or dropping the runtime allowance?

packages/agent-bff/src/data/agent-query.ts
```diff
@@ -1,105 +1,124 @@
...
   return typeof value === 'object' && value !== null && !Array.isArray(value);
 }
 
-function assertNoNodeReadableAsBothLeafAndBranch(node: unknown, depth = 0): void {
-  if (depth > MAX_FILTER_DEPTH) throw filterTooDeep(MAX_FILTER_DEPTH);
-  if (typeof node !== 'object' || node === null) return;
+const LEAF_KEYS = ['field', 'operator', 'value'];
+const BRANCH_KEYS = ['aggregator', 'conditions'];
+
+function rejectBody(logger: Logger, error: BffHttpError): never {
```
nbouliol: `serializePage` at `:143` is the one body rejection still throwing without this, so Observability's "covers every throw site" is one short. It is also the only message carrying submitted values, so routing it through here needs the message reworded first.

packages/agent-bff/src/data/agent-query.ts
```diff
@@ -1,105 +1,124 @@
...
-  if (!isPlainObject(filter)) throw invalidRequest('filter must be an object');
+  if (!isPlainObject(filter)) rejectBody(logger, invalidRequest('filter must be an object'));
 
-  assertNoNodeReadableAsBothLeafAndBranch(filter);
+  assertFilterNode(filter, logger);
 }
 
-export function parseListRequest(body: unknown): ListRequestBody {
-  if (!isPlainObject(body)) throw invalidRequest('Request body must be an object');
+function parseRequest<S extends ZodType>(schema: S, body: unknown, logger: Logger): z.output<S> {
```
nbouliol: `logger` is now threaded through six functions, including the exported `parseParentId`, only to emit one fixed message. A try/catch here and in the two relation parsers logging `error.message` would remove all six parameters, keep the exported signature, and would have caught the `serializePage` site.

packages/agent-bff/src/data/request-schemas.ts
```diff
@@ -25,21 +16,30 @@ export const SearchInput = z.string();
 
 export const SearchExtendedInput = z.boolean();
 
-export const TimezoneInput = z.string();
+export const TimezoneInput = z.string().regex(/\S/);
```
nbouliol: `/\S/` is now the second inline non-blank rule in this file, and it is hard-coded twice more as `'\\S'` in `unfolded-paths.ts`. One named `NonBlankString` would carry the rule and one message.

packages/agent-bff/test/data/agent-query.test.ts
```diff
@@ -384,17 +449,164 @@ describe('parseRelationListRequest', () => {
...
+  it.each(FLAT_PARSERS)(
+    'should reject a non-string timezone on %s rather than resolve one the caller did not send',
+    (_label, parse) => {
+      expect(() => parse({ timezone: null })).toThrow(REJECTED);
+      expect(() => parse({ timezone: 42 })).toThrow(REJECTED);
+    },
+  );
+
+  it.each(FLAT_PARSERS)(
+    'should reject a blank timezone on %s rather than silently resolve another one',
```
nbouliol, line 524: This asserts the same `{type, status}` as the non-string test above, so it cannot tell a blank-timezone rejection from any other 400 — worth asserting the reason names `timezone`.

nbouliol (commented):
Eight of the ten points from the two earlier rounds are fully closed. Worth calling out that the filter-tree closure avoids the JSON Schema trap the `allOf` composition hit: `additionalProperties: false` sits inside each `anyOf` member rather than beside it, so it actually bites, and the published key sets match `LEAF_KEYS`/`BRANCH_KEYS` exactly.

Two are partial: the blank-timezone rejection holds on the four data routes but not on the action ones, and `serializePage` is the one body rejection still throwing without a log. Nothing blocking.

One line in the description to narrow: "Every request body in both document forms carries `additionalProperties: false`" — action bodies deliberately do not (`schemas.ts:152` is a plain `z.object`, per the hook rationale at `unfolded-paths.ts:404`).

packages/agent-bff/src/openapi/schemas.ts
Tonours: Fixed in an earlier round: `CLOSED_BODY_NOTE` now says the leaf typo is rejected, and the Known limitations bullet no longer lists it.

packages/agent-bff/test/openapi/openapi-unfolded.test.ts
Tonours: Fixed: `FilterLeaf_*` and the `Filter_*` branch both carry `additionalProperties: false` (unfolded-paths.ts:144,218).

packages/agent-bff/src/openapi/unfolded-paths.ts
Tonours: Fixed: the relation bodies take their own subject line — `Filter, sort, projection and search` on list, `Filter and search` on count (unfolded-paths.ts:385-387).

packages/agent-bff/src/data/request-schemas.ts
Tonours: Fixed: `TimezoneInput` is the shared `NonBlankString`, so a blank is a 400.

packages/agent-bff/src/data/agent-query.ts
Tonours: Fixed: every rejection now throws and `loggingRejections` logs it once at the entry point, so the filter tree, the non-object body, the non-object filter and `parseParentId` all log.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Reworded: the header wins when both are sent, but the key is still validated, so a blank body timezone 400s even when the header resolves.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Right. The Timezone description now says an action body is read without that validation and falls back to the header or the default, so the shared component no longer claims something `/actions/*` does not do.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Added an empty-object member to `ConditionTree` and to the unfolded `Filter_*`, with the description saying it is how an absent filter is spelled. The runtime allowance stays.

packages/agent-bff/src/data/agent-query.ts
Tonours: Fixed at the root: the offset/limit rule is a rule on the body, so it moved into `PageInput` as a refine. `serializePage` no longer throws, and the rejection flows through the logged parse path like the others.

packages/agent-bff/src/data/agent-query.ts
Tonours: Done: one `loggingRejections(logger, fn)` wrapper at the four parsers and `parseParentId`; every inner helper lost its `logger` parameter and plain `throw`s again.

packages/agent-bff/src/data/request-schemas.ts
Tonours: Extracted `NonBlankString` and `NON_BLANK_PATTERN` in request-schemas.ts; `TimezoneInput`, `ParentIdInput` and the two raw patterns in unfolded-paths.ts now share them.

packages/agent-bff/test/data/agent-query.test.ts
Tonours: The assertion now requires the message to name `timezone`, so it cannot pass on an unrelated 400.

packages/agent-bff/src/data/agent-query.ts
```diff
@@ -114,16 +139,9 @@ function serializeSort(sort: BffSortClause[]): string {
...
-// page-model rule. The agent paginates by page number/size, so an arbitrary offset that is not a
-// whole multiple of the limit cannot be expressed — reject it rather than return a shifted window.
 function serializePage(page: BffPage): Record<string, number> {
   const { limit, offset } = page;
 
-  if (offset % limit !== 0) {
-    throw invalidRequest(`page.offset (${offset}) must be a multiple of page.limit (${limit})`);
-  }
-
   return { 'page[size]': limit, 'page[number]': offset / limit + 1 };
```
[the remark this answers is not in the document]
Tonours: Restored. The rule now lives in both places on purpose: `PageInput` rejects it during parsing, so an HTTP request is refused with a logged 400 before reaching the builder, and this guard covers a direct `buildListAgentQuery` call, which the parser never sees. A test pins the direct-call path.

PR #1856, opened 2026-08-28
fix(agent-bff): reject non-json content types with 415

Tonours:
## What

Agent routes (POST/PUT/PATCH under `/agent`) answer `415 unsupported_media_type` when the body cannot be read as JSON:

- a non-JSON `Content-Type`, form-urlencoded included;
- **no** `Content-Type` at all on a request that carries a body.

`application/json` and `application/*+json` parse. A bodyless POST needs no `Content-Type`.

fixes PRD-1099

## Why

`@koa/bodyparser` skips a content type that matches no enabled type, without an error. The body became `{}`, so `/User/list` ran unfiltered and answered **200** — a wrong result set with a success status. `fetch(url, { method: 'POST', body: JSON.stringify(...) })` sends `text/plain;charset=UTF-8` when you forget the header, so this was one missing line away in any browser client.

The document declared 415 on 91 operations. It was unreachable.

## How

One guard, `agentScoped(createJsonOnlyGuard())`, mounted before the parser and gated on the agent chain being up — the same shape as the agent error middleware. It throws before any parsing, so the 415 fires before auth, next to the existing 413.

The guard asks the same question the parser asks — `ctx.is(JSON_BODY_TYPES)`, which is koa's wrapper over the `type-is` the parser itself calls — instead of matching the header by hand. Hand-matching left a hole: koa's `request.type` is a bare `header.split(';')[0]`, while the parser goes through `content-type.parse`, which **throws** on a parameter with no value. So `application/json; charset` and `application/json;;` looked like `application/json` to a hand-rolled guard, were skipped by the parser, and became `{}` — the original bug, on a header a client can produce by accident.

Both sides read the same constant:

```ts
const JSON_BODY_TYPES = ['application/json', 'application/*+json'];
```

That matters. `@koa/bodyparser` only knows six fixed JSON types, so a guard promising `application/*+json` would have let `application/ld+json` through to the silent drop this PR is closing. Passing the list explicitly is also deliberate: `extendTypes` is merged with lodash `merge`, which merges arrays **index by index**, so `{ json: ['application/*+json'] }` alone would overwrite `application/json` at index 0 — and `type-is` does not match a bare `application/json` against the wildcard. Every ordinary JSON body would have drawn a 415.

Going through `type-is` also handles the parameter forms for free: `application/json; charset=utf-8` and the legal-but-odd `application/json ; charset=utf-8` both parse, where a `split(';')[0]` comparison rejected the second on a trailing space.

## Breaking

- Form-urlencoded on agent routes used to parse as JSON with string values. It now gets 415. The document only ever declared `application/json`.
- A request carrying a body with no `Content-Type` used to have that body dropped. It now gets 415, which is what RFC 9110 allows a recipient to do with a payload it cannot identify.

A **bodyless** POST is never rejected, whatever it declares: with nothing to parse there is nothing to drop. The guard checks `Content-Length > 0` or a `Transfer-Encoding` header before it looks at the type.

`/oauth/token` keeps its form and JSON bodies — it is outside the agent scope.

## How to test

- POST `/agent/v1/<collection>/list` with `Content-Type: text/plain` and a filter body: 415. Same body as `application/json`: filtered rows.

The parse-side tests send an oversized body and expect **413** rather than 401: the size limit only fires once the parser has read the body, so it is the cheapest proof that the content type reached a parser instead of being dropped.

### General

### Security

### DIFF
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -80,6 +80,9 @@ function agentScoped(middleware: Middleware): Middleware {
 function createBodyParser(hasAiQueryRoute: boolean): Middleware {
   const extendTypes = { json: JSON_BODY_TYPES };
   const parseBody = bodyParser({ jsonLimit: BODY_LIMIT, extendTypes });
+
+  if (!hasAiQueryRoute) return parseBody;
+
   const parseAiBody = bodyParser({
     jsonLimit: AI_BODY_LIMIT,
     enableTypes: ['json'],
@@ -87,7 +90,7 @@ function createBodyParser(hasAiQueryRoute: boolean): Middleware {
   });
 
   return async function selectedBodyParser(ctx, next) {
-    if (hasAiQueryRoute && ctx.path === AI_QUERY_ROUTE) {
+    if (ctx.path === AI_QUERY_ROUTE) {
       await parseAiBody(ctx, next);
 
       return;
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -72,12 +72,12 @@ function agentScoped(middleware: Middleware): Middleware {
   };
 }
 
-function createBodyParser(hasAiQueryRoute: boolean): Middleware {
+function createBodyParser(hasAiQueryRoute: boolean, hasAgentEdge: boolean): Middleware {
   const parseBody = bodyParser({ jsonLimit: BODY_LIMIT });
   const parseAiBody = bodyParser({ jsonLimit: AI_BODY_LIMIT, enableTypes: ['json'] });
 
   return async function selectedBodyParser(ctx, next) {
-    if (isAgentPath(ctx.path)) rejectNonJsonBody(ctx);
+    if (hasAgentEdge && isAgentPath(ctx.path)) rejectNonJsonBody(ctx);
 
     if (hasAiQueryRoute && ctx.path === AI_QUERY_ROUTE) {
       await parseAiBody(ctx, next);
@@ -401,7 +401,7 @@ export default async function runCli(
   const middlewares = [
     createCorsMiddleware({ allowedOrigins: config.allowedOrigins }),
     ...agentErrorMiddleware,
-    createBodyParser(aiMiddlewares.length > 0),
+    createBodyParser(aiMiddlewares.length > 0, agentMiddlewares.length > 0),
     ...oauth.middlewares,
     // Outside the agent-scoped chain on purpose: the viewer is a public page, the document it fetches
     // is not. Gated on the edge being mounted too, like the error middleware above: with no agent
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -55,21 +55,11 @@ function hasBody(ctx: Parameters<Middleware>[0]): boolean {
   return (ctx.request.length ?? 0) > 0 || ctx.get('transfer-encoding') !== '';
 }
 
-function isJsonType(mime: string): boolean {
-  return mime === 'application/json' || (mime.startsWith('application/') && mime.endsWith('+json'));
-}
-
 function createJsonOnlyGuard(): Middleware {
   return async function jsonOnlyGuard(ctx, next) {
-    if (!BODY_METHODS.has(ctx.method)) {
-      await next();
-
-      return;
+    if (BODY_METHODS.has(ctx.method) && hasBody(ctx) && !ctx.is(JSON_BODY_TYPES)) {
+      throw unsupportedMediaType();
     }
-
-    const mime = ctx.request.type.trim().toLowerCase();
-
-    if (mime === '' ? hasBody(ctx) : !isJsonType(mime)) throw unsupportedMediaType();
 
     await next();
   };
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -49,15 +49,30 @@ function isAgentPath(path: string): boolean {
 
 const BODY_METHODS = new Set(['POST', 'PUT', 'PATCH']);
 
-function rejectNonJsonBody(ctx: Parameters<Middleware>[0]): void {
-  if (!BODY_METHODS.has(ctx.method)) return;
-  if (ctx.request.type === '') return;
+const JSON_BODY_TYPES = ['application/json', 'application/*+json'];
 
-  const mime = ctx.request.type.toLowerCase();
+function hasBody(ctx: Parameters<Middleware>[0]): boolean {
+  return (ctx.request.length ?? 0) > 0 || ctx.get('transfer-encoding') !== '';
+}
 
-  if (mime !== 'application/json' && !(mime.startsWith('application/') && mime.endsWith('+json'))) {
-    throw unsupportedMediaType();
-  }
+function isJsonType(mime: string): boolean {
+  return mime === 'application/json' || (mime.startsWith('application/') && mime.endsWith('+json'));
+}
+
+function createJsonOnlyGuard(): Middleware {
+  return async function jsonOnlyGuard(ctx, next) {
+    if (!BODY_METHODS.has(ctx.method)) {
+      await next();
+
+      return;
+    }
+
+    const mime = ctx.request.type.trim().toLowerCase();
+
+    if (mime === '' ? hasBody(ctx) : !isJsonType(mime)) throw unsupportedMediaType();
+
+    await next();
+  };
 }
 
 function agentScoped(middleware: Middleware): Middleware {
@@ -72,13 +87,16 @@ function agentScoped(middleware: Middleware): Middleware {
   };
 }
 
-function createBodyParser(hasAiQueryRoute: boolean, hasAgentEdge: boolean): Middleware {
-  const parseBody = bodyParser({ jsonLimit: BODY_LIMIT });
-  const parseAiBody = bodyParser({ jsonLimit: AI_BODY_LIMIT, enableTypes: ['json'] });
+function createBodyParser(hasAiQueryRoute: boolean): Middleware {
+  const extendTypes = { json: JSON_BODY_TYPES };
+  const parseBody = bodyParser({ jsonLimit: BODY_LIMIT, extendTypes });
+  const parseAiBody = bodyParser({
+    jsonLimit: AI_BODY_LIMIT,
+    enableTypes: ['json'],
+    extendTypes,
+  });
 
   return async function selectedBodyParser(ctx, next) {
-    if (hasAgentEdge && isAgentPath(ctx.path)) rejectNonJsonBody(ctx);
-
     if (hasAiQueryRoute && ctx.path === AI_QUERY_ROUTE) {
       await parseAiBody(ctx, next);
 
@@ -396,12 +414,14 @@ export default async function runCli(
   const oauth = await buildOAuthMiddlewares(config, logger);
   const aiMiddlewares = buildAiMiddlewares(config, oauth, logger);
   const agentMiddlewares = buildAgentMiddlewares(config, logger, oauth, aiMiddlewares);
-  const agentErrorMiddleware =
-    agentMiddlewares.length > 0 ? [agentScoped(createErrorMiddleware({ logger }))] : [];
+  const hasAgentEdge = agentMiddlewares.length > 0;
+  const agentErrorMiddleware = hasAgentEdge ? [agentScoped(createErrorMiddleware({ logger }))] : [];
+  const agentJsonOnlyGuard = hasAgentEdge ? [agentScoped(createJsonOnlyGuard())] : [];
   const middlewares = [
     createCorsMiddleware({ allowedOrigins: config.allowedOrigins }),
     ...agentErrorMiddleware,
-    createBodyParser(aiMiddlewares.length > 0, agentMiddlewares.length > 0),
+    ...agentJsonOnlyGuard,
+    createBodyParser(aiMiddlewares.length > 0),
     ...oauth.middlewares,
     // Outside the agent-scoped chain on purpose: the viewer is a public page, the document it fetches
     // is not. Gated on the edge being mounted too, like the error middleware above: with no agent
--- a/packages/agent-bff/src/cli-core.ts
+++ b/packages/agent-bff/src/cli-core.ts
@@ -25,7 +25,7 @@ import createPerKeyOriginMiddleware from './cors/per-key-origin';
 import createDataRoutesMiddleware from './data/data-routes-middleware';
 import createDocsRoutes from './docs/docs-routes';
 import { extractErrorMessage } from './errors';
-import { unauthorized } from './http/bff-http-error';
+import { unauthorized, unsupportedMediaType } from './http/bff-http-error';
 import BFFHttpServer from './http/bff-http-server';
 import BODY_LIMIT, { AI_BODY_LIMIT } from './http/body-limit';
 import createErrorMiddleware from './http/error-middleware';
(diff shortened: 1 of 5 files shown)


packages/agent-bff/src/cli-core.ts
```diff
@@ -47,6 +47,19 @@ function isAgentPath(path: string): boolean {
...
 
+const BODY_METHODS = new Set(['POST', 'PUT', 'PATCH']);
+
+function rejectNonJsonBody(ctx: Parameters<Middleware>[0]): void {
+  if (!BODY_METHODS.has(ctx.method)) return;
+  if (ctx.request.type === '') return;
+
+  const mime = ctx.request.type.toLowerCase();
+
+  if (mime !== 'application/json' && !(mime.startsWith('application/') && mime.endsWith('+json'))) {
```
nbouliol: `@koa/bodyparser` only parses a fixed json allowlist (`json`, `json-patch+json`, `vnd.api+json`, `csp-report`, `reports+json`, `scim+json`), so `application/ld+json` clears this guard and is then silently dropped — the original bug, on a type the document now promises works. What about `extendTypes: { json: ['application/*+json'] }` on the parser so both sides agree?

nbouliol (commented):
Reviewed against PRD-1099. The 415 path itself is solid and the `text/plain` regression is covered. One blocking gap: the guard accepts a wider `+json` set than the parser can actually parse, so the original silent drop survives on those types.

packages/agent-bff/src/cli-core.ts
```diff
@@ -47,6 +47,19 @@ function isAgentPath(path: string): boolean {
   return path === '/agent' || path.startsWith('/agent/');
 }
 
+const BODY_METHODS = new Set(['POST', 'PUT', 'PATCH']);
+
+function rejectNonJsonBody(ctx: Parameters<Middleware>[0]): void {
+  if (!BODY_METHODS.has(ctx.method)) return;
+  if (ctx.request.type === '') return;
+
+  const mime = ctx.request.type.toLowerCase();
```
nbouliol: Koa's `request.type` is `header.split(';')[0]` with no trim, so a legal `application/json ; charset=utf-8` arrives as `'application/json '` and now gets 415 — `.trim()` here?

packages/agent-bff/src/cli-core.ts
```diff
@@ -59,15 +72,14 @@ function agentScoped(middleware: Middleware): Middleware {
...
+function createBodyParser(hasAiQueryRoute: boolean, hasAgentEdge: boolean): Middleware {
   const parseBody = bodyParser({ jsonLimit: BODY_LIMIT });
-
-  if (!hasAiQueryRoute) return parseBody;
-
   const parseAiBody = bodyParser({ jsonLimit: AI_BODY_LIMIT, enableTypes: ['json'] });
 
   return async function selectedBodyParser(ctx, next) {
-    if (ctx.path === AI_QUERY_ROUTE) {
+    if (hasAgentEdge && isAgentPath(ctx.path)) rejectNonJsonBody(ctx);
```
nbouliol: A standalone `agentScoped(createJsonOnlyGuard())` inserted before the parser would drop the `hasAgentEdge` param, this re-check, and the two positional booleans at the call site — same shape as `agentErrorMiddleware` above.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -185,7 +185,112 @@ describe('runCli', () => {
...
+    it('should still parse an application/*+json body on a data route', async () => {
+      const server = await runCli(OAUTH_ENV, noopLogger);
+
+      try {
+        const response = await request(server.callback)
+          .post('/agent/v1/books/list')
+          .set('Content-Type', 'application/vnd.api+json')
+          .send(JSON.stringify({ page: { limit: 5 } }));
+
+        expect(response.status).toBe(401);
```
nbouliol: `toBe(401)` is also what a dropped body returns, so this doesn't prove the body was parsed — an oversized body expecting 413, like the test above, would. Same for the charset case below.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -165,7 +165,7 @@ const AI_DUAL_SHAPED_ERRORS: Record<string, string> = {
   401: `The BFF session is missing, invalid or expired, or the Forest server refused the access token this route forwards. ${AI_RELAYED_OR_ENVELOPE}`,
   403: `The request presented an API key instead of a session (type oauth_required), or the Forest server refused the query. ${AI_RELAYED_OR_ENVELOPE}`,
   413: `The request body exceeds the AI query limit of ${AI_BODY_LIMIT}, or the Forest AI proxy refused it as too large. ${AI_RELAYED_OR_ENVELOPE}`,
-  415: `The request declares a character set the BFF cannot decode. ${AI_RELAYED_OR_ENVELOPE}`,
+  415: `The request Content-Type is not application/json, or declares a character set the BFF cannot decode. ${AI_RELAYED_OR_ENVELOPE}`,
```
nbouliol: The guard lets `application/*+json` through on this route too, so "not application/json" over-rejects here — mirror the data-route wording?

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -316,10 +316,9 @@ const TIMEZONE_HEADER = z
 const SHARED_DESCRIPTION =
   'The timezone is resolved from the `X-Forest-Timezone` header first, then a `timezone` body ' +
   'field, then the BFF default when one is configured. A deployment without a default rejects a ' +
-  'request carrying neither with 400 missing_timezone, so send one of the two to be safe. Sending ' +
-  'a content type other than application/json is not an error: a form-urlencoded body is parsed ' +
-  'like JSON, while any other content type is read as absent, which silently drops any filter, ' +
-  'sort, or page.';
+  'request carrying neither with 400 missing_timezone, so send one of the two to be safe. Agent ' +
+  'routes read application/json (and application/*+json) bodies only: any other Content-Type, ' +
```
nbouliol, line 320: A POST with no `Content-Type` at all still reads as an absent body, so filters can still drop silently — worth one clause here, since this is now the only silent path left and the test forbids the word.

packages/agent-bff/src/cli-core.ts
Tonours: Right, and it was the real hole: the guard promised `application/*+json` while the parser knew six fixed types, so `application/ld+json` cleared the 415 and then hit the exact silent drop this PR exists to close.

`extendTypes` alone would have made it worse, though. `@koa/bodyparser` merges with lodash `merge`, which merges arrays **index by index**:

```
merge(defaults, { json: ['application/*+json'] }).json
  -> ["application/*+json", "application/json-patch+json", ...]
```

`application/json` is gone — index 0 was overwritten — and `type-is` does not match a bare `application/json` against `application/*+json` (the wildcard requires a suffix). That one-liner would have 415'd every ordinary JSON body. Measured, not reasoned about.

So it is passed explicitly, and the same constant feeds the guard:

```ts
const JSON_BODY_TYPES = ['application/json', 'application/*+json'];
```

Merged list: `application/json`, the wildcard, then the remaining defaults. Both parsers get it, so the AI route agrees too.

Test: the `it.each` block now sends an **oversized** body and expects 413 — the size limit only fires once the parser has read the body, so it proves the type reached a parser rather than being dropped. `application/ld+json` is one of the cases.

packages/agent-bff/src/cli-core.ts
Tonours: Fixed — `.trim()` before the comparison. `application/json ; charset=utf-8` is legal (RFC 9110 allows OWS before the parameter separator) and koa's `request.type` is a bare `split(';')[0]`, so it arrived as `'application/json '` and drew a 415 on a valid request.

Covered as its own case in the `it.each` block, which expects 413 rather than 401 so it proves the body was parsed.

packages/agent-bff/src/cli-core.ts
Tonours: Done. `agentScoped(createJsonOnlyGuard())` sits before the parser, gated on `hasAgentEdge` exactly like `agentErrorMiddleware`. `createBodyParser` is back to one parameter, the `isAgentPath` re-check inside it is gone, and the two positional booleans at the call site are down to one.

packages/agent-bff/test/cli-core.test.ts
Tonours: Correct, `toBe(401)` proved nothing — a dropped body reaches auth just the same. Both cases now send a 20kb body and expect **413**: the limit only fires once the parser has read the body, so it is the cheapest available proof that the content type reached a parser.

They are one `it.each` now, covering the four types that have to parse: `application/vnd.api+json` (in the parser's default list), `application/ld+json` (only the wildcard covers it — this is the case that failed before this push), `application/json; charset=utf-8`, and `application/json ; charset=utf-8` with the space.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Mirrored. The AI 415 now reads "neither application/json nor an application/*+json type, or the request carries a body with no Content-Type at all, or it declares a character set the BFF cannot decode", which is what the shared guard actually does — the AI route goes through the same `agentScoped(createJsonOnlyGuard())`.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Closed rather than documented — it was the last silent path, and leaving it as prose would have kept the ticket's own symptom alive on the one request shape nobody would think to test.

The guard now rejects a request that carries a body with no `Content-Type` at all:

```ts
if (mime === '' ? hasBody(ctx) : !isJsonType(mime)) throw unsupportedMediaType();
```

`hasBody` reads `Content-Length > 0` or a `Transfer-Encoding` header. A bodyless POST with no Content-Type keeps working — that is the case the existing test pins, and it stays green.

RFC 9110 says a sender SHOULD declare a Content-Type and lets a recipient reject what it cannot identify, so 415 is the right status rather than a guess at JSON.

Both descriptions say it now (`ERROR_STATUSES` and `SHARED_DESCRIPTION`), including the "a bodyless POST needs no Content-Type" half, and there is a test: "should reject a body sent with no Content-Type at all, the last silent-drop path".

Tonours:
Third push. The review pass found the guard still had a hole, and it was the ticket's own bug.

**The guard matched the header by hand; the parser does not.** koa's `request.type` is a bare `header.split(';')[0]`, while `@koa/bodyparser` goes through `type-is` → `content-type.parse`, which **throws** on a parameter with no value and is swallowed into a `false`. So:

```
Content-Type: application/json; charset      guard: "application/json" -> pass
                                             parser: content-type.parse throws -> {} 
```

Body dropped, `/list` runs unfiltered, **200**. Same for `application/json;;`. Measured against the installed packages, not reasoned about.

Fixed by asking the same question the parser asks: `ctx.is(JSON_BODY_TYPES)`, koa's wrapper over that same `type-is`. The two can no longer disagree in the dangerous direction. Three cases added to the tests.

**Bodyless POSTs are no longer rejected on their declared type.** The previous shape checked the type first, so `Content-Length: 0` with `text/plain` — what .NET's `StringContent("")` sends — went from a legitimate 200 to a 415. It was the only newly-415 request whose old behaviour was correct. The guard now checks that a body exists before it looks at the type: nothing to parse, nothing to drop, nothing to reject.

**One asymmetry left, in the safe direction.** The parser also accepts `application/csp-report`, `application/reports+json` and `application/scim+json` from its own defaults; the guard rejects them, since the document only ever promised `application/json` and `application/*+json`. Guard stricter than parser costs a clear 415; the reverse is what produced this whole ticket.

Also from the pass: the `it.each` rows for `vnd.api+json` and the two charset forms pass on revert — the parser already handled those. Only the `ld+json` row proves the `extendTypes` change. Left as regression guards rather than removed, since they pin the published contract, but worth knowing they are not what proves this push.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -165,7 +165,7 @@ const AI_DUAL_SHAPED_ERRORS: Record<string, string> = {
   401: `The BFF session is missing, invalid or expired, or the Forest server refused the access token this route forwards. ${AI_RELAYED_OR_ENVELOPE}`,
   403: `The request presented an API key instead of a session (type oauth_required), or the Forest server refused the query. ${AI_RELAYED_OR_ENVELOPE}`,
   413: `The request body exceeds the AI query limit of ${AI_BODY_LIMIT}, or the Forest AI proxy refused it as too large. ${AI_RELAYED_OR_ENVELOPE}`,
-  415: `The request declares a character set the BFF cannot decode. ${AI_RELAYED_OR_ENVELOPE}`,
+  415: `The request Content-Type is neither application/json nor an application/*+json type, or the request carries a body with no Content-Type at all, or it declares a character set the BFF cannot decode. ${AI_RELAYED_OR_ENVELOPE}`,
```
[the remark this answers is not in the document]
Tonours: Fixed — the contradiction was mine, introduced when the 415 wording changed and the path description did not follow.

`registerAiQueryPath` now reads "only application/json and application/*+json are read: any other content type, and a body sent with none at all, is rejected with 415 before the relay", which is what `agentScoped(createJsonOnlyGuard())` actually does — the AI route sits inside the same agent scope as the data routes.

Pinned by a test so the two cannot drift again: "should say a non-json body is rejected here, matching the 415 the guard raises", asserting the new clause is present and `relayed as {}` is gone.

packages/agent-bff/test/cli-core.test.ts
```diff
@@ -185,7 +185,159 @@ describe('runCli', () => {
...
+      }
+    });
+
+    it('should reject a body sent with no Content-Type at all, the last silent-drop path', async () => {
+      const server = await runCli(OAUTH_ENV, noopLogger);
+
+      try {
+        const response = await request(server.callback)
+          .post('/agent/v1/books/list')
+          .set('Content-Type', '')
```
nbouliol: `.send(string)` with an empty `Content-Type` makes superagent default to `application/x-www-form-urlencoded`, so this re-tests the form path above and leaves the no-Content-Type branch uncovered — `.unset('Content-Type')` after `.send()` would send it for real.

packages/agent-bff/src/cli-core.ts
```diff
@@ -47,6 +47,24 @@ function isAgentPath(path: string): boolean {
...
+
+const JSON_BODY_TYPES = ['application/json', 'application/*+json'];
+
+function hasBody(ctx: Parameters<Middleware>[0]): boolean {
+  return (ctx.request.length ?? 0) > 0 || ctx.get('transfer-encoding') !== '';
+}
+
+function createJsonOnlyGuard(): Middleware {
+  return async function jsonOnlyGuard(ctx, next) {
+    if (BODY_METHODS.has(ctx.method) && hasBody(ctx) && !ctx.is(JSON_BODY_TYPES)) {
```
nbouliol, line 60: `ctx.is()` already returns `null` when there is no body, so `ctx.is(JSON_BODY_TYPES) === false` would drop `hasBody` entirely — the only behaviour change is a zero-length body with a non-JSON type, which would then 415 too.

packages/agent-bff/src/cli-core.ts
```diff
@@ -60,14 +78,16 @@ function agentScoped(middleware: Middleware): Middleware {
 }
 
 function createBodyParser(hasAiQueryRoute: boolean): Middleware {
```
nbouliol, line 80: Now that `hasAgentEdge` is gone, `if (!hasAiQueryRoute) return parseBody;` fits again and skips building the second parser.

nbouliol (commented):
Re-reviewed after the rework. The six earlier points are all addressed: `ctx.is()` plus `extendTypes` make the guard a subset of what the parser actually parses, so the silent drop is genuinely closed, and proving the parse through the size limit is a much stronger fixture than the old 401. One test does not send what its name claims, which leaves the new no-Content-Type branch uncovered.

nbouliol (approved):
Three rounds in, the guard and the parser now agree on exactly one set, and the fixtures prove the parse instead of inferring it from a 401. The no-Content-Type body was closed rather than documented, which is the shape the ticket would have kept bleeding on. Remaining suggestion on hasBody is not blocking. The red ai-proxy check is a provider-side tool-call failure, unrelated to this diff.

PR #1857, opened 2026-08-28
fix(agent-bff): type the action form and execute 200 responses

Tonours:
## What

The action `/form` and `/execute` 200 responses were published with an empty schema, so a generated client got no type for either. Both the generic and the unfolded documents now reference shared components:

- `ActionFormResponse` — the loaded form, its fields, `canExecute` and the relayed `layout`.
- `ActionResult` — a union discriminated on `type` with the three bodies the BFF returns at 200: `success`, `webhook`, `redirect`.

`ErrorResponse` also gained a description: `details` is polymorphic, and until now nothing said what it carries per error `type`.

fixes PRD-1097

## Why

Actions are the product of an admin app. Making every consumer reverse-engineer the shapes by calling the routes is the most expensive typing gap in the document.

## Why three branches and not five

The ticket lists five result forms. Two of them are not 200 bodies:

- `error` — the agent answers 400 and the BFF re-throws it as `action_error` (`action-routes-middleware.ts`), so it is an `ErrorResponse`.
- `file` — a binary stream with no JSON marker, so it falls through to the 501 `unsupported_action_result` the execute response already documents.

Publishing them inside `ActionResult` would say they can arrive with a 200. They cannot.

## Requiredness follows the wire, not the interfaces

`value`, `headers` and `body` are optional because serialization drops an `undefined` key. `message` and `html` are required and nullable because the mapper always assigns them. `ActionExecuteWebhookBody.url` / `.method` and `ActionExecuteRedirectBody.path` were `unknown` on the TypeScript side while the published schema said `string` — the schema was right (`isWebhook` and the `redirectTo` guard prove it before the body is built), so the interfaces were tightened to match.

## The `details` catalogue

Enumerated per `type` in prose rather than typed: `error.type` is an open string (agent errors are relayed with their own payload), so there is no discriminator OpenAPI can key a `oneOf` on. The list is exhaustive — `invalidRequest` is the only other factory that accepts `details` and no caller passes it — and the description says so.

The `action_error` entry carries the untrusted-html warning, in the same words as the execute success `html`.

**Expected conflict with #1859 (PRD-1095):** that branch sanitizes this html at the BFF. This one is off `main`, so "untrusted, sanitize before rendering" is accurate here. Whichever lands second flips both descriptions to say it is sanitized server-side.

## How to test

`yarn workspace @forestadmin/agent-bff test` — the openapi suite asserts the emitted schemas and runs a live `@hey-api/openapi-ts` codegen round-trip against the real routes.

No runtime change.

### General

### Security

### DIFF
--- a/packages/agent-bff/src/openapi/schemas.ts
+++ b/packages/agent-bff/src/openapi/schemas.ts
@@ -145,6 +145,95 @@ export const ActionRequestSchema = z
       'targets no record. Every id is coerced to a string before reaching the agent.',
   });
 
+const ActionFormResponseFieldSchema = z
+  .object({
+    name: z.string(),
+    // Verbatim from the agent, so a list field carries `['String']` rather than `'StringList'`.
+    type: z.union([z.string(), z.array(z.string())]),
+    value: z.unknown().optional(),
+    isRequired: z.boolean(),
+    enumValues: z.union([z.array(z.string()), z.null()]).optional(),
+  })
+  .openapi('ActionFormResponseField', {
+    description:
+      'A form field with its current value. `value` is the resolved value at load time and is ' +
+      'absent when the field carries none; `enumValues` is present only on an Enum field, and ' +
+      'null when the agent declares no options.',
+  });
+
+export const ActionFormResponseSchema = z
+  .object({
+    fields: z.array(ActionFormResponseFieldSchema),
+    canExecute: z.boolean(),
+    requiredFields: z.array(z.string()),
+    skippedFields: z.array(z.string()),
+    layout: z.array(z.unknown()),
+  })
+  .openapi('ActionFormResponse', {
+    description:
+      'The loaded form. `canExecute` is true only when every required field already carries a ' +
+      'value: an explicit empty string or 0 counts as present, only a missing one does not. ' +
+      '`requiredFields` names the required fields still missing a value. `skippedFields` names ' +
+      'the submitted fields the static form does not carry; the same names are rejected with 400 ' +
+      'on execute. `layout` is the agent layout tree relayed verbatim (pages, rows, separators, ' +
+      'HTML blocks and field references, discriminated by `component`); it is the agent own ' +
+      'contract, so it is left untyped here.',
+  });
+
+export const ActionResultSuccessSchema = z
+  .object({
+    type: z.literal('success'),
+    message: z.union([z.string(), z.null()]),
+    invalidated: z.array(z.string()),
+    html: z
+      .union([z.string(), z.null()])
+      .describe(
+        'Untrusted HTML relayed verbatim from the agent result: sanitize it before rendering ' +
+          '(stored/reflected XSS risk). Null when the result carries none.',
+      ),
+  })
+  .openapi('ActionResultSuccess', {
+    description:
+      'The action ran. `invalidated` lists the collections whose records changed and should be ' +
+      're-fetched; `message` is the agent wording, null when the result carries none.',
+  });
+
+export const ActionResultWebhookSchema = z
+  .object({
+    type: z.literal('webhook'),
+    url: z.string(),
+    method: z.string(),
+    headers: z.unknown().optional(),
+    body: z.unknown().optional(),
+  })
+  .openapi('ActionResultWebhook', {
+    description:
+      'The action asks the caller to fire an HTTP request. `headers` and `body` are relayed ' +
+      'from the agent payload and are absent when it carried none.',
+  });
+
+export const ActionResultRedirectSchema = z
+  .object({
+    type: z.literal('redirect'),
+    path: z.string(),
+  })
+  .openapi('ActionResultRedirect', {
+    description: 'The action asks the caller to navigate to `path`, relayed verbatim.',
+  });
+
+export const ActionResultSchema = z
+  .discriminatedUnion('type', [
+    ActionResultSuccessSchema,
+    ActionResultWebhookSchema,
+    ActionResultRedirectSchema,
+  ])
+  .openapi('ActionResult', {
+    description:
+      'The normalized execute result, discriminated by `type`. These three are the only 200 ' +
+      'bodies: a result the BFF cannot normalize answers 501 instead (see the execute 501 ' +
+      'response), and a form the agent rejects answers 400 action_error.',
+  });
+
 const ForestRecordMetaSchema = z
   .object({
     collection: z.string(),
--- a/packages/agent-bff/src/openapi/schemas.ts
+++ b/packages/agent-bff/src/openapi/schemas.ts
@@ -148,9 +148,6 @@ export const ActionRequestSchema = z
 const ActionFormResponseFieldSchema = z
   .object({
     name: z.string(),
-    // Verbatim from the agent, so a list field carries exactly `['String']` rather than
-    // `'StringList'`: an array of exactly one string. Length is pinned (zod-to-openapi maps
-    // min/max but not `.length`) because `prefixItems` alone would still admit extra elements.
     type: z.union([z.string(), z.array(z.string()).min(1).max(1)]),
     value: z.unknown().optional(),
     isRequired: z.boolean(),
@@ -183,7 +180,7 @@ export const ActionFormResponseSchema = z
       'contract, so it is left untyped here.',
   });
 
-export const ActionResultSuccessSchema = z
+const ActionResultSuccessSchema = z
   .object({
     type: z.literal('success'),
     message: z.union([z.string(), z.null()]),
@@ -201,7 +198,7 @@ export const ActionResultSuccessSchema = z
       're-fetched; `message` is the agent wording, null when the result carries none.',
   });
 
-export const ActionResultWebhookSchema = z
+const ActionResultWebhookSchema = z
   .object({
     type: z.literal('webhook'),
     url: z.string(),
@@ -215,7 +212,7 @@ export const ActionResultWebhookSchema = z
       'from the agent payload: absent when it omitted them, an explicit null relayed as null.',
   });
 
-export const ActionResultRedirectSchema = z
+const ActionResultRedirectSchema = z
   .object({
     type: z.literal('redirect'),
     path: z.string(),
--- a/packages/agent-bff/src/openapi/schemas.ts
+++ b/packages/agent-bff/src/openapi/schemas.ts
@@ -149,8 +149,9 @@ const ActionFormResponseFieldSchema = z
   .object({
     name: z.string(),
     // Verbatim from the agent, so a list field carries exactly `['String']` rather than
-    // `'StringList'`: a one-element tuple, never a free-form array.
-    type: z.union([z.string(), z.tuple([z.string()])]),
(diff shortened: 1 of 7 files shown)


packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -140,6 +140,95 @@ export const ActionRequestSchema = z
...
+    html: z
+      .union([z.string(), z.null()])
+      .describe(
+        'Untrusted HTML relayed verbatim from the agent result: sanitize it before rendering ' +
+          '(stored/reflected XSS risk). Null when the result carries none.',
+      ),
+  })
+  .openapi('ActionResultSuccess', {
+    description:
+      'The action ran. `invalidated` lists the collections whose records changed and should be ' +
```
nbouliol: `invalidated` is relayed from `refresh.relationships`, which the agent fills from `ResultBuilder.success(msg, { invalidated })` — relation names rather than collections?

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -140,6 +140,95 @@ export const ActionRequestSchema = z
...
+    fields: z.array(ActionFormResponseFieldSchema),
+    canExecute: z.boolean(),
+    requiredFields: z.array(z.string()),
+    skippedFields: z.array(z.string()),
+    layout: z.array(z.unknown()),
+  })
+  .openapi('ActionFormResponse', {
+    description:
+      'The loaded form. `canExecute` is true only when every required field already carries a ' +
+      'value: an explicit empty string or 0 counts as present, only a missing one does not. ' +
```
nbouliol: The mapper also counts an explicit `null` as missing (`action-form-mapper.ts:34`), which reads against the `value` description above — what about "a null or absent value counts as missing, an empty string or 0 as present"?

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -140,6 +140,95 @@ export const ActionRequestSchema = z
...
+  .object({
+    type: z.literal('success'),
+    message: z.union([z.string(), z.null()]),
+    invalidated: z.array(z.string()),
+    html: z
+      .union([z.string(), z.null()])
+      .describe(
+        'Untrusted HTML relayed verbatim from the agent result: sanitize it before rendering ' +
+          '(stored/reflected XSS risk). Null when the result carries none.',
+      ),
```
nbouliol, line 196: The same untrusted html reaches the 400 `action_error` through `error.details` (`action-routes-middleware.ts:140`), which stays `z.unknown()` with no warning — worth typing that branch too, or calling it out as deliberately out of scope?

nbouliol (commented):
Three notes on the published descriptions, all on the prose rather than the shapes — the schemas themselves check out against the mappers.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Relations, and the description was wrong. Fixed.

The docs are explicit — `invalidated` refreshes the Related Data sections of the acted-on record's Summary View, and the example is `invalidated: ['emitted_transactions']`, a relation name on the collection the action runs on (https://docs.forest.app/product/process/actions/custom-actions/related-data-invalidation). The wire name says the same thing: the agent relays it as `refresh.relationships` (`packages/agent/src/routes/modification/action/action.ts:146`).

Worth noting the API reference table itself says "Array of collection names to refresh" (`/reference/agent-api/nodejs`), so the doc site contradicts its own guide — that is a docs bug, not something to mirror here.

New text: "`invalidated` names the relations of the acted-on collection whose Related Data should be re-fetched — the agent fills it from `resultBuilder.success(message, { invalidated })` and the BFF relays it from the agent `refresh.relationships`." Asserted in `openapi-document.test.ts`.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Fixed, on the `canExecute` sentence rather than on `value` — the two describe different things and only one was wrong.

`ActionFormResponseField.value` is about **serialization**: an unresolved value is absent from the JSON, an explicit null is serialized as null. That stays accurate.

`ActionFormResponse.canExecute` is about **the missing-value rule**, and it read "an explicit empty string or 0 counts as present, only a missing one does not" — which hid that `null` counts as missing too (`action-form-mapper.ts:34`, `getValue() === undefined || getValue() === null`). It now says "a null or absent value counts as missing, an explicit empty string or 0 as present", your wording. Asserted in `openapi-document.test.ts`.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Called out rather than typed, and the call-out now covers every branch instead of only that one.

`ErrorResponse` is shared by every error the BFF emits, and `details` is genuinely polymorphic — `{ field }`, `{ field, validOperators }`, `{ maxDepth }`, `{ fields }`, `{ roleIdsAllowedToApprove }`, `{ html }`. Typing it would mean either a `oneOf` keyed on a `type` string with a dozen values (which OpenAPI cannot discriminate on a free-form string), or a separate response component per route that carries a different `details` — a lot of surface for a field whose consumers branch on `error.type` anyway.

So the component description enumerates the shapes per `type`, and the `action_error` entry carries the same untrusted-html warning as the execute success `html`, in the same words. Asserted in `openapi-document.test.ts`.

One thing this collides with: PRD-1095 (#1859) sanitizes that html at the BFF, on both surfaces. This branch is off `main`, so "untrusted, sanitize before rendering" is accurate here — whichever of the two lands second has to flip both descriptions to say the html is sanitized server-side. Flagging it so it is an expected conflict rather than a surprise.

Tonours:
Second push, from the review passes.

**A real divergence, fixed.** `ActionExecuteWebhookBody.url` / `.method` and `ActionExecuteRedirectBody.path` were `unknown` on the TypeScript side while the published schema says `string`. The schema was right — `isWebhook` checks `typeof url === 'string' && typeof method === 'string'` and the redirect branch checks `typeof body.redirectTo === 'string'` before the body is built. Interfaces tightened to what the guards already prove.

**The `details` catalogue is now stated as exhaustive.** It was a list that read as authoritative without saying so. Checked: `invalidRequest` is the only other factory whose signature accepts `details`, and none of its 12 call sites passes it. The description says an unlisted type carries no `details`, and that an agent-forwarded error carries the agent's own payload.

**One test name fixed** — `should keep a missing value and a non-enum field honest on the wire` asserted two unrelated published facts and named neither, so it would have survived either assertion being inverted. Split in two.

**One thing I am not doing, and why.** Two reviewers pointed at the same structure: the zod components restate shapes that `action-form-mapper.ts` and `action-execute-mapper.ts` also declare as TypeScript interfaces, where the repo elsewhere declares a shape once (`data/request-schemas.ts`) and lets `openapi/schemas.ts` only name and describe it.

The precedent does not transfer. `request-schemas.ts` shapes are the *parsers* — the same object validates the incoming body and documents it, which is why one declaration matters there and why its own docstring says "the document and the runtime cannot drift apart". These are response schemas: nothing parses through them, they are declaration-only. Deriving the mapper types from them would bind the published document to an internal type no runtime checks, and it costs three concrete losses of precision — `layout` drops from `ForestServerActionFormLayoutElement[]` to `unknown[]`, `type` from `string | [string]` to `string | string[]` (`.min(1).max(1)` is the correct JSON Schema here, but it infers a plain array), and `value` becomes optional.

What the tests pin instead is the document against reality: the emitted `required` sets, the union's `oneOf` and discriminator mapping, and a live codegen round-trip through the real routes. That is the drift the ticket is about.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -299,7 +393,17 @@ export const ErrorResponseSchema = z
...
-  .openapi('ErrorResponse');
+  .openapi('ErrorResponse', {
+    description:
+      'The error envelope. `details` is left untyped because its shape depends on `type`: ' +
+      '`{ field }` on unknown_field and field_not_filterable, `{ field, validOperators }` on ' +
+      'invalid_filter_operator, `{ maxDepth }` on filter_too_deep, `{ fields }` on ' +
+      'relation_field_not_supported, `{ roleIdsAllowedToApprove }` on ' +
+      'action_requires_approval, and `{ html }` on action_error. An error forwarded from the ' +
+      'agent carries the agent own payload instead, and any other type carries no `details` at ' +
+      `all. The action_error html is untrusted agent output relayed verbatim: ${UNTRUSTED_HTML_NOTE} ` +
```
nbouliol: `UNTRUSTED_HTML_NOTE` carries its own period, so this renders as "...relayed verbatim: sanitize it before rendering (stored/reflected XSS risk). exactly like the execute success html." — dropping the period from the constant and adding it at each call site would fix this without breaking the `html` description above.

nbouliol (commented):
The three earlier notes are all addressed, thanks. Three more on the new descriptions.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -299,7 +393,17 @@ export const ErrorResponseSchema = z
...
     }),
   })
-  .openapi('ErrorResponse');
+  .openapi('ErrorResponse', {
+    description:
+      'The error envelope. `details` is left untyped because its shape depends on `type`: ' +
+      '`{ field }` on unknown_field and field_not_filterable, `{ field, validOperators }` on ' +
+      'invalid_filter_operator, `{ maxDepth }` on filter_too_deep, `{ fields }` on ' +
+      'relation_field_not_supported, `{ roleIdsAllowedToApprove }` on ' +
+      'action_requires_approval, and `{ html }` on action_error. An error forwarded from the ' +
```
nbouliol: These last two are conditional — `action-routes-middleware.ts:140` passes `undefined` when the agent omits `html`, and `actionRequiresApproval` likewise — so "and, when the agent supplies it, `{ html }` on action_error" would be truer?

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -140,6 +140,100 @@ export const ActionRequestSchema = z
...
+        `Untrusted HTML relayed verbatim from the agent result: ${UNTRUSTED_HTML_NOTE} Null ` +
+          'when the result carries none.',
+      ),
+  })
+  .openapi('ActionResultSuccess', {
+    description:
+      'The action ran. `invalidated` names the relations of the acted-on collection whose ' +
+      'Related Data should be re-fetched — the agent fills it from ' +
+      '`resultBuilder.success(message, { invalidated })` and the BFF relays it from the agent ' +
+      '`refresh.relationships`. `message` is the agent wording, null when the result carries none.',
```
nbouliol: With agent-nodejs a message-less success serializes `success: ''` (`ResultBuilder.success` does `message ?? ''`) and the mapper keeps the empty string, so `null` only happens when an agent omits `success` entirely — worth saying that explicitly?

nbouliol (approved):
Schemas check out against the mappers property by property, and every claim in the published prose now matches the real construction sites. The three description fixes each ship with a regression test. Only CI red is the unrelated ai-proxy live-API job.

PR #1858, opened 2026-08-28
fix(bff): cap page.limit and rate-limit agent requests

Tonours:
## What

Two guards on the agent BFF edge:

- **`page.limit` capped at 1000.** Above it: 400 `invalid_request` — rejected, not clamped, so a caller cannot believe it got what it asked for. The cap is published as `Page.properties.limit.maximum`.
- **Per-identity rate limiting** on every authenticated `/agent/*` request: 429 `too_many_requests` with an integer `Retry-After`. Defaults 300 requests / 60 s, tunable through `BFF_RATE_LIMIT_MAX_REQUESTS` and `BFF_RATE_LIMIT_WINDOW_MS`; an invalid value fails boot rather than falling back.

fixes PRD-1100

## Why

A valid key could pull an entire collection in one request: `limit: 1000000000` was forwarded verbatim as `page[size]`. And nothing throttled the cadence — 400 concurrent requests, zero 429.

## How

- Buckets key on the resolved identity, mode-prefixed (`api-key:` / `oauth:`). Rejected keys and anonymous requests die at 401 before the limiter and never allocate a bucket.
- The counter increments **before** the chain runs, so a concurrent burst cannot slip past the limit. Covered by a 10-parallel test: exactly `maxRequests` pass.
- At the 10 000 bucket ceiling a new identity fails closed rather than evicting a live window.
- A request the limiter cannot bucket — no identity on the context — is rejected with 401 rather than passed through. The limiter's contract is that every authenticated agent request is counted; letting an unidentifiable one through would make any future route with a looser auth step silently unlimited.

## Two different 429s

The same code emits 429 for two reasons, and the document now says so instead of blaming the caller for both:

- **Quota** — this identity exceeded its window. `Retry-After` is when *its* window resets.
- **Saturation** — the bucket table is full and no window can be opened. The caller exceeded nothing. `Retry-After` is a lower bound: the earliest reset among the identities currently holding a window, after which a slot *may* free up — the freed slot goes to whoever asks first.

A client that backs off on the wrong signal is the reason this distinction is published rather than left in the message alone.

## How to test

```bash
cd packages/agent-bff
yarn jest test/data test/rate-limit test/http test/config test/openapi test/cli-core.test.ts
yarn lint && yarn build
```

- `POST /agent/v1/users/list` with `{"page":{"limit":1000000000,"offset":0}}` returns 400 `invalid_request`.
- With `BFF_RATE_LIMIT_MAX_REQUESTS=2`, the third authenticated request in the window returns 429 with `Retry-After`.

## Known limitation

Requests rejected by the body parser (400/413) do not count toward the quota. They are bounded at 16 KB and never reach the agent.

### General

### Security


packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -37,7 +37,7 @@ const ERROR_STATUSES: Record<string, string> = {
   413: `The request body exceeds the BFF limit of ${BODY_LIMIT}`,
   415: 'The request declares a character set the server cannot decode. Other content types are NOT rejected: a form-urlencoded body is parsed and validated like JSON (its values arrive as strings, so typed fields such as page.limit fail with 400), while any other non-JSON content type is read as an absent body, silently dropping filters and pagination',
   422: 'A field is unknown, not filterable, or is a nested relation path',
-  429: 'The agent rate-limited the request',
+  429: 'The BFF rate-limited the request (the caller identity exceeded its per-window budget, and Retry-After carries the seconds to wait); on data and action routes the agent may also rate-limit the request itself',
```
nbouliol: This says the caller exceeded its budget, but the saturation 429 at `rate-limit-middleware.ts:80` fires when the bucket table is full — that caller exceeded nothing. What about splitting the two cases, so a client doesn't back off on the wrong signal?

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -53,13 +53,22 @@ const MESSAGELESS_ERROR_RESPONSE_REF = '#/components/schemas/MessagelessErrorRes
...
   },
-};
+});
+
+const RETRY_AFTER_HEADER = retryAfterHeader(
+  'Seconds to wait before retrying. Set when the API key could not be resolved.',
+);
+
+const RETRY_AFTER_RATE_LIMIT_HEADER = retryAfterHeader(
+  'Seconds until the BFF rate-limit window of the caller resets. Present when the BFF ' +
```
nbouliol: At saturation this is the earliest reset across *other* identities' buckets — the caller has no window yet. Could we word it as a lower bound on when a retry might be admitted?

packages/agent-bff/src/rate-limit/rate-limit-middleware.ts
```diff
@@ -0,0 +1,102 @@
...
+  return async function rateLimitMiddleware(ctx, next) {
+    if (!isAgentPath(ctx.path)) {
+      await next();
+
+      return;
+    }
+
+    const key = bucketKeyOf(ctx.state as AuthEdgeState);
+
+    if (key === undefined) {
```
nbouliol: This fails open, and its safety rests entirely on `resolveAuthMode` throwing upstream — any future route with a looser auth step becomes silently unlimited. Throw here instead, or at least record why the bypass is safe?

packages/agent-bff/src/config/env-config.ts
```diff
@@ -50,19 +56,31 @@ function normalize(value: string | undefined): string | undefined {
   return value === undefined || value.trim() === '' ? undefined : value;
 }
 
-function parsePort(raw?: string): number {
-  const trimmed = raw?.trim();
-  if (trimmed === undefined || trimmed === '') return DEFAULT_BFF_PORT;
+function parseBoundedInteger(
```
nbouliol, line 60: `parseTimeoutMs` (106) is now a near-verbatim copy of this — exactly `parseBoundedInteger(raw, envName, { min: 1, max: MAX_TIMEOUT_MS, fallback: defaultMs })`, and no test pins its message wording.

packages/agent-bff/src/rate-limit/rate-limit-middleware.ts
```diff
@@ -0,0 +1,102 @@
...
+      }
+
+      entry = { count: 0, resetAt: current + windowMs };
+      buckets.set(key, entry);
+    }
+
+    entry.count += 1;
+
+    if (entry.count > maxRequests) {
+      const retryAfter = Math.max(1, Math.ceil((entry.resetAt - current) / 1000));
```
nbouliol: Same formula as line 78, with `1000` magic in both — a `retryAfterSeconds(resetAt, current)` helper and an `MS_PER_SECOND` constant would cover both sites.

nbouliol (commented):
Reviewed the two guards. The cap side is clean: `MAX_PAGE_LIMIT` is defined once and reused by the published prose, so the document and the runtime can't drift, and I confirmed the limiter really does sit after `createAuthModeMiddleware`/`apiKeyStep`, so anonymous and rejected-key traffic 401s without ever allocating a bucket. Test coverage on the limiter is well above the usual bar.

The main thread below is the published contract: the 429 and `Retry-After` descriptions hold for the quota path but not for the saturation path the same code can produce.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Split. The 429 description now names both causes and says the message distinguishes them:

> The BFF rate-limited the request, for one of two reasons: the caller identity exceeded its per-window budget, or the limiter is saturated and cannot open a window for a new identity.

Test: "should name both 429 causes, since a saturation 429 blames a caller that exceeded nothing".

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Reworded as a lower bound:

> Seconds until the BFF rate-limit window of the caller resets. When the limiter is saturated the caller has no window yet, so the value is a lower bound instead: the earliest reset among the identities currently holding one, after which a slot may free up.

"May" is doing real work there — the freed slot goes to whoever asks first, so the retry is admitted, not guaranteed. Test: "should call the saturation Retry-After a lower bound, not the caller own window".

packages/agent-bff/src/rate-limit/rate-limit-middleware.ts
Tonours: Fails closed now — `throw unauthorized()` instead of `await next()`.

Recording why in a comment was not an option (the repo bans them), and a comment would not have stopped the future route you describe anyway. 401 is the honest status: from the limiter's seat the request carries no identity, which is exactly what the auth chain would have answered had it run. If a route ever does need to sit on an agent path without an identity, it now fails loudly at boot-time traffic rather than becoming silently unlimited.

The test that pinned the old behaviour is inverted rather than deleted: "rejects an agent request the limiter cannot bucket, rather than letting it through unlimited".

packages/agent-bff/src/config/env-config.ts
Tonours: Collapsed — `parseTimeoutMs` is now one line:

```ts
return parseBoundedInteger(raw, envName, { min: 1, max: MAX_TIMEOUT_MS, fallback: defaultMs });
```

Equivalent on every input: `DECIMAL_INTEGER` already rejected negatives, so `timeout === 0 || timeout > MAX_TIMEOUT_MS` was `min: 1, max: MAX_TIMEOUT_MS`. The message changes from "must be a positive integer of at most 120000 milliseconds" to "must be an integer between 1 and 120000" — checked, nothing asserts either wording, and both env vars end in `_MS` so the unit is not lost.

packages/agent-bff/src/rate-limit/rate-limit-middleware.ts
Tonours: Both extracted:

```ts
const MS_PER_SECOND = 1000;

function retryAfterSeconds(resetAt: number, current: number): number {
  return Math.max(1, Math.ceil((resetAt - current) / MS_PER_SECOND));
}
```

Used by the saturation and the quota branch, and `MS_PER_SECOND` also replaces the third `1000` in the quota message.

Tonours:
Second push, from the review pass.

**The two new env vars were undocumented.** `packages/agent-bff/README.md` carries the exhaustive env table and neither `BFF_RATE_LIMIT_MAX_REQUESTS` nor `BFF_RATE_LIMIT_WINDOW_MS` was in it — so an operator hitting 429 at the default 300/60s had no documented knob, on a feature that changes production behaviour on day one. Both rows added, along with the saturation case and the `Retry-After` semantics, and the config-validation bullet now names `BFF_RATE_LIMIT_*` among the values that fail boot.

**Two notes I am leaving as they are, with the reasoning:**

The `throw unauthorized()` added last push is unreachable in the mounted chain — `auth-mode.ts` already 401s when neither a bearer nor a key is present, the OAuth step sets `principal`, the api-key step sets `apiKeyIdentity` or rethrows, and the docs viewer is mounted outside the agent chain. That is the point: it is defence in depth against the future route the thread above describes, not a live path. Verified no current route regresses — `/agent/openapi.json`, `/agent/v1/context` and the AI route all sit behind the same auth step, and `/health` is not an agent path.

`cli-core.test.ts` "should leave a small burst untouched at the default configuration" passes with the limiter removed. It guards against an over-eager default rather than proving the limiter works, which the dedicated suite does. Kept, worth knowing it is not a regression test.

Commit titles on this branch: one is 73 chars and one uses the `agent-bff` scope where the others use `bff`. The repo squashes on the PR title, so neither reaches `main`.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -37,7 +37,7 @@ const ERROR_STATUSES: Record<string, string> = {
   413: `The request body exceeds the BFF limit of ${BODY_LIMIT}`,
   415: 'The request declares a character set the server cannot decode. Other content types are NOT rejected: a form-urlencoded body is parsed and validated like JSON (its values arrive as strings, so typed fields such as page.limit fail with 400), while any other non-JSON content type is read as an absent body, silently dropping filters and pagination',
   422: 'A field is unknown, not filterable, or is a nested relation path',
-  429: 'The agent rate-limited the request',
+  429: 'The BFF rate-limited the request, for one of two reasons: the caller identity exceeded its per-window budget, or the limiter is saturated and cannot open a window for a new identity. The message distinguishes them, and Retry-After carries the seconds to wait. On data and action routes the agent may also rate-limit the request itself',
```
nbouliol: `The message distinguishes them` asks callers to branch on message text, which README:116 rules out ("consumers branch on `error.type`, never on message text") — and both throw sites carry the same `too_many_requests`. Give saturation its own `type` or a `details` discriminator, or say plainly that the two are indistinguishable to a caller?

packages/agent-bff/README.md
```diff
@@ -94,13 +94,16 @@ yarn start:dev         # node --env-file=.env dist/cli.js
...
 | `BFF_OPENAPI_ENABLED` | no       | Serve `GET/HEAD /agent/openapi.json` (auth-gated) when `true`. Defaults to `true`. Set to `false` for customers who do not want the HTTP surface exposed: an authenticated `GET`/`HEAD` then gets `404 openapi_disabled`, other methods fall through to the agent routes exactly as they do when enabled, and `forest-bff openapi` keeps working either way. Accepted values: `true`/`false`. **The served document is unfolded and is not filtered per caller**: any authenticated caller, whatever their role, reads the name of every exposed collection, relation and field. Set this to `false` if that surface must not be reachable over HTTP. |
 
 ### Config validation
 
 - A malformed value (a non-http(s) `*_URL`, a `HTTP_PORT` that is not a decimal integer in 0–65535,
-  a non-IANA `BFF_DEFAULT_TIMEZONE`, a non-boolean `BFF_OPENAPI_ENABLED`) fails fast at boot: the
-  process exits with a clear error and never echoes the offending value.
+  a non-IANA `BFF_DEFAULT_TIMEZONE`, a non-boolean `BFF_OPENAPI_ENABLED`, a `BFF_RATE_LIMIT_*` that is
```
nbouliol: `BFF_RATE_LIMIT_MAX_REQUESTS=20000` and `BFF_RATE_LIMIT_WINDOW_MS=500` are positive decimal integers that still fail boot (max 10 000, min 1 000). Worth stating the ranges here and in the two rows above, the way `HTTP_PORT` and `BFF_AI_TIMEOUT_MS` do?

packages/agent-bff/src/rate-limit/rate-limit-middleware.ts
```diff
@@ -0,0 +1,105 @@
...
+  return async function rateLimitMiddleware(ctx, next) {
+    if (!isAgentPath(ctx.path)) {
+      await next();
+
+      return;
+    }
+
+    const key = bucketKeyOf(ctx.state as AuthEdgeState);
+
+    if (key === undefined) throw unauthorized();
```
nbouliol: `unauthorized()` takes the default "Missing or invalid credentials", which reads like a real credential failure on a path that should be unreachable — `permissions-routes-middleware.ts:42` passes `'No caller identity for this request'` for the same condition.

nbouliol (commented):
Re-reviewed at `fe244e5`. All five comments from the last pass are addressed, and the fail-closed test was properly renamed and re-asserted rather than left pinning the old behaviour — I checked that nothing legitimate now hits the new `throw`, since `createCorsMiddleware` short-circuits `OPTIONS` at 204 before the chain.

Three things below. The first is the one that matters: the 429 split tells callers to branch on message text, which this package's own error contract rules out.

One more that falls outside the diff so I couldn't anchor it: `README.md:114` still says the request edge "enforces three cross-cutting concerns" (`Auth-mode` / `CORS` / `Timezone`) — the limiter makes four, and it's currently documented only as env rows.

Unrelated to this PR: the red check is `packages/ai-proxy/test/llm.integration.test.ts:503` (39 passed, 1 failed), a live-provider test in a package this branch doesn't touch.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -37,7 +37,7 @@ const ERROR_STATUSES: Record<string, string> = {
   413: `The request body exceeds the BFF limit of ${BODY_LIMIT}`,
   415: 'The request declares a character set the server cannot decode. Other content types are NOT rejected: a form-urlencoded body is parsed and validated like JSON (its values arrive as strings, so typed fields such as page.limit fail with 400), while any other non-JSON content type is read as an absent body, silently dropping filters and pagination',
   422: 'A field is unknown, not filterable, or is a nested relation path',
-  429: 'The agent rate-limited the request',
+  429: 'The BFF rate-limited the request, for one of two reasons: the caller identity exceeded its per-window budget, or the limiter is saturated and cannot open a window for a new identity. The `details.cause` field of the error body distinguishes them (`limit_exceeded` or `limiter_saturated`), and Retry-After carries the seconds to wait. On data and action routes the agent may also rate-limit the request itself',
```
nbouliol: A relayed agent 429 never carries `cause` — `mapAgentError` sets `details` to the agent's own payload — so a caller branching on it sees `undefined` or arbitrary data. The `Retry-After` description just below (72-73) already spells out the parallel case; worth the same clause here?

packages/agent-bff/src/rate-limit/rate-limit-middleware.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (entry === undefined) {
+      purgeExpired(current);
+
+      if (buckets.size >= maxEntries) {
+        const earliestReset = Math.min(...[...buckets.values()].map(live => live.resetAt));
+
+        throw tooManyRequests(
+          retryAfterSeconds(earliestReset, current),
+          'Too many requests: the rate limiter is saturated',
+          { cause: 'limiter_saturated' },
```
nbouliol: Now that clients branch on these, the two values live in 7 places across 5 files with no shared source (here, `:101`, the OpenAPI prose at `:40`, `README.md:98`, and both tests) — a typo in any one silently breaks the contract. Export a `RateLimitCause` union next to `tooManyRequests` and reference it?

nbouliol (commented):
Third pass, at `2637468`. All three comments from the last round are fixed, and the 429 one properly — `details.cause` is a real machine-readable discriminator rather than a reworded sentence, which is the right answer to the type-first contract. I checked the README ranges against `env-config.ts:176-187` and they agree exactly on both bounds, and `tooManyRequests` has only the two callers so the new parameter regressed nothing.

Two things left, both about the new `cause` contract rather than the limiter itself.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Fixed: the two throw sites carry `details.cause` (`limit_exceeded` / `limiter_saturated`) and the description says callers branch on that, not on the message.

packages/agent-bff/README.md
Tonours: Fixed: both rows now state the ranges (1–10 000, 1 000–2 147 483 647) and say an out-of-range value fails the boot.

packages/agent-bff/src/rate-limit/rate-limit-middleware.ts
Tonours: Fixed: `unauthorized('No caller identity for this request')`, same wording as permissions-routes-middleware.ts:42.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Added: the 429 description now says a relayed agent 429 carries the agent's payload as `details`, so neither `cause` nor Retry-After, and to read them only when present. Covered by a test.

packages/agent-bff/src/rate-limit/rate-limit-middleware.ts
Tonours: Done: `RATE_LIMIT_CAUSES` and the `RateLimitCause` union sit next to `tooManyRequests`, whose `cause` argument is now required and typed. The middleware and the OpenAPI description read from that const; the two test assertions stay literal on purpose, since they pin the wire value.

nbouliol (approved):
Fourth pass, at f3ef344. All ten comments across the review are addressed and verified against the code.

The last two landed well beyond the ask: cause is now a required typed argument backed by RATE_LIMIT_CAUSES, so the compiler enforces it at every call site and the OpenAPI description interpolates the constants instead of restating them — the value can no longer drift between code, document and tests. The relayed-429 caveat also tells callers to read cause and Retry-After only when present rather than branching on them, which is the right guidance.

I also checked the three main merges for integration damage on the overlapping files: the page cap survives inside the new strictObject with the offset refine running after it, and the Page description merges cleanly with the default-page-size prose. CI is green apart from the ai-proxy live-provider test, which this branch does not touch.

One thing for later, not blocking and not this PR's code: the JSON-only guard added in #1856 sits ahead of the agent chain like the body parser, so 415 now escapes the quota alongside the 400/413 the Known limit paragraph already lists.

PR #1859, opened 2026-08-28
fix(agent-bff): sanitize action result html before relaying

Tonours:
## What

The BFF sanitizes the agent's HTML at its own boundary, on the three surfaces that relayed it verbatim:

- the execute success body `html`,
- the 400 `action_error` `details.html`,
- every `htmlBlock` `content` in the `/form` layout, including inside a `page`.

One helper, three call sites. The `/form` and `/execute` 200 descriptions say so, in both the generic and the unfolded OpenAPI document.

fixes PRD-1095

## Why

Rendering the old verbatim html through `innerHTML` executed arbitrary script in any consumer. A contract note alone leaves that live for whoever does not read it, and PRD-1097 owns the response typing anyway. So we sanitize at the boundary and no consumer has to know.

The pin is exact on purpose. 2.17.6+ resolves htmlparser2 `^12`, which is ESM-only with no CJS build, and Jest 29 cannot run it. Below 2.17.2 the zero-padded numeric reference fix (a `javascript:` URL detection bypass) is missing. 2.17.5 is the only version with both.

## The policy

Default tag allowlist — rich text (p, b, strong, a, tables, lists) survives; `script`, `svg`, `iframe`, `img`, `on*` attributes and `javascript:` URLs do not.

On top of it, `style` is allowed on any tag but filtered: ~38 presentational properties (color, background, font, padding, margin, border, width, text-align…) with values matching `/^[^;{}()]*$/`. No parentheses means no `url(...)` and no `expression(...)`; `position`, `top`, `left` and `z-index` are not on the list, so styled output cannot overlay the host page.

The default allowlist alone was not viable: it drops `style` entirely, and this repo's own demo (`datasource-demo-fintech/src/customizations/kyc_cases.ts`) builds its whole result html out of inline styles.

## Size cap

Parsing is super-linear in nesting depth, and the agent's html was previously relayed without being parsed at all — so sanitizing introduces a CPU cost the agent controls. Measured on 2.17.5, `<div>` nested N deep:

| input | wall time |
|---|---|
| 110 KB (10k deep) | 23 ms |
| 275 KB (25k deep) | 210 ms |
| 1.1 MB (100k deep) | 1 872 ms |

An action interpolating a record field into its html — the exact pattern behind this ticket — would let a user write that field and block the event loop. So html over 256 KB is dropped rather than sanitized, and the drop is logged `Warn` with the length. At the cap the parse costs ~120 ms.

The layout traversal is bounded the same way: past 10 nested `page` levels the inner `elements` are dropped to `[]` and logged, so a deep layout cannot blow the stack. Dropping rather than relaying the subtree is deliberate — an unrecursed subtree would ship unsanitized `htmlBlock` content.

`message` is not sanitized, on either surface. It is plain text by contract; sanitizing it would mangle a legitimate `<` or `&`.

## Contract changes a consumer can see

- An html that sanitizes to nothing collapses to "no html": the success body sends `html: null` (it sent `""` before), and the error omits `details` rather than relaying `{ html: "" }`.
- A non-string html is still `null` / no details, as before.
- `` is dropped, element and all. Nothing in the repo needs it and allowing it buys a tracking-pixel surface for no demonstrated use.

## Scope and safety

- No route added, no schema changed: `ActionExecuteSuccessBody.html` was already `string | null` and `details` was already optional.
- A malformed agent layout (a non-array `layout`, a null element, a `page` whose `elements` is not an array) passes through instead of throwing. The sanitizer walks agent-controlled JSON, so it does not trust its own types.
- Lockfile delta is +85 / -0: `sanitize-html@2.17.5` and its transitive deps, nothing re-resolved.
- Every drop is logged: over the cap (`Warn`), and a sanitizer throw (`Error`, dead at the pinned version but the execute path runs *after* the action committed, so a 500 there means a retry and a double execution).

## How to test

- `cd packages/agent-bff && yarn jest test/action test/openapi`
- By hand: execute an action returning `<p>ok</p>` — it comes back as `<p>ok</p>`. Run the fintech demo's *KYC Approved* action — the green banner and the table keep their styling.

Full package suite: 1367 green.

## Known limitation

Agent html size stays unbounded, as before. The exact pin stops Dependabot until Jest has an ESM story.

### General

### Security

### DIFF
--- a/packages/agent-bff/src/action/action-routes-middleware.ts
+++ b/packages/agent-bff/src/action/action-routes-middleware.ts
@@ -138,8 +138,6 @@ async function handleExecute({
     }
 
     if (error instanceof ActionFormValidationError) {
-      // The agent's Error-result html is the same untrusted output as the success html, so it
-      // goes through the same sanitizer; a non-string value yields no details rather than a 500.
       const html = sanitizeActionHtml(error.html);
 
       throw actionError(error.message, html === null ? undefined : { html });
--- a/packages/agent-bff/src/action/action-routes-middleware.ts
+++ b/packages/agent-bff/src/action/action-routes-middleware.ts
@@ -17,6 +17,7 @@ import {
 import { mapActionExecuteResult } from './action-execute-mapper';
 import { mapActionForm } from './action-form-mapper';
 import defaultCreateAgentActionClient, { extractRawLayout } from './agent-action-client';
+import sanitizeActionHtml from './sanitize-action-html';
 import { mapAgentError } from '../http/agent-error-mapper';
 import {
   callAgent,
@@ -137,7 +138,11 @@ async function handleExecute({
     }
 
     if (error instanceof ActionFormValidationError) {
-      throw actionError(error.message, error.html !== undefined ? { html: error.html } : undefined);
+      // The agent's Error-result html is the same untrusted output as the success html, so it
+      // goes through the same sanitizer; a non-string value yields no details rather than a 500.
+      const html = sanitizeActionHtml(error.html);
+
+      throw actionError(error.message, html === null ? undefined : { html });
     }
 
     throw mapAgentError(error, { logger });
--- a/packages/agent-bff/src/action/action-routes-middleware.ts
+++ b/packages/agent-bff/src/action/action-routes-middleware.ts
@@ -102,7 +102,7 @@ async function handleForm({
   const skippedFields = await callAgent(() => action.tryToSetFields(values), logger);
 
   ctx.status = 200;
-  ctx.body = mapActionForm(action, skippedFields, extractRawLayout(action));
+  ctx.body = mapActionForm(action, skippedFields, extractRawLayout(action), logger);
 }
 
 async function handleExecute({
@@ -138,7 +138,7 @@ async function handleExecute({
     }
 
     if (error instanceof ActionFormValidationError) {
-      const html = sanitizeActionHtml(error.html);
+      const html = sanitizeActionHtml(error.html, logger);
 
       throw actionError(error.message, html === null ? undefined : { html });
     }
@@ -146,7 +146,7 @@ async function handleExecute({
     throw mapAgentError(error, { logger });
   }
 
-  const { status, body } = mapActionExecuteResult(raw);
+  const { status, body } = mapActionExecuteResult(raw, logger);
 
   // An unrecognized payload (a File stream, or a new agent result type) maps to a generic 501 with
   // no trace of what it was; log a short shape hint so the case can be diagnosed without the body.
--- a/packages/agent-bff/src/action/sanitize-action-html.ts
+++ b/packages/agent-bff/src/action/sanitize-action-html.ts
@@ -2,23 +2,23 @@ import type { ForestServerActionFormLayoutElement } from '@forestadmin/forestadm
 
 import sanitizeHtml from 'sanitize-html';
 
+function sanitize(html: unknown): string {
+  return typeof html === 'string' ? sanitizeHtml(html) : '';
+}
+
 export default function sanitizeActionHtml(html: unknown): string | null {
-  if (typeof html !== 'string') return null;
-
-  const sanitized = sanitizeHtml(html);
-
-  return sanitized === '' ? null : sanitized;
+  return sanitize(html) || null;
 }
 
 export function sanitizeActionLayout(
   layout: ForestServerActionFormLayoutElement[],
 ): ForestServerActionFormLayoutElement[] {
   return layout.map(element => {
-    if (element.component === 'htmlBlock') {
-      return { ...element, content: sanitizeActionHtml(element.content) ?? '' };
+    if (element?.component === 'htmlBlock') {
+      return { ...element, content: sanitize(element.content) };
     }
 
-    if (element.component === 'page') {
+    if (element?.component === 'page' && Array.isArray(element.elements)) {
       return { ...element, elements: sanitizeActionLayout(element.elements) };
     }
 
--- a/packages/agent-bff/src/action/sanitize-action-html.ts
+++ b/packages/agent-bff/src/action/sanitize-action-html.ts
@@ -1,13 +1,5 @@
 import sanitizeHtml from 'sanitize-html';
 
-/**
- * The agent's action-result html is untrusted output: an action can interpolate record data into
- * it, so the BFF sanitizes it at this trust boundary before relaying (PRD-1095). The library's
- * default allowlist keeps rich text (paragraphs, emphasis, links, tables, lists) and drops active
- * content: script/style/iframe/svg/img tags, on* attributes, javascript:/data: URLs. A non-string
- * value is an unchecked agent-JSON cast, not html: null is relayed instead of throwing inside the
- * sanitizer.
- */
 export default function sanitizeActionHtml(html: unknown): string | null {
   return typeof html === 'string' ? sanitizeHtml(html) : null;
 }
--- a/packages/agent-bff/src/action/sanitize-action-html.ts
+++ b/packages/agent-bff/src/action/sanitize-action-html.ts
@@ -60,17 +60,19 @@ const OPTIONS: sanitizeHtml.IOptions = {
 function sanitize(html: unknown, logger: Logger): string | null {
   if (typeof html !== 'string') return null;
 
-  if (html.length > MAX_HTML_CHARACTERS) {
-    logger('Warn', 'Action html dropped: longer than the sanitizable size', {
-      characters: html.length,
+  let bounded = html;
+
+  if (bounded.length > MAX_HTML_CHARACTERS) {
+    logger('Warn', 'Action html truncated: longer than the sanitizable size', {
+      characters: bounded.length,
       limit: MAX_HTML_CHARACTERS,
     });
 
-    return null;
+    bounded = bounded.slice(0, MAX_HTML_CHARACTERS);
   }
 
   try {
-    return sanitizeHtml(html, OPTIONS) || null;
+    return sanitizeHtml(bounded, OPTIONS) || null;
   } catch (error) {
     logger('Error', 'Action html dropped: sanitization failed', { cause: String(error) });
 
--- a/packages/agent-bff/src/action/sanitize-action-html.ts
+++ b/packages/agent-bff/src/action/sanitize-action-html.ts
@@ -90,10 +90,33 @@ export function sanitizeActionLayout(
   layout: ForestServerActionFormLayoutElement[],
   logger: Logger,
   depth = 0,
+  budget: { remaining: number } = { remaining: MAX_HTML_CHARACTERS },
 ): ForestServerActionFormLayoutElement[] {
   return layout.map(element => {
     if (element?.component === 'htmlBlock') {
-      return { ...element, content: sanitize(element.content, logger) ?? '' };
+      const { content } = element;
+
+      if (typeof content !== 'string') {
+        return { ...element, content: sanitize(content, logger) ?? '' };
+      }
+
+      if (content.length > budget.remaining) {
+        logger(
+          'Warn',
+          'Action layout html truncated: total layout html longer than the sanitizable size',
+          {
+            characters: content.length,
+            remaining: budget.remaining,
+            limit: MAX_HTML_CHARACTERS,
+          },
+        );
+      }
+
+      const bounded = content.slice(0, budget.remaining);
+
+      budget.remaining -= bounded.length;
+
+      return { ...element, content: sanitize(bounded, logger) ?? '' };
     }
 
     if (element?.component === 'page' && Array.isArray(element.elements)) {
@@ -105,7 +128,10 @@ export function sanitizeActionLayout(
         return { ...element, elements: [] };
       }
 
-      return { ...element, elements: sanitizeActionLayout(element.elements, logger, depth + 1) };
+      return {
+        ...element,
+        elements: sanitizeActionLayout(element.elements, logger, depth + 1, budget),
+      };
     }
 
     return element;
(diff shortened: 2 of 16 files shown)


packages/agent-bff/src/action/action-routes-middleware.ts
```diff
@@ -137,7 +138,9 @@ async function handleExecute({
     }
 
     if (error instanceof ActionFormValidationError) {
-      throw actionError(error.message, error.html !== undefined ? { html: error.html } : undefined);
+      const html = sanitizeActionHtml(error.html);
+
+      throw actionError(error.message, html === null ? undefined : { html });
```
nbouliol, line 143: An html that is entirely active markup sanitizes to `''`, so this relays `details: { html: '' }` where a missing html omits details - should an empty result collapse to `undefined` too?

packages/agent-bff/src/action/sanitize-action-html.ts
```diff
@@ -0,0 +1,5 @@
+import sanitizeHtml from 'sanitize-html';
+
+export default function sanitizeActionHtml(html: unknown): string | null {
+  return typeof html === 'string' ? sanitizeHtml(html) : null;
```
nbouliol: The allowlist is whatever the library defaults to, so a bump silently changes the contract the OpenAPI description advertises - what about passing an explicit options constant?

nbouliol (commented):
The `/form` route still relays `layout` verbatim, and `htmlBlock.content` (`action-form-mapper.ts:56`, via `agent-action-client.ts:58`) is the same untrusted agent HTML - same vector, still open. Sanitize it here too, or a follow-up ticket?

packages/agent-bff/src/action/action-routes-middleware.ts
Tonours: Fixed. `sanitizeActionHtml` now returns `null` when the sanitized string is empty, so an html that is entirely active markup omits `details` exactly like a missing html. Same collapse on the success surface (`html: null` instead of `""`). Covered by `action-routes-middleware.test.ts` ("omits details when the action Error html is entirely active markup") and `action-execute-mapper.test.ts` ("maps an html that is entirely active markup to null").

packages/agent-bff/src/action/sanitize-action-html.ts
Tonours: Not doing it. A bump cannot happen silently here: the dep is pinned exact (`"sanitize-html": "2.17.5"` in `packages/agent-bff/package.json`, no caret), and the tests assert the exact sanitized string, not "something was stripped" — `action-execute-mapper.test.ts` expects `<p>Safe <b>markup</b></p><a>click</a>` and `action-form-mapper.test.ts` expects `<p>Safe <b>markup</b></p>`. Any allowlist change, in a bump or otherwise, fails CI.

An explicit constant would have to restate ~70 tags and the attribute/URL-scheme rules, then be kept in sync by hand — a copy that drifts *toward* being wrong, and drifts silently since the tests would still pass. The OpenAPI text advertises "sanitized server-side against an allowlist", not a tag list, so nothing in the published contract is pinned to a specific set either.

Tonours:
Fixed, in this PR rather than a follow-up: same trust boundary, same untrusted agent string, and the fix is one call.

`mapActionForm` now runs the layout through `sanitizeActionLayout` (`sanitize-action-html.ts`), which walks `htmlBlock.content` including the ones nested in a `page`, and leaves every other component untouched. Single caller, so `/form` on both the folded and unfolded routes is covered. The `/form` 200 description says so now, mirroring `/execute`. Tests: three cases in `action-form-mapper.test.ts` (top-level block, nested-in-page block, block that is entirely active markup → empty content).

Two other things on this push:

- `sanitizeActionHtml` collapses a fully stripped html to `null`, so `details` is omitted instead of relaying `{ html: "" }` (your inline comment).
- The lockfile was reverted to a minimal delta. The previous push had re-resolved the whole tree: 1703 insertions / 1124 deletions, 209 version bumps on unrelated dev deps (`@actions/*`, `@octokit/*`, `semantic-release`, `sigstore`). It is now +85 / -0 — `sanitize-html@2.17.5` and its transitive deps, nothing else.

On the qlty comment (`mapActionExecuteResult`, 4 returns): pre-existing, not introduced here. `git show origin/main:packages/agent-bff/src/action/action-execute-mapper.ts` has the same 4 returns; this PR only changed the `html:` line inside one of them. They are guard clauses, which is what the repo asks for. Not touching it.

packages/agent-bff/src/action/sanitize-action-html.ts
Tonours: Reversing this, for a different reason than the one on the thread. The bump argument still does not hold (exact pin, exact-string tests), but the library defaults turned out to be wrong for what Forest actions actually emit: `style` is in no `allowedAttributes` entry, so every inline style is dropped. `packages/datasource-demo-fintech/src/customizations/kyc_cases.ts:60` builds its whole result html out of inline styles — this repo's own demo would have come back as unstyled text.

So there is an explicit options constant now, and it is narrow: the default tag allowlist untouched, `style` allowed on any tag, and `allowedStyles` restricted to ~38 presentational properties with values matching `/^[^;{}()]*$/`. That regex is the point — no parentheses means no `url(...)` and no `expression(...)` — and `position`, `top`, `left` and `z-index` are simply not on the list, so styled output cannot overlay the host page.

Measured on 2.17.5:

```
<div style="background:#16a34a;color:#fff;padding:12px">OK</div>
  -> unchanged
<div style="position:fixed;top:0;z-index:9999">overlay</div>
  -> <div>overlay</div>
<div style="background:url(javascript:alert(1))">x</div>
  -> <div>x</div>
```

`img` stays off the allowlist: nothing in the repo needs it, and adding it buys a tracking-pixel surface for no demonstrated use. Tests: two new cases in `action-form-mapper.test.ts` and one in `action-execute-mapper.test.ts` pinning both halves — styles kept, overlay and `url()` properties dropped.

Tonours:
Two more pushes, both from things this PR introduced rather than from the ticket.

**Explicit sanitize options.** The library defaults drop `style` on every tag. `packages/datasource-demo-fintech/src/customizations/kyc_cases.ts:60` builds its entire result html out of inline styles, so the default policy would have shipped this repo's own demo as unstyled text. There is a constant now: default tags untouched, `style` allowed, `allowedStyles` restricted to ~38 presentational properties with values matching `/^[^;{}()]*$/` — no parentheses means no `url(...)` and no `expression(...)`, and `position` / `top` / `left` / `z-index` are off the list, so styled output cannot overlay the host page. `img` stays off. Details on the `sanitize-action-html.ts` thread.

**Size cap.** Before this PR the agent's html was relayed without being parsed at all; sanitizing hands the agent a CPU cost. Parsing is super-linear in nesting depth — measured on 2.17.5, `<div>` nested N deep: 10k → 23 ms, 25k → 210 ms, 100k → 1 872 ms. An action interpolating a record field into its html (the pattern behind this ticket) lets a user write that field and block the event loop. Html over 256 KB is now dropped instead of sanitized, logged `Warn` with the length; at the cap the parse costs ~120 ms.

The drop is logged rather than silent, which is why `mapActionForm` and `mapActionExecuteResult` now take the `Logger` the middleware already had in scope — same shape as the existing `Warn` on the 501 path.

Also: a malformed agent layout (non-array `layout`, a null element, a `page` whose `elements` is not an array) now passes through instead of throwing. `sanitizeActionLayout` walks agent-controlled JSON, so it does not trust its own types.

`message` is deliberately not sanitized on either surface — it is plain text by contract, and sanitizing it would mangle a legitimate `<` or `&`.

packages/agent-bff/src/action/sanitize-action-html.ts
```diff
@@ -0,0 +1,98 @@
...
+  layout: ForestServerActionFormLayoutElement[],
+  logger: Logger,
+): ForestServerActionFormLayoutElement[] {
+  return layout.map(element => {
+    if (element?.component === 'htmlBlock') {
+      return { ...element, content: sanitize(element.content, logger) };
+    }
+
+    if (element?.component === 'page' && Array.isArray(element.elements)) {
+      return { ...element, elements: sanitizeActionLayout(element.elements, logger) };
```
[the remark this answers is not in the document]
Tonours: Fixed. `sanitizeActionLayout` carries a depth counter and stops at 10: past that a `page`'s `elements` are dropped to `[]` and the drop is logged `Warn`, rather than recursing until the stack goes. Dropping rather than passing the subtree through is deliberate — relaying it unrecursed would ship unsanitized `htmlBlock` content, which is the hole this PR closes.

10 is well past any real form; a Forest layout nests one page level.

Test in `action-form-mapper.test.ts`: a 12-deep chain of pages ending in a `<script>` block asserts the inner `elements` come back empty and the `Warn` fires.

Same class as the 256 KB html cap on this branch: before this PR the layout was relayed without being walked, so the traversal cost is one the change introduces.

packages/agent-bff/src/action/sanitize-action-html.ts
```diff
@@ -0,0 +1,109 @@
+import type { Logger } from '../ports/logger-port';
+import type { ForestServerActionFormLayoutElement } from '@forestadmin/forestadmin-client';
+
+import sanitizeHtml from 'sanitize-html';
+
+const MAX_HTML_CHARACTERS = 256 * 1024;
+
+const MAX_LAYOUT_DEPTH = 10;
+
+const SAFE_STYLE_VALUE = /^[^;{}()]*$/;
```
nbouliol: This blocks every parenthesised value, so `rgb()`, `rgba()`, `hsl()`, `calc()` and `var()` all vanish while the fixture only uses hex - allow parens but reject `url(` / `expression(`?

packages/agent-bff/src/action/sanitize-action-html.ts
```diff
@@ -0,0 +1,109 @@
...
+  'text-align',
+  'text-decoration',
+  'text-transform',
+  'vertical-align',
+  'white-space',
+  'width',
+];
+
+const OPTIONS: sanitizeHtml.IOptions = {
+  allowedAttributes: { ...sanitizeHtml.defaults.allowedAttributes, '*': ['style'] },
```
nbouliol, line 55: `allowedTags` is still the library default, so `img` is dropped and `class`/`id` stripped - an existing htmlBlock logo or Forest class regresses here. Allow `img` and `class` explicitly?

packages/agent-bff/src/action/sanitize-action-html.ts
```diff
@@ -0,0 +1,109 @@
...
+  allowedStyles: {
+    '*': Object.fromEntries(ALLOWED_STYLE_PROPERTIES.map(name => [name, [SAFE_STYLE_VALUE]])),
+  },
+};
+
+function sanitize(html: unknown, logger: Logger): string {
+  if (typeof html !== 'string') return '';
```
nbouliol: All three `return ''` mean "dropped", but line 82 then decodes `''` as null via truthiness while line 92 keeps it - could `sanitize` return `string | null` so the divergence is explicit?

packages/agent-bff/src/action/sanitize-action-html.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  if (html.length > MAX_HTML_CHARACTERS) {
```
nbouliol: An oversize html is dropped whole, so a client cannot tell it from the agent sending none - truncate instead, or surface the drop in the response?

packages/agent-bff/src/action/sanitize-action-html.ts
```diff
@@ -0,0 +1,109 @@
...
+  logger: Logger,
+  depth = 0,
+): ForestServerActionFormLayoutElement[] {
+  return layout.map(element => {
+    if (element?.component === 'htmlBlock') {
+      return { ...element, content: sanitize(element.content, logger) };
+    }
+
+    if (element?.component === 'page' && Array.isArray(element.elements)) {
+      if (depth >= MAX_LAYOUT_DEPTH) {
```
nbouliol, line 123: `DynamicLayoutElementPage.elements` is `DynamicFormElement`, which excludes `Page` (`datasource-customizer/src/decorators/actions/types/fields.ts:277`), so a page cannot nest a page - is this cap guarding a reachable shape?

nbouliol (commented):
Round-2 pass. The three round-1 findings are addressed: htmlBlock content is sanitized (recursion included), a fully stripped html now relays as no html, and the options are explicit. The third is only half closed though - see the `allowedTags` comment. Five findings on the new code below.

packages/agent-bff/src/action/sanitize-action-html.ts
nbouliol: Still open - a page cannot nest a page, so this cap guards an unreachable shape. Drop it, or is there a path I am missing?

nbouliol (approved):
Sanitizes all three surfaces at the boundary, and the style policy now names the URL carriers instead of banning parentheses - verified on 2.17.5 that rgb()/calc()/var() survive while URL(, url (, \75 rl(, image-set( and -moz-binding( do not. Boundary tests are exact-valued. The only thing left is the layout depth cap guarding an unreachable shape, which is dead code rather than a defect.

packages/agent-bff/src/action/sanitize-action-html.ts
Tonours: The customizer type is right — a typed Node form cannot nest Page. This helper does not see that type. It walks `ForestServerActionFormLayoutElement` from the agent's hook JSON, and that union is recursive (`page.elements` includes `page`). Toolkit `ActionLayoutElement` is the same. `ensureFormIsCorrect` only forbids mixing pages at the root; `parseLayout` / `buildLayoutSchema` will serialize a nested page if one shows up. Rails / Python / a custom `getForm` never go through the customizer types. Dropping the cap makes a deep page chain a `RangeError` on `/form`, which is why it is here.

PR #1860, opened 2026-08-28
fix(agent-bff): enforce the action input contract on execute

Tonours:
## What

`/execute` validates the submitted `values` against the live form before running the action.

- A required field left empty answers **400** `missing_required_fields`. `canExecute` was a form-time hint nobody re-read at execute; the gate and the hint now come from the same `missingRequiredFieldNames` helper, so they cannot disagree.
- An out-of-enum or wrongly typed value answers **422** `invalid_action_value`, list items included — the message names the failing index (`index 1 to be a number`).
- A record-picker field takes a packed id string, whatever PK type the schema publishes.

Validation runs after `setFields`, on the form the change hooks rebuilt: what is checked is exactly what `execute()` would submit. A loaded default satisfies a required field.

fixes PRD-1096

## Why

The document promised required / enum / type and the BFF forwarded everything unvalidated, so a destructive action gated on `canExecute` ran anyway.

## What is *not* checked

The OpenAPI text says so now, because the gap is invisible otherwise:

- A declared string `format` (`date-time`, `uuid`) — a `Date` field takes any string.
- A widget's own option list. `Dropdown` / `RadioGroup` / `CheckboxGroup` carry their values in `widgetEdit.parameters.static.options`, not in `enums`, and a searchable dropdown computes them per query — a static list is not an exhaustive one, so enforcing it would reject valid values.
- `Json` and `File` fields, by design. A File value is already a data-uri string by the time the validator runs.

## Observability

Every rejection logs a `Warn` with the field **names** — never the values, which carry customer data. An enum field the BFF had to accept without options logs too: that is a change hook returning an empty or malformed `enums`, and it silently disables the check.

## Cross-SDK check

The validator branches on `field.getReference()`, so an agent that omits `reference` on a record picker would see packed ids rejected against the PK's numeric type. Checked, not assumed: the Python agent emits it (`agent-python/.../generator_action.py:76`, `output["reference"] = f"{collection.name}.{pk}"`) and so does Ruby (`agent-ruby/.../generator_action.rb:110`), both alongside the PK column type, exactly like Node (`generator-actions.ts:145-149`). No SDK gap found.

## Shared rules

The column-type algebra lives in one place, `read-model/field-type.ts`: `FieldType`, `normalizeFieldType`, `isEnumFieldType`, `primitiveShapeOf`, `enumOptionsOf`. The OpenAPI document builder and the runtime validator both read it, so a primitive a newer agent introduces is unconstrained on both sides or constrained on both — never one. `NumberList` and the array form `['Number']` normalize to the same thing, so agents that differ on that encoding are judged identically.

## Known limitation

Breaking for JSON callers that sent coerced values: a string in a Number field is the ticket's own repro and now answers 422.

## How to test

- `yarn workspace @forestadmin/agent-bff test` and `yarn workspace @forestadmin/agent-client test`.
- Repros pinned in `test/action/action-execute-validation.test.ts`: `values: {}` with a required field → 400, `Tier: "NOPE"` → 422, `Amount: "nan"` → 422, and `canExecute:false` → 400 instead of executing.

### General

### Security


packages/agent-bff/src/action/action-routes-middleware.ts
```diff
@@ -110,15 +112,26 @@ async function handleExecute({
...
+      throw invalidRequest(error.message);
+    }
+
     throw mapAgentError(error, { logger });
   }
 
+  // The published field contract is enforced here, on the live form: setFields has applied the
+  // submitted values and the change hooks have rebuilt the fields, so this validates exactly the
+  // state execute() would submit. canExecute at form time was only ever a hint.
+  assertActionValuesExecutable(action);
```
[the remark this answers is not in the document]
Tonours: Addressed in db76db4b7. The validator now normalizes field types before recursing: `normalizeFieldType` (packages/agent-bff/src/read-model/capabilities-cache.ts) maps the legacy collapsed names (`NumberList` → `[Number]`, `StringList` → `[String]`), so `expectedWhenViolated` takes the list branch for them as well. A non-array value now gets the promised 422 `invalid_action_value` ("an array of a number"), and list items are strictly type-checked — a string inside a NumberList is rejected before `execute()` runs. Covered by the execute-validation tests extracted in 0c3c9756 and extended in 9d3486c5 ("rejects a wrongly typed item in a legacy NumberList field", "checks a legacy EnumList field against its options", plus the non-array list cases); re-ran the suite on the PR head: 20/20 green.

packages/agent-bff/src/action/action-values-validator.ts
```diff
@@ -0,0 +1,119 @@
...
+import type { FieldType } from '../read-model/capabilities-cache';
+
+import { isEnumFieldType, normalizeFieldType } from '../read-model/capabilities-cache';
+import { invalidActionValue, missingRequiredActionFields } from '../validation/validation-errors';
+
+const PACKED_ID_EXPECTATION = 'a string holding a packed record id';
+
+type JsonShape = 'array' | 'boolean' | 'number' | 'string' | 'any';
+
+const STRING_SHAPES = new Set([
```
nbouliol: `STRING_SHAPES` restates the primitive table `openapi/field-schemas.ts` already holds — a new agent primitive would silently fall to `any` here and `{}` there. Could this derive from `toFieldSchema(type).type` instead?

packages/agent-bff/src/action/action-values-validator.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export default function assertActionValuesExecutable(action: ActionForm): void {
+  const missing: string[] = [];
+  const violations: { field: string; expected: string }[] = [];
+
+  for (const field of action.getFields()) {
+    const name = field.getName();
+    const value = field.getValue();
+
+    if (value === undefined || value === null) {
```
nbouliol: This re-derives `action-form-mapper.ts:32-35` rather than sharing it — if the two drift, `canExecute` starts lying again, which is the ticket's own bug.

packages/agent-bff/src/action/action-values-validator.ts
```diff
@@ -0,0 +1,119 @@
...
+      if (field.isRequired()) missing.push(name);
+    } else if (field.getReference()) {
+      if (typeof value !== 'string') {
+        violations.push({ field: name, expected: PACKED_ID_EXPECTATION });
+      }
+    } else {
+      const type = normalizeFieldType(field.getType());
+      const rawOptions = isEnumFieldType(type) ? action.getEnumField(name).getOptions() : undefined;
```
nbouliol: `isEnumFieldType` accepts `EnumList`, but `mapActionForm` only emits `enumValues` when `getType() === 'Enum'` — a legacy `EnumList` field gets a 422 on options the `/form` response never exposed.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -364,9 +365,24 @@ function registerRelationRequests(
 }
 
 function actionFieldSchema(field: UnfoldedAction['fields'][number]): SchemaObject {
-  const base = field.enums
-    ? { type: 'string' as const, enum: field.enums }
-    : toFieldSchema(field.type);
+  if (field.reference !== null) {
```
nbouliol, line 368: This early return skips the `isRequired` description added at line 387, so a required record-picker field is no longer documented as required.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -364,9 +365,24 @@ function registerRelationRequests(
...
+    return { type: 'string', description: `A packed record id for ${field.reference}.` };
+  }
+
+  const options =
```
nbouliol: Same predicate as `action-values-validator.ts:109-110` — a shared `enumOptionsOf(type, enums)` would keep the document and the runtime from drifting apart.

nbouliol (commented):
Solid fix: the three repro cases and the `canExecute` gate are each gated and tested before `execute()`. Main concern is that the same rule now lives in two places in a few spots — document vs runtime — which is the drift class the ticket is about.

packages/agent-bff/src/action/action-values-validator.ts
Tonours: Done, though not through `toFieldSchema` — `test/openapi/openapi-mount-invariant.test.ts` forbids `src/openapi` imports outside the two known mount points, so the validator cannot reach into it.

The table moved down instead. `PRIMITIVE_SHAPES` and `jsonShapeOf` now live in `read-model/capabilities-cache.ts`, next to `FieldType` and `normalizeFieldType`, and both sides read it: `toFieldSchema` builds its schema on top of the shape (formats and the File description are the only things left in `openapi/field-schemas.ts`), and the validator calls `jsonShapeOf` directly. A new agent primitive is now unconstrained on both sides or constrained on both, never one.

One behaviour delta, and it is unreachable: the old `STRING_SHAPES` omitted `File`, so a File field fell to `any`. It is `string` in the shared table. `encodeFileFieldValue` runs inside `setFields`, before the validator, so by then a File value is already a data-uri string or null — the validator never sees a non-string one.

packages/agent-bff/src/action/action-values-validator.ts
Tonours: Shared now. `missingRequiredFieldNames(fields)` is exported from `action-form-mapper.ts` and used by both: `mapActionForm` builds `requiredFields` (and therefore `canExecute`) from it, and `assertActionValuesExecutable` throws its 400 from the same call. One rule, one place — they cannot answer differently.

packages/agent-bff/src/action/action-values-validator.ts
Tonours: Right, and the fix is on the form side: `mapActionForm` now emits `enumValues` for every enum field via `isEnumFieldType(field.getType())`, so `EnumList` and `['Enum']` publish their options like `Enum` does. Narrowing the validator instead would have left a legacy `EnumList` unvalidated, which is what macroscope flagged earlier on this PR.

Test in `action-form-mapper.test.ts`: "emits enumValues for a legacy EnumList field, which the execute validator checks".

packages/agent-bff/src/openapi/unfolded-paths.ts
Tonours: Fixed. `actionFieldBaseSchema` builds the base (record picker included) and `actionFieldSchema` appends the required sentence on top, concatenating when the base already carries a description. A required record picker now documents as `A packed record id for users.id. The agent declares this field required on the static form.`

The unfolded fixture grew a required reference field (`owner`), so the existing `toEqual` on the whole `values` schema asserts it.

packages/agent-bff/src/openapi/unfolded-paths.ts
Tonours: Shared. `enumOptionsOf(type, options)` sits in `read-model/capabilities-cache.ts` next to `isEnumFieldType`: it returns the options only for an enum field type, and only when they are a non-empty list of strings. `unfolded-paths.ts` passes `field.enums`, the validator passes `action.getEnumField(name).getOptions()`. `isEnumFieldType` no longer has a second caller applying its own emptiness rule.

Tonours:
Second push, from the review passes rather than from the thread comments.

**The validation path was invisible.** `BffHttpError` is serialized and returned before any logging, and there is no access log middleware — so every 400 and 422 this PR introduces left zero server-side trace. An agent shipping a change hook that starts marking a field required would produce a wave of 400s with silent logs. Each rejection now logs a `Warn` carrying the field **names** only; action values carry customer data and never enter a log line, which a test asserts.

**An enum with no usable options silently stops being validated.** `enums: []` is legitimate (a hook can compute an empty list) but `enums: 'gold'` is an agent bug, and both ended in the same place: the field falls back to a plain string check and accepts anything. It logs a `Warn` now.

**The list message was ambiguous.** `tags (expected a string)` meant either "send an array" or "item 3 is wrong". It names the index: `index 1 to be a number`, and reports every bad index rather than stopping at the first.

**Two things the OpenAPI text claimed more broadly than the code delivers**, now stated:

- A declared string `format` is not enforced. A `Date` field takes any string.
- A widget's option list is not checked. `Dropdown` / `RadioGroup` / `CheckboxGroup` carry their values in `widgetEdit.parameters.static.options`, not `enums`. Enforcing them is the wrong call, not just missing work: a searchable dropdown computes its options per query, so the static list is not exhaustive and rejecting against it would refuse valid values.

**Cross-SDK check on the record-picker branch**, since it decides between "packed id string" and the PK's numeric type. Python emits `reference` (`agent-python/.../generator_action.py:76`) and so does Ruby (`agent-ruby/.../generator_action.rb:110`), both alongside the PK column type, like Node. No gap.

Also: `test/action/action-execute-validation.test.ts:158` claimed "before executing" without asserting the spy — fixed, both branches assert it now.

Adjacent bug found and deliberately left alone: `agent-client/src/action-fields/file-value.ts:69` tests `isFile(value)` on a single object, so an *array* of File objects sent to a `['String']` or `Json` field passes through and serializes as `{"buffer":{"type":"Buffer",...}}`. Pre-existing, and through the BFF the new validator now happens to catch it — but mcp-server calls the same `setFields` with no validator, so that path still writes corrupt data and reports success. Worth its own ticket.

packages/agent-bff/src/action/action-values-validator.ts
```diff
@@ -0,0 +1,115 @@
...
+): string | undefined {
+  if (Array.isArray(type)) {
+    const [itemType] = type;
+
+    if (!Array.isArray(value)) return `an array of ${checkFor(itemType, options).expected}`;
+
+    const bad = value
+      .map((item, index) => ({ index, expected: expectedWhenViolated(itemType, item, options) }))
+      .filter((item): item is { index: number; expected: string } => item.expected !== undefined);
+
```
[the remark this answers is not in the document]
Tonours: Not doing this: the composite branch is unreachable on the execute path.

A `{ fields }` type can only reach an action field through one route — a Collection field, where the agent sets `output.type = pkSchema.columnType` (`packages/agent/src/utils/forest-schema/generator-actions.ts:145-149`). Every other branch of `buildFieldSchema` emits an `ActionFieldType` string or `[Primitive]`. Python and Ruby build it the same way (`agent-python/.../generator_action.py:74-76`, `agent-ruby/.../generator_action.rb:108-110`).

And that same route always sets `reference` alongside the type — the two lines are adjacent in all three SDKs. The validator branches on `field.getReference()` **before** it looks at the type at all, so a Collection field is checked as a packed id string and never reaches `expectedWhenViolated`. There is no payload where a composite type gets to `shapeOf`.

Recursing into composite fields would therefore be code with no caller, guarded by a test that has to fabricate a form no agent emits.

Separately, `Json` fields do accept any value, deliberately — that one is real, tested (`action-execute-validation.test.ts`, "does not type-check Json or File fields at the BFF") and now stated in the OpenAPI response text.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -379,10 +397,12 @@ function actionValuesSchema(action: UnfoldedAction): SchemaObject {
     // No property is required and no additional one is forbidden on purpose: a load or change hook
     // rebuilds the form at call time, so the schema's static fields are an indication, not the
     // contract — and the read-model cannot tell a hookless action from a legacy schema that simply
```
nbouliol, line 399: This dropped the closing line the comment had on `main` (`// omitted its hooks, so "static" is never certain.`), so the sentence now ends on "that simply".

nbouliol (commented):
Round 1 findings all addressed — the shared field-type module is the right seam. Remaining points below are minor; CI's only red is the unrelated ai-proxy LLM integration test.

nbouliol (approved):
All six threads from the earlier rounds are addressed in code: the primitive table, the enum-options predicate and the required-field predicate now share `read-model/field-type.ts` and `action-form-mapper.ts`, `mapActionForm` covers a legacy `EnumList`, the record-picker schema keeps its required description, and a malformed `reference` publishes as null. The three repros and the `canExecute` gate are each gated and tested before `execute()`.

PR #1861, opened 2026-08-31
fix(bff): let a key with allowedOrigins serve non-browser clients

Tonours:
## What

Layer 2 (per-key `allowedOrigins`) now only rejects an `Origin` the client actually sent. A request with no `Origin` header — curl, a Node backend, CI — passes.

One condition in `packages/agent-bff/src/cors/per-key-origin.ts`: read the header once, skip the allow-list check when the client sent nothing.

fixes PRD-1102

## Why

A key created with `allowedOrigins: ["http://localhost:4567"]` was usable from a browser only. Koa returns `''` for an absent header, `normalizeOrigin('')` returns `null`, `originAllowed` then returns `false` — so every server-side client got `403 origin_not_allowed`, with nothing in the error or the docs to explain it. The workaround was provisioning a second, origins-less key.

The `Origin` is not the boundary for a server client; the API key is. The API key middleware runs before this guard (`cli-core.ts`) and is the only writer of `ctx.state.apiKeyIdentity`, so no unauthenticated request reaches the relaxed path.

## Scope and safety

- A present, unlisted `Origin` still gets 403. Unchanged, still tested.
- `Origin: null` (sandboxed iframe, cross-origin redirect) is a **present** origin and still gets 403. The ticket suggested `normalizeOrigin(...) !== null` as the skip condition; that would have let the opaque origin through, and a browser *can* produce it. Asking whether the client sent anything at all keeps every browser-producible origin gated.
- A present but unparseable `Origin` still gets 403, as before.
- Layer 1 (`BFF_ALLOWED_ORIGINS`, `cors-middleware.ts`) is untouched — byte-identical to `main`. The two layers need opposite absent-`Origin` semantics, which is why the fix is at the caller and not in `originAllowed`.
- An empty per-key list stays a no-op; OAuth requests carry no per-key identity and stay unrestricted.

A blank `Origin:` header counts as absent and passes. That is not a new decision so much as a fact about HTTP: Node strips leading and trailing whitespace from header values, so `Origin:` and `Origin:    ` both reach koa as the empty string, indistinguishable from no header at all. No browser sends either — an opaque origin serializes to the literal string `null`, which stays rejected.

## Documented, not just fixed

`bffApiKey` in the OpenAPI document said only "a BFF API key". It now carries the rule, so a consumer reading the document alone learns that a key with `allowedOrigins` gates browsers and not server-side callers, and that `Origin: null` is rejected. A test pins it against the README.

## How to test

`yarn workspace @forestadmin/agent-bff test`

The no-Origin case in `test/cors/per-key-origin.test.ts` was flipped from 403 to 200 as the ticket asked, and two route cases were added: `Origin: null` gets 403, an empty header gets 200. The blank-input contract of `hasOrigin` is pinned in `test/cors/origin.test.ts` rather than through the route, since HTTP cannot deliver whitespace to it.

### General

### Security

### DIFF
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -146,7 +146,8 @@ normalized away, no trailing slash, no wildcard, no subdomain matching):
   `403 origin_not_allowed`. A request with no `Origin` at all — a server-side client such as curl,
   a Node backend or CI, an empty header counting as none — passes: there the API key is the
   boundary, not the origin. An opaque `Origin: null` (sandboxed iframe, cross-origin redirect) is a
-  present origin and is rejected. An empty per-key list is a no-op.
+  present origin and is rejected. An empty per-key list is a no-op. The OpenAPI document says the
+  same on the `bffApiKey` security scheme, so a consumer reading the document alone learns it too.
 
 **Local development:** browsers still enforce CORS against `localhost`, so add your dev origin(s) to
 `BFF_ALLOWED_ORIGINS` (e.g. `BFF_ALLOWED_ORIGINS=http://localhost:4200`) — there is no dev bypass.
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -146,8 +146,7 @@ normalized away, no trailing slash, no wildcard, no subdomain matching):
   `403 origin_not_allowed`. A request with no `Origin` at all — a server-side client such as curl,
   a Node backend or CI, an empty header counting as none — passes: there the API key is the
   boundary, not the origin. An opaque `Origin: null` (sandboxed iframe, cross-origin redirect) is a
-  present origin and is rejected. An empty per-key list is a no-op. The OpenAPI document says the
-  same on the `bffApiKey` security scheme, so a consumer reading the document alone learns it too.
+  present origin and is rejected. An empty per-key list is a no-op.
 
 **Local development:** browsers still enforce CORS against `localhost`, so add your dev origin(s) to
 `BFF_ALLOWED_ORIGINS` (e.g. `BFF_ALLOWED_ORIGINS=http://localhost:4200`) — there is no dev bypass.
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -142,8 +142,11 @@ normalized away, no trailing slash, no wildcard, no subdomain matching):
   blocks). Applies to `POST /oauth/token` too. Preflight `OPTIONS` from an allow-listed origin gets
   the allowed methods + headers; credentials are never enabled.
 - **Layer 2 (per-key authorization, Mode 2 only)** — when the resolved key has a non-empty
-  `allowedOrigins`, the request `Origin` must also be in that list (a missing `Origin` is rejected),
-  else `403 origin_not_allowed`. An empty per-key list is a no-op.
+  `allowedOrigins`, an `Origin` sent by the client must be in that list, else
+  `403 origin_not_allowed`. A request with no `Origin` at all — a server-side client such as curl,
+  a Node backend or CI, an empty header counting as none — passes: there the API key is the
+  boundary, not the origin. An opaque `Origin: null` (sandboxed iframe, cross-origin redirect) is a
+  present origin and is rejected. An empty per-key list is a no-op.
 
 **Local development:** browsers still enforce CORS against `localhost`, so add your dev origin(s) to
 `BFF_ALLOWED_ORIGINS` (e.g. `BFF_ALLOWED_ORIGINS=http://localhost:4200`) — there is no dev bypass.
--- a/packages/agent-bff/src/cors/cors-middleware.ts
+++ b/packages/agent-bff/src/cors/cors-middleware.ts
@@ -1,6 +1,6 @@
 import type { Middleware } from 'koa';
 
-import { originAllowed } from './origin';
+import { hasOrigin, originAllowed } from './origin';
 
 export const ALLOWED_METHODS = 'GET, POST, PUT, PATCH, DELETE, OPTIONS';
 export const ALLOWED_HEADERS =
@@ -17,9 +17,9 @@ export default function createCorsMiddleware({
   return async function corsMiddleware(ctx, next) {
     const origin = ctx.get('Origin');
 
-    if (origin) ctx.vary('Origin');
+    if (hasOrigin(origin)) ctx.vary('Origin');
 
-    const allowed = origin ? originAllowed(origin, allowedOrigins) : false;
+    const allowed = hasOrigin(origin) && originAllowed(origin, allowedOrigins);
 
     if (allowed) ctx.set('Access-Control-Allow-Origin', origin);
 
--- a/packages/agent-bff/src/cors/cors-middleware.ts
+++ b/packages/agent-bff/src/cors/cors-middleware.ts
@@ -1,6 +1,6 @@
 import type { Middleware } from 'koa';
 
-import { hasOrigin, originAllowed } from './origin';
+import { originAllowed } from './origin';
 
 export const ALLOWED_METHODS = 'GET, POST, PUT, PATCH, DELETE, OPTIONS';
 export const ALLOWED_HEADERS =
@@ -17,9 +17,9 @@ export default function createCorsMiddleware({
   return async function corsMiddleware(ctx, next) {
     const origin = ctx.get('Origin');
 
-    if (hasOrigin(origin)) ctx.vary('Origin');
+    if (origin) ctx.vary('Origin');
 
-    const allowed = hasOrigin(origin) && originAllowed(origin, allowedOrigins);
+    const allowed = origin ? originAllowed(origin, allowedOrigins) : false;
 
     if (allowed) ctx.set('Access-Control-Allow-Origin', origin);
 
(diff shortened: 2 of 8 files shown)


Tonours (commented):
removed

packages/agent-bff/src/cors/cors-middleware.ts
```diff
@@ -17,9 +17,9 @@ export default function createCorsMiddleware({
   return async function corsMiddleware(ctx, next) {
     const origin = ctx.get('Origin');
 
-    if (origin) ctx.vary('Origin');
+    if (hasOrigin(origin)) ctx.vary('Origin');
 
-    const allowed = origin ? originAllowed(origin, allowedOrigins) : false;
+    const allowed = hasOrigin(origin) && originAllowed(origin, allowedOrigins);
```
nbouliol: `ctx.get` only returns `''` when the header is absent and `originAllowed('')` is already `false`, so this rewrite is behaviour-identical and the guard on line 22 is dead — keep layer 1 untouched?

packages/agent-bff/src/cors/origin.ts
```diff
@@ -1,3 +1,7 @@
+export function hasOrigin(raw: string): boolean {
+  return raw !== '';
+}
```
nbouliol: `normalizeOrigin` next door takes `string | undefined | null` and trims; here `hasOrigin(undefined)` is `true` and `'  '` counts as present. Align the contract?

packages/agent-bff/README.md
```diff
@@ -142,8 +142,11 @@ normalized away, no trailing slash, no wildcard, no subdomain matching):
   blocks). Applies to `POST /oauth/token` too. Preflight `OPTIONS` from an allow-listed origin gets
   the allowed methods + headers; credentials are never enabled.
 - **Layer 2 (per-key authorization, Mode 2 only)** — when the resolved key has a non-empty
-  `allowedOrigins`, the request `Origin` must also be in that list (a missing `Origin` is rejected),
-  else `403 origin_not_allowed`. An empty per-key list is a no-op.
+  `allowedOrigins`, an `Origin` sent by the client must be in that list, else
+  `403 origin_not_allowed`. A request with no `Origin` at all — a server-side client such as curl,
+  a Node backend or CI, an empty header counting as none — passes: there the API key is the
+  boundary, not the origin. An opaque `Origin: null` (sandboxed iframe, cross-origin redirect) is a
+  present origin and is rejected. An empty per-key list is a no-op.
```
nbouliol, line 149: PRD-1102 also flagged the document (`bffApiKey`: "a BFF API key"); `openapi-document.ts:404` is still unchanged — worth a line there too?

nbouliol (commented):
Core fix looks right, and the deviation from the ticket's suggested `normalizeOrigin(...) !== null` is the better call — that snippet would have let `Origin: null` and garbage through, against the ticket's own Expected. Three non-blocking notes below.

packages/agent-bff/src/cors/cors-middleware.ts
Tonours: Reverted — layer 1 is back to its original two lines. You are right on both counts: `ctx.get` returns `''` only when the header is absent, and `originAllowed('')` is already false, so the rewrite bought nothing and only made the two layers look like they share a rule they deliberately do not. The whole diff is now the one condition in `per-key-origin.ts` plus the helper.

packages/agent-bff/src/cors/origin.ts
Tonours: Aligned:

```ts
export function hasOrigin(raw: string | undefined | null): boolean {
  return raw !== undefined && raw !== null && raw.trim() !== '';
}
```

Same input contract as `normalizeOrigin`, and the same trim. It changes one case: a whitespace-only `Origin` used to draw a 403 (`normalizeOrigin('  ')` is null, so `originAllowed` was false) and now passes as absent. That is the right side to land on — it is the same argument the PR already makes for the empty header: no browser serializes it (an opaque origin serializes to the literal string `null`, which stays rejected), and any client able to send `Origin:    ` can omit the header entirely.

Where the two helpers still differ is deliberate: `normalizeOrigin` returns null for `'null'` and for garbage, because those are *present* origins that must be gated. `hasOrigin` only answers "did the client send anything at all".

Test added: "treats a whitespace-only Origin header like an absent one and proceeds".

packages/agent-bff/README.md
Tonours: Added. `bffApiKey` was a single line that said nothing about origins, so a consumer reading the document alone could not know a key might 403 them from a browser and work from curl:

> Mode 2: a BFF API key. Never send both this and an Authorization header. A key created with allowedOrigins restricts browser callers only: an Origin the client sends must be in that list, else 403 origin_not_allowed, while a request with no Origin at all — curl, a server-side client, CI — passes, because there the key is the boundary rather than the origin. The opaque Origin null is a present origin and is rejected.

Pinned by a test so it cannot drift from the README ("should say in the api key scheme that allowedOrigins gates browsers only"), and the README now points at it.

packages/agent-bff/src/cors/origin.ts
Tonours: Correction to what I wrote above: the whitespace case never drew a 403, and it cannot.

Node's HTTP parser strips leading and trailing OWS from header values before koa sees them. Measured on this Node:

```
client sends  Origin: '   '
server reads  req.headers.origin === ''
```

So `Origin:    ` has always arrived as the empty string and has always taken the absent-header path — before this PR that meant 403, after it means 200, exactly like sending no header at all. My "used to draw a 403 and now passes" was wrong; the `trim()` changes nothing reachable over HTTP.

Adjusted rather than dropped. The route-level whitespace test is gone — it asserted the same runtime input as the no-Origin case one test above, so it proved nothing — and the contract is pinned where it actually lives, in the unit test:

```ts
it('is false for blank or absent input, the same contract normalizeOrigin takes', () => {
  expect(hasOrigin('   ')).toBe(false);
  expect(hasOrigin(undefined)).toBe(false);
  expect(hasOrigin(null)).toBe(false);
});
```

The widened signature and the trim stay for the reason you raised — the two helpers should take the same input contract — and that test is what stops the trim from being silently deleted later as dead code.

nbouliol (commented):
Round 1 all addressed — layer 1 reverted, `hasOrigin` widened, and the OpenAPI scheme now carries the rule with a test. Two follow-ups below, neither blocking. Note the red `ai-proxy` check is unrelated to this branch (live Anthropic API rejecting `thinking.type.disabled` for `claude-fable-5-1`).

packages/agent-bff/src/cors/origin.ts
```diff
@@ -1,3 +1,7 @@
+export function hasOrigin(raw: string | undefined | null): boolean {
+  return raw !== undefined && raw !== null && raw.trim() !== '';
+}
```
nbouliol: `normalizeOrigin` re-does this blank test on lines 6-9. What about `raw is string` here, then opening `normalizeOrigin` with `if (!hasOrigin(raw)) return null;` so the blank rule lives in one place?

packages/agent-bff/README.md
```diff
@@ -142,8 +142,12 @@ normalized away, no trailing slash, no wildcard, no subdomain matching):
...
   the allowed methods + headers; credentials are never enabled.
 - **Layer 2 (per-key authorization, Mode 2 only)** — when the resolved key has a non-empty
-  `allowedOrigins`, the request `Origin` must also be in that list (a missing `Origin` is rejected),
-  else `403 origin_not_allowed`. An empty per-key list is a no-op.
+  `allowedOrigins`, an `Origin` sent by the client must be in that list, else
+  `403 origin_not_allowed`. A request with no `Origin` at all — a server-side client such as curl,
+  a Node backend or CI, an empty header counting as none — passes: there the API key is the
+  boundary, not the origin. An opaque `Origin: null` (sandboxed iframe, cross-origin redirect) is a
+  present origin and is rejected. An empty per-key list is a no-op. The OpenAPI document says the
+  same on the `bffApiKey` security scheme, so a consumer reading the document alone learns it too.
```
nbouliol: The OpenAPI text omits the empty-header and empty-per-key-list rules, so "says the same" isn't quite true — and a doc asserting another doc matches will rot. Drop this sentence?

packages/agent-bff/src/cors/origin.ts
Tonours: Done: `hasOrigin` is a `raw is string` predicate and `normalizeOrigin` opens with `if (!hasOrigin(raw)) return null;`, so the blank rule lives in one place (origin.ts:1-6).

packages/agent-bff/README.md
Tonours: Dropped: the README no longer asserts the OpenAPI text matches, and the scheme description states the empty-header and empty-per-key-list rules itself.

nbouliol (approved):
Approved. The whole diff is one condition in `per-key-origin.ts` plus the helper, and header presence rather than the ticket's suggested `normalizeOrigin(...) !== null` is the right call — that snippet would have let the opaque `Origin: null` and malformed origins through, against the ticket's own Expected.

Checked the `normalizeOrigin` fold differentially rather than by eye: 0 divergences over 33 inputs, including whitespace-only, ` null `, IPv6, and non-special schemes. Layer 1 is untouched, and `apiKeyStep` still runs before the guard (`cli-core.ts:354-356`), so the relaxed path needs a resolved key.

The red `ai-proxy` check is unrelated to this branch — `claude-fable-5-1` rejecting `thinking.type.disabled` against the live Anthropic API. Worth its own ticket, like the `capacitor://`/`file://` allow-list gap you flagged.

PR #1862, opened 2026-08-31
feat(agent-bff): document permissions and openapi routes

Tonours:
## What

Registers the two live routes the served OpenAPI document was missing, so a consumer building from the document alone can fetch permission hints and discover the document endpoint. Both are registered before the generic/unfolded fork, so both document variants carry them.

- **`GET /agent/v1/permissions`** — the full `crud` + `actions` response schema (`PermissionHints`, `ActionHints`, `CrudHints`), the `collections` query filter, the shared `X-Forest-Timezone` header parameter, both auth modes, and the display-hint contract: hints gray out buttons, they never authorize.
- **`GET` / `HEAD /agent/openapi.json`** — self-documenting, with a typed `404 openapi_disabled`.

fixes PRD-1107

## What the document now says that it did not

Three claims were wrong or misleading before review and are corrected here — the whole value of this PR is that a consumer can trust the document, so they matter more than the routes themselves:

- **The hints lag; the agent does not.** The BFF caches hints for 15 minutes. The agent runs with `instantCacheRefresh` by default (`options-validator.ts`, `?? true`) and invalidates on Forest events, so it enforces a new permission while these hints are still stale. Only an agent explicitly configured `instantCacheRefresh: false` caches 15 minutes of its own, independently — and then the two can disagree in either direction until both expire. Both cases are stated, and both are pinned by tests.

- **A 503 here carries `Retry-After` on two of its three types.** `permissions_unavailable` and `key_resolution_unavailable` always set it; only `schema_unavailable` does not — a client backing off on a header that is not there would have looped immediately.
- **The timezone middleware does not cover every `/agent` route.** `createOpenApiRoutes`, the context routes and the AI relay are all mounted before it. The permissions route is behind it, which is why it 400s `missing_timezone` on a deployment with no default — the description now says "unlike the context and document routes" instead of "like every route under `/agent`".
- **`crud` is not a role fact on a development environment.** `canRolePerform` short-circuits there and all six flags come back true whatever the role. `ActionHints` already said this; `CrudHints` now does too.

Smaller accuracy fixes: `collections` is published as `string | string[]` because a repeated parameter is joined and re-split, so both forms are one request; and the shared `Error503` / `Retry-After` descriptions name the failure *class* (the BFF could not reach the Forest server) rather than the single instance they listed.

## Why `/health` stays out

PRD-1107 accepts either outcome — "documented, **or** a deliberate decision is recorded that they are intentionally out of the agent contract" — and its regression check is "at minimum `/agent/v1/permissions`".

The decision is the boundary, not tooling: this document describes the auth-gated `/agent` surface, and `/health` is answered in `bff-http-server` before CORS and before the auth edge. A consumer generating a client from this document should not get a `/health` method that plays by different rules than every other one it holds.

`/oauth/*` is out for the same reason, and so is **`GET /docs`** — the unauthenticated viewer that renders this very document, which was neither described nor recorded before. `info.description` records all three.

## Scope and safety

Documentation only. No middleware, mounting condition, status or body change. A drift guard in `test/permissions/build-permission-hints.test.ts` parses a real `buildPermissionHints` output with the documented schema, so the documented shape cannot silently diverge from the runtime one.

## How to test

```
yarn workspace @forestadmin/agent-bff test   # redocly lint included
yarn workspace @forestadmin/agent-bff lint && yarn workspace @forestadmin/agent-bff build
```

### General

### Security


packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +349,95 @@ const UNFOLDED_DESCRIPTION =
...
+function registerDocumentPath(registry: OpenAPIRegistry, errorRefs: ErrorResponseRefs): void {
+  const disabled = registerErrorComponent(registry, OPENAPI_DISABLED_COMPONENT, {
+    description:
+      'The deployment runs with `BFF_OPENAPI_ENABLED=false`, so the document is not served over ' +
+      'HTTP. The body is typed `openapi_disabled`; a bare 404 with no typed body means the agent ' +
+      'edge is not mounted at all.',
+    schema: { $ref: ERROR_RESPONSE_REF },
+  });
+
+  registry.registerPath({
```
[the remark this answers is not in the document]
Tonours: Fixed in ec8405a: `registerDocumentPath` now also registers a dedicated `head` operation (`headOpenApiDocument`) with the same security schemes, the same error responses including the typed `404 openapi_disabled`, and a body-less 200 — matching the runtime, where GET and HEAD share one handler. Covered by a test pinning the HEAD operation inventory and its 404 ref to the GET one.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -321,6 +326,12 @@ const SHARED_DESCRIPTION =
   'like JSON, while any other content type is read as absent, which silently drops any filter, ' +
   'sort, or page.';
 
+const SURFACE_DESCRIPTION =
```
nbouliol, line 381: No `redocly.yaml` here, so `operation-4xx-response` is only fatal because `openapi-spec-validity.test.ts:50-53` asserts zero warnings — a per-operation override would let `/health` in with its real 200/503, which is what the ticket asked for.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +349,117 @@ const UNFOLDED_DESCRIPTION =
...
+    path: `${ROUTE_PREFIX}/permissions`,
+    operationId: 'getPermissionHints',
+    summary: 'Read what the caller may see and do, as display hints',
+    description:
+      'Answers for the caller behind the credentials, so two callers get two different payloads. ' +
+      'The hints come from the Forest permissions, cached for ' +
+      `${
+        PERMISSIONS_CACHE_TTL_MS / 60_000
+      } minutes: a permission changed in Forest can take that ` +
+      'long to show up here, while the agent enforces the new one immediately. A 503 means the ' +
```
nbouliol: The agent's own permission cache defaults to the same 15 minutes (`options-validator.ts:74`, `DEFAULT_MINIMUM_CACHE_DURATION * 15`), so enforcement isn't immediate unless `instantCacheRefresh` is on — could we say the two caches can disagree in either direction instead?

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +349,117 @@ const UNFOLDED_DESCRIPTION =
...
+      'permissions could not be fetched and no fresh cache was left (type ' +
+      '`permissions_unavailable`); it carries Retry-After. Like every route under `/agent`, this ' +
```
nbouliol: `createOpenApiRoutes`, the context routes and the AI relay are all registered before `createTimezoneMiddleware` in `cli-core.ts`, so "every route under `/agent`" isn't accurate — what about "unlike the context and document routes, this one sits behind the timezone middleware"?

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +349,117 @@ const UNFOLDED_DESCRIPTION =
...
+      'one goes through the timezone middleware even though it reads no timezone, so a deployment ' +
+      'with no configured default answers 400 `missing_timezone` unless `X-Forest-Timezone` is ' +
+      'sent.',
+    security: SECURITY,
+    request: {
+      query: z.object({
+        collections: z
+          .string()
```
nbouliol: `resolveRequestedCollections` joins a repeated `collections` param (`Array.isArray(raw) ? raw.join(',')`), so `z.union([z.string(), z.array(z.string())])` would match runtime.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +349,117 @@ const UNFOLDED_DESCRIPTION =
...
+      200: {
+        description: 'The display hints for the caller role',
+        content: { 'application/json': { schema: PermissionHintsSchema } },
+      },
+      400: errorRefs.byStatus['400'],
+      401: errorRefs.byStatus['401'],
+      403: errorRefs.byStatus['403'],
+      500: errorRefs.byStatus['500'],
+      501: errorRefs.byStatus['501'],
+      503: errorRefs.byStatus['503'],
```
nbouliol, line 466: The shared `Error503` and `Retry-After` descriptions only mention the unresolvable API key, but this route's 503 is `permissions_unavailable` — could we widen both, or register a dedicated component the way `Error404OpenapiDisabled` does?

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +349,117 @@ const UNFOLDED_DESCRIPTION =
...
+    },
+  });
+}
+
+function registerDocumentPath(registry: OpenAPIRegistry, errorRefs: ErrorResponseRefs): void {
+  const disabled = registerErrorComponent(registry, OPENAPI_DISABLED_COMPONENT, {
```
nbouliol: Every other error component is registered by `registerErrorResponses`; this is the only path-registering function that also mutates the component registry — move it next to its siblings and pass the ref in?

nbouliol (commented):
Reviewed the documented contract against the runtime. Two descriptions state something the code does not do (the agent-side cache, and the timezone middleware covering every `/agent` route) — those are worth fixing before merge since the whole value of the PR is that a consumer can trust the document. The rest are smaller accuracy points.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -280,6 +281,62 @@ export const ContextResponseSchema = z
...
+    browse: z.boolean(),
+    read: z.boolean(),
+    edit: z.boolean(),
+    add: z.boolean(),
+    delete: z.boolean(),
+    export: z.boolean(),
+  })
+  .openapi('CrudHints', {
+    description:
+      'The six collection-level rights of the caller role, as Forest states them. `browse` false ' +
```
nbouliol, line 431: On a development environment `canRolePerform` short-circuits and all six flags are true whatever the role — `ActionHints` says this, `CrudHints` should too, or a client reads an all-true block as a role fact.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: You are right that the lint failure is self-imposed — `openapi-spec-validity.test.ts` asserting zero warnings is what makes `operation-4xx-response` fatal, and an override would lift it.

Keeping `/health` out anyway, and rewriting the reason rather than the decision. PRD-1107 accepts either outcome explicitly: "The three routes are documented **or a deliberate decision is recorded that they are intentionally out of the agent contract**", and its regression check is "at minimum `/agent/v1/permissions`". So the ticket is satisfied either way, and the choice is a boundary question, not a tooling one.

The boundary: this document describes the auth-gated `/agent` surface. `/health` is answered in `bff-http-server` before CORS and before the auth edge — it is not part of that surface, and a consumer generating a client from this document should not get a `/health` method that behaves under different rules than every other one it holds. `/oauth/*` is out for the same reason.

What was wrong was giving redocly as the reason, which made a deliberate boundary look like a tooling workaround. `info.description` says it plainly now, and the PR description no longer leans on the lint rule.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Fixed — and it was worse than a missing nuance, since "the agent enforces the new one immediately" is simply false by default. Both caches are 15 minutes (`options-validator.ts:74`, `DEFAULT_MINIMUM_CACHE_DURATION * 15`), and they are independent, so they expire at unrelated moments.

> The agent caches its own copy for the same 15 minutes by default, and the two are independent: after a change in Forest these hints and what the agent enforces can disagree in either direction until both expire. An agent running with `instantCacheRefresh` enforces immediately while these hints stay stale.

"Either direction" is the load-bearing part: a right that was granted can be visible here and still refused by the agent, and a right that was revoked can be hidden here and still enforced — a client that assumed one direction would build the wrong fallback. Pinned by a test.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Fixed, with your wording. `cli-core.ts:356-365` puts `createOpenApiRoutes`, the context routes and the AI relay before `createTimezoneMiddleware`, so "like every route under `/agent`" was false for three of them.

> Unlike the context and document routes, this one sits behind the timezone middleware even though it reads no timezone, so a deployment with no configured default answers 400 `missing_timezone` unless `X-Forest-Timezone` is sent.

The test asserts both the new clause and the absence of the old one.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Done — `z.union([z.string(), z.array(z.string())])`, so the published schema accepts both forms the runtime accepts. `resolveRequestedCollections` joins a repeated parameter with commas and splits it again, so `?collections=a&collections=b` and `?collections=a,b` are the same request; the description says so now, and the test asserts the emitted `anyOf`.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Widened both rather than registering a fourth component — the two shared descriptions were simply incomplete, and a dedicated `Error503Permissions` would duplicate a body that is identical in shape.

- **503**: "…the API key could not be resolved, **or the Forest permissions could not be fetched and no fresh cache was left (type `permissions_unavailable`)**".
- **Retry-After**: "Set when the BFF could not reach the Forest server — an unresolvable API key, or permissions it could not fetch."

The second is the more useful correction: it now names the *class* of failure rather than one instance, so the next 503 that sets the header does not require a third edit.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Moved. `Error404OpenapiDisabled` is registered inside `registerErrorResponses` with its siblings and travels on `ErrorResponseRefs` as `openapiDisabled`; `registerDocumentPath` now only reads a ref, like every other path-registering function.

That also removes an ordering hazard the old shape carried: the component only existed if `registerDocumentPath` happened to run, so any future caller registering the path conditionally would have produced a document with a dangling `$ref`.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Added, in the same words as `ActionHints`:

> On a development environment all six are true whatever the role, so an all-true block is not a statement about the role.

`canRolePerform` short-circuits on a development environment, so a client reading `crud` as a role fact would gray out nothing in dev and everything would look permitted — exactly the misreading the `ActionHints` sentence already guards against. Pinned by a test.

Tonours:
Correction — my previous push got the cache claim backwards, and the review pass caught it.

`instantCacheRefresh` defaults to **true** (`options-validator.ts:44`, `?? true`), and when it is on the agent takes `DEFAULT_CACHE_DURATION_WITH_EVENTS` and invalidates on Forest events. The 15-minute fallback only applies to an agent explicitly configured `instantCacheRefresh: false`. So the original text was right by default, and "the agent caches its own copy for the same 15 minutes by default" — which I wrote in response to the thread above — was wrong.

What it says now:

> …cached for 15 minutes, so they lag a change made in Forest. The agent does not: it runs with `instantCacheRefresh` by default and invalidates on Forest events, so it enforces the new permission while these hints are still stale. An agent explicitly configured with `instantCacheRefresh: false` caches for 15 minutes of its own, independently, and the two can then disagree in either direction until both expire.

Both halves are pinned by tests now, the default and the non-default one, so neither can be flipped again without a red suite.

**Two more published claims corrected on the same push:**

- **The 503 does not always carry Retry-After.** Only `permissions_unavailable` does. The same route also answers `schema_unavailable` (via `agent-route-helpers.ts`) and `key_resolution_unavailable` from the API-key step, neither of which sets the header. A client backing off on a header that is not there would have looped immediately. All three types are named now.
- **`/docs` was missing from the surface statement.** `cli-core.ts` mounts `createDocsRoutes`, which serves `GET /docs` and its bundle **unauthenticated** — a live public route the document neither described nor recorded as excluded, which is the gap PRD-1107 is about. It is recorded with `/health` and `/oauth/*`.

Also: the HEAD description said a 404 "carries the same `openapi_disabled` type as the GET", two sentences after saying HEAD answers with no body — koa ends a HEAD response without one, so the type is unobservable there. It now says to read the status, not the type. And the comment added to `build-permission-hints.test.ts` is gone; the test name already carried it.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -321,6 +338,13 @@ const SHARED_DESCRIPTION =
   'like JSON, while any other content type is read as absent, which silently drops any filter, ' +
   'sort, or page.';
 
+const SURFACE_DESCRIPTION =
+  'Only the auth-gated `/agent` surface is described here. Three live route families are ' +
+  'deliberately absent because they sit outside that surface and outside its auth edge: ' +
+  '`GET /health`, which answers 200 `ok` or 503 `degraded` to an unauthenticated probe; the ' +
+  '`/oauth/*` login routes; and `GET /docs`, the unauthenticated viewer that renders this very ' +
+  'document. The package README documents all three.';
```
nbouliol: `README.md:24` covers `/health` and `/oauth/*` only — there's no `/docs` row — so "all three" doesn't hold yet; while here, `/health` and `/docs` both answer `HEAD` too, and `/docs/redoc.standalone.js` is a second public route.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +362,116 @@ const UNFOLDED_DESCRIPTION =
...
+    summary: 'Read what the caller may see and do, as display hints',
+    description:
+      'Answers for the caller behind the credentials, so two callers get two different payloads. ' +
+      'The hints come from the Forest permissions, cached for ' +
+      `${
+        PERMISSIONS_CACHE_TTL_MS / 60_000
+      } minutes, so they lag a change made in Forest. The agent does not: it runs with ` +
+      '`instantCacheRefresh` by default and invalidates on Forest events, so it enforces the new ' +
+      'permission while these hints are still stale. An agent explicitly configured with ' +
+      '`instantCacheRefresh: false` caches for 15 minutes of its own, independently, and the two ' +
```
nbouliol: With `instantCacheRefresh` on, the agent's TTL becomes `DEFAULT_CACHE_DURATION_WITH_EVENTS` (~1 year) and freshness rests entirely on the SSE stream, which `events-subscription/index.ts` logs a dedicated error for when a reverse proxy swallows it — there the hints are the fresher of the two, not the staler. Also `permissionsCacheDurationInSeconds` is configurable (floor 60s), so the `false` branch is a default rather than a fixed 15 minutes.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +362,116 @@ const UNFOLDED_DESCRIPTION =
...
+      'can then disagree in either direction until both expire. A 503 here is either ' +
+      '`permissions_unavailable` — they could not be fetched and no fresh cache was left, and it ' +
+      'carries Retry-After — or `schema_unavailable` / `key_resolution_unavailable`, which do ' +
```
nbouliol: `keyResolutionUnavailable` takes a required `retryAfter` and all three call sites pass one (`api-key-authenticator.ts:35,45,47`), so this 503 always carries the header — only `schema_unavailable` doesn't, and line 64 right above already says the opposite.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +362,116 @@ const UNFOLDED_DESCRIPTION =
...
+      'not. Unlike the context and document routes, this one sits behind the timezone middleware ' +
```
nbouliol: The AI relay sits ahead of the timezone middleware too (`cli-core.ts:364-365`) and `/agent/v1/ai/query` is in this document — the PR body already lists it, so "unlike the context, document and AI-query routes"?

nbouliol (commented):
Re-reviewed at 7b77508. Four of the seven round-1 points landed clean: the timezone contrast, `collections` repeatability, the dev-mode `CrudHints` warning and moving the `Error404OpenapiDisabled` registration. My round-1 comment on the agent cache was wrong on one point and you were right to correct it — `options-validator.ts:44` defaults `instantCacheRefresh` to `true`, which I missed by reading line 74 alone.

The remaining findings are all in the replacement prose rather than the routes. The `Retry-After` split is the one to fix before merge: it contradicts the header description edited in the same commit, and the PR body repeats it, so a consumer implementing backoff would skip the one 503 that always tells it how long to wait.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +363,127 @@ const UNFOLDED_DESCRIPTION =
...
+    description:
+      'Answers for the caller behind the credentials, so two callers get two different payloads. ' +
+      'The hints come from the Forest permissions, cached for ' +
+      `${
+        PERMISSIONS_CACHE_TTL_MS / 60_000
+      } minutes, so they lag a change made in Forest. The agent runs with ` +
+      '`instantCacheRefresh` by default, so its own cache has no meaningful expiry — about a year — ' +
+      'and freshness rides on the Forest event stream: as long as those events reach it, the agent ' +
+      'enforces the new permission while these hints are still stale. If the stream is cut — a ' +
+      'reverse proxy that swallows it, which the agent logs — the agent can hold the old permission ' +
```
nbouliol: The heartbeat listener is `{ once: true }` and `detectBuffering` runs once at subscribe, so this only fires when the first heartbeat never arrives — "logs at startup", or drop the parenthetical rather than implying continuous detection?

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +363,127 @@ const UNFOLDED_DESCRIPTION =
...
+      'far longer than these hints, and the hints are then the fresher of the two. An agent ' +
+      'explicitly configured with `instantCacheRefresh: false` caches for ' +
+      '`permissionsCacheDurationInSeconds` — 15 minutes by default, configurable with a 60-second ' +
+      'floor — independently of these hints, and the two can then disagree in either direction ' +
```
nbouliol: With `instantCacheRefresh: false` the agent refetches on a denial (`action-permission.ts:48-56`), so a newly granted permission lands on the next call and only a revocation lingers — the disagreement is one-directional there.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +363,127 @@ const UNFOLDED_DESCRIPTION =
...
+      'until both expire. A 503 here is `permissions_unavailable` or ' +
+      '`key_resolution_unavailable`, which always carry Retry-After, or `schema_unavailable`, ' +
+      'which does not. Unlike the context, document and AI-query routes, this one sits behind the ' +
```
nbouliol: `hasAiQueryRoute` defaults to false and `registerAiQueryPath` is conditional (`:577`), so on those deployments this points at a path the document doesn't carry — "and the AI-query relay where published"?

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +363,127 @@ const UNFOLDED_DESCRIPTION =
...
+      401: errorRefs.byStatus['401'],
+      403: errorRefs.byStatus['403'],
+      500: errorRefs.byStatus['500'],
+      501: errorRefs.byStatus['501'],
+      503: errorRefs.byStatus['503'],
+    },
+  });
+}
+
+function headDocumentResponses(): Record<string, { description: string }> {
```
nbouliol: A parameterless function returning a static literal, called once — a module const beside `ERROR_STATUSES` and `RETRY_AFTER_HEADER` matches the file's convention and removes the ordering constraint the commit title hints at.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +363,127 @@ const UNFOLDED_DESCRIPTION =
...
+  return {
+    200: { description: 'The document exists and is readable; no body is returned for HEAD' },
+    400: { description: ERROR_STATUSES['400'] },
+    401: { description: ERROR_STATUSES['401'] },
+    403: { description: ERROR_STATUSES['403'] },
```
nbouliol: These reuse descriptions written for a JSON-bodied response — `['403']` says "the body carries the approving roles" — while the 404 was reworded for exactly that reason; one "read the status, not the type" caveat on the HEAD description would cover all five.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +363,127 @@ const UNFOLDED_DESCRIPTION =
...
+    404: {
+      description:
+        'The deployment runs with `BFF_OPENAPI_ENABLED=false`, so the document is not served ' +
```
nbouliol: Word-for-word from `OPENAPI_DISABLED_COMPONENT` at `:110`, and the comment at `:97-99` argues against inlining — a shared const both concatenate, plus a line saying why HEAD is the exception?

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +363,127 @@ const UNFOLDED_DESCRIPTION =
...
+        'over HTTP. The status is the same as the GET one, but the typed `openapi_disabled` body ' +
+        'never comes back on a HEAD: read the status, not the type.',
+    },
+    500: { description: ERROR_STATUSES['500'] },
+    503: { description: ERROR_STATUSES['503'] },
```
nbouliol: Dropping the `$ref`s also dropped the `Retry-After` the `Error503` component carried (`:124`), but line 479 still promises header parity and the runtime does send it (`error-middleware.ts:30` is method-agnostic) — `headers: RETRY_AFTER_HEADER` here, with content still absent?

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +363,127 @@ const UNFOLDED_DESCRIPTION =
...
+  };
+}
+
+function registerDocumentPath(registry: OpenAPIRegistry, errorRefs: ErrorResponseRefs): void {
+  const disabled = errorRefs.openapiDisabled;
```
nbouliol: Single use now that HEAD's 404 is inline — inline `errorRefs.openapiDisabled` at the GET and drop the local?

packages/agent-bff/test/openapi/openapi-document.test.ts
```diff
@@ -453,6 +471,179 @@ describe('generateOpenApiDocument', () => {
...
+
+    expect(description).toContain('`instantCacheRefresh` by default');
+    expect(description).toContain('enforces the new permission while these hints are still stale');
+    expect(description).toContain('freshness rides on the Forest event stream');
+    expect(description).toContain('the hints are then the fresher of the two');
+    expect(description).toContain('`permissionsCacheDurationInSeconds`');
+    expect(description).toContain('60-second floor');
+    expect(description).toContain('disagree in either direction');
+    expect(description).toContain('Unlike the context, document and AI-query routes');
+    expect(description).not.toContain('Unlike the context and document routes');
```
nbouliol: This can't fail while `:501` passes — it pins the previous round's wording rather than a behaviour.

packages/agent-bff/test/openapi/openapi-document.test.ts
```diff
@@ -453,6 +471,179 @@ describe('generateOpenApiDocument', () => {
...
+      get: { security: unknown; description: string };
+    };
+
+    expect(path.get.security).toEqual([{ bffSession: [] }, { bffApiKey: [] }]);
+    expect(path.get.description).toContain('`HEAD` is served identically');
+  });
+
+  it('should document HEAD on the document path, so a generated client can probe before fetching', () => {
+    const path = (document.paths ?? {})[DOCUMENT_PATH] as {
+      get: { responses: Record<string, unknown> };
```
nbouliol: `path.get` lost its only consumer when the `$ref` comparison went — droppable from the cast.

nbouliol (commented):
Re-reviewed at fc97225e. All four round-2 findings are genuinely fixed, and I checked the numbers rather than the prose: `instantCacheRefresh ?? true`, `DEFAULT_CACHE_DURATION_WITH_EVENTS` = 31560000s (~365 days), the 900s default and the 60s clamp all hold; both 503 constructors take a required `retryAfter`, so "always carry Retry-After" is now literally true and the three-type list is exhaustive for the route; `/health`, `/docs` and its bundle all answer GET+HEAD, oauth is only `authorize` + `token`, and nothing else is mounted outside `/agent` — "three live route families" is an accurate count and the README now covers all three.

One regression to fix before merge: the commit that stopped HEAD advertising bodies also dropped the `Retry-After` its 503 `$ref` carried, so the operation now promises header parity it no longer expresses. The rest are smaller accuracy and structure points.

packages/agent-bff/test/openapi/openapi-document.test.ts
```diff
@@ -453,6 +471,179 @@ describe('generateOpenApiDocument', () => {
...
+      '500',
+      '503',
+    ]);
+    expect(path.head.responses['200'].content).toBeUndefined();
+    expect(path.head.responses['200'].description).toContain('no body');
+
+    for (const status of ['400', '401', '403', '404', '500', '503']) {
+      expect(path.head.responses[status].content).toBeUndefined();
+      expect(path.head.responses[status].$ref).toBeUndefined();
+      expect(path.head.responses[status].description).toBeTruthy();
```
nbouliol: The loop and the `content`/`$ref` checks are right, but `toBeTruthy` passes on any non-empty string — comparing against the GET component's description is what "same statuses" means, and it would have caught the missing `Retry-After`.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Fixed: README.md:24-25 now lists `/docs` (page and `redoc.standalone.js` bundle) alongside `/health` and `/oauth/*`, and says `/health` and `/docs` answer HEAD.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Rewritten: the description now says the agent's own TTL is about a year under `instantCacheRefresh`, that freshness rides on the event stream and the hints are then the fresher of the two when a proxy swallows it, and that `permissionsCacheDurationInSeconds` is a configurable default with a 60-second floor.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Fixed: `permissions_unavailable` and `key_resolution_unavailable` always carry Retry-After, `schema_unavailable` does not — that is what the prose says now.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Fixed: the sentence reads "Unlike the context and document routes, and the AI-query relay where it is published".

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Reworded: "which the agent logs once at subscribe time when the first heartbeat never arrives", so it no longer implies continuous detection.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Fixed, one-directional now: a newly GRANTED permission lands on the next call because the agent refetches on a denial; only a REVOKED one lingers agent-side until the cache expires.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Fixed: "and the AI-query relay where it is published".

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Done: `HEAD_DOCUMENT_RESPONSES` is a module const beside `ERROR_STATUSES` and `RETRY_AFTER_HEADER`, so the ordering constraint is gone.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Added once on the HEAD operation description: every status below reuses the GET wording, which describes a JSON body no HEAD returns — read the status, not the type. The 404 no longer repeats it.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Extracted `OPENAPI_DISABLED_DESCRIPTION`; the component and the HEAD 404 both concatenate it, each adding its own tail. The component note about inlining stands — this is one shared string, not an inlined component.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Fixed: the HEAD 503 carries `headers: RETRY_AFTER_HEADER`, content still absent, so header parity holds.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Inlined: `errorRefs.openapiDisabled` at the GET, local dropped.

packages/agent-bff/test/openapi/openapi-document.test.ts
Tonours: Dropped: that assertion pinned the previous wording. The test now asserts the claims that can actually break — "only a REVOKED one lingers agent-side" and "the AI-query relay where it is published".

packages/agent-bff/test/openapi/openapi-document.test.ts
Tonours: Dropped from the cast.

packages/agent-bff/test/openapi/openapi-document.test.ts
Tonours: Replaced: a new test compares each HEAD status against the resolved GET component — same description, same headers, no content — which is what "same statuses and headers" means, and it is what caught the missing Retry-After.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -338,6 +373,113 @@ const UNFOLDED_DESCRIPTION =
...
+
+function registerDocumentPath(registry: OpenAPIRegistry, errorRefs: ErrorResponseRefs): void {
+  registry.registerPath({
+    method: 'get',
+    path: DOCUMENT_PATH,
+    operationId: 'getOpenApiDocument',
+    summary: 'Read this document',
+    description:
+      'Serves this very document, behind the same credentials as every other `/agent` route: ' +
+      'without them the schema it describes is not readable. `HEAD` is served identically. The ' +
```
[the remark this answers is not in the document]
Tonours: Good catch. Reworded rather than taken verbatim, to keep the original point (the document is credential-gated) and add the exception: "Both auth modes reach it — a session or an API key — as they do on every data, action, context and permissions route; the AI-query relay is the one exception, and takes a session only." A test now pins that against the relay's actual `security` block.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -344,6 +379,115 @@ const UNFOLDED_DESCRIPTION =
...
+  registry.registerPath({
+    method: 'get',
+    path: DOCUMENT_PATH,
+    operationId: 'getOpenApiDocument',
+    summary: 'Read this document',
+    description:
+      'Serves this very document behind credentials, like every other `/agent` route: without ' +
+      'them the schema it describes is not readable. Both auth modes reach it — a session or an ' +
+      'API key — as they do on every data, action, context and permissions route; the AI-query ' +
+      'relay is the one exception, and takes a session only. `HEAD` is served identically. The ' +
```
nbouliol: Same conditional as `:581`, so an OAuth-less deployment names an exception its document doesn't carry — the permissions prose already hedges this at `:411`, worth the same wording here.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -344,6 +379,115 @@ const UNFOLDED_DESCRIPTION =
...
+
+  registry.registerPath({
+    method: 'head',
+    path: DOCUMENT_PATH,
+    operationId: 'headOpenApiDocument',
+    summary: 'Probe this document without fetching it',
+    description:
+      'Same route as `GET`, answering the same statuses and headers with no body, so a client can ' +
+      'probe whether the document is enabled and reachable before re-fetching it. Every status ' +
+      'below reuses the GET wording, which describes a JSON body no HEAD ever returns: read the ' +
```
nbouliol: The 200 and 404 have their own wording (`:78`, `:82`) and the 404 deliberately drops the typed-body sentence — the claim holds for the five error statuses that do reuse GET, but not for those two.

packages/agent-bff/test/openapi/openapi-document.test.ts
```diff
@@ -21,12 +24,15 @@ const responseComponents = (document.components?.responses ?? {}) as unknown as
   ResolvedResponse
 >;
 
-function responsesOf(path: string): Record<string, ResolvedResponse> {
+function responsesOf(path: string, method?: 'head'): Record<string, ResolvedResponse> {
```
nbouliol: `method?: 'get' | 'post' | 'head'` would match what the body actually does.

nbouliol (commented):
Re-reviewed at 098eae2e, against the new merge-base after main came in. All eleven round-3 findings are genuinely fixed, and I checked the two things most likely to have gone wrong rather than taking the commits at their word:

- `HEAD_DOCUMENT_RESPONSES` becoming a shared module object is safe — `generatePath` rebuilds `responses` through `mapValues` into a fresh object and `getResponse` uses `__rest` + `Object.assign({}, …)`, so nothing writes back into it. No cross-document contamination.
- The declared status sets are still complete after the merge: 415 only fires in `createJsonOnlyGuard` for POST/PUT/PATCH with a body, 502/504 belong to the AI relay, and an upstream 429 during key resolution maps to the 503 `key_resolution_unavailable` that is already declared.
- The refetch-on-denial and startup-log claims both hold, and the refetch one covers `crud` as well as `actions` since `buildPermissionHints` derives both from `ActionPermissions`.

Two prose points left, both in sentences round 4 itself added. Nothing blocking.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -365,6 +402,116 @@ const UNFOLDED_DESCRIPTION =
...
+      headers: timezoneHeader,
+    },
+    responses: {
+      200: {
+        description: 'The display hints for the caller role',
+        content: { 'application/json': { schema: PermissionHintsSchema } },
+      },
+      400: errorRefs.byStatus['400'],
+      401: errorRefs.byStatus['401'],
+      403: errorRefs.byStatus['403'],
```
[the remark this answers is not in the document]
Tonours: Fixed in 445b6a6ef — and on the document route too, which is the same defect one line over. `createOpenApiRoutes` sits after `createRateLimitMiddleware` in the same chain (`cli-core.ts:377-382`), so `GET` and `HEAD /agent/openapi.json` can answer 429 as well and neither documented it. All three now carry `429: errorRefs.byStatus['429']`, and the HEAD entry carries `Retry-After` like its GET counterpart.

Side effect worth naming: the HEAD description read "The five error statuses below reuse the GET wording". With 429 they are six, so the prose and its assertion moved too.

Each addition is mutation-checked on its own — dropping the permissions 429 fails 1 test, the document GET one 2, the HEAD one 2.

packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -460,4 +519,20 @@ export const MessagelessErrorResponseSchema = z
       'The unsupported-action-result body, which carries no message field unlike every other error.',
   });
 
+export const OpenapiDisabledErrorResponseSchema = z
+  .object({
+    error: z.object({
+      type: z.literal('openapi_disabled'),
+      status: z.literal(404),
+      message: z.string(),
+      details: z.unknown().optional(),
```
nbouliol: `openapiDisabled()` passes no options and `toErrorBody` only writes `details` when defined, so this one can never carry it — the pinned schema invites a client to look for a field the runtime never sends.

nbouliol (commented):
Re-reviewed at 445b6a6e, against the new merge-base after main came in a second time. All three round-4 points are fixed, and the two bot findings with them.

I checked the round-5 additions against the runtime rather than the prose, and they hold: the pinned 404 schema matches what `openapiDisabled()` actually emits (`message` always set, status always 404, and the disabled path really does go through the typed error since `agentScoped(createErrorMiddleware)` is mounted ahead of the agent middlewares); the README does say consumers branch on `error.type` and never on message text; `createRateLimitMiddleware` sits at `cli-core.ts:377-380`, ahead of both `createOpenApiRoutes` and the permissions route, and is path-scoped so HEAD hits it too, with `Retry-After` always set on these routes; `'429'` is in `DATA_ERRORS` so no variant can emit a dangling `$ref`; and the HEAD count ("six error statuses … the 200 and the 404 carry their own", eight declared) is right.

One small thing left, nothing blocking.

packages/agent-bff/src/openapi/schemas.ts
Tonours: Dropped in dc2527339. `details` is gone from the pinned schema, so `error` declares exactly `type`, `status`, `message` — the three fields `openapiDisabled()` really produces.

You are right on the mechanism: `bff-local-errors.ts:47-49` throws `new BffHttpError(404, 'openapi_disabled', message)` with no options, and `toErrorBody` only writes the key when it is defined (`bff-http-error.ts:57`), so the field could never appear. Pinning `type` to gain a discriminator while leaving an optional field the thrower cannot fill was working against itself — a client generated from it would carry a branch for a shape that never arrives.

The description now says so rather than leaving the absence to be noticed: it carries no `details`, unlike the shared envelope, because the thrower passes none. The test asserts the exact key list, not just the presence of the pinned values, so re-adding the field fails it — checked.

nbouliol (approved):
Approving at dc252733.

Six rounds, 26 comments, every one either fixed or answered with a rationale I accept. The two routes are documented, /health is out with the decision recorded in `info.description` rather than in a lint rule, and the drift guard in `build-permission-hints.test.ts` keeps the documented shape tied to the runtime one.

What sold it is that the contract text was checked against the code rather than asserted: the cache paragraph now matches `instantCacheRefresh`'s real default and its one-year TTL, the Retry-After split matches the three 503 constructors, the 429 matches the rate-limit middleware's position in the chain, and the pinned 404 declares exactly the three fields `openapiDisabled()` emits. I was wrong on the cache default in round 1 and the correction was right.

Two cosmetic points I raised and dropped: `z.literal(404)` renders `number` where its two sibling schemas render `integer`, and the 429 assertion sits in the test named for the 503. Neither blocks.

PR #1863, opened 2026-08-31
fix(agent-bff): expose search on unfolded request schemas

Tonours:
## What

The unfolded (per-collection) OpenAPI document exposes `search` and `searchExtended` on every list and count request. Generated clients were missing both inputs although the runtime honours them on all four path families — list, count, relation list, relation count.

- `ListRequest_<key>` and `CountRequest_<key>` carry `search` / `searchExtended` as `$ref`s to one shared `Search` / `SearchExtended` pair, registered through `ComponentPool.reuse` exactly like `Page` and `Timezone`.
- Relation requests inherit both through their `allOf` head.
- The relation descriptions are split by shape: list says "Filter, sort, projection and search", count says only "Filter and search". The previous shared string promised sort and projection to count, which carries neither.
- The count description no longer says "matching a filter and a search" — both keys are optional, and the useful statement is the one the generic document already made: it accepts the same search inputs as list, so a count can be trusted against the list it describes.

fixes PRD-1103

## Overlaps with #1855

That branch closes the request bodies (`additionalProperties: false`) and, to do it, had to publish `search` / `searchExtended` on the unfolded bodies too — a closed body that omits an input the runtime accepts would declare valid requests invalid. So the two PRs touch the same lines, deliberately, from opposite directions.

Whichever lands second will conflict on `registerRequests`. #1855 also **flattens the relation bodies out of their `allOf`**, so the "relations inherit through the `allOf` head" sentence above becomes stale once it merges — the inputs then arrive by spread instead, and the outcome for a consumer is identical.

## Known limitation

Same as the generic document: a collection that is not searchable still shows a documented `search` input, and using it answers 400 "Collection is not searchable". Gating on a searchable flag needs a signal the BFF capabilities do not carry today.

## How to test

- `yarn workspace @forestadmin/agent-bff test test/openapi` — redocly lint over the served document included
- Two consecutive builds of the unfolded document produce identical bytes.

### General

### Security

### DIFF
--- a/packages/agent-bff/src/openapi/unfolded-paths.ts
+++ b/packages/agent-bff/src/openapi/unfolded-paths.ts
@@ -21,6 +21,8 @@ import {
   OPERATORS,
   PageSchema,
   ParentIdSchema,
+  SearchExtendedSchema,
+  SearchSchema,
   SortClauseSchema,
   TimezoneSchema,
 } from './schemas';
@@ -265,19 +267,23 @@ function registerRequests(deps: Deps, plan: Omit<CollectionPlan, 'requests'>): R
   const { collection, key } = plan;
   const refs = fieldRefs(deps, plan);
   const timezone = pool.reuse('Timezone', TimezoneSchema);
+  const search = pool.reuse('Search', SearchSchema);
+  const searchExtended = pool.reuse('SearchExtended', SearchExtendedSchema);
 
   return {
     list: pool.add(`ListRequest_${key}`, {
       type: 'object',
       description: requestDescription(
         collection,
-        `Filter, sort and project records of ${quoted(collection.name)}.`,
+        `Filter, search, sort and project records of ${quoted(collection.name)}.`,
       ),
       properties: {
         filter: refs.filter,
         projection: { type: 'array', items: refs.projectable },
         sort: { type: 'array', items: refs.sort },
         page: pool.reuse('Page', PageSchema),
+        search,
+        searchExtended,
         timezone,
       },
     }),
@@ -285,9 +291,9 @@ function registerRequests(deps: Deps, plan: Omit<CollectionPlan, 'requests'>): R
       type: 'object',
       description: requestDescription(
         collection,
-        `Count records of ${quoted(collection.name)} matching a filter.`,
+        `Count records of ${quoted(collection.name)} matching a filter and a search.`,
       ),
-      properties: { filter: refs.filter, timezone },
+      properties: { filter: refs.filter, search, searchExtended, timezone },
     }),
   };
 }
@@ -346,18 +352,18 @@ function registerRelationRequests(
     properties: { parentId },
     required: ['parentId'],
   };
-  const description =
-    `Filter, sort and projection apply to ${quoted(foreign.collection.name)}, the foreign ` +
+  const describe = (inputs: string) =>
+    `${inputs} apply to ${quoted(foreign.collection.name)}, the foreign ` +
     `collection of ${quoted(plan.collection.name)}.${quoted(relation.name)}; the parent only ` +
     'resolves which records are related.';
 
   return {
     list: pool.add(`RelationListRequest_${relationKey}`, {
-      description,
+      description: describe('Filter, sort, projection and search'),
       allOf: [foreign.requests.list, parentProperties],
     }),
     count: pool.add(`RelationCountRequest_${relationKey}`, {
-      description,
+      description: describe('Filter and search'),
       allOf: [foreign.requests.count, parentProperties],
     }),
   };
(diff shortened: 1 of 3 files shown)


packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -319,7 +319,7 @@ const SHARED_DESCRIPTION =
   'request carrying neither with 400 missing_timezone, so send one of the two to be safe. Sending ' +
   'a content type other than application/json is not an error: a form-urlencoded body is parsed ' +
   'like JSON, while any other content type is read as absent, which silently drops any filter, ' +
-  'sort, or page.';
+  'search, sort, or page.';
```
nbouliol: `packages/agent-bff/README.md:14` still says "the filter, projection and paging all travel in the body" — worth naming search there too, for consistency with this and the 415 prose.

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -265,29 +267,33 @@ function registerRequests(deps: Deps, plan: Omit<CollectionPlan, 'requests'>): R
...
+        searchExtended,
         timezone,
       },
     }),
     count: pool.add(`CountRequest_${key}`, {
       type: 'object',
       description: requestDescription(
         collection,
-        `Count records of ${quoted(collection.name)} matching a filter.`,
+        `Count records of ${quoted(collection.name)} matching a filter and a search.`,
```
nbouliol: Both are optional, so "a filter and a search" reads more binding than it is — the generic `CountRequest` wording ("accepts the same search inputs as list") says the useful part.

nbouliol (commented):
Read against PRD-1103 and the repo standards. The fix itself is right: `search` / `searchExtended` land on the unfolded `ListRequest_*` / `CountRequest_*` as `$ref`s to shared components, the property order matches the generic `ListRequestSchema` / `CountRequestSchema` so the new parity test passes for the right reason, and relations inherit both through the `allOf` head. `ActionRequest` is the only other generic request shape and it has no missing input, so nothing else of this bug class is open. Two wording notes below.

packages/agent-bff/src/openapi/openapi-document.ts
Tonours: Added — `README.md:14` now reads "the filter, projection, search and paging all travel in the body". It was the first sentence describing the routes, so it was the worst place to omit it.

packages/agent-bff/src/openapi/unfolded-paths.ts
Tonours: Agreed, and taken verbatim from the generic one:

> Count records of "My Coll". Accepts the same search inputs as list, so a client can count exactly the rows its search returns.

"matching a filter and a search" read as a requirement on a body where both keys are optional — and it also failed to say the useful thing, which is that the two routes take the same inputs so a count can be trusted against the list it describes.

Pinned by a test asserting the new phrasing and the absence of the old one, since nothing else guards this string.

Tonours:
Review pass done — GO, no findings on the change itself. Every claim checked against the runtime: all four path families really do honour `search`/`searchExtended` (`request-schemas.ts`, both relation parsers spreading the same flat inputs, all four handlers reaching `applySearch`), the unfolded keys match the generic ones in order, and relations inherit through the `allOf` head.

It did surface two pre-existing omissions in the same prose I was already editing, so they are fixed here rather than left for a third pass:

- `README.md:14` listed "filter, projection and paging" — `sort` was missing too, not just `search`.
- The 415 dropped-inputs sentence listed "filter, search, sort, or page" and omitted `projection`, which a non-JSON content type drops exactly the same way.

Both now name every input that travels in the body.

One thing worth flagging for whoever merges: this PR and **#1855** touch the same lines from opposite directions. #1855 closes the request bodies with `additionalProperties: false`, and to do that it had to publish `search`/`searchExtended` on the unfolded bodies too — a closed body omitting an input the runtime accepts would declare valid requests invalid. It also flattens the relation bodies out of their `allOf`, so the "relations inherit through the `allOf` head" sentence in this description goes stale once it lands; the inputs then arrive by spread and a consumer sees no difference. Whichever merges second resolves `registerRequests`.

nbouliol (approved):
Approving. The fix is right where it matters: `search` / `searchExtended` land on the unfolded `ListRequest_*` / `CountRequest_*` as `$ref`s to one shared component pair registered like `Page` and `Timezone`, the property order matches the generic `ListRequestSchema` / `CountRequestSchema` so the new parity test passes for the right reason, and relations inherit both through the `allOf` head. `ActionRequest` is the only other generic request shape and it carries no missing input, so nothing else of this bug class is open.

Both wording notes addressed — the README sentence now names search (and sort), and the count description reuses the generic phrasing.

PR #1865, opened 2026-08-31
fix(agent-bff): document the default page size in the openapi document

Tonours:
## What

The `Page` and `ListResponse` component descriptions now say what a list request **without** `page` returns: the agent's default page. On the Node agent that is the first page (offset 0) with a limit of 15 records.

The text also says what the body does not: the response is not guaranteed to hold the whole collection, a response of exactly that length is probably truncated, and nothing in the payload flags it. So a consumer is told to send `page` and walk it.

Description strings only. `page` stays optional, no request is newly rejected, no wire change.

fixes PRD-1101

## Why

A consumer building a client from the document alone had no way to learn that omitting `page` silently truncates the list — it looked like "the collection", and for any collection over 15 rows it was not.

The number is attributed to the Node agent rather than stated flatly, because agent-bff can front other agent implementations with their own defaults. `count` is mentioned with its own caveat: a collection can deactivate it.

## How to test

```bash
cd packages/agent-bff && yarn test test/openapi
```

Redocly lint on the generic, unfolded and ai-less documents, plus assertions on the generated descriptions — `offset 0`, `limit of 15 records`, `not guaranteed to be the whole collection`, `probably truncated` — and a separate test naming the `limit`/`offset` requirement so it is discoverable by whoever changes it.

### General

### Security

### DIFF
--- a/packages/agent-bff/test/openapi/openapi-document.test.ts
+++ b/packages/agent-bff/test/openapi/openapi-document.test.ts
@@ -528,12 +528,17 @@ describe('the documented search inputs', () => {
   it('should warn that a search query can filter on a collection the BFF does not expose', () => {
     expect(schemas.Search.description).toContain('does not expose');
   });
+});
 
+describe('the documented pagination inputs', () => {
   it('should publish the default page applied when page is absent', () => {
     expect(schemas.Page.description).toContain('the object itself is optional');
     expect(schemas.Page.description).toContain('the first page (offset 0)');
     expect(schemas.Page.description).toContain('a limit of 15 records on the Node agent');
     expect(schemas.Page.description).toContain('silently missing');
+  });
+
+  it('should require both limit and offset once page is sent', () => {
     expect(schemas.Page.required).toEqual(['limit', 'offset']);
   });
 
--- a/packages/agent-bff/test/openapi/openapi-document.test.ts
+++ b/packages/agent-bff/test/openapi/openapi-document.test.ts
@@ -528,6 +528,20 @@ describe('the documented search inputs', () => {
   it('should warn that a search query can filter on a collection the BFF does not expose', () => {
     expect(schemas.Search.description).toContain('does not expose');
   });
+
+  it('should publish the default page applied when page is absent', () => {
+    expect(schemas.Page.description).toContain('the object itself is optional');
+    expect(schemas.Page.description).toContain('the first page (offset 0)');
+    expect(schemas.Page.description).toContain('a limit of 15 records on the Node agent');
+    expect(schemas.Page.description).toContain('silently missing');
+    expect(schemas.Page.required).toEqual(['limit', 'offset']);
+  });
+
+  it('should warn on the response that a page-less list is one page, not the collection', () => {
+    expect(schemas.ListResponse.description).toContain('not guaranteed to be the whole collection');
+    expect(schemas.ListResponse.description).toContain('up to 15 records from the start');
+    expect(schemas.ListResponse.description).toContain('probably truncated');
+  });
 });
 
 describe('serializeOpenApi', () => {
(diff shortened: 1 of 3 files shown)


packages/agent-bff/test/openapi/openapi-document.test.ts
```diff
@@ -528,6 +528,20 @@ describe('the documented search inputs', () => {
   it('should warn that a search query can filter on a collection the BFF does not expose', () => {
     expect(schemas.Search.description).toContain('does not expose');
   });
+
+  it('should publish the default page applied when page is absent', () => {
+    expect(schemas.Page.description).toContain('the object itself is optional');
+    expect(schemas.Page.description).toContain('the first page (offset 0)');
+    expect(schemas.Page.description).toContain('a limit of 15 records on the Node agent');
+    expect(schemas.Page.description).toContain('silently missing');
+    expect(schemas.Page.required).toEqual(['limit', 'offset']);
```
Tonours, line 542: Claude Opus 5 (claude-opus-5): Violates conventions — the `limit`/`offset`-required-when-`page`-is-sent contract has no discoverable test name: this assertion lives under a test named for `page` *absence*, so anyone relaxing `PageInput`'s required fields will not find a test named for the behavior they are changing. Rule: `testing.md#Test name states the exact behavior it asserts` ("the name reads as the behavior under test; when the assertion changes, the name must change with it") — four of the five assertions here are the page-absence behavior, this fifth is the opposite condition. Fix: split it into its own test, e.g. `should require both limit and offset once page is sent`.

Tonours (commented):
Spec (PRD-1101): conforms — the document now declares `page` optional and publishes the default page (first page, offset 0, limit of 15 records on the Node agent), which is the ticket's second acceptable outcome; its regression check ("omitting `page` either 400s or the document states the default page size") is satisfied on the document side, asserted against the generated document.

nbouliol (approved):
Description strings only, and every claim checks out against the runtime: offset 0 / limit 15 from DEFAULT_PAGE_TO_SKIP and DEFAULT_ITEMS_PER_PAGE, pagination forwarded only when `page` is set, and the `deactivated` count caveat. Relation list reuses the same `Page` and `ListResponse`, so it is covered too.

packages/agent-bff/test/openapi/openapi-document.test.ts
Tonours: Already split, in `2c427c9b5`. The page-absence assertions and the required-fields one are two named tests now:

```
should publish the default page applied when page is absent
should require both limit and offset once page is sent
```

Anyone relaxing `PageInput`'s required fields finds the second by name.

Tonours:
Review pass done, nothing to change. Since this PR is nothing but published claims, each one was checked against the code that produces the behaviour rather than against the ticket:

- **No pagination forwarded when `page` is absent** — `agent-query.ts:165`, `if (body.page)` is the only writer of `page[...]`, and both list and relation list go through `buildListAgentQuery`.
- **offset 0 / limit 15** — `packages/agent/src/utils/query-string.ts:21-22` (`DEFAULT_PAGE_TO_SKIP`, `DEFAULT_ITEMS_PER_PAGE`) resolving to `new Page(0, 15)`, applied from `routes/access/list.ts` and `list-related.ts`.
- **"nothing in the body says so"** — `response-mappers.ts:23-43` emits `countStatus` and no length or total marker, so there really is no truncation signal.
- **"up to 15" / "probably"** — correct for a collection with fewer than, or exactly, 15 rows.
- **Count caveat** — `schema.countable` at `routes/access/count.ts:18,38`.
- **Relation list covered** — it reuses the same `Page` and `ListResponse` components in both document forms.
- **"on the Node agent"** — justified rather than hedging: the BFF speaks the generic `/forest/{collection}` wire and nothing pins the upstream to `@forestadmin/agent`, so another liana can set a different default.

The `Page.required` assertion asked for in the thread above was already split into its own named test in `2c427c9b5`.

PR #1866, opened 2026-08-31
fix(agent-bff): publish BFF_PUBLIC_URL as the openapi server url

Tonours:
fixes PRD-1109

## What changed

Optional `BFF_PUBLIC_URL` now carries the deployment's external base URL into `servers[0].url` of every document shape: the generic document, the served unfolded document, and the `forest-bff openapi` export. Unset, `servers[0].url` stays `/`. A malformed value fails the boot. Trailing slashes, query strings, and fragments are rejected, because generated clients concatenate `servers[0].url + path` and anything behind a `?` or `#` swallows the path.

## Why these choices

- Not a templated server variable. `@hey-api/openapi-ts` (the reference generator in `test/openapi/openapi-generated-client.test.ts`) returns no base URL at all when a server url holds a `{variable}`. The `/` it resolves today is worth more than nothing.
- Not derived from the request. The unfolded document is memoized on the read model, so one caller's `Host` header would end up in every other caller's document.
- Not `FOREST_SERVER_URL`. That one points at the Forest SaaS API, not at this BFF.
- Unset keeps `/` on purpose. Per OAS, a relative server url resolves against the retrieval url, so Swagger UI or Postman importing the served document resolves the right host today. Only the offline export has no origin to resolve against, and that is the case `BFF_PUBLIC_URL` exists for.

## How to test

```bash
BFF_PUBLIC_URL=https://bff.example.com yarn workspace @forestadmin/agent-bff start:dev &
curl -s -H "Authorization: Bearer $TOKEN" localhost:3450/agent/openapi.json | jq .servers
BFF_PUBLIC_URL='https://x?tenant=1' forest-bff openapi   # exits with a ConfigurationError
```

`yarn workspace @forestadmin/agent-bff test` passes 79 suites / 1363 tests, lint and build are clean. New cases cover parsing (valid, malformed, trailing slash, query, fragment, unset) and `servers[0].url` in the document builder, the served route, and both export branches.

## Known limitation

Without `BFF_PUBLIC_URL`, a client generated from the offline export still has no base URL. The fallback server entry names the variable to set.

### General

### Security


packages/agent-bff/src/openapi/openapi-routes.ts
```diff
@@ -57,6 +59,7 @@ export default function createOpenApiRoutes({
     const { document, unfolding } = await buildUnfoldedDocument(source, readModel, token, {
       version,
       hasAiQueryRoute,
+      publicUrl,
```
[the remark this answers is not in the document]
Tonours: Fixed in f74a9df0b: `parsePublicUrl` now rejects authority userinfo (`https://user@host`, `https://user:pass@host`) with a dedicated ConfigurationError, so credentials cannot reach `servers[0].url` in any document shape (served or exported).

packages/agent-bff/src/cli-dispatch.ts
```diff
@@ -91,11 +91,14 @@ export async function renderOpenApi(env: NodeJS.ProcessEnv, logger: Logger): Pro
       : undefined;
 
   const hasAiQueryRoute = publishesAiQuery(env, logger);
+  const publicUrl = parsePublicUrl(env.BFF_PUBLIC_URL);
```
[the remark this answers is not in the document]
Tonours: Fixed in 3e8eeec81: the guard is now a raw delimiter scan (`value.includes('?') || value.includes('#')`), which also catches the bare delimiters that `new URL` reports as empty components. Covered by tests in `env-config.test.ts`.

packages/agent-bff/test/openapi/openapi-cli.test.ts
```diff
@@ -201,6 +201,30 @@ describe('renderOpenApi', () => {
       /HTTP_PORT/,
     );
   });
+
+  // The export is the case BFF_PUBLIC_URL exists for: an offline document has no retrieval URL to
+  // resolve `/` against, so a client generated from it would have no base URL at all.
```
nbouliol: CLAUDE.md asks for no comments in tests — the `it` name already carries this intent.

packages/agent-bff/test/openapi/openapi-document.test.ts
```diff
@@ -65,6 +65,35 @@ function leafOperators(): string[] {
...
+    it('should publish the configured public URL as an absolute base URL', () => {
+      expect(
+        generateOpenApiDocument('9.9.9', { publicUrl: 'https://bff.example.com' }).servers,
+      ).toEqual([
+        { url: 'https://bff.example.com', description: expect.stringContaining('public base URL') },
+      ]);
+    });
+
+    // A generated client concatenates `servers[0].url` with the operation path, so anything holding
+    // a template placeholder leaves it with no base URL at all — worse than the relative fallback.
```
nbouliol: Same rule here: no comments in tests, and `expect(server.variables).toBeUndefined()` already states it.

packages/agent-bff/src/cli-dispatch.ts
nbouliol, line 93: `publishesAiQuery` swallows the `parseConfig` failure, so a bad `BFF_PUBLIC_URL` logs a misleading "Omitting POST /agent/v1/ai/query" warning before the real error — could this line move above it?

nbouliol (commented):
`BFF_PUBLIC_URL` is missing from `.env.example`, which lists the other optional BFF vars, and the `forest-bff openapi` export section of the README doesn't point at it — that export being the case the variable exists for.

Tonours:
Both added.

`.env.example` gains `# BFF_PUBLIC_URL=https://bff.example.com`, commented like the other optional vars.

The `forest-bff openapi` section now says why it matters there specifically:

> Set `BFF_PUBLIC_URL` for this command in particular. A document fetched over HTTP resolves its relative `servers[0].url` against the URL it came from; an exported file has no such origin, so without it a client generated from the export has no base URL at all.

I also widened the env-table row, which said only "a malformed value fails the boot" and did not name the two rejections that matter: a URL carrying credentials would publish them to every reader of the document, and anything behind a `?` or `#` swallows the path a generated client appends.

packages/agent-bff/test/openapi/openapi-cli.test.ts
Tonours: Removed, and folded into the name: `should publish BFF_PUBLIC_URL in the generic export, which has no retrieval url to resolve against`. The "why" was the only thing the comment carried, and the name is where a reader looks first anyway.

packages/agent-bff/test/openapi/openapi-document.test.ts
Tonours: Removed. Renamed to `should fall back to a root-relative url with no template variable, which a generator cannot resolve`, so the `variables` assertion and the reason for it read together.

packages/agent-bff/src/cli-dispatch.ts
Tonours: Moved. `parsePublicUrl` now runs before `publishesAiQuery`, so `BFF_PUBLIC_URL=https://x?tenant=1` throws its own `ConfigurationError` instead of printing "Omitting POST /agent/v1/ai/query" first and sending the operator after the wrong variable.

The ordering is pinned rather than left to chance — a test asserts the export rejects a bad public url, and it would fail if the two lines swapped back, since `publishesAiQuery` would swallow the `parseConfig` failure again.

Tonours:
The credential check I added last push did not hold. Fixed properly now.

It was a raw-string regex, `/^https?:\/\/[^\/?#]*@/`, while `isHttpUrl` validates through zod, which checks the **normalised** `url.protocol`. Anything WHATWG parses as `https://user:pass@host` without literally starting with lowercase `http://` slipped through both. Measured:

```
HTTPS://admin:s3cret@bff.example.com   protocol https:  user "admin"  regex blocks: false
https:\\admin:s3cret@bff.example.com   protocol https:  user "admin"  regex blocks: false
https:/admin:s3cret@bff.example.com    protocol https:  user "admin"  regex blocks: false
https:admin:s3cret@bff.example.com     protocol https:  user "admin"  regex blocks: false
https://admin:s3cret@bff.example.com   protocol https:  user "admin"  regex blocks: true
```

So `BFF_PUBLIC_URL=HTTPS://admin:s3cret@bff.example.com` booted and published the password in `servers[0].url` of every served document and of the `forest-bff openapi` export — the export being the artefact most likely to end up committed to a repository. Exactly the class the previous commit claimed to close.

The check now runs on the parsed URL, which is the only thing that agrees with the validator:

```ts
const url = new URL(value);

if (url.username !== '' || url.password !== '') throw …;

return url.href.replace(/\/+$/, '');
```

Emitting `url.href` rather than the raw string closes the second half of the same problem: `https:/bff.example.com` was being published verbatim, so a generated client concatenated onto a single-slash URL. It now normalises to `https://bff.example.com`, and `HTTPS://BFF.Example.COM` to `https://bff.example.com`.

Four bypass forms added to the rejection table, plus a canonicalisation case. The `?`/`#` check stays a string check on purpose — that one has to catch bare delimiters, which the parsed object reports as empty.

packages/agent-bff/src/cli-dispatch.ts
nbouliol: Measured it: swapping the two lines back, `test/openapi/openapi-cli.test.ts:205` still passes — it asserts only `rejects.toThrow(/BFF_PUBLIC_URL/)`, and `publishesAiQuery` swallows the `parseConfig` failure without rethrowing, so the same error surfaces either way.

The only difference is the logger, which the test passes as `noopLogger`:

| `cli-dispatch.ts:93-94` | the 4 `BFF_PUBLIC_URL` tests | logger calls |
|---|---|---|
| current | pass | `[]` |
| swapped back | pass | `Warn "Omitting /agent/v1/ai/query from the document: the configuration could not be read (Invalid configuration: BFF_PUBLIC_URL must not carry a query string or fragment.)"` |

A `jest.fn()` logger plus `expect(logger).not.toHaveBeenCalled()` would pin the ordering the name claims.

packages/agent-bff/test/config/env-config.test.ts
```diff
@@ -326,4 +326,87 @@ describe('parseConfig', () => {
...
+      ).toBe('http://localhost:3450/bff');
+    });
+
+    it('should strip trailing slashes, which a client would otherwise concatenate into //', () => {
+      expect(
+        parseConfig({ ...VALID_ENV, BFF_PUBLIC_URL: 'https://bff.example.com/' }).publicUrl,
+      ).toBe('https://bff.example.com');
+      expect(
+        parseConfig({ ...VALID_ENV, BFF_PUBLIC_URL: 'HTTPS://BFF.Example.COM' }).publicUrl,
+      ).toBe('https://bff.example.com');
```
nbouliol: Case normalisation isn't trailing-slash stripping — its own `it`?

Tonours: Split: case normalisation has its own `it` now (env-config.test.ts:355).

nbouliol (approved):
All six findings addressed. I verified the two test fixes rather than took them on trust: reverting the `cli-dispatch.ts` swap now fails `should fail on a bad BFF_PUBLIC_URL before warning about an unrelated omitted route` on the `not.toHaveBeenCalled` assertion, so the ordering is genuinely pinned this time, and case normalisation has its own `it`.

Also re-checked after the main merge: the `servers` block is untouched, all three `generateOpenApiDocument` call sites still pass `publicUrl`, the unfolded memo stays closure-local per `createOpenApiRoutes` call, and nothing main brought adds a document path that bypasses it. 433 tests green locally.

The credentials bypass you caught and documented mid-review (`HTTPS://admin:s3cret@host` slipping past the raw regex while zod validated the normalised protocol) was the right call to fix on the parsed URL.

The remaining `/` default without `BFF_PUBLIC_URL` reads as an accepted limitation, not a gap.

PR #1867, opened 2026-08-31
feat(agent-bff): publish typed record schemas in the openapi document

Tonours:
## What

The unfolded OpenAPI document emits a `Record_<collection>` schema per collection with a known field set, typed from the capabilities field types and published nullable — the capabilities never report nullability — and a `ListResponse_<collection>` wrapper referencing it. A relation list operation answers with the FOREIGN collection's response schema. A degraded collection keeps the shared untyped `ListResponse`, for its own list and for relations pointing at it.

No runtime behaviour change: the record shape, the key transform and the id string form are untouched. This documents them.

fixes PRD-1105

## Three traps the document now names

- **Fields are published under the key the response carries, not the schema name.** `agent-client` deserializes JSON:API with `keyForAttribute: 'camelCase'`, so a `first_name` column is *projected* under that name and *returned* as `firstName`. The transform comes from `inflected` — the same library `jsonapi-serializer` uses — so parity holds by construction rather than through a hand-rolled mirror that would drift silently.
- **`id` is a string on every record, even when the key column is a Number.** It is the JSON:API resource id. `__forest.primaryKey` carries the same id *typed*, so the two forms of one record disagree by construction, and comparing them without coercion fails. Stated on the shared `ForestRecordMeta` and `ListResponse` too, so the generic document carries it, including the `|`-joined packed form of a composite key.
- **Every field is nullable.** The capabilities report a column type and never its nullability, so a nullable column answers `null` against a type a non-null schema would reject. `firstName` is `['string', 'null']`; a `Json` column stays unconstrained because it already accepted null, and nested `items` and sub-properties are widened too, since the same holds at every depth. Publishing non-null would have been a claim the runtime breaks on the first empty column.

## Verification

- `yarn workspace @forestadmin/agent-bff test` — includes redocly spec validity, the generated-client suite, and a real-agent integration suite pinning `id: "8"` (string) against `__forest.primaryKey.id: 8` (number), and `first_name` returned as `firstName`.
- `lint`, `build`, `yarn install --frozen-lockfile`: clean.

The live-agent harness those integration suites share (`findFreePort`, `buildApp`, the schema transport) lives in `test/data/fixtures/live-agent-harness.ts` rather than being copied per suite.

## Known limitations

- The camelCase transform is lossy: fields differing only by case or separator (`first_name` vs `firstName`) collapse onto one response key. The record schema publishes that key unconstrained and names the fields that reach it. The underlying data loss is out of scope and reported separately.
- The transform recurses into attribute values, so the inner keys of a `Json` column are camelized too.
- `packages/workflow-executor/src/adapters/agent-client-agent-port.ts` transcribes this transform by hand; the monorepo now holds both that transcription and a direct use of `inflected`. Unifying them is tracked as PRD-1113.

### General

### Security


packages/agent-bff/src/openapi/record-schemas.ts
```diff
@@ -0,0 +1,116 @@
...
+      description:
+        `${names} all reach the response under this single key, so which one it holds is not ` +
+        'determined here. Project one of them at a time to know.',
+    };
+  }
+
+  const [field] = fields;
+  const schema = toFieldSchema(field.type);
+
+  return field.name === key ? schema : { ...schema, description: `The ${names} field.` };
```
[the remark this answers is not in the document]
Tonours: Fixed in 6a1d8968f: the renamed-key branch now joins the base schema description with the field-name note instead of overwriting it, so a snake_case `File` column keeps its `A data URI.` contract. Regression test added in `test/openapi/record-schemas.test.ts` (asserts `A data URI. The "profile_file" field.` under `profileFile`).

packages/agent-bff/src/openapi/record-schemas.ts
```diff
@@ -0,0 +1,120 @@
...
+function propertySchema(key: string, fields: ProjectableField[]): SchemaObject {
+  const names = fields.map(field => quoted(field.name)).join(', ');
+
+  if (fields.length > 1) {
+    return {
+      description:
+        `${names} all reach the response under this single key, so which one it holds is not ` +
+        'determined here. Project one of them at a time to know.',
+    };
+  }
```
nbouliol, line 102: Capabilities never report nullability, so a nullable column returns `firstName: null` against a published `{ type: 'string' }` — should this be `type: ['string', 'null']`, to keep the record schema from rejecting what the runtime really sends?

packages/agent-bff/test/data/record-contract.integration.test.ts
```diff
@@ -0,0 +1,141 @@
...
+  app.use(bodyParser());
+  app.use(async (ctx, next) => {
+    ctx.state.timezone = TIMEZONE;
+    ctx.state.agentToken = token;
+    await next();
+  });
+  app.use(createDataRoutesMiddleware({ store, agentUrl, logger: noopLogger }));
+
+  return app;
+}
```
nbouliol: Lines 23-84 are byte-identical to `search-agent.integration.test.ts`, and the copy dropped its two rationale comments — what about extracting the harness into `test/data/fixtures/`?

global.d.ts
```diff
@@ -1,5 +1,16 @@
 /// <reference types="jest-extended" />
 
 declare module 'eventsource';
+// Typed rather than left as `any`: `agent-bff` publishes the response record keys these calls
+// produce, so a wrong signature there would go unnoticed. Exported as one object because the
+// methods read `this.inflections` — destructuring them breaks at runtime, not at compile time.
+declare module 'inflected' {
```
nbouliol: `@types/inflected@1.1.29` covers the pinned 1.x — any reason to prefer a hand-written declaration over it?

nbouliol (commented):
Two-axis review (standards + spec against PRD-1105). The spec side holds up: parity with `jsonapi-serializer`'s camelCase branch is real, the id string-vs-number contract is pinned by a genuine real-agent test, degraded collections are covered on both halves, and "no runtime change" checks out. One blocking question on nullability, one on DRY in the new integration test.

packages/agent-bff/src/openapi/record-schemas.ts
Tonours: Right, and it was the one thing here a generated client would actually break on. Every published field is nullable now:

```ts
function nullable(schema: SchemaObject): SchemaObject {
  if (typeof schema.type !== 'string') return schema;

  return { ...schema, type: [schema.type, 'null'] };
}
```

`firstName` is `['string', 'null']`, `age` is `['number', 'null']`, and a `Json` column stays unconstrained (it already accepted null). The record description says why, so it does not read as an oversight: the capabilities report a column type and never its nullability, so a null is always possible whatever the type says.

Publishing non-null would have been a claim the runtime breaks on the first empty column — the same class as the `id` string-vs-number trap this PR exists to document. Test: "should publish every field as nullable, since capabilities never report nullability".

packages/agent-bff/test/data/record-contract.integration.test.ts
Tonours: Extracted to `test/data/fixtures/live-agent-harness.ts`: `findFreePort`, `agentToken`, `schemaFetcherFromFile`, `buildApp`, the shared constants and `noopLogger`. Both suites import from it, and the two rationale comments the copy had dropped are back on the harness — the `RESERVED_SCHEMA_PREFIX` one about `agent.stop()` cleanup, and the `buildApp` one about only the schema transport being swapped.

Each suite keeps what is genuinely its own: its datasource fixture, its `createTestableAgent` customizations, and its schema path. `search-agent` also keeps its `disableSearch` and `addManyToOneRelation` setup, which has nothing to do with the record contract.

Both suites still pass against a real booted agent.

global.d.ts
Tonours: No reason — `@types/inflected@^1.1.29` is a devDependency now and the hand-written block is gone from `global.d.ts`. Lockfile delta is +5 / -0.

The published types are strictly better here: the declaration only covered the two methods this file calls, so any third call would have silently gone `any`, which is exactly the drift the comment above it was worried about. `tsc` is clean and the record-key tests still pass, so the signatures agree.

Tonours:
Review pass done. Two cleanups from it, both mine from the last push:

- The collapse rationale ("several fields can collapse to the same key… the type is left open") had ended up stacked above `nullable()` instead of `propertySchema`, so it read as an explanation of the nullability helper. Moved back.
- `global.d.ts` was left with a stray blank line after the hand-written `inflected` declaration came out. It is byte-identical to `main` again, which is the right end state now that `@types/inflected` covers it.

Worth recording from the pass, since it is not obvious from the diff: **`record-contract.integration.test.ts` passes on revert, by design.** It pins the *runtime* the schema mirrors — `id: "8"` against `__forest.primaryKey.id: 8`, and `first_name` returned as `firstName` — and never reads the OpenAPI output. That is what makes it useful: if the agent or `agent-client` ever changes either behaviour, the test fails and the published schema is then known to be wrong. The schema side is pinned separately by `record-schemas.test.ts` and the unfolded document tests, which do fail on revert.

Two ceilings named rather than fixed:

- `nullable()` only wraps the top-level `type`. A nullable *item* inside an array column, or a nested property of an object column, is still published non-null. The capabilities carry no nullability at any depth, so the same argument applies one level down — but a `Json` column is already unconstrained, and the array/object cases are rare enough that widening them now would be guesswork rather than a correction.
- The degraded branch keys on `projectable.length === 0` rather than `degraded !== null`. Equivalent for everything `collect-unfolding` produces, since a degraded collection reports no projectable field; a hand-built `Unfolding` with `degraded` set *and* fields would get a typed record. No caller does that.

`id` stays non-nullable and required on purpose: `response-mappers.ts` throws on a missing or null id, so a record without one never reaches a response.

nbouliol (commented):
Round 2. The three findings from the first pass are confirmed fixed: the harness extraction is clean (the search suite keeps the same secrets, timezone and middleware order), `@types/inflected` sits correctly in `devDependencies` and matches the pinned 1.x, and the nullable fix is the right call. One follow-up on how deep it goes.

One thing outside the diff: the PR body still says properties are "typed from the capabilities field types" and never mentions nullability, though a generated client's field types went from `string` to `string | null` since the first revision — worth a line.

packages/agent-bff/src/openapi/record-schemas.ts
```diff
@@ -0,0 +1,134 @@
...
+};
+
+/**
+ * Every published field is nullable. The capabilities report a column type and never its
+ * nullability, so a nullable column answers `null` against a type this schema would otherwise
+ * declare non-null — a generated client validating the response would reject what the runtime
+ * really sends. An unconstrained schema already accepts null and is left alone.
+ */
+function nullable(schema: SchemaObject): SchemaObject {
+  if (typeof schema.type !== 'string') return schema;
```
nbouliol: `nullable()` stops at the container, so `items` and a nested object's `properties` stay non-null — the same failure mode one level down, while the record prose says a null is always possible whatever the type says. Recurse, or narrow the prose to the top level?

packages/agent-bff/test/data/fixtures/live-agent-harness.ts
```diff
@@ -0,0 +1,86 @@
...
+import SchemaCache from '../../../src/read-model/schema-cache';
+
+export const TIMEZONE = 'Europe/Paris';
+export const AUTH_SECRET = 'b0bdf0a639c16bae8851dd24ee3d79ef0a352e957c5b86cb';
+export const ENV_SECRET = 'ceba742f5bc73946b34da192816a4d7177b3233fee7769955c29c0e90fd584f2';
+export const BOOT_TIMEOUT_MS = 60_000;
+
+// agent-testing only deletes a schema file whose name carries this prefix, so reusing it keeps the
+// temporary schema cleaned up by `agent.stop()` even though the path is chosen here.
+export const RESERVED_SCHEMA_PREFIX = 'reserved-forestadmin-schema-test-';
```
nbouliol: `TIMEZONE` and `noopLogger` are exported but neither suite imports them.

packages/agent-bff/src/openapi/record-schemas.ts
Tonours: Fixed: `nullable` recurses into `items` and into a nested object's `properties` (record-schemas.ts:49-60), so the prose holds at every level. A `$ref` is left alone, and an unconstrained schema already accepts null.

packages/agent-bff/test/data/fixtures/live-agent-harness.ts
Tonours: Unexported: both are plain module consts now (live-agent-harness.ts:17,26).

packages/agent-bff/src/openapi/record-schemas.ts
```diff
@@ -0,0 +1,153 @@
...
+  if (fields.length > 1) {
+    return {
+      description:
+        `${names} all reach the response under this single key, so which one it holds is not ` +
+        'determined here. Project one of them at a time to know.',
+    };
+  }
+
+  const [field] = fields;
+  const schema = nullable(toFieldSchema(field.type));
```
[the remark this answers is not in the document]
Tonours: Confirmed and fixed. `keyForAttribute` recurses into complex values (`jsonapi-serializer/lib/deserializer-utils.js:65-66`), so a composite member declared `postal_code` does come back as `postalCode`. A `withRecordKeys` pass now maps nested property names through `recordKey`, including object schemas inside arrays, before `nullable` widens them. Two assertions cover the nested-object and array-of-object cases.

packages/agent-bff/src/openapi/record-schemas.ts
```diff
@@ -0,0 +1,172 @@
...
+    mapped.properties = Object.fromEntries(
+      Object.entries(schema.properties).map(([key, nested]) => [
+        recordKey(key),
+        isReference(nested) ? nested : withRecordKeys(nested),
+      ]),
+    );
+  }
+
+  return mapped;
+}
```
nbouliol: `nullable` and `withRecordKeys` are the same traversal twice — same `items` guard, same `properties` map, same `isReference` checks — and line 102 runs them back to back over the same structure; worth one pass?

packages/agent-bff/src/openapi/record-schemas.ts
```diff
@@ -0,0 +1,172 @@
...
+    mapped.items = withRecordKeys(schema.items);
+  }
+
+  if (schema.properties !== undefined) {
+    mapped.properties = Object.fromEntries(
+      Object.entries(schema.properties).map(([key, nested]) => [
+        recordKey(key),
+        isReference(nested) ? nested : withRecordKeys(nested),
+      ]),
+    );
```
nbouliol: Two nested members colliding after `recordKey` (`postal_code` + `postalCode`) drop one silently, where the top level publishes an unconstrained key naming both — same lossy transform, no note one level down.

nbouliol (commented):
Round 3. Both round-2 findings verified fixed in the code, not just claimed: `nullable` now recurses into `items` and `properties` and leaves a `$ref` alone, and the harness constants are unexported. The new nested tests are solid — exact `toEqual`, and both the nested-object and array-of-object cases. Two notes on `withRecordKeys` itself.

Unrelated to this PR: `LLM Integration Tests (ai-proxy)` is red, but it is red on `main` too.

nbouliol (approved):
The contract this documents now holds at every depth: fields published under the key the response really carries, `id` as a string against the typed `__forest.primaryKey`, and nullability the capabilities never report — each one pinned by a test rather than by prose alone. The two notes on `withRecordKeys` are non-blocking.

PR #1868, opened 2026-08-31
feat(agent-bff): log origin guard rejections

Tonours:
## What

Both CORS guards log one `Warn` per rejection.

- **Layer 2** (per-key `allowedOrigins`) logs `{ origin, path, keyHash, renderingId }` right before the 403.
- **Layer 1** (`BFF_ALLOWED_ORIGINS`) logs `{ origin, path }` when a preflight carries an origin that is not allow-listed.

Identifiers only, never the key or the payload. `keyHash` is the same truncated sha256 `api-key-middleware` already logs, so a rejection joins straight back to its `Resolved BFF API key` line — `renderingId` alone is shared by every key on a rendering, and `allowedOrigins` is per-key, so it could not identify what to fix.

fixes PRD-1114

## Why

Every `403 origin_not_allowed` and every rejected preflight left zero server-side evidence. The browser reports a generic "blocked by CORS policy" and the BFF logged nothing, so an operator troubleshooting a broken `allowedOrigins` had nothing to start from. It is the most common operator-facing failure of this stack.

## What is deliberately not logged

The preflight classification requires `Access-Control-Request-Method` alongside OPTIONS and Origin, per the fetch spec. A plain OPTIONS probe carrying a forwarded Origin — curl, a monitoring check — stays silent, so an operator grepping the logs is not sent hunting a browser preflight problem that never happened.

## Scope and safety

No behaviour change on pass paths, no change to response bodies or status codes. The signature change stays internal: neither middleware is exported from the package barrel, and every call site moves in the same commit.

No dedup or rate limit: a misconfigured deployment emits one Warn per rejected request, which is the telemetry the ticket asks for.

Three rejection shapes still leave no trace, all pre-existing and outside this ticket: an OAuth request with a disallowed origin (no per-key identity), an API key with an empty `allowedOrigins` list, and an origin allowed per-key but rejected by `BFF_ALLOWED_ORIGINS` on a simple (non-preflight) request. That last one is the awkward case — the server runs the request and the browser discards the response — and a follow-up could log layer 1 on simple requests too, distinguishing preflight from actual request.

## Overlaps with #1861

That branch changes the same guard: layer 2 stops rejecting a request that carries **no** `Origin`, so a server-side client with a restricted key works. Whichever lands second conflicts on the condition in `per-key-origin.ts`. The two changes compose cleanly once merged — an absent origin passes, and a present-but-unlisted one is rejected *and* logged.

## How to test

- `yarn build` at the repo root.

The rejection and pass behaviours are pinned by the supertest suites in `test/cors/`, which mount the real Koa middleware stack, including the zero-log assertions on the pass paths.

### General

### Security


packages/agent-bff/src/cors/cors-middleware.ts
```diff
@@ -28,6 +31,8 @@ export default function createCorsMiddleware({
         ctx.set('Access-Control-Allow-Methods', ALLOWED_METHODS);
         ctx.set('Access-Control-Allow-Headers', ALLOWED_HEADERS);
         ctx.set('Access-Control-Max-Age', String(PREFLIGHT_MAX_AGE_SECONDS));
+      } else if (origin) {
+        logger('Warn', 'BFF preflight origin rejected', { origin, path: ctx.path });
```
Tonours: Preferential. The preflight classification keys on OPTIONS + Origin only. A spec-strict preflight also carries Access-Control-Request-Method (fetch spec), so a plain OPTIONS probe with a forwarded Origin (curl, monitoring) is logged as a preflight rejection. An operator grepping logs then hunts a browser preflight problem that never happened. Consider keying the classification on Access-Control-Request-Method as well. The existing preflight tests in this suite use OPTIONS + Origin without it, so this matches the established test contract; flagged as hardening, not a defect.

Tonours (commented):
**Spec (PRD-1114)**: conforms. Both guards emit one Warn per rejection carrying identifiers only (origin, path, renderingId where the identity exists), wired in cli-core.ts, pinned by tests including the zero-log pass paths.

**Applies to:** the PR's scope. Preferential. Layer-1 telemetry fires for preflights only, as the ticket scopes it, so three rejection shapes still leave no server-side trace: OAuth requests with a disallowed origin (no per-key identity), API keys with an empty allowedOrigins list, and origins allowed per-key but rejected by BFF_ALLOWED_ORIGINS. For a non-preflight request the browser then blocks the response after the server has run it, with nothing in the logs. Pre-existing and out of this ticket's scope; a follow-up could log layer-1 disallowed origins on simple requests too, distinguishing preflight from actual request.

packages/agent-bff/src/cors/cors-middleware.ts
Tonours: Fixed in faeda530d: the rejection log now requires Access-Control-Request-Method alongside OPTIONS + Origin, so a plain OPTIONS probe with only an Origin stays silent. A test pins that shape.

packages/agent-bff/src/cors/per-key-origin.ts
```diff
@@ -1,14 +1,30 @@
...
+  logger: Logger;
+}
+
+export default function createPerKeyOriginMiddleware({
+  logger,
+}: PerKeyOriginMiddlewareOptions): Middleware {
   return async function perKeyOriginMiddleware(ctx, next) {
-    const identity = ctx.state.apiKeyIdentity as { allowedOrigins?: string[] } | undefined;
+    const origin = ctx.get('Origin');
+    const identity = ctx.state.apiKeyIdentity as
```
nbouliol: `permissions-routes-middleware.ts:38` already casts this same state slot to `ResolvedApiKeyIdentity` — reusing it keeps one shape, and lets `identity.renderingId` drop the `?`.

packages/agent-bff/src/cors/per-key-origin.ts
```diff
@@ -1,14 +1,30 @@
...
+      | { allowedOrigins?: string[]; renderingId?: number }
+      | undefined;
     const allowedOrigins = identity?.allowedOrigins ?? [];
 
-    if (allowedOrigins.length > 0 && !originAllowed(ctx.get('Origin'), allowedOrigins)) {
+    if (allowedOrigins.length > 0 && !originAllowed(origin, allowedOrigins)) {
+      logger('Warn', 'BFF per-key origin rejected', {
+        origin,
+        path: ctx.path,
+        renderingId: identity?.renderingId,
```
nbouliol: `renderingId` is not key-unique and the console logger emits no request id, so this can't be joined back to the `Resolved BFF API key` line. What about `keyHash: fingerprintApiKey(ctx.get(BFF_KEY_HEADER))` alongside, like `api-key-middleware.ts` does?

packages/agent-bff/src/cors/per-key-origin.ts
Tonours: Done — `ctx.state.apiKeyIdentity as ResolvedApiKeyIdentity | undefined`, same as `permissions-routes-middleware.ts`. The inline shape is gone, and since `allowedOrigins` and `renderingId` are both required on that type, narrowing on `identity` in the condition lets `identity.renderingId` drop the `?`:

```ts
if (identity && allowedOrigins.length > 0 && !originAllowed(origin, allowedOrigins)) {
```

The narrowing is honest rather than cosmetic: `allowedOrigins.length > 0` already implies an identity, so the optional chain was covering a state the branch cannot be in.

packages/agent-bff/src/cors/per-key-origin.ts
Tonours: Added, same call as `api-key-middleware`:

```ts
keyHash: fingerprintApiKey(ctx.get(BFF_KEY_HEADER)),
```

You are right that the log was half-useless without it. `renderingId` is shared by every key on a rendering, and with no request id the operator could see *that* an origin was rejected but not *which key* to fix — which is the one thing they need, since `allowedOrigins` is per-key. The `keyHash` joins the line straight back to `Resolved BFF API key`, which already logs the same fingerprint.

Safe to emit: `fingerprintApiKey` is a truncated sha256, so it identifies the key without carrying it, which is what the logging convention asks for.

Both telemetry tests now send the header and assert the fingerprint, so the join cannot silently break.

Tonours:
Review pass done — no must-fix. One note taken anyway, because it is the kind that only shows up in production.

**The logged `origin` is caller-controlled and was unbounded.** `Origin` can carry any string up to Node's ~16 KB header limit, and this log fires once per rejected request with nothing rate-limiting it above — so a misconfigured (or hostile) client could pump kilobytes per request into the log pipeline. `loggableOrigin` caps it at 256 characters with a `…` marker so a truncated value cannot be mistaken for the real one. No real origin comes close, so the debugging value is unchanged.

Checked and clean in the pass, worth recording since this is a logging change:

- `fingerprintApiKey` is the first 12 hex of a sha256 over the full `fbff_<16hex>_<64hex>` key — 320 bits of entropy behind a 48-bit digest, not reversible, and the raw key never reaches a log line.
- `ctx.path` is `parse(req).pathname`, so no query string reaches the log.
- `fingerprintApiKey('')` is unreachable in production: the only writer of `ctx.state.apiKeyIdentity` runs after `if (!rawKey) return`, so an identity implies the header was present.
- Max one Warn per request: layer 1 returns on every OPTIONS before layer 2 runs, layer 2 only fires on non-OPTIONS, and `error-middleware` returns silently for a serializable error rather than logging it again.
- No import cycle from `cors/` reaching into `api-key/`.

On the three untraced shapes in the description: two of them (OAuth with a disallowed origin, per-key-allowed but blocked by `BFF_ALLOWED_ORIGINS` on a simple request) are browser-side blocks of a **200** the server already ran — there is no rejection to log. The right tool for the third would be a startup warning when a key's `allowedOrigins` cannot pass `BFF_ALLOWED_ORIGINS`, not a per-request line. Worth its own ticket rather than widening this one.

packages/agent-bff/src/cors/per-key-origin.ts
```diff
@@ -1,14 +1,61 @@
...
-import { originAllowed } from './origin';
+import { loggableOrigin, originAllowed } from './origin';
+import { fingerprintApiKey } from '../api-key/api-key';
+import { BFF_KEY_HEADER } from '../api-key/api-key-middleware';
 import { originNotAllowed } from '../http/bff-http-error';
 
-export default function createPerKeyOriginMiddleware(): Middleware {
+export interface PerKeyOriginMiddlewareOptions {
+  logger: Logger;
+  serverAllowedOrigins?: string[];
```
nbouliol: `logger` was made required and both test call sites updated in the same commit — leaving this one optional lets the new diagnostic silently no-op, which is what `context-routes-middleware.test.ts:75` and `error-contract.test.ts:51` currently do. Make it required too?

packages/agent-bff/src/cors/per-key-origin.ts
```diff
@@ -1,14 +1,61 @@
...
+    logger('Warn', 'BFF key origins are all outside BFF_ALLOWED_ORIGINS', {
+      keyHash,
+      keyOrigins: allowedOrigins.map(loggableOrigin),
+    });
+  }
+
   return async function perKeyOriginMiddleware(ctx, next) {
-    const identity = ctx.state.apiKeyIdentity as { allowedOrigins?: string[] } | undefined;
+    const origin = ctx.get('Origin');
+    const identity = ctx.state.apiKeyIdentity as ResolvedApiKeyIdentity | undefined;
```
nbouliol, line 44: Every request with a restricted key hashes the key and re-runs the whole `allowedOrigins × serverAllowedOrigins` normalization, then discards it once the key is known-good — could the verdict be memoized per key, or computed at resolve time where the identity is already cached? An early `if (!identity || identity.allowedOrigins.length === 0) return next()` would also drop the repeated guard and let the hash be hoisted to a single local.

packages/agent-bff/src/cors/per-key-origin.ts
Tonours: Made required: `logger: Logger` in `PerKeyOriginMiddlewareOptions` (per-key-origin.ts:11), and both test call sites pass one.

packages/agent-bff/src/cors/per-key-origin.ts
Tonours: Done, both halves: an early `if (!identity || allowedOrigins.length === 0) return next()` drops the repeated guard and hoists the hash to a single local, and the `allowedOrigins × serverAllowedOrigins` verdict is memoized per key hash in `assessedKeys` (bounded at 1 000 entries).

packages/agent-bff/src/cors/per-key-origin.ts
```diff
@@ -1,14 +1,67 @@
...
+  serverAllowedOrigins: string[];
+}
+
+// A key whose allowedOrigins share nothing with BFF_ALLOWED_ORIGINS can never serve a browser: this
+// middleware lets the request through, the server-level CORS layer omits Access-Control-Allow-Origin,
+// and the browser drops a response the BFF answered normally — no rejection anywhere, no trace. The
+// misconfiguration belongs to the key, not to the request, so it is reported once per key rather
+// than once per request; the cap keeps the seen set bounded on a rotating key population.
+const MAX_ASSESSED_KEYS = 1000;
+
```
nbouliol, line 21: Nothing reads this value back — only `.has`, `.size`, `.clear` and `.set`, never `.get`, and `canNeverPass` is consumed via the local on line 30. `Set<string>` would say the same thing.

Tonours: Right, the value was dead. Now a `Set<string>`, with `add` in place of `set`; `canNeverPass` stays the local it always was.

packages/agent-bff/src/cors/per-key-origin.ts
```diff
@@ -1,14 +1,67 @@
...
+    if (!identity || allowedOrigins.length === 0) {
+      await next();
+
+      return;
+    }
+
+    const keyHash = fingerprintApiKey(ctx.get(BFF_KEY_HEADER));
+    reportOriginsThatCanNeverPass(allowedOrigins, keyHash);
+
+    if (!originAllowed(origin, allowedOrigins)) {
```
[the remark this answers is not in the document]
Tonours: Fixed at the trust boundary instead of in the middleware: `ApiKeyClient.isResolvedIdentity` now requires every `allowedOrigins` entry to be a string, so a malformed SaaS body is rejected as `unreachable` before it ever reaches `ctx.state.apiKeyIdentity`. Filtering inside `per-key-origin.ts` would have left the other consumers of `ResolvedApiKeyIdentity` exposed to the same `TypeError`, and would have silently downgraded a malformed key to a narrower origin list rather than failing. Covered by a new case in `api-key-client.test.ts`.

nbouliol (approved):
Conflict with #1861 resolved the right way: main's `hasOrigin` gate adopted rather than overwritten, `loggableOrigin` kept alongside it, and the test that pinned 403 for a no-Origin request inverted instead of carried over — that one auto-merged cleanly, so it would only have turned red at CI. All review threads closed and verified in the code.

PR #1869, opened 2026-08-31
fix(mcp-server): accept file uploads on v1 String fields carrying the file picker widget

Scra3:
Legacy v1 lianas declare action file inputs as `type: 'String'` with `widget: 'file picker'` — the pattern [our own docs](https://docs.forest.app/legacy/ruby-agent/reference-guide/actions/smart-action-examples/upload-several-files-with-the-file-picker) still recommend for multi-file upload. The front dispatches on the **widget** and works. The MCP dispatches on the **declared type** and does not:

- `getActionForm` reports `"String"`, so the model never calls `requestActionFileUpload`
- `encodeFileFieldValue` throws when a `File` object reaches a non-`File` field

Reported by a customer on `forest-rails` 9.15.8 whose KYC/KYB/claim upload actions are all built this way.

## Why the combination exists

It is a validation hole on our side, not a customer mistake. In `private-api`, `generate-widget-properties` runs the type/widget validator **only** on the v2 `widgetEdit` path; the v1 `widget` path just checks the widget name against an allowlist and stores it on any type. Verified in production: those fields carry `widgetEdit: { name: "file picker", parameters: {} }`.

v2 makes the combination impossible at the TypeScript level (only `FileDynamicField` accepts `FilePicker`), so **this is a v1-only compat shim**, in the same family as the ones `agent-client` already carries (the Ruby `/hooks/load` 404, the snake_case operators).

## What changes

`FieldGetter.getEffectiveTypeName()` reads `String` / `StringList` carrying that widget as `File` / `FileList`, and feeds the two existing dispatch points:

| | before | after |
|---|---|---|
| type `getActionForm` reports | `String` | `File` → the model calls `requestActionFileUpload` |
| type `setFieldValue` encodes with | `String` → throws | `File` → data uri |

`getTypeName()` is **untouched** — it still drives the `Enum` branch and reports the wire type.

## One behaviour change worth knowing

Value-shape validation on file-picker fields gets stricter, because they now take the `File`/`FileList` branch of `encodeFileFieldValue`:

- a non-array on a `['String']` + file-picker field now throws `expects a list of files: pass an array.` — it used to be forwarded verbatim
- a non-file *object* on a scalar `String` + file-picker field now throws instead of being stored as is

Plain strings still pass through untouched, so the common case (a data uri or a URL) is unaffected. Erroring matches what native `File`/`FileList` fields already did, and replaces a silently wrong-shaped payload with a message — but `['String']` + file picker is the shape the legacy docs recommend for multi-file upload, so it is called out rather than buried.

## What does not change

- A `File` landing on a `String` **without** the widget still throws. That guard prevents a file being JSON-serialized into a column as `{"buffer":{"type":"Buffer",...}}` with no error. Covered by a test.
- Nothing on the customer side. On the wire both types carry a data uri, which is exactly what their handler already receives when a human drops a file in the front.
- `getType()` still returns the verbatim wire value, so what `loadChanges` echoes back to the agent is unchanged.

## Tests

`agent-client` 400/400 · `mcp-server` 1015/1015 · lint 0 error.

New coverage: the two mappings and every negative case (no widget, other widget, non-String type); the guard still firing; list encoding; and the real customer shape — a file field carrying `hook: "onFieldChanged"`, so the `/hooks/change` round-trip is exercised with the data uri in the body.

`get-action-form.test.ts` grew ~55 lines of mock churn: its ~20 hand-rolled field mocks each needed a `getEffectiveTypeName` alongside their existing `getTypeName`, mirroring the same literal. No behaviour change in those tests.

## Noted, not fixed here

`ActionFieldString.fill()` calls `value.toString()` before the guard can see the value, so a `File` routed through that public wrapper would be stored as `'[object Object]'`. Unreachable from the MCP (both entry points go through `FieldFormStates.setFieldValue`), so it is out of scope — worth its own ticket.

fixes PRD-1116

> [!NOTE]
> ### Accept file uploads on v1 String fields with `file picker` widget via `getEffectiveTypeName()`
> - Adds `getEffectiveTypeName()` to `FieldGetter` and `ActionField` to map `String`/`StringList` fields carrying the `file picker` widget to `File`/`FileList`; other types are unchanged.
> - `FieldFormStates.setFieldValue` and the `get-action-form` MCP tool now use the effective type instead of the declared type for encoding and serialization, so v1 file picker fields produce data URIs and report `File`/`FileList` in tool responses.
> - Updates error text in `encodeFileFieldValue` to suggest declaring `File` or using the `file picker` widget.
> - Behavioral Change: `get-action-form` tool responses now report `File`/`FileList` for v1 file picker String fields; consumers relying on the previous `String`/`StringList` type for those fields need to handle the new type.
>
> 
> #### Changes since #1869 opened
>
> - Introduced distinction between declared type and effective type for field type resolution in `FieldGetter` class [a690374]
> - Updated documentation in `CLAUDE.md` to explain the type name distinction for file picker widget handling [a690374]
> - Updated test fixtures across `packages/agent-client` to use empty object for file picker widget parameters [a690374]
> - Refined error message in `encodeFileFieldValue` function for invalid file assignment [a690374]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a550341.</sup>
> 
>


packages/agent-client/src/action-fields/action-field.ts
```diff
@@ -26,6 +26,10 @@ export default abstract class ActionField {
     return this.field?.getTypeName();
   }
 
+  getEffectiveTypeName(): string {
```
Scra3, line 29: Claude Opus 5 (claude-opus-5[1m]): **Should fix** — this delegation is on the hot path of the fix and nothing pins it, so the read half can silently regress to the bug this PR closes.

`get-action-form.ts:128` calls `getEffectiveTypeName()` on **`ActionField`** instances, not on `FieldGetter`: `Action.getFields()` (`domains/action.ts:219-220`) maps the `FieldGetter`s into typed `ActionField` subclasses, which is also why line 113 can call `isRequired()`. So this method — not `FieldGetter`'s — is what decides the type the model is told.

No test ever calls it on a real `ActionField`. The 22 occurrences in `get-action-form.test.ts` are hand-rolled literals (`getEffectiveTypeName: () => 'File'`), and the agent-client tests all go through `FieldFormStates.getField()`, which returns a `FieldGetter`.

I measured it rather than reasoning about it. Replacing this body with `return this.field?.getTypeName();` leaves **397/397 agent-client and 38/38 get-action-form tests green**, lint and tsc clean — while `getActionForm` goes back to reporting `"String"` for the v1 field. The bug ships.

The contrast with the write half is the tell: mutating `field-form-states.ts:89` the same way fails 5 tests. That side is pinned; this one is not.

Smallest fix: one case in the existing `ActionField base class methods` block (`test/action-fields/action-fields.test.ts:835`, which already covers `getType`/`getValue`/`isRequired` this way) — build a `{ type: 'String', widgetEdit: { name: 'file picker', parameters: {} } }` field and assert `new ActionFieldString(...).getEffectiveTypeName() === 'File'` while `getType() === 'String'`.

While you are there: `file-value.test.ts:205` ("leaves a plain string untouched") also survives a revert of the write dispatch — it characterises pre-existing pass-through and pins nothing about this change. Its neighbour at `:230` looks similar but earns its place: it is the negative control that would fail if the widget match were broadened.

packages/agent-client/test/action-fields/field-getter.test.ts
```diff
@@ -71,4 +71,75 @@ describe('FieldGetter', () => {
...
+    it('should leave an Enum untouched', () => {
+      const fieldGetter = new FieldGetter(
+        createPlainField({ type: 'Enum', enums: ['passport', 'id_card'] }),
+      );
+
+      expect(fieldGetter.getEffectiveTypeName()).toBe('Enum');
+      expect(fieldGetter.getTypeName()).toBe('Enum');
+    });
+
+    it('should delegate to getTypeName for native file types', () => {
```
Scra3: Claude Opus 5 (claude-opus-5[1m]): **Violates conventions** — `testing.md#Test name states the exact behavior it asserts`: *"The name reads as the behavior under test […], not the method name."*

The name states an internal call relationship — that `getEffectiveTypeName` delegates to `getTypeName` — rather than what the two assertions prove: a native `File` reports `File`, a native `['File']` reports `FileList`. Every sibling in this `describe` is named behaviourally ("should map a String carrying the file picker widget to File", "should leave an Enum untouched"), so this one is the outlier.

The consequence the rule guards against is live here: rewrite `getEffectiveTypeName` to read `plainField.type` directly — behaviour-identical — and the name becomes false while the test stays green.

Suggested: `'should report a native File as File and a native File list as FileList'`.

Scra3 (commented):
**Spec (PRD-1116)**: conforms. The mapping is exactly the two cases the ticket prescribes, both dispatch points are the ones it names, `getTypeName()` is untouched, the anti-corruption guard still fires for a `String` without the widget, and the two tests the ticket asked for by name are present and strong — the `/hooks/change` round-trip pins the request body *and* the returned `file_0: pdfDataUri`, and the Ruby static-form case proves `widgetEdit` survives `applyFallbackForm`.

I verified the two claims the change rests on rather than taking them from the description: `resolveUploadedFileValues` substitutes upload handles without ever reading the field type, so the handle flow is preserved and the file still never enters the model's context; and the ~55 lines of mock churn in `get-action-form.test.ts` are behaviour-neutral — every mock mirrors its `getTypeName` literal except the three new cases, checked mechanically.

---

Claude Opus 5 (claude-opus-5[1m]): **Preferential** — the PR description says nothing about a behaviour change it introduces.

**Applies to:** the PR description

A caller that passes a non-array to a `['String']` + file-picker field used to have the value forwarded verbatim (`type: 'StringList'` matched neither file branch, so `encodeFileFieldValue` returned it); it now throws `expects a list of files: pass an array.` Same shape for a non-file *object* on a scalar `String` + file-picker field. Plain strings still pass through, so the common URL case is unaffected.

Erroring is the better behaviour — it matches what native `File`/`FileList` fields already did. But `['String']` + file picker is precisely the pattern Forest's own legacy docs recommend for multi-file upload, so this tightens validation on the exact shape the PR targets, and "no customer-side change" in the description currently reads as covering it. Worth one sentence.



---

Claude Opus 5 (claude-opus-5[1m]): **Preferential** — a documented statement goes stale with this PR.

**Applies to:** `docs.forest.app` → `/product/embed/mcp-server.mdx`

That page scopes the upload path to File fields in two places — the tool table ("Get an upload destination for an action's `File` or `FileList` field") and the section intro ("Actions with **File fields** work over MCP out of the box"). After this PR a `String` + file-picker field also participates, so both sentences under-describe the surface and a reader has no way to learn the combination is supported.

Distinct from the legacy `['String']` doc pages PRD-1116 already scopes out to a follow-up: this staleness is created by this change, not inherited.

packages/agent-client/src/action-fields/action-field.ts
Scra3: Fixed in `ba41e33`.

Three cases in the existing `ActionField base class methods` block, mirroring how it already covers `getType`/`getValue`/`isRequired`: a `String` + file-picker field reports `File` while `getType()` still reports `String`, a `[String]` one reports `FileList`, and a plain `String` stays `String`.

Verified the same way the finding was made: mutating the delegation to `return this.field?.getTypeName();` now fails 2 tests, where before it left all 435 green.

Qlty had spotted the same hole from the other side — 0% diff coverage on `action-field.ts`, dragging the file from 92.3% to 85.7%. Both are closed by the same three tests.

On the secondary note in this finding — `file-value.test.ts:205` (`leaves a plain string untouched`) — I checked and I am keeping it. It does survive a revert of the write dispatch, but it is not characterisation: removing the string pass-through in `encodeFileValue` fails it. Before this change a `String` field never entered `encodeFileValue` at all, so that path is newly reachable and this is the only test covering it on a file-picker field. The neighbour at `:230` stays for the reason you gave.

packages/agent-client/test/action-fields/field-getter.test.ts
Scra3: Fixed in `ba41e33`. Renamed to `should report a native File as File and a native File list as FileList`, which is what the two assertions actually prove.

packages/mcp-server/src/tools/get-action-form.ts
```diff
@@ -125,7 +125,7 @@ The response includes:
                 const description = field.getPlainField()?.description;
                 const baseField = {
                   name: field.getName(),
-                  type: field.getTypeName(),
+                  type: field.getEffectiveTypeName(),
```
hercemer42, line 128: Claude Opus 5 (claude-opus-5[1m]): **Should fix**

**Applies to:** `packages/mcp-server/CLAUDE.md:21` (not in this diff)

The next person or agent editing this package is told to do the opposite of this line. `CLAUDE.md` loads as project instructions and states: *"`getTypeName()` is the collapsed `'FileList'`, for dispatch and for what this server reports to a model."* After this change neither half holds — the encoding dispatch is `getEffectiveTypeName()` (`field-form-states.ts:89`) and so is what the server reports; `getTypeName()` survives only as the `Enum` branch three lines below.

The failure mode is concrete: a cleanup that "restores consistency" between line 128 and line 134 by reverting this line silently reinstates PRD-1116 for every v1 file-picker field, and nothing in `mcp-server` fails — the 22 `getEffectiveTypeName` occurrences in `get-action-form.test.ts` are hand-rolled literals, so they follow whichever accessor the tool calls.

Naming `getEffectiveTypeName()` as the reporting path and `getTypeName()` as Enum-dispatch-only in that sentence is the whole fix.

packages/agent-client/src/action-fields/file-value.ts
```diff
@@ -61,7 +61,10 @@ export default function encodeFileFieldValue(
   // A file reaching a field that is not declared as one is never intentional, and it would be
   // JSON-serialized into the column as {"buffer":{"type":"Buffer",...}} without any error.
   if (isFile(value)) {
-    throw fileError(fieldName, `is a ${type} field and cannot hold a file.`);
+    throw fileError(
+      fieldName,
+      `is a ${type} field and cannot hold a file: declare it as type File, or give it the "file picker" widget.`,
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): **Should fix**

The remediation reaches the LLM verbatim and tells it to do something only the agent's author can do. A plain `Error` thrown here becomes the tool result text unchanged — `parseAgentError` returns `error.message` for any object carrying one (`error-parser.ts:29-39`), and `registerToolWithLogging` puts it straight into `content` with `isError: true` (`tool-with-logging.ts:118-128`).

Reachable without any misconfiguration: `resolveUploadedFileValues` substitutes by **value shape**, not by field type (`resolve.ts:19-30`), so a `$uploadedFile:` handle the model puts in the wrong field — a `comment`, say — arrives here as a real `File` on a `String` field and trips this throw. The model's actual fix is "send the file to the file field"; the message instead tells it the schema is wrong, so it reports a configuration bug to the user or retries the upload.

The neighbouring messages in this flow set the bar — `resolve.ts:70`, `:147` each name the field, the likely cause and the recovery the caller can perform. Leading with the caller-actionable half ("this field takes text, not a file — send the file to a File field") and keeping the declaration advice second would match it.

packages/agent-client/src/action-fields/field-getter.ts
```diff
@@ -34,4 +34,14 @@ export default class FieldGetter {
 
     return Array.isArray(type) ? `${type[0]}List` : type;
   }
+
+  getEffectiveTypeName(): string {
```
hercemer42, line 44: Claude Opus 5 (claude-opus-5[1m]): **Should fix**

This is the only method in the file with no comment, and the one whose rule is least inferable from its body: that a v1 liana declares a file field as `String` and marks it *only* through `widgetEdit.name`. Its two neighbours both document their slice of the same distinction (`:22`, `:27-31`).

Worse, `getTypeName()`'s comment directly above is now half-false — *"For dispatching on the type and for reporting it to a reader"* — since reporting to a reader goes through this method (`get-action-form.ts:128`) and only the `Enum` dispatch stayed behind. A file that exists to disambiguate three accessors now mis-signposts which one feeds what.

One line of why here, plus trimming `getTypeName()`'s comment to the `Enum` dispatch and the wire-echo warning it still owns.

packages/agent-client/src/action-fields/types.ts
```diff
@@ -20,6 +20,7 @@ export type PlainField = {
   isReadOnly: boolean;
   hook?: string;
   widgetEdit?: {
+    name?: string;
```
hercemer42, line 23: Claude Opus 5 (claude-opus-5[1m]): **Preferential**

The widget this line teaches the code to recognize cannot satisfy the type it sits in: `parameters.static` is required two lines below, and a `file picker` widgetEdit never carries it. v1 lianas emit `parameters: {}` (`forest_liana/widgets_helper.rb`, `forest-express/src/utils/widgets.js`); v2 emits `{ prefix, filesExtensions, filesSizeLimit, filesCountLimit }` (`forestadmin-client/src/schema/types.ts:231-240`).

The divergence is already visible in this PR's own fixtures: `field-getter.test.ts:73` and `file-value.test.ts:173` invent `parameters: { static: {} }` to typecheck, while `field-form-states.test.ts:707` — the one written to mirror the real customer payload — uses `parameters: {}`. So the fixtures that satisfy the compiler are the ones that do not match production.

Runtime is unaffected (`ActionFieldMultipleChoice.getOptions()` optional-chains the whole path), so this is type accuracy, not a break. `forestadmin-client` already exports `WidgetEditConfiguration` (`schema/types.ts:59-61`) as `{ name: string; parameters: Record<string, unknown> }` — which is both correct here and stricter about `name` than the `name?` this line adds.

hercemer42 (commented):
Claude Opus 5 (claude-opus-5[1m]): **Spec (`PRD-1116`)**: conforms. `getEffectiveTypeName()` implements the ticket's mapping table verbatim and is wired at exactly the two dispatch points it names — `get-action-form.ts:128` (read) and `field-form-states.ts:89` (write) — with `getTypeName()` left driving the `Enum` branch and `getType()` left as the wire value. I checked the premise against the two v1 lianas rather than the stored apimap alone: `forest-rails` (`app/helpers/forest_liana/widgets_helper.rb`, `set_field_widget`) and `forest-express` (`src/utils/widgets.js`, `setFieldWidget`) both rewrite `widget: 'file picker'` into `widgetEdit: { name: 'file picker', parameters: {} }` and delete the `widget` key, on the `/hooks/load` **and** `/hooks/change` responses. So the key this reads is present on both dynamic and static form paths, and it survives the change-hook round trip — the mid-form reversion this design would otherwise be exposed to does not occur.

---

Claude Opus 5 (claude-opus-5[1m]): **Preferential**

**Applies to:** the PR as a whole (`packages/workflow-executor/src/adapters/agent-client-agent-port.ts:319`, not in this diff)

The same v1 field is now `File` to an MCP model and still `String` to the workflow AI, so a file-picker field stays unfillable from a `trigger-record-action` step. `agent-client-agent-port.ts:319` reports `field.getType()`, and that value goes into the form-fill prompt at `trigger-record-action-step-executor.ts:308`; under its never-guess rule the model either omits a required file field — the operator sees a generic form-validation failure with no mention of files — or invents a plausible string.

Not a defect in this PR: `getType()` is deliberately the wire value there, because agent-bff and workflow-executor put it straight into API responses, and changing it would change those response shapes. But the description's "what does not change" section reads as though the agent-client change reaches every consumer, and this is the one place it deliberately does not. Worth a line in the body or a follow-up ticket so it is inherited on purpose.

packages/mcp-server/src/tools/get-action-form.ts
Scra3: Fixed in `a690374`. `mcp-server/CLAUDE.md:21` now reads:

> `getTypeName()` is the collapsed `'FileList'`, and is now only the Enum dispatch; what this server reports to a model is `getEffectiveTypeName()`, which additionally reads a v1 `String`/`StringList` carrying the `file picker` widget as `File`/`FileList` — the encoding in `setFieldValue` dispatches on it too.

You are right that the file was the load-bearing part: it loads as project instructions, so the sentence was actively telling the next agent to revert line 128. Trimmed `getTypeName()`'s own doc comment in `field-getter.ts` for the same reason — it still claimed "for reporting it to a reader", which moved.

packages/agent-client/src/action-fields/file-value.ts
Scra3: Fixed in `a690374`. The caller-actionable half now leads:

> Field "comment" takes String, not a file: send the file to a File field instead. If this field is meant to take one, the action must declare it as type File or carry the "file picker" widget.

The reachability argument is the part that changed my mind. I had considered the message's accuracy earlier and dropped it as low-probability, because I was picturing a `File` object on a `Number` field — nobody does that. Substitution by value shape in `resolve.ts` is a much more ordinary path: the model puts the handle on the wrong field and gets told the schema is broken. That is a plausible mistake with a misleading answer, which is a different thing from an implausible one.

packages/agent-client/src/action-fields/field-getter.ts
Scra3: Fixed in `a690374`. `getEffectiveTypeName` now carries the rule that is not inferable from its body — that a v1 liana has no File type in action forms and marks a file input only through the widget — and `getTypeName`'s comment is trimmed to what it still owns: the Enum dispatch and the wire-echo warning.

Worth recording why this one needed your push. This repo's standing instruction here is zero comments by default; the author of the change wrote a doc comment for this method, then deleted it to honour that rule, and flagged the deletion. The rule does carve out exactly this case — a non-obvious *why*, a workaround, an external constraint — and a v1 liana's widget-only encoding is squarely that. Two independent reviews landing on the same line is what settled it.

packages/agent-client/src/action-fields/types.ts
Scra3: Fixed in `a690374`. `parameters.static` is optional now, in both this type and the `ForestSchemaAction` twin in `forestadmin-client` — the latter is what types `fallbackFields`, so leaving it out made the static-form suite fail to compile. All nine file-picker fixtures across the three test files now state `parameters: {}`, which is what v1 actually emits.

Kept the inline shape rather than switching to `WidgetEditConfiguration`: its `parameters: Record<string, unknown>` would lose the typing that `getMultipleChoiceField()` relies on to read `parameters.static.options`. Your point about `name` being stricter there stands, and is the better long-term answer — single-sourcing the two declarations is a change worth making on its own rather than inside this fix.

One thing your finding got exactly right and I want to make explicit: the tell was in the fixtures. Six invented `{ static: {} }` to satisfy the compiler while the one written from the real payload used `{}`. A type that makes correct data fail to typecheck teaches every fixture author to write something production never sends.

hercemer42 (approved):

PR #1870, opened 2026-09-01
refactor(agent-bff): assemble the server behind a single buildBff

nbouliol:
First of the stack that lets an agent embed the BFF in-process (`agent.addBff()`).

## Why

`runCli` owned the middleware order — cors, error, body parser, oauth, docs, then the agent-scoped chain — so a host wanting the same stack without a listener had no entry point. Rebuilding that order on the agent side would have made the two deployment modes drift silently, and a divergence there is CORS or auth quietly wrong, not a failing test.

## What

- `buildBff({ config, logger })` returns the request handler and owns the whole assembly. `runCli` becomes `parseConfig` + `buildBff` + `listen`.
- `/health` and `X-Forest-Bff-Version` move into the handler, where they belong: they are part of what the BFF serves, not of how it listens. Both are extracted into `createHealthRoute` and `createVersionHeaderMiddleware`, used by `buildBff` and by the legacy `BFFHttpServer` path — one implementation, two call sites.
- `BFFHttpServer` is a published export, so its `{ port, version, config, logger, middlewares }` constructor keeps working unchanged; it just gains an optional `callback`.
- `resolveOAuthConfig` and `resolveUnfoldSource` move alongside the rest, so `cli-core.ts` no longer reaches into `openapi/` — the mount-invariant allow-list gets tighter, not looser.

No behavior change: same middlewares, same order, same responses.

## Tests

`yarn workspace @forestadmin/agent-bff test` — 80 suites, 1350 tests. The existing `cli-core` suite still exercises the composed stack end to end through `runCli`; the new `build-bff` suite covers what moved (health ok/degraded, version header, the agent edge sitting behind the health route, the malformed-origins warning).

Fixes PRD-1076

> [!NOTE]
> ### Assemble the agent BFF server behind a single `buildBff` callback
> - Moves all middleware assembly (health, version, CORS, error handling, body parsing, OAuth, documentation, agent routes) from `cli-core.ts` and `BFFHttpServer` into the new `buildBff` function in [build-bff.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1870/files#diff-44122a133e23acfd87a9ef4d51935c661e42d5145a703f50e58b3ce1acb1b81e), which returns a ready-to-serve Koa callback
> - `BFFHttpServer` in [bff-http-server.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1870/files#diff-883f86539854a7042f72689b857c7561e2876a7b8ec1e8fd028aee8892758569) now accepts either a prebuilt callback or the legacy internally-assembled options; a prebuilt callback is served without adding extra health, version, or route middleware
> - Adds reusable `createHealthRoute` and `createVersionHeaderMiddleware` middleware, and a shared `warnMissingConfig` helper that logs missing required keys during construction
> - Exports `buildBff` and `BffCallback`/`BuildBffOptions` types from the package entrypoint in [index.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1870/files#diff-f90e0b1dca1f38dae6914b1be178eac8622802ac17e471aee507b6b274a2ef75)
> - Behavioral Change: missing-configuration warnings now fire during handler construction rather than at server `listen` time; consumers using `buildBff` get the callback directly and bypass `BFFHttpServer`'s internal assembly path entirely
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 3326cd4.</sup>
> 
>


packages/agent-bff/test/build-bff.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  it('should set the version header on every response', async () => {
+    const callback = await buildCallback(VALID_ENV);
+
+    const response = await request(callback).get('/unknown-path');
+
+    expect(response.headers['x-forest-bff-version']).toBe(version);
+  });
+
+  it('should mount the agent edge behind the health route', async () => {
```
Tonours: Claude Opus 5 (claude-opus-5): **Violates conventions** — `skills/conventions/testing.md#Test name states the exact behavior it asserts`

A reader of this file will believe the health/agent mount order is covered. It is not: swap the first two entries of `buildBff`'s middleware array and this test still passes green, so the ordering invariant the commit message calls out is unguarded.

A 401 on `/agent/...` proves the auth chain is mounted; it says nothing about the health route sitting in front of it, because `agentScoped` never lets the agent chain touch `/health` and the two paths never collide.

Either rename it to what it verifies (`should answer 401 on an unauthenticated agent route`), or make the ordering observable — assert that `GET /health` returns 200 **and** carries `x-forest-bff-version`, which only holds if the version header is mounted ahead of the health route, since the health route short-circuits without calling `next()`.

packages/agent-bff/src/http/bff-http-server.ts
```diff
@@ -14,46 +17,44 @@ export interface BFFHttpServerOptions {
...
-      ctx.set('X-Forest-Bff-Version', this.options.version);
-      await next();
-    });
-
-    this.app.use(async (ctx, next) => {
-      if ((ctx.method === 'GET' || ctx.method === 'HEAD') && ctx.path === '/health') {
-        const { config, version } = this.options;
-        ctx.status = config.hasAllRequired ? 200 : 503;
-        ctx.body = { status: config.hasAllRequired ? 'ok' : 'degraded', version };
+    this.handler = options.callback ?? BFFHttpServer.buildHandler(options);
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**

A host that adds `callback` without removing `middlewares` boots successfully, logs `Forest BFF started`, and then 404s every one of its own routes from bare Koa — with no warning and nothing correlating the 404s to the dropped array. The combination could not exist before this PR, since `callback` is new.

Both fields are optional on `BFFHttpServerOptions`, so passing both compiles, and `options.callback ?? buildHandler(options)` silently discards the array. The doc comment says so; nothing at runtime does. Same shape of problem one field over: `version` stays required but is dead in the `callback` branch, so a caller of the new public API has to pass a value that does nothing.

A discriminated union on the options type would make the invalid combination unrepresentable instead of documented. Failing that, a `Warn` naming the number of dropped middlewares — plus a test pinning the documented behaviour, since `test/http/bff-http-server.test.ts` only exercises the `middlewares` branch today.

packages/agent-bff/src/build-bff.ts
```diff
@@ -0,0 +1,428 @@
...
+
+  return chain.map(agentScoped);
+}
+
+/**
+ * Assemble the whole BFF — `/health`, the version header, and every middleware in the one order both
+ * deployment modes must share — and hand back the request handler. `runCli` puts it behind a
+ * listener; an embedding host mounts it on its own server.
+ */
+export default async function buildBff({
```
Tonours, line 419: Claude Opus 5 (claude-opus-5): **Should fix**

**Applies to:** `packages/agent-bff/src/http/bff-http-server.ts:67` (not in this diff) — anchored here because `buildBff` is where the fix belongs.

The log line that names *which* required keys are missing lives in `BFFHttpServer.start()`, which a host mounting `callback` on its own server never calls. `buildBff`'s own warnings say `required configuration is missing` without naming a variable, so the two deployment modes now give different diagnostics for the same misconfiguration — in the PR whose stated purpose is that the modes cannot diverge.

Bounded today: the caller builds `config` itself and has `config.presence` in hand, so this is a lost convenience rather than blindness. It stops being bounded at `addBff()`, where the agent calls `parseConfig` internally and the host never sees `presence` at all.

The `presence` diff needs only `config.presence` and the logger, both already in hand here. Move it into `buildBff` and drop it from `start()`, otherwise the CLI logs it twice.

packages/agent-bff/src/build-bff.ts
```diff
@@ -0,0 +1,428 @@
...
+    createTimezoneMiddleware({ defaultTimezone }),
+    ...buildAgentRouteMiddlewares(bundle, config, logger),
+  ];
+
+  return chain.map(agentScoped);
+}
+
+/**
+ * Assemble the whole BFF — `/health`, the version header, and every middleware in the one order both
+ * deployment modes must share — and hand back the request handler. `runCli` puts it behind a
```
Tonours, line 414: Claude Opus 5 (claude-opus-5): **Preferential**

"an embedding host mounts it on its own server" promises more than the handler delivers, and `buildBff` is published in `index.ts` as of this PR — so the first embedder discovers the constraint rather than reading it.

The callback is a terminal Koa handler: it answers 404 itself and never yields to a host `next()`. A host that routes every path into it loses its own routes; a host that mounts it under a prefix gets `/health` and `/agent/...` at the wrong absolute paths. The `basePath` support that makes the sentence true arrives in a later PR of this stack, but the export ships now.

Worth one clause naming the constraint — mounted at the root, absolute paths, terminal handler.

Tonours (commented):
**Spec (PRD-1076)**: conforms. Step 1 is delivered as written — `buildBff({ config, logger }) -> { callback }` exists, `runCli` decomposes into `parseConfig` + `buildBff` + `listen`, `/health` and `X-Forest-Bff-Version` moved into the handler, `cli-dispatch` follows `resolveOAuthConfig` / `resolveUnfoldSource` to their new module, and the public surface is additive only. The `invalidate` half of the ticket's `{ callback, invalidate }` signature belongs to step 6 and is not expected here.

Iso-behaviour verified against the merge base: all eleven extracted helpers are byte-identical, and the effective middleware order is unchanged (version header, `/health`, CORS, scoped error middleware, body parser, OAuth, docs, agent chain). Nothing dropped. `BFFHttpServer`'s existing constructor is not broken — `version` was already required before this PR, and `callback` is new, so an existing consumer never reaches the ignored-`middlewares` path.

Four findings inline, none blocking.

packages/agent-bff/test/build-bff.test.ts
nbouliol: Fair — `agentScoped` skips `/health` and the health route skips `/agent/...`, so the two never collide and the old name claimed coverage the assertion could not give. Renamed to `should answer 401 on an unauthenticated agent route`, and the ordering invariant is now pinned by a separate test asserting `x-forest-bff-version` on the `/health` 200: I verified it goes red (assertion failure, not timeout) when the header middleware and the health route are swapped.

packages/agent-bff/src/http/bff-http-server.ts
nbouliol: Made unrepresentable rather than documented: `BFFHttpServerOptions` is now a union of `AssembledOptions` (`version` required, `middlewares` optional, `callback?: never`) and `PrebuiltOptions` (`callback` required, `version?: never`, `middlewares?: never`). Both invalid combinations — `callback` + `middlewares` and `callback` + `version` — are compile errors now, verified with a throwaway `tsc` probe, so the dead-`version` half is gone too and `runCli` no longer passes it. Existing consumers passing `{ port, version, config, logger, middlewares }` still fit the assembled variant unchanged. Added a test that the prebuilt handler is served as-is.

packages/agent-bff/src/build-bff.ts
nbouliol: Agreed — extracted to `src/config/missing-config-warning.ts` and dropped from `start()`. One difference from your suggestion: I kept a call site in `BFFHttpServer`, in the assembled branch of the constructor, otherwise the legacy `middlewares` path (a published export) would have lost the diagnostic entirely. The branches are exclusive so the CLI still logs it once, and both modes now warn at assembly time rather than after `listen`. Tests cover both: the assembled branch warns naming the keys, the prebuilt branch stays silent.

packages/agent-bff/src/build-bff.ts
nbouliol: Right, `app.callback()` is terminal. The doc comment now names the constraint: mounted at the root of the host server, terminal handler that answers 404 itself instead of yielding to a host `next()`, absolute paths, so a prefix mount would move `/health` and `/agent/...` off the paths the frontend calls — until `basePath` lands later in the stack.

Tonours (approved):
All four findings verified on 23ccdc5: the ordering test now pins `x-forest-bff-version` on the /health 200 and the old test is renamed to what it asserts; `BFFHttpServerOptions` is a union so `callback` + `middlewares`/`version` no longer compiles; `warnMissingConfig` is shared by both modes and gone from `start()`; the `buildBff` doc names the terminal-handler and absolute-path constraint.

PR #1871, opened 2026-09-01
fix(agent-bff): resolve the environment id on demand, not at boot

nbouliol:
Stacked on #1870.

## Why

`buildOAuthMiddlewares` awaited `serverClient.fetchEnvironmentId()` while assembling the edge. That made one network call the only fatal boot error of a server whose documented posture is the opposite — "a required var that is absent does not crash the server, it boots and reports the gap through /health".

Worse, the failure mode is silent and permanent: a Forest server unreachable for five seconds at startup leaves OAuth and the AI relay unmounted for the life of the process, with nothing to restart since the process is otherwise healthy.

## What

`createEnvironmentIdResolver` resolves on first use, caches on success, and never caches a failure — concurrent callers share the in-flight fetch. A transient outage now heals itself on the next request.

The three consumers take a resolver instead of a number:

- `/oauth/authorize` — a failed resolution is turned into `server_error` and redirected back to the client, which is what RFC 6749 §4.1.2.1 asks for once the redirect_uri is validated (it is, several lines earlier), rather than a bare 500.
- the AI relay resolves inside the try that already maps Forest server failures.
- `/agent/v1/context` keeps answering and drops the field, with a `Warn`: the id decorates the payload, it is not what the route is for.

This changes the standalone BFF too, on purpose: one behavior, one code path. `createOAuthRoutes` is a public export and its options change accordingly.

## Tests

81 suites, 1355 tests. Two `cli-core` tests asserted the old eager fetch and were rewritten to the new intent — the boot makes no network call, and an unreachable Forest server no longer prevents booting. New `environment-id` suite covers cache-on-success, retry-after-failure and in-flight sharing; `oauth-routes` covers the `server_error` redirect.

Fixes PRD-1076

> [!NOTE]
> ### Resolve Forest environment id lazily at request time instead of boot time
> - Environment id is now fetched on demand via `createEnvironmentIdResolver` in [environment-id.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1871/files#diff-76704ca6c9a7c873fd05c70cf7c018cec00c0f9feeccbedf2c35a865a0b9c089), with indefinite success caching, 5-second failure caching, and shared concurrent fetches
> - The AI route, context route, and OAuth authorization handler all resolve the environment id per-request instead of using a value captured at startup; resolution failures return typed `environment_unresolved` errors (502) or timeout-mapped errors (504)
> - Adds `buildBff` in [build-bff.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1871/files#diff-44122a133e23acfd87a9ef4d51935c661e42d5145a703f50e58b3ce1acb1b81e) as a public async BFF callback builder, and refactors `BFFHttpServer` in [bff-http-server.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1871/files#diff-883f86539854a7042f72689b857c7561e2876a7b8ec1e8fd028aee8892758569) to accept either a prebuilt callback or legacy options
> - CLI startup now succeeds even when the Forest server is unreachable; environment fetch uses a 5-second timeout and classifies 4xx responses (except 429) as permanent failures
> - Risk: `BFFHttpServer` prebuilt callbacks no longer receive server-added health or version middleware — those must be included in the callback itself via `buildBff`; consumers using the legacy constructor path still get them automatically
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 05eea7a.</sup>
> 
>

### DIFF
--- a/packages/agent-bff/src/ai/ai-routes-middleware.ts
+++ b/packages/agent-bff/src/ai/ai-routes-middleware.ts
@@ -11,11 +11,13 @@ import { AiProxyTimeoutError } from './ai-proxy-client';
 import { requireRenderingId } from '../auth/auth-mode';
 import { unauthorized } from '../http/bff-http-error';
 import {
+  environmentUnresolved,
   oauthRequired,
   upstreamError,
   upstreamTimeout,
   upstreamUnreachable,
 } from '../http/bff-local-errors';
+import { environmentFailureLevel } from '../oauth/environment-fetch-error';
 import { OAuthRequestError } from '../oauth/oauth-error';
 import ensureFreshServerAccess from '../oauth/session-lifecycle';
 
@@ -54,6 +56,32 @@ function describeFailure(error: unknown, depthLeft = MAX_CAUSE_DEPTH): string {
   const tail = walkable ? ` (cause: ${describeFailure(cause, depthLeft - 1)})` : '';
 
   return `${error.name}: ${clamp(error.message)}${tail}`;
+}
+
+/**
+ * Resolved outside the `try` that maps AI proxy failures: a Forest server refusing the environment
+ * read is not the AI proxy being unreachable, and telling the operator otherwise sends them
+ * debugging a component that was never contacted.
+ */
+async function resolveEnvironmentIdOrRefuse(
+  resolveEnvironmentId: EnvironmentIdResolver | undefined,
+  logger: Logger,
+): Promise<number | undefined> {
+  if (!resolveEnvironmentId) return undefined;
+
+  try {
+    return await resolveEnvironmentId();
+  } catch (error) {
+    logger(
+      environmentFailureLevel(error),
+      'AI query refused: the Forest environment id could not be resolved',
+      { cause: describeFailure(error) },
+    );
+
+    if (error instanceof Error && error.name === 'TimeoutError') throw upstreamTimeout();
+
+    throw environmentUnresolved();
+  }
 }
 
 async function resolveSessionAccessToken(
@@ -125,12 +153,14 @@ export default function createAiRoutesMiddleware({
       logger,
     );
 
+    const environmentId = await resolveEnvironmentIdOrRefuse(resolveEnvironmentId, logger);
+
     let response: AiProxyResponse;
 
     try {
       response = await client.query({
         saasAccessToken,
-        environmentId: await resolveEnvironmentId?.(),
+        environmentId,
         renderingId,
         body: ctx.request.body,
       });
--- a/packages/agent-bff/src/ai/ai-routes-middleware.ts
+++ b/packages/agent-bff/src/ai/ai-routes-middleware.ts
@@ -1,6 +1,7 @@
 import type AiProxyClient from './ai-proxy-client';
 import type { AiProxyResponse } from './ai-proxy-client';
 import type { BffAccessTokenPayload } from '../oauth/bff-token';
+import type { EnvironmentIdResolver } from '../oauth/environment-id';
 import type ForestServerClient from '../oauth/forest-server-client';
 import type { SessionStore } from '../oauth/session-store';
 import type { Logger } from '../ports/logger-port';
@@ -28,7 +29,7 @@ export interface AiRoutesMiddlewareOptions {
   client: AiProxyClient;
   sessionStore: SessionStore;
   serverClient: ForestServerClient;
-  environmentId?: number;
+  resolveEnvironmentId?: EnvironmentIdResolver;
   logger: Logger;
 }
 
@@ -104,7 +105,7 @@ export default function createAiRoutesMiddleware({
   client,
   sessionStore,
   serverClient,
-  environmentId,
+  resolveEnvironmentId,
   logger,
 }: AiRoutesMiddlewareOptions): Middleware {
   return async function aiRoutesMiddleware(ctx, next) {
@@ -129,7 +130,7 @@ export default function createAiRoutesMiddleware({
     try {
       response = await client.query({
         saasAccessToken,
-        environmentId,
+        environmentId: await resolveEnvironmentId?.(),
         renderingId,
         body: ctx.request.body,
       });
--- a/packages/agent-bff/src/build-bff.ts
+++ b/packages/agent-bff/src/build-bff.ts
@@ -1,4 +1,5 @@
 import type { BFFConfig } from './config/env-config';
+import type { EnvironmentIdResolver } from './oauth/environment-id';
 import type { SessionStore } from './oauth/session-store';
 import type { UnfoldSource } from './openapi/unfolded-document';
 import type { Logger } from './ports/logger-port';
@@ -31,6 +32,7 @@ import BODY_LIMIT, { AI_BODY_LIMIT } from './http/body-limit';
 import createErrorMiddleware from './http/error-middleware';
 import createHealthRoute from './http/health-route';
 import createVersionHeaderMiddleware from './http/version-header-middleware';
+import createEnvironmentIdResolver from './oauth/environment-id';
 import ForestServerClient from './oauth/forest-server-client';
 import createOAuthRoutes from './oauth/oauth-routes';
 import createInMemorySessionStore from './oauth/session-store';
@@ -156,11 +158,11 @@ interface OAuthSession {
 
 interface OAuthEdge {
   middlewares: Middleware[];
-  environmentId?: number;
+  resolveEnvironmentId?: EnvironmentIdResolver;
   session?: OAuthSession;
 }
 
-async function buildOAuthMiddlewares(config: BFFConfig, logger: Logger): Promise<OAuthEdge> {
+function buildOAuthMiddlewares(config: BFFConfig, logger: Logger): OAuthEdge {
   const oauthConfig = resolveOAuthConfig(config);
 
   if (!oauthConfig) {
@@ -173,7 +175,7 @@ async function buildOAuthMiddlewares(config: BFFConfig, logger: Logger): Promise
     oauthConfig;
 
   const serverClient = new ForestServerClient({ forestServerUrl, envSecret: forestEnvSecret });
-  const environmentId = await serverClient.fetchEnvironmentId();
+  const resolveEnvironmentId = createEnvironmentIdResolver(serverClient);
 
   const sessionStore = createInMemorySessionStore({
     cipher: createTokenCipher(tokenEncryptionKey),
@@ -186,13 +188,13 @@ async function buildOAuthMiddlewares(config: BFFConfig, logger: Logger): Promise
     sessionStore,
     forestAppUrl,
     authSecret: forestAuthSecret,
-    environmentId,
+    resolveEnvironmentId,
     logger,
   });
 
   return {
     middlewares: [oauthRoutes],
-    environmentId,
+    resolveEnvironmentId,
     session: { store: sessionStore, serverClient, forestServerUrl },
   };
 }
@@ -340,7 +342,7 @@ function buildAgentRouteMiddlewares(
 }
 
 function buildAiMiddlewares(config: BFFConfig, oauth: OAuthEdge, logger: Logger): Middleware[] {
-  const { session, environmentId } = oauth;
+  const { session, resolveEnvironmentId } = oauth;
 
   if (!session) {
     logger('Warn', 'AI query route disabled: the deployment carries no OAuth session');
@@ -358,10 +360,35 @@ function buildAiMiddlewares(config: BFFConfig, oauth: OAuthEdge, logger: Logger)
       client,
       sessionStore: session.store,
       serverClient: session.serverClient,
-      environmentId,
+      resolveEnvironmentId,
       logger,
     }),
   ];
+}
+
+/**
+ * The context payload carries the environment id when there is one, and drops it when the Forest
+ * server cannot be reached — the route itself must keep answering either way.
+ */
+function toleratedEnvironmentId(
+  oauth: OAuthEdge,
+  logger: Logger,
+): (() => Promise<number | undefined>) | undefined {
+  const { resolveEnvironmentId } = oauth;
+
+  if (!resolveEnvironmentId) return undefined;
+
+  return async function resolveOrForget(): Promise<number | undefined> {
+    try {
+      return await resolveEnvironmentId();
+    } catch (error) {
+      logger('Warn', 'Serving the context without an environment id', {
+        cause: error instanceof Error ? error.message : String(error),
+      });
+
+      return undefined;
+    }
+  };
 }
 
 function buildAgentMiddlewares(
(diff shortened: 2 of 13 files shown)


packages/agent-bff/src/oauth/environment-id.ts
```diff
@@ -0,0 +1,33 @@
...
+    if (environmentId !== undefined) return environmentId;
+
+    inFlight ??= client
+      .fetchEnvironmentId()
+      .then(resolved => {
+        environmentId = resolved;
+
+        return resolved;
+      })
+      .finally(() => {
```
Tonours, line 56: Claude Opus 5 (claude-opus-5): **Must fix**

While the Forest server *hangs* — a black-holed connection, a load balancer that never answers, exactly the "briefly unreachable" case the comment above targets — every `GET /agent/v1/context` blocks for up to 60 s before answering 200 without the id, every `/oauth/authorize` blocks 60 s before redirecting, and every AI query burns 60 s before the AI proxy is even contacted. The route this design exists to keep answering takes a minute to answer.

Mechanism: the resolver has no deadline of its own and inherits `REQUEST_TIMEOUT_MS = 60_000` (`forest-server-client.ts:37`). Never caching a failure is the right call against bricking, but with no failure memory the cost is paid per request, and the in-flight sharing only covers *concurrent* callers — two sequential bootstraps pay 60 s each. Each one also emits another identical `Warn`, so the log fills at request rate.

`ECONNREFUSED` fails fast, so this is invisible in the common case; nothing bounds the case that hangs.

Two fixes, either works: give this lookup its own short deadline (it is a configuration read, not a user operation — a few seconds), or add a short negative TTL (5–10 s) so a burst pays one round trip instead of one each. The negative TTL additionally protects the SaaS from one retry per request.

packages/agent-bff/src/ai/ai-routes-middleware.ts
```diff
@@ -129,7 +130,7 @@ export default function createAiRoutesMiddleware({
     try {
       response = await client.query({
         saasAccessToken,
-        environmentId,
+        environmentId: await resolveEnvironmentId?.(),
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**

With a wrong `FOREST_ENV_SECRET`, the client is told `The Forest server could not be reached` and the log line names the AI query — so the operator debugs the AI proxy, which was never contacted. The real failure is `Failed to fetch environment: 401 Unauthorized` (`forest-server-client.ts:65`): the Forest server *is* reachable and is refusing. The message is not imprecise, it is wrong.

Mechanism: this `await` sits in `client.query`'s argument list, therefore inside the `try` opened on the line above. A resolver rejection lands in the `catch` below, which logs `AI query failed against the Forest server` and throws `upstreamUnreachable()` — `502 network_error` (`bff-local-errors.ts:72`). At the merge base `environmentId` was a plain number and this `try` covered only `client.query`.

Secondary, same cause: a resolver timeout is a `DOMException` named `TimeoutError`, not an `AiProxyTimeoutError`, so the `instanceof` check misses it and a 60 s environment-id stall returns `502 network_error` instead of `504 upstream_timeout`.

Resolve before the `try` (or in its own `try`) and give it its own message, the way the OAuth path already does.

packages/agent-bff/src/build-bff.ts
```diff
@@ -336,20 +338,44 @@ function buildAiMiddlewares(config: BFFConfig, oauth: OAuthEdge, logger: Logger)
...
+  oauth: OAuthEdge,
+  logger: Logger,
+): (() => Promise<number | undefined>) | undefined {
+  const { resolveEnvironmentId } = oauth;
+
+  if (!resolveEnvironmentId) return undefined;
+
+  return async function resolveOrForget(): Promise<number | undefined> {
+    try {
+      return await resolveEnvironmentId();
```
Tonours: Claude Opus 5 (claude-opus-5): **Violates conventions** — `skills/conventions/testing.md#Cover error and edge paths, not only the happy path`

This `catch` is the only thing standing between a Forest server outage and a broken front bootstrap, and nothing exercises it. `git grep 'Serving the context without an environment id'` at this SHA hits this definition only.

The branch is reachable: `fetchEnvironmentId` rejects on an unreachable fetch and on `!response.ok`, the resolver propagates it with no internal catch, and `test/cli-core.test.ts` now proves the process boots with `global.fetch` rejecting — so the context route is mounted and live with a rejecting resolver. The `catch` fires on the first `GET /agent/v1/context` in that state.

What no test asserts: that the route answers 200 with `meta` carrying `schemaRevision` and *no* `environmentId`, and that the `Warn` carries the cause. `test/context/context-routes-middleware.test.ts:36` always wires a resolver that resolves, and `cli-core.test.ts` stops at `expect(server).toBeDefined()` without issuing a request. That suite already wires `createErrorMiddleware`, so the missing case is a few lines — `test/oauth/oauth-routes.test.ts:160-176` does exactly this for the OAuth path and is the model.

packages/agent-bff/src/build-bff.ts
```diff
@@ -151,7 +153,7 @@ async function buildOAuthMiddlewares(config: BFFConfig, logger: Logger): Promise
     oauthConfig;
 
   const serverClient = new ForestServerClient({ forestServerUrl, envSecret: forestEnvSecret });
-  const environmentId = await serverClient.fetchEnvironmentId();
+  const resolveEnvironmentId = createEnvironmentIdResolver(serverClient);
```
Tonours, line 178: Claude Opus 5 (claude-opus-5): **Should fix**

To be clear about what this is *not*: that boot no longer fails is deliberate and the ticket argues it well, so no objection there. The gap is that nothing replaced the diagnostic. A permanently wrong `FOREST_ENV_SECRET` (a 401, which will never succeed) and a 30-second SaaS blip produce the **same `Warn` line, at request rate**, and telling them apart means reading the `cause` field of each occurrence. The ticket is silent on these semantics, so this is a choice this PR makes rather than one it inherits.

An operator watching a deployment therefore has no signal that distinguishes "this environment will never resolve" from "the SaaS wobbled".

Proportionate fix that keeps the lazy behaviour intact: fire one eager resolution at boot whose *only* job is to log — `.catch` attached so it cannot become an unhandled rejection, `Error` level on a 4xx, `Warn` on a network failure. Nothing becomes fatal; the deployment just says once what it found.

packages/agent-bff/test/context/context-routes-middleware.test.ts
```diff
@@ -32,7 +32,9 @@ function makeRouteOnlyApp(fetchSchema: jest.Mock, environmentId?: number, now?:
 
   const app = new Koa();
   app.use(createErrorMiddleware({ logger: () => {} }));
-  app.use(createContextRoutesMiddleware({ store, environmentId }));
+  app.use(
+    createContextRoutesMiddleware({ store, resolveEnvironmentId: async () => environmentId }),
```
Tonours, line 37: Claude Opus 5 (claude-opus-5): **Violates conventions** — `skills/conventions/testing.md#Test name states the exact behavior it asserts`

**Applies to:** `packages/agent-bff/test/context/context-routes-middleware.test.ts:107` (not in this diff) — anchored on the wiring this PR changed, which is what makes the name below stale.

Line 107 still reads `it('should carry the environment id the deployment resolved at boot', …)`, in the PR titled *not at boot*. This line changed its wiring from a boot-time value to a per-request resolver, so "resolved at boot" is the one thing the test no longer proves.

Consequence: the file someone opens to learn how the context payload gets its environment id documents the semantics this PR deletes — and the next person reasoning about caching or refresh reads "at boot" as the contract. Line 115 (`when the deployment resolved none`) carries the same flavour, less wrongly.

packages/agent-bff/src/oauth/oauth-routes.ts
```diff
@@ -110,6 +112,22 @@ function mapIdentityError(error: unknown): OAuthRequestError {
...
 
+/**
+ * The id is only needed once the redirect_uri is trusted, so a resolution failure travels back to
+ * the client as `server_error` (RFC 6749 §4.1.2.1) instead of a bare 500.
+ */
+async function resolveEnvironmentId(options: OAuthRoutesOptions): Promise<number> {
+  try {
+    return await options.resolveEnvironmentId();
+  } catch (error) {
+    options.logger('Warn', 'Could not resolve the Forest environment id', {
```
Tonours: Claude Opus 5 (claude-opus-5): **Preferential**

The redirect itself is right — RFC 6749 §4.1.2.1, and the id is only needed once the `redirect_uri` is trusted, exactly as the comment says. The consequence worth accepting knowingly: `/oauth/authorize` then answers **302**, so any uptime check or error-rate alert keyed on status codes sees a healthy route while every login is broken, and the only server-side trace is this `Warn`.

A login that cannot start is not a warning. `Error` here costs nothing and is the difference between an incident someone notices and one a user reports.

Tonours (commented):
**Spec (PRD-1076)**: conforms to step 2. The eager `await serverClient.fetchEnvironmentId()` is gone from every path — `buildOAuthMiddlewares` is no longer `async` and `fetchEnvironmentId` has exactly one caller left — and the resolver is threaded through all three consumers. Two clarifications on things that look like gaps and are not: the ticket's "answers 502 on that request" and this PR's `server_error` redirect are the *same object* (`serverError()` builds an `OAuthRequestError(502, 'server_error', …)`), so there is no divergence there; and the per-consumer failure policy for the AI and context routes is this PR's own design — the ticket only says those two "already take it optional" and is silent on what they do when resolution fails.

The resolver itself is right, and I checked the three things that usually go wrong in this shape: `inFlight ??=` is assigned synchronously before any `await`, so concurrent callers share one fetch; `.finally` clears the slot, so no rejection is ever cached; and the guard is `!== undefined` rather than truthiness, which combined with `forest-server-client.ts:77` rejecting `<= 0` means no falsy id can reach a JWT payload or the authorize URL. No unhandled rejection on any path.

What the findings are about is not the mechanism but its cost under failure, and the fact that the one genuinely new branch has no test.

Six findings inline, one below.

---

Claude Opus 5 (claude-opus-5): **Should fix**

This PR decides a failure policy that now differs per consumer — boot no longer fails at all, the context route degrades by dropping the id, `/oauth/authorize` surfaces `server_error`, and the AI route relays an upstream error. Three policies for one value, decided here, recorded in no ADR: the ADR search returns nothing applicable for this area.

All three gates are met — it is hard to reverse (operators and front clients build on these semantics), surprising without context (nothing in the code explains why the three consumers differ), and a real trade-off that the ticket itself argues explicitly (self-healing against a loud boot failure). Worth a pass of `/adr` so the next person changing one of the three knows the other two were deliberate.

packages/agent-bff/src/oauth/environment-id.ts
nbouliol: Fixed both ways in 2718d1b: `fetchEnvironmentId` now carries its own `ENVIRONMENT_TIMEOUT_MS = 5_000` instead of inheriting the 60 s request deadline, and the resolver remembers a failure for `FAILURE_TTL_MS = 5_000` so a burst pays one round trip instead of one each while still healing on the next request — the log line stays per refused request on purpose, since a request that was refused deserves a trace.

packages/agent-bff/src/ai/ai-routes-middleware.ts
nbouliol: Fixed in 2718d1b: the resolution moved out of that `try` into `resolveEnvironmentIdOrRefuse`, which logs `AI query refused: the Forest environment id could not be resolved` and throws `502 environment_unresolved`, or `504 upstream_timeout` when the failure is a `TimeoutError` — so the AI proxy is no longer blamed for a Forest server refusal.

packages/agent-bff/src/build-bff.ts
nbouliol: Fixed in 2718d1b: the wrapper moved to `environment-id.ts` as `tolerateEnvironmentIdFailure`, and `context-routes-middleware.test.ts` now wires it over a rejecting resolver and asserts the route answers 200 with `meta` carrying only `schemaRevision`, the four collections, and the log line with its cause — plus three level-mapping cases in `environment-id.test.ts`.

packages/agent-bff/src/build-bff.ts
nbouliol: The diagnostic gap is real and fixed in 2718d1b, but not with an eager boot resolution: that reintroduces the boot-time network call this PR exists to remove, and `cli-core.test.ts:133-138` asserts the boot makes none. Instead the failure now carries its own verdict — `fetchEnvironmentId` throws `EnvironmentFetchError` marked permanent on a 4xx other than 429 and on an unparseable body — so a wrong `FOREST_ENV_SECRET` logs at `Error` and a SaaS blip stays a `Warn`, which is the signal an operator was missing, and the negative TTL keeps it from repeating at request rate.

packages/agent-bff/test/context/context-routes-middleware.test.ts
nbouliol: Fixed in 2718d1b: renamed to `should carry the environment id the resolver returns` and `should omit the environment id when the resolver returns none`.

packages/agent-bff/src/oauth/oauth-routes.ts
nbouliol: Agreed and changed in 2718d1b: `oauth-routes.ts:125` logs at `Error` whatever the cause, with the comment saying why — the 302 means no status-code alert can fire while every login is broken.

nbouliol:
On the ADR finding: the repository has no ADR practice to add to — `find . -iname "*adr*"` outside `node_modules` returns nothing, and there is no `docs/adr`, so this PR would have to establish the convention rather than follow it. Standing up that infrastructure inside one of eight stacked PRs is out of scope here.

The substance of the finding is addressed in code instead, in 2718d1b: the three failure policies are now documented together in `environment-id.ts` (the doc comment on `createEnvironmentIdResolver`) rather than being three unexplained decisions in three files — `/oauth/authorize` cannot mint a login without the id, the AI relay cannot address the proxy without it, and `/agent/v1/context` merely decorates its payload with it. That is the file anyone changing one of the three opens first.

Tonours (approved):
All six findings verified on 2718d1b: `ENVIRONMENT_TIMEOUT_MS = 5_000` plus a 5s negative TTL bound the hang; the AI route resolves outside the proxy try and answers `environment_unresolved` / `upstream_timeout`; `tolerateEnvironmentIdFailure` is covered on the context route; `EnvironmentFetchError.permanent` gives the Error/Warn split I asked for without a boot-time call; the context test names match their assertions; the OAuth failure logs at Error.

Tonours (approved):
Delta since my last review (2718d1b → cf5cc6d0): clean rebase onto #1870's fix plus one real change — `warnMissingConfig` now also fires inside `buildBff`, with `buildOAuthMiddlewares` no longer awaited. Verified: no double warning (the `BFFHttpServer` call site sits behind the prebuilt early-return, so the assembled-without-buildBff flow and the buildBff flow each warn exactly once), and the sync call matches the now non-async signature. Approving.

PR #1873, opened 2026-09-01
feat(agent-bff): let the emitted paths carry a mount prefix

nbouliol:
Stacked on #1872.

## Why

Two places assume the BFF owns the origin root:

- `renderDocsPage` embeds absolute paths — `/agent/openapi.json` and `/docs/redoc.standalone.js` — and a browser resolves those against the origin, not against the page.
- the OpenAPI document declares `servers: [{ url: '/' }]`, which is what a generated client uses as its base url.

Serve the BFF under `/bff` and both are wrong: the viewer fetches a document that is not there, and a generated client targets paths that 404.

## What

`buildBff({ basePath })`, default `''`, threaded into the docs page and the document's `servers`.

`basePath` is **emitted, never routed on**: the host strips its own prefix before the request reaches this app, so every route stays exactly where it was — there is a test for that, since it is the part that would be easy to get wrong later.

`basePath` is normalized once, at the entry (`normalizeBasePath`, modelled on `mcp-server`'s `normalizeMountPath`): `''` and `'/'` mean the origin root, a leading slash is optional, duplicate and trailing slashes collapse, and anything that is not a plain path prefix throws at boot. `'/'` is the input that mattered — interpolated raw it emits `//agent/openapi.json`, which a browser resolves with `agent` as the host.

The docs page's code samples read the prefix off the document's own `servers[0].url` rather than assuming `location.origin` is the base, so the snippets and a generated client cannot disagree.

## Tests

83 suites, 1390 tests. `build-bff` covers the prefixed and unprefixed page (delimiters anchored, so a doubled prefix fails), the `'/'` case, a rejected prefix, and the invariant that routing is unaffected; `openapi-routes` asserts `servers` on the *served* document in both the unfolded and generic branches; `base-path` covers normalization and rejection; `docs-page` asserts the emitted sample urls for a prefix, the root, and a server url that already carries a host.

Fixes PRD-1076

> [!NOTE]
> ### Add mount prefix to `agent-bff` emitted paths and introduce `AgentTransport` abstraction
> - Adds `normalizeBasePath` and `resolveEmittedBase` utilities to derive the externally visible BFF mount prefix from the request URL and a configured `basePath` option on `buildBff`.
> - Passes the mount prefix into OpenAPI document generation ([openapi-document.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1873/files#diff-fe205dfba3d9163361b5101e76cdd78dc3bd2c2eb7a03e54d0bdc112b2b5584c)), OpenAPI route caching ([openapi-routes.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1873/files#diff-98b5421ed1d92a87c274e0812f3f8611271019c97f94cc5a913b39f6603a3b06)), docs route rendering ([docs-routes.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1873/files#diff-3dcba5d66b2d959c4b9880d5839f46457c1ce973af10e1fef317c2d9603312bf)), and browser sample URL construction ([docs-samples.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1873/files#diff-df99f9972657ae5a6fd5da1c6cde3856ecca90542fe76d8ce83fb20709bcf61c)).
> - Replaces per-component `agentUrl` and `timeoutMs` fields with a shared `AgentTransport` interface across data, action, and capability-fetcher middleware ([agent-transport.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1873/files#diff-94f244014974fc44b816827aa84f69a67e80b9a1b60c556753af7bb3b5fb7add)).
> - Adds `createInProcessTransport` and `InProcessRequester` to dispatch agent calls without a network socket, including path/query parity with the HTTP requester and a 501 `streaming_unsupported` error for stream calls ([in-process-transport.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1873/files#diff-2f0a4f464a7053c092f1ab0cc5e113ce50ebc32cdaa0ecf68945fa00a8c1634f)).
> - Risk: `buildBff` now rejects invalid `basePath` values (absolute URLs, query strings, fragments) with `ConfigurationError` at construction time; the OpenAPI and docs route caches are now keyed by base path, so consumers relying on a single cached document per process will see per-prefix memoization instead.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0671c9f.</sup>
> 
>

### DIFF
--- a/packages/agent-bff/src/build-bff.ts
+++ b/packages/agent-bff/src/build-bff.ts
@@ -22,6 +22,7 @@ import ApiKeyClient from './api-key/api-key-client';
 import createApiKeyMiddleware from './api-key/api-key-middleware';
 import createResolveCache from './api-key/resolve-cache';
 import createAuthModeMiddleware from './auth/auth-mode-middleware';
+import normalizeBasePath from './base-path';
 import warnMissingConfig from './config/missing-config-warning';
 import createContextRoutesMiddleware from './context/context-routes-middleware';
 import createCorsMiddleware from './cors/cors-middleware';
@@ -58,6 +59,9 @@ export interface BuildBffOptions {
    * Prefix the host serves this BFF under, `/bff` for an embedded one. Routing is the host's job —
    * it strips the prefix before the request lands here — but the paths the BFF *emits* (the OpenAPI
    * `servers` entry, the docs page's asset and document urls) must carry it.
+   *
+   * A plain path prefix, with or without its leading slash; `'/'` and `''` both mean the BFF owns the
+   * origin root. Normalized by `normalizeBasePath`, which throws on anything else.
    */
   basePath?: string;
 }
@@ -442,8 +446,12 @@ function buildAgentMiddlewares(
 export default async function buildBff({
   config,
   logger = createConsoleLogger(),
-  basePath = '',
+  basePath,
 }: BuildBffOptions): Promise<Bff> {
+  // Before anything is assembled: a mount the host does not serve must fail at boot, not surface as
+  // a docs page that cannot load itself.
+  const mountPath = normalizeBasePath(basePath);
+
   if (config.invalidAllowedOrigins.length > 0) {
     logger('Warn', 'Ignoring malformed BFF_ALLOWED_ORIGINS entries', {
       entries: config.invalidAllowedOrigins,
@@ -454,7 +462,7 @@ export default async function buildBff({
 
   const oauth = buildOAuthMiddlewares(config, logger);
   const aiMiddlewares = buildAiMiddlewares(config, oauth, logger);
-  const agentMiddlewares = buildAgentMiddlewares(config, logger, oauth, aiMiddlewares, basePath);
+  const agentMiddlewares = buildAgentMiddlewares(config, logger, oauth, aiMiddlewares, mountPath);
   const hasAgentEdge = agentMiddlewares.length > 0;
   const agentErrorMiddleware = hasAgentEdge ? [agentScoped(createErrorMiddleware({ logger }))] : [];
   const agentJsonOnlyGuard = hasAgentEdge ? [agentScoped(createJsonOnlyGuard())] : [];
@@ -473,7 +481,7 @@ export default async function buildBff({
     createDocsRoutes({
       enabled: config.openapiEnabled && agentMiddlewares.length > 0,
       documentPath: OPENAPI_PATH,
-      basePath,
+      basePath: mountPath,
       logger,
     }),
     ...agentMiddlewares,
--- a/packages/agent-bff/src/build-bff.ts
+++ b/packages/agent-bff/src/build-bff.ts
@@ -54,6 +54,12 @@ export type BffCallback = (req: IncomingMessage, res: ServerResponse) => void | 
 export interface BuildBffOptions {
   config: BFFConfig;
   logger?: Logger;
+  /**
+   * Prefix the host serves this BFF under, `/bff` for an embedded one. Routing is the host's job —
+   * it strips the prefix before the request lands here — but the paths the BFF *emits* (the OpenAPI
+   * `servers` entry, the docs page's asset and document urls) must carry it.
+   */
+  basePath?: string;
 }
 
 export interface Bff {
@@ -376,6 +382,7 @@ function buildAgentMiddlewares(
   logger: Logger,
   oauth: OAuthEdge,
   aiMiddlewares: Middleware[],
+  basePath: string,
 ): Middleware[] {
   const { forestAuthSecret, defaultTimezone } = config;
 
@@ -405,6 +412,7 @@ function buildAgentMiddlewares(
       source,
       hasAiQueryRoute: aiMiddlewares.length > 0,
       publicUrl: config.publicUrl,
+      basePath,
     }),
     ...(bundle
       ? [
@@ -434,6 +442,7 @@ function buildAgentMiddlewares(
 export default async function buildBff({
   config,
   logger = createConsoleLogger(),
+  basePath = '',
 }: BuildBffOptions): Promise<Bff> {
   if (config.invalidAllowedOrigins.length > 0) {
     logger('Warn', 'Ignoring malformed BFF_ALLOWED_ORIGINS entries', {
@@ -445,7 +454,7 @@ export default async function buildBff({
 
   const oauth = buildOAuthMiddlewares(config, logger);
   const aiMiddlewares = buildAiMiddlewares(config, oauth, logger);
-  const agentMiddlewares = buildAgentMiddlewares(config, logger, oauth, aiMiddlewares);
+  const agentMiddlewares = buildAgentMiddlewares(config, logger, oauth, aiMiddlewares, basePath);
   const hasAgentEdge = agentMiddlewares.length > 0;
   const agentErrorMiddleware = hasAgentEdge ? [agentScoped(createErrorMiddleware({ logger }))] : [];
   const agentJsonOnlyGuard = hasAgentEdge ? [agentScoped(createJsonOnlyGuard())] : [];
@@ -464,6 +473,7 @@ export default async function buildBff({
     createDocsRoutes({
       enabled: config.openapiEnabled && agentMiddlewares.length > 0,
       documentPath: OPENAPI_PATH,
+      basePath,
       logger,
     }),
     ...agentMiddlewares,
--- a/packages/agent-bff/src/docs/docs-routes.ts
+++ b/packages/agent-bff/src/docs/docs-routes.ts
@@ -16,6 +16,11 @@ export interface DocsRoutesOptions {
   enabled: boolean;
   /** Where the shell fetches the document. Passed in so this module never reaches into `src/openapi`. */
   documentPath: string;
+  /**
+   * Prefix the host serves the BFF under. The page carries absolute paths — the browser resolves
+   * them against the origin, not against the page — so they must carry the prefix too.
+   */
+  basePath?: string;
   logger: Logger;
   /** The bundle lookup, as a seam: an install that shipped without the asset is a real state to serve. */
   resolveBundlePath?: () => string | undefined;
@@ -57,6 +62,7 @@ export function resolveBundle(directory: string = __dirname): string | undefined
 export default function createDocsRoutes({
   enabled,
   documentPath,
+  basePath = '',
   logger,
   resolveBundlePath = resolveBundle,
 }: DocsRoutesOptions): Middleware {
@@ -66,7 +72,9 @@ export default function createDocsRoutes({
     logger('Warn', `API documentation page disabled: ${BUNDLE_FILE} is missing from this install`);
   }
 
-  const page = bundle ? renderDocsPage(documentPath, DOCS_BUNDLE_PATH) : undefined;
+  const page = bundle
+    ? renderDocsPage(`${basePath}${documentPath}`, `${basePath}${DOCS_BUNDLE_PATH}`)
+    : undefined;
   let script: string | undefined;
 
   return async function docsRoutes(ctx, next) {
--- a/packages/agent-bff/src/docs/docs-routes.ts
+++ b/packages/agent-bff/src/docs/docs-routes.ts
@@ -5,6 +5,7 @@ import { existsSync, readFileSync } from 'fs';
 import path from 'path';
 
 import renderDocsPage from './docs-page';
+import { emittedBaseOf } from '../base-path';
 
 export const DOCS_PATH = '/docs';
 export const DOCS_BUNDLE_PATH = '/docs/redoc.standalone.js';
@@ -72,9 +73,20 @@ export default function createDocsRoutes({
     logger('Warn', `API documentation page disabled: ${BUNDLE_FILE} is missing from this install`);
   }
 
-  const page = bundle
-    ? renderDocsPage(`${basePath}${documentPath}`, `${basePath}${DOCS_BUNDLE_PATH}`)
-    : undefined;
+  // Rendered per request rather than once: the prefix depends on where the host serves the BFF,
+  // which only the request knows. Memoized per prefix, of which a deployment has one in practice.
+  const pages = new Map<string, string>();
+
+  const pageFor = (base: string): string => {
+    const cached = pages.get(base);
+    if (cached !== undefined) return cached;
+
+    const rendered = renderDocsPage(`${base}${documentPath}`, `${base}${DOCS_BUNDLE_PATH}`);
+    pages.set(base, rendered);
+
+    return rendered;
+  };
+
   let script: string | undefined;
 
   return async function docsRoutes(ctx, next) {
@@ -112,6 +124,6 @@ export default function createDocsRoutes({
     ctx.status = 200;
     ctx.type = 'text/html';
     ctx.set('Cache-Control', 'no-store');
-    ctx.body = page;
+    ctx.body = pageFor(emittedBaseOf(ctx, basePath));
   };
 }
--- a/packages/agent-bff/src/openapi/openapi-document.ts
+++ b/packages/agent-bff/src/openapi/openapi-document.ts
@@ -562,13 +562,20 @@ const RELATIVE_SERVER_DESCRIPTION =
 export interface GenerateOpenApiDocumentOptions {
   unfolding?: Unfolding;
   hasAiQueryRoute?: boolean;
-  /** The deployment's own external base URL, from `BFF_PUBLIC_URL`. Absent falls back to `/`. */
+  /** The deployment's own external base URL, from `BFF_PUBLIC_URL`. Takes precedence. */
   publicUrl?: string;
+  /** Where the BFF answers from the caller's point of view. Empty when it owns the origin root. */
+  basePath?: string;
 }
 
 export function generateOpenApiDocument(
   version: string,
-  { unfolding, hasAiQueryRoute = false, publicUrl }: GenerateOpenApiDocumentOptions = {},
+  {
+    unfolding,
+    hasAiQueryRoute = false,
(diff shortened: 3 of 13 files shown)


packages/agent-bff/src/docs/docs-routes.ts
```diff
@@ -66,7 +72,9 @@ export default function createDocsRoutes({
     logger('Warn', `API documentation page disabled: ${BUNDLE_FILE} is missing from this install`);
   }
 
-  const page = bundle ? renderDocsPage(documentPath, DOCS_BUNDLE_PATH) : undefined;
+  const page = bundle
+    ? renderDocsPage(`${basePath}${documentPath}`, `${basePath}${DOCS_BUNDLE_PATH}`)
```
Tonours: Claude Opus 5 (claude-opus-5): **Must fix**

**Applies to:** `packages/agent-bff/src/docs/docs-samples.ts:330` and `:309` (not in this diff) — anchored here because this is the line that prefixes the *other* emitted URLs on the same page, and the samples are the one site the prefix was not threaded into.

Every copy-pasteable snippet on the docs page of a prefixed deployment targets the wrong URL. Under `basePath: '/bff'` the page is served at `https://host/bff/docs`, renders a document whose paths are right, and emits `curl -X POST 'https://host/agent/v1/<collection>/list'` — 404, or worse a 200 from whatever unrelated app owns the origin root. All three languages, since `curlSample` / `nodeSample` / `rubySample` share the origin.

Mechanism: `decorateWithSamples(spec, window.location.origin)` — `location.origin` is scheme + host and never a path — composed with the operation path, which this PR deliberately leaves unprefixed. `spec.servers[0].url`, the one place the prefix now lives, is never read. Redoc's own base-URL display is correct while the snippet next to it is wrong, which is the worst version of this.

The module's own header comment is the reasoning that hides it: *"the document declares `servers: [{ url: '/' }]`, so a sample built into it could only hold a placeholder host, while the page knows where it is served from"*. After this PR the document does not declare `/`, and the page does not know where it is served from. That comment needs updating with the fix.

One line, and it makes the samples self-consistent with the `servers` entry this PR adds rather than adding plumbing:

```js
var mount = ((spec.servers || [])[0] || {}).url || '/';
return decorateWithSamples(spec, window.location.origin + (mount === '/' ? '' : mount));
```

Found independently by three separate review passes; the test gap that let it through is that `test/build-bff.test.ts` asserts the document and bundle URLs carry `/bff` and nothing asserts a sample URL.

packages/agent-bff/src/build-bff.ts
```diff
@@ -428,6 +436,7 @@ function buildAgentMiddlewares(
 export default async function buildBff({
   config,
   logger = createConsoleLogger(),
+  basePath = '',
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**

`basePath` is a public option on the exported `BuildBffOptions`, so its value is whatever an embedding host — often an env var — hands over, and it is interpolated raw at every site with no trim, no leading-slash enforcement and no trailing-slash collapse. Executed, not reasoned (resolved from a page at `https://host/docs`):

```
basePath   emitted doc url               resolves to
"/bff"     /bff/agent/openapi.json       https://host/bff/agent/openapi.json   correct
"/"        //agent/openapi.json          https://agent/openapi.json            <-- "agent" read as the HOST
"/bff/"    /bff//agent/openapi.json      https://host/bff//agent/openapi.json  404 after the host strips /bff
"bff"      bff/agent/openapi.json        https://host/bff/agent/openapi.json   page ok, but servers="bff" is relative
```

`'/'` is the input that hurts most: `//agent/openapi.json` is an RFC 3986 network-path reference, so the browser resolves `agent` as a hostname. Same for `<script src="//docs/redoc.standalone.js">` → `https://docs/...`. The page renders blank with a cross-origin failure. `'bff'` is the likeliest typo: the page happens to resolve, but `servers: [{ url: 'bff' }]` is a *relative* server URL that generators resolve against the document location, so every generated client targets a base nobody chose — and the two disagree silently.

The precedent is in this monorepo, for this same concept: `packages/mcp-server/src/mcp-paths.ts:3-23`. `normalizeMountPath` trims, maps `'/'` to `''`, prepends the leading slash, collapses duplicate and trailing slashes, allowlists the characters, and **throws** `Invalid MCP mount path "…": use a plain path prefix like "/mcp"`. Its comment gives the reason directly: the prefix is interpolated raw into route patterns and resolved through `new URL()`, so both interpretations must agree. That is exactly the failure set above.

It also holds the new input to a lower bar than the old one in the same function body: `build-bff.ts` warns and lists the malformed `BFF_ALLOWED_ORIGINS` entries a few lines down. Normalise once here, before anything is interpolated, and reject with a message naming the value.

packages/agent-bff/src/openapi/openapi-document.ts
```diff
@@ -485,7 +487,7 @@ export function generateOpenApiDocument(
         unfolding ? UNFOLDED_DESCRIPTION : GENERIC_DESCRIPTION
       } ${SHARED_DESCRIPTION}`,
     },
-    servers: [{ url: '/' }],
+    servers: [{ url: basePath || '/' }],
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**

**Applies to:** `packages/agent-bff/src/cli-dispatch.ts:99` and `:104` (not in this diff) — anchored on the line whose value the export omits.

The CLI export and the live route now describe the same deployment differently, and both succeed. `forest-bff openapi` calls `generateOpenApiDocument(version, { hasAiQueryRoute })` and `buildUnfoldedDocument(…)` without `basePath`, and no env key carries one — `basePath` exists only as a code-level option. So for a deployment served under `/bff`, the live `/agent/openapi.json` emits `servers: [{ url: '/bff' }]` while the document a CI job commits and generates an SDK from emits `servers: [{ url: '/' }]`.

Who finds out: not whoever ran the export — exit 0, no warning — but whoever consumes the generated client later and gets 404s against `/agent/…`. New with this PR: before it there was no prefix to disagree about.

Either thread a `BFF_BASE_PATH` env key through the export, or have the export state in its output that it assumes the origin root.

packages/agent-bff/test/build-bff.test.ts
```diff
@@ -69,6 +69,45 @@ describe('buildBff', () => {
...
 
+  describe('when the host serves the BFF under a prefix', () => {
+    it('should point the docs page at the prefixed document and bundle', async () => {
+      const { callback } = await buildBff({
+        config: parseConfig(VALID_ENV),
+        logger: noopLogger,
+        basePath: '/bff',
+      });
+
+      const response = await request(callback).get('/docs');
```
Tonours, line 121: Claude Opus 5 (claude-opus-5): **Should fix**

This assertion cannot fail on the bug it exists to catch. `toContain('/bff/agent/openapi.json')` is satisfied by `"/bff/bff/agent/openapi.json"` — verified, `includes` returns `true` — so a regression that applies the prefix twice passes green.

The sibling case for the unprefixed deployment gets this right by anchoring the delimiters (`'"/agent/openapi.json"'`, `'src="/docs/redoc.standalone.js"'`), which is exactly why it *can* fail on a stray prefix. Use the same anchored form here: `'"/bff/agent/openapi.json"'` and `'src="/bff/docs/redoc.standalone.js"'`.

packages/agent-bff/src/openapi/openapi-routes.ts
```diff
@@ -57,6 +60,7 @@ export default function createOpenApiRoutes({
     const { document, unfolding } = await buildUnfoldedDocument(source, readModel, token, {
       version,
       hasAiQueryRoute,
+      basePath,
```
Tonours: Claude Opus 5 (claude-opus-5): **Violates conventions** — `skills/conventions/testing.md#Cover error and edge paths, not only the happy path`

The prefixed variant of the *served* document has no case, and the untested branch is the production one. The only `basePath` assertion on `servers` is the pure-function unit test in `test/openapi/openapi-document.test.ts`; nothing at the routes level asserts the emitted `servers` under a prefix, and the new prefixed block in `test/build-bff.test.ts` only requests `/docs`, never `/agent/openapi.json`.

So dropping or mistyping this `basePath` forward — the unfolded branch, which is what a deployment with an agent URL actually serves — leaves the whole suite green while an embedded deployment publishes a document whose `servers` is `/`. Every client generated from it then targets the host root, and this repo treats generated clients as the contract (`test/openapi/openapi-generated-client.test.ts` exists for that reason), so all data and action calls 404.

Smallest fix: one case asserting `JSON.parse(response.text).servers` equals `[{ url: '/bff' }]` for an authenticated caller with `basePath: '/bff'`.

Same shape one line up in `unfolded-document.ts`: deleting `basePath` from that destructure also leaves every assertion in this PR green.

packages/agent-bff/src/build-bff.ts
```diff
@@ -53,6 +53,12 @@ export type BffCallback = (req: IncomingMessage, res: ServerResponse) => void |
 export interface BuildBffOptions {
   config: BFFConfig;
   logger?: Logger;
+  /**
+   * Prefix the host serves this BFF under, `/bff` for an embedded one. Routing is the host's job —
+   * it strips the prefix before the request lands here — but the paths the BFF *emits* (the OpenAPI
+   * `servers` entry, the docs page's asset and document urls) must carry it.
+   */
+  basePath?: string;
```
Tonours, line 66: Claude Opus 5 (claude-opus-5): **Preferential**

`basePath` is now part of the published surface of `@forestadmin/agent-bff` and is documented nowhere outside this doc comment — not in the package README, not in the docs. Given the normalisation findings above, the two things a host needs stated are the accepted shape (leading slash, no trailing slash) and that it changes only what the BFF *emits*, never what it serves, because the host is expected to strip the prefix before dispatching.

The README lands in a later PR of this stack, so this is a note rather than a request — worth making sure it covers this option when it does.

Tonours (commented):
**Spec (PRD-1076)**: step 4 is half implemented. Its first sentence — thread `basePath` into the two absolute paths in `docs-routes.ts` and into the OpenAPI `servers` — landed, and landed well. Its second sentence did not:

> Derive the emitted base from the request when the host framework exposes it (`req.originalUrl` under Express), falling back to `basePath`: an agent mounted under `app.use('/api', router)` routes correctly (Express strips the mount path) but would otherwise emit `/bff/...` where the browser needs `/api/bff/...`.

`git grep originalUrl -- packages/agent-bff/src` is empty at this SHA and the diff mentions neither `req.` nor `originalUrl`. So the case the ticket wrote that requirement for — a host mounting the agent under its own prefix — still emits the wrong base. Worth saying whether it is deferred on purpose; the static half is a prerequisite for it either way, so this is not a blocker for the PR, just a gap the ticket named.

**What is right, and it is the part I expected to be wrong.** The prefix lands in `servers[].url` exactly once and the operation paths stay unprefixed, so the absolute URL composes correctly with no double-count and no drop. Both document forms get it symmetrically (generic via `openapi-routes.ts`, unfolded via `unfolded-document.ts`). Route *matching* is untouched — `ctx.path` comparisons still use the unprefixed constants — which is what the host-strips design requires, and the test at `test/build-bff.test.ts` guards it.

---

**Applies to:** the PR as a whole — coordination with #1866.

#1866 (`publish BFF_PUBLIC_URL as the openapi server url`, still open on main) writes the same `servers[0].url` this PR writes. They are not just a textual conflict: the two values must be **composed**, not chosen. For a BFF embedded under `/bff` behind `https://app.example.com`, the correct value is `https://app.example.com/bff` — `basePath` alone loses the host, `BFF_PUBLIC_URL` alone loses the mount prefix.

Neither PR's tests exercise both options at once, so whichever merges second can resolve the conflict by keeping its own value and no test will fail. PRD-1076 never mentions `BFF_PUBLIC_URL` at all, so the ticket does not arbitrate it either. Whoever lands second wants one assertion: `basePath: '/bff'` + `BFF_PUBLIC_URL: 'https://host'` → `servers[0].url === 'https://host/bff'`.

packages/agent-bff/src/docs/docs-routes.ts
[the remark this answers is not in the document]
nbouliol: Fixed, but at the entry rather than at the concatenation: `basePath` is now normalized once in `buildBff` through `src/base-path.ts`, so `'/'`, `'/bff/'`, `'//bff//'` and `'bff'` all collapse to a single well-formed prefix (or to `''` for the origin root) and `//agent/openapi.json` can no longer be emitted — `build-bff.test.ts` asserts the `'/'` case renders the unprefixed page.

nbouliol: Fixed: `withSamples` now builds on `sampleBase(spec)`, which composes `window.location.origin` with the document's own `servers[0].url` (a server url that already carries a host is taken as it stands, `/` means the origin root), the module header comment now states that composition instead of the old `servers: [{ url: '/' }]` reasoning, and `docs-page.test.ts` asserts the emitted sample urls for `/bff`, `/` and `https://public.example.com/bff/`.

packages/agent-bff/src/build-bff.ts
nbouliol: Fixed: `normalizeBasePath` (`src/base-path.ts`, modelled on `mcp-paths.ts`) trims, maps `''`/`'/'` to `''`, prepends the leading slash, collapses duplicate and trailing slashes, and throws `ConfigurationError('Invalid BFF base path "…": use a plain path prefix like "/bff" …')` on anything else — called as the first statement of `buildBff`, before any value is interpolated, so a bad mount fails at boot rather than as a docs page that cannot load itself.

packages/agent-bff/test/build-bff.test.ts
nbouliol: Fixed: both assertions now anchor their delimiters (`'"/bff/agent/openapi.json"'` and `'src="/bff/docs/redoc.standalone.js"'`), so a doubled prefix fails instead of passing.

packages/agent-bff/src/openapi/openapi-routes.ts
nbouliol: Fixed: `openapi-routes.test.ts` now asserts `servers` equals `[{ url: '/bff' }]` on the *served* document in both branches — unfolded through the read-model store, and generic — plus that the operation paths stay unprefixed, so dropping the `basePath` forward in either `openapi-routes.ts` or `unfolded-document.ts` now fails.

packages/agent-bff/src/openapi/openapi-document.ts
nbouliol: Not changing this, and I don't think the two disagree: `basePath` reaches the BFF only through `buildBff({ basePath })`, and the CLI's single build site is `cli-core.ts:15` — `buildBff({ config, logger })`, no prefix — so every deployment `forest-bff` can *serve* owns the origin root, and the document it exports (`servers: [{ url: '/' }]`) is exactly what that same binary serves. The host that does set a prefix sets it in code (the embedded agent passes the constant `/bff`), so no env key could carry its value anyway, and adding `BFF_BASE_PATH` for the export alone would create the divergence it is meant to close — that only makes sense as one key threaded through both `cli-core` and `renderOpenApi`, which is a separate change.

packages/agent-bff/src/build-bff.ts
nbouliol: Noted, and no change here: I'll make sure the docs PR of the stack (#1877) states the accepted shape (a plain path prefix, leading slash optional, trailing slash and `/` collapsed, anything else rejected at boot) and that `basePath` only changes what the BFF emits, never what it routes on.

nbouliol:
On the two points in the review body:

**`req.originalUrl` (spec step 4, second sentence)** — deliberately not in this PR, and you're right that it is not implemented anywhere in the stack. Two reasons: the only consumer in the stack passes a static prefix in code (`embedded-bff.ts` hands `buildBff` the constant `/bff`), and a request-derived base is not a one-liner on top of this — the docs page is rendered once at construction and the unfolded document is memoized on the read-model identity, so both would have to be keyed per prefix, with the cache-growth question that comes with a value taken from the request. I'd rather land that as its own change than bolt it onto the static half, so a host that mounts the agent under its own prefix still needs to pass `basePath: '/api/bff'` for now.

**`BFF_PUBLIC_URL` / #1866** — agreed that the two compose rather than compete. Nothing to compose on this branch yet (`git grep BFF_PUBLIC_URL` is empty here), so this PR leaves the value a bare prefix; whoever lands second owns the `basePath: '/bff'` + `BFF_PUBLIC_URL: 'https://host'` → `servers[0].url === 'https://host/bff'` assertion. One thing this round already does for it: the docs-page samples no longer assume `location.origin` is the base — they read `servers[0].url` and take it as-is when it already carries a host, with a test for exactly that shape, so the snippets stay correct whichever composition lands.

Tonours (approved):
All six findings verified on 951f8ba: `normalizeBasePath` runs as the first statement of `buildBff` and rejects anything that is not a plain prefix, so '/', '/bff/' and 'bff' can no longer be emitted; the samples build on `sampleBase(spec)` with cases for /bff, / and an absolute server url; the docs assertions are delimiter-anchored; `openapi-routes.test.ts` pins `servers` on the served document in both branches. The CLI-export argument holds — that binary owns the origin root — and the `basePath` documentation landed in #1877.

Tonours (approved):
Delta since 951f8ba: rebase plus 90001a0 (emit the prefix the caller reached, not the one configured). Verified `resolveEmittedBase` against its edges — query string carried by both urls, root ownership, trailing slash, unrelated rewrite falls back to configured — and the docs/openapi wiring memoizes per prefix off a normalized `mountPath`. The `/api` host-prefix case is asserted end-to-end by the e2e suite added later in this stack. Approving.

PR #1876, opened 2026-09-01
feat(agent): serve a BFF in-process with addBff()

nbouliol:
Stacked on #1875. This is where the stack becomes a feature.

## Why

Running a BFF meant a second deployment: another process, another port, another set of secrets to keep in sync with the agent's — for a component whose only job is to sit in front of that same agent.

## What

```ts
createAgent(options)
  .addDataSource(/* … */)
  .addBff({ allowedOrigins: ['https://my-app.com'] })
  .start();
```

Served at `/bff` on the agent's own port, on every mount target. The BFF reaches the agent through the in-process dispatcher the embedded MCP server already uses: no socket, no agent url to guess per host framework, no second listener.

`authSecret`, `envSecret`, `forestServerUrl`, `forestAppUrl` and the logger are inherited — a divergent `authSecret` would make the agent reject the very tokens the BFF mints, as an opaque 401. What is left are features the BFF switches on: `tokenEncryptionKey` (the OAuth login/refresh flow, and with it the AI relay — it gates that flow, not the data surface, which answers to any `bff_access` bearer signed with `authSecret`), `allowedOrigins` (browser access), `openapiEnabled` (the docs, **off by default** when embedded because the document is not filtered per caller).

Three lifecycle details that are easy to get wrong, and are tested:

- **The dispatcher is registered in `addBff()`, not at `start()`.** `getInProcessDispatcher()` pushes its hook the first time it is called, and `mount()` only runs the hooks registered before it — asked for later, every BFF call would throw `not mounted yet` until the first restart.
- **`/bff` answers 503 while the agent is starting**, rather than falling through to the host's 404, which would read as "wrong url" instead of "not started".
- **`stop()` stops answering.** The host application keeps whatever middleware it registered, so without it a stopped agent would keep serving BFF data through a dispatcher pointing at a dead stack.

`restart()` invalidates what the BFF read from the SaaS — a restart means the customizations changed.

`addBff()` refuses a second call, and refuses to coexist with an MCP server mounted under `/bff` in either order — including the spellings `bff`, `/bff/` and `/bff/ai`, which the MCP server normalizes onto the same prefix. The BFF is registered at builder time and wins the root middleware's first-match, so the MCP surface would boot, log success and never be reachable.

## The build cycle

`agent` gains an optional peer dependency (exact pin, like every internal dep — multi-semantic-release rewrites those ranges on release) plus a dev dependency on `@forestadmin/agent-bff`, which would close the cycle `agent → agent-bff → agent-testing → agent` that `lerna run build` sorts on. It is broken by moving the search integration suite out of agent-bff, which loses its dev dependency on the agent.

`engines: { node: ">=22.12.0" }` is declared on the agent too: `addBff()` pulls in a package that requires it, and the agent declared nothing.

## Tests

agent-bff 82 suites / 1368 tests, agent 87 suites / 1587 tests.

The moved suite becomes `test/bff/embedded-bff.e2e.test.ts`: a real `Agent` over a real datasource, mounted on Express, with a real BFF in front — list, search, relation-extended search, count, and an agent-side refusal surfacing as the BFF error contract. It also pins what the unit tests cannot see: `/forest` still answers next to it, `/bffalo` is not claimed, and a stopped agent answers 503. Its data contract runs twice — over the in-process dispatcher, and over a real socket against a listening agent, so the HTTP transport the standalone deployment uses keeps a gate. The CI job that ran the old suite now runs this one, and the unit job ignores it.

Fixes PRD-1076

> [!NOTE]
> ### Add in-process BFF support to `Agent` via `addBff()`
> - Adds `Agent.addBff()` to configure an in-process embedded BFF, dynamically loading the optional `@forestadmin/agent-bff` package.
> - Introduces the `EmbeddedBff` class to handle BFF build, lifecycle, and request routing under `/bff`.
> - Updates `resolveTransport` to prefer an in-process dispatcher over HTTP when available.
> - Hardens CORS middleware to reject disallowed cross-origin requests with a 403 error before downstream handling.
> - Updates `/health` to report explicit `configured` flags for `oauth`, `ai`, `cors`, and `openapi`.
> - Risk: `Agent.mountAiMcpServer()` and `addBff()` throw if MCP paths overlap `/bff`; `parseConfig` `hasAllRequired` no longer requires `BFF_TOKEN_ENCRYPTION_KEY`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e05ddeb.</sup>
> 
>


packages/agent/src/agent.ts
```diff
@@ -363,6 +386,51 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
...
+   *   .addDataSource(...)
+   *   .addBff({ allowedOrigins: ['https://my-app.com'] })
+   *   .start();
+   */
+  addBff(options: BffEmbedOptions = {}): this {
+    if (this.embeddedBff) {
+      throw new Error('addBff can only be called once.');
+    }
+
+    if (this.mcpBasePath === BFF_PREFIX) {
```
Tonours: Claude Opus 5 (claude-opus-5): **Must fix**

The collision guard is defeated by any non-literal spelling of `/bff`, and when it is, the **entire MCP surface goes silently dark** — the server initialises, logs `Server initialized successfully`, and is never reachable.

Both guards compare the raw option string to `'/bff'`, while `mcp-server` normalises its `basePath` (trim, add the leading slash, collapse and strip trailing slashes) before building its paths. Executed against the shipped helpers:

```
basePath="/bff"     guard trips=true
basePath="/bff/"    guard trips=false   <-- MCP shadowed silently
basePath="/bff/ai"  guard trips=false   <-- MCP shadowed silently
basePath="bff"      guard trips=false   <-- MCP shadowed silently
```

`'bff'` and `'/bff/'` both normalise to `/bff` inside the MCP server, so it claims exactly `/bff/oauth/` and `/bff/mcp` — the collision this error exists to reject. `'/bff/ai'` is accepted by `normalizeMountPath`'s allowlist (it permits multi-segment prefixes) and lands wholly under `/bff`. In every case `isBffRoute` claims the path, so `POST /bff/mcp` returns the **BFF's** 404.

And the comment above `BFF_MCP_COLLISION` has the direction backwards: *"the MCP server … is consulted first, so it would shadow two of the BFF's own routes"*. It is the other way round. `addBff()` calls `setBffCallback` at builder time; `setMcpCallback` runs inside `start()`; `handlerFor` iterates a `Map` in insertion order and returns the first match, so **`'bff'` wins**. `Map.set` on an existing key preserves position, so `restart()` does not reorder it. The throw is still right — it is the stated victim that is wrong, which matters because the next person reasoning about a new handler will reason from a false premise.

Normalise before comparing, and test containment rather than equality — either path being a prefix of the other is a collision, whichever is consulted first. `isBffRoute(`${normalized}/mcp`)` covers all four rows above.

packages/agent/src/types.ts
```diff
@@ -139,6 +139,37 @@ export type WorkflowExecutorEmbedOptions = Omit<WorkflowExecutorTuningOptions, '
...
 
+/**
+ * Options of an embedded BFF. Everything the BFF shares with the agent — the secrets, the Forest
+ * urls, the logger — is inherited rather than repeated: a divergent `authSecret` would make the
+ * agent reject the very tokens the BFF mints, as an opaque 401.
+ */
+export type BffEmbedOptions = {
+  /**
+   * Base64-encoded 32-byte AES-256 key encrypting stored refresh tokens. Enables OAuth (mode 1),
+   * and with it the AI relay. Absent, the BFF serves API-key callers only.
```
Tonours: Claude Opus 5 (claude-opus-5): **Must fix**

This sentence makes a security claim that the code does not honour, on public API surface.

> Absent, the BFF serves API-key callers only.

It does not. `createAuthModeMiddleware` is mounted whenever `forestAuthSecret` exists, and it verifies a `bff_access` JWT with `authSecret` **alone** — no session store, no `tokenEncryptionKey` — then issues the agent token from that principal. So with the key omitted, every data route still answers to anyone who can sign a `bff_access` JWT with the agent's `authSecret`. This PR's own e2e demonstrates it: `addBff({ allowedOrigins: [...] })` with no key reports `features.oauth: false`, and `POST /bff/agent/v1/books/list` with a `bff_access` token returns 200.

What the key actually gates is the login/refresh flow and the AI relay. The consequence of the current wording is the dangerous direction: an integrator who omits it *believing they have closed session access* has in fact left the data surface open.

Reword to say what it gates, and say plainly that the session-bearer path is governed by `authSecret`, not by this key.

Tonours (commented):
**Spec (PRD-1076)**: steps 8, 9 (its `/bff` half), 10, 11, 12, 15 and 16 are all delivered, and the two packaging decisions the ticket asked for by name — `engines: { node: ">=22.12.0" }` and the exact-pinned optional peer — match step 15 word for word, so neither is in question here. Two steps landed partially:

- **Step 13** asked that a `buildBff` failure not leave a half-started agent, *"the host is already serving `/forest` when `start()` rejects, and a retry could duplicate hooks and subscriptions"*. The `mounted` flag and the audit-trail close are in, well reasoned. The retry half is not: `subscribeToServerEvents()` and `onRefreshCustomizations()` run before the failure point and there is no guard against a second `start()`.
- **Step 17** asked to *"document that the `Promise.race` cancels nothing: an action cut at 10s keeps running agent-side, and a retry can double the mutation"*. The propagation itself is complete and I traced it end to end — `agentTimeoutMs` → config → `resolveTransport` → `InProcessRequester` → `injectWithTimeout`. The documentation is absent: `git grep -i 'cancels nothing|keeps running|double the mutation'` over both this PR and the docs PR returns nothing. It is a safety caveat about a doubled mutation, and embedded mode is what makes it reachable.

**What I checked and found sound**, so it does not get re-litigated: the `bff-routes.ts` helpers are correct on every input I threw at them, including `/bff` and `/bff/` both stripping to `/` and never to `''`, `/bffalo` falling through, and query/fragment handling; the agent's auth chain genuinely still runs on the in-process path (the dispatcher injects into the mounted `/forest` router whose first root route installs `jwt({ secret: authSecret })`, and `types.ts` states the shared-secret invariant explicitly); `getInProcessDispatcher()` at builder time really is necessary and the comment explaining why is right; `stop()` leaks nothing, since `agent-bff/src` holds no timer; and the dependency direction between the two packages is now cleanly one-way.

One packaging mechanic worth a decision rather than a fix: `engines: ">=22.12.0"` is **blocking under yarn 1** — an install failure, not a warning — and it ships here as a `feat`, so a minor bump. Any consumer still on Node 20 fails to install on a minor. The ticket asked for the field; it did not weigh that. Your call whether it wants a release note.

Eleven findings inline, three of them Must fix.

.github/workflows/build.yml
```diff
@@ -153,7 +153,7 @@ jobs:
   # Boots a real agent on a local HTTP port, so it stays out of the unit job. Unlike the LLM suite
```
Tonours: Claude Opus 5 (claude-opus-5): **Must fix**

**Applies to:** `.github/workflows/build.yml:114` (the unit job's ignore pattern, not in this diff) — anchored on the comment this diff makes false.

This comment now says the opposite of what happens. The unit job runs `yarn test --coverage --testPathIgnorePatterns='llm.integration|search-agent.integration'`. The deleted file matched `search-agent.integration`; the new one is `test/bff/embedded-bff.e2e.test.ts`, which matches **neither** pattern, and `packages/agent/jest.config.ts` collects `test/**/*.test.ts`.

So the suite — two full agents booted, a mock Forest server, a 30s `beforeAll` budget — now runs inside the fail-fast matrix's `agent` cell at `timeout-minutes: 10`, **in addition** to its own dedicated job. A slow runner takes the whole matrix down with it, and the failure will look like an unrelated unit regression.

One word: `--testPathIgnorePatterns='llm.integration|embedded-bff.e2e'`.

packages/agent/src/embedded-bff.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    req.url = stripBffPrefix(req.url ?? BFF_PREFIX);
+    this.bff.callback(req, res);
+  };
+
+  /**
+   * Dynamically load the optional @forestadmin/agent-bff package. Deferred so agents that embed no
+   * BFF never load its code at startup.
+   */
+  private async importPackage() {
```
Tonours, line 198: Claude Opus 5 (claude-opus-5): **Should fix**

`error` is bound and never used, so every way of failing to load a package that **is** installed collapses into "install it". The developer installs, reinstalls, and the message does not move.

Realistic causes here, since the package loads from the customer's own `node_modules`: a missing or ESM-only *transitive* dependency of agent-bff (the message then names the wrong package), a `SyntaxError` from a partially written install or a bundler that rewrote the dist, a throw during agent-bff's own module evaluation, ESM/CJS interop under a `type: module` host, or a Node version below the newly declared `engines` — which is a resolution failure, not a helpful error.

This is the class PR #1834 already fixed in this repo for `mcp-server`, and the MCP path sits one file over doing it right: it logs `Failed to initialize MCP server: ${message}` and rethrows the original.

Keep the hint, keep the cause — `engines` now guarantees Node ≥22.12, so `cause` is available:

```ts
const { message } = error as Error;
throw new Error(
  `Cannot load '@forestadmin/agent-bff' (required by addBff()): ${message}. ` +
    'If it is not installed: `npm install @forestadmin/agent-bff`.',
  { cause: error },
);
```

packages/agent/src/embedded-bff.ts
```diff
@@ -0,0 +1,128 @@
...
+   * Build the BFF. Called from agent.start() after mount(), because the dispatcher only reaches the
+   * agent's own stack once that stack is mounted.
+   */
+  async start(dispatcher: AgentDispatcher): Promise<void> {
+    const { embedOptions } = this;
+    if (!embedOptions) return;
+
+    const { buildBff, parseConfig, IN_PROCESS_AGENT_URL } = await this.importPackage();
+
+    const config = parseConfig({
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**

Everything this line validates is data the caller already handed to `addBff()`, but it is validated at `start()` — **after** `mount()`. So a mistyped key does not degrade the BFF, it kills the agent, with the host left half-serving.

`parseConfig` throws synchronously on a `tokenEncryptionKey` that is not base64/32 bytes (the likeliest typo), a non-IANA `defaultTimezone`, a zero or out-of-range timeout, and a malformed Forest URL. Resulting state for the documented Express pattern, where `app.listen()` precedes `agent.start()`: `/forest` live and serving; `/bff/*` claimed since builder time and answering `503 bff_not_started` for the rest of the process; `start()` rejected. There is no path back without a process restart — `embedOptions` is never cleared and `setBffCallback(null)` is never called.

Retrying `start()` is worse, and this is step 13's unaddressed half: `subscribeToServerEvents()` and `onRefreshCustomizations()` run before the failure point, with no guard against a second `start()`, so a retry opens a second SSE subscription and adds a second restart listener (the handler is a bare `eventEmitter.on`, no dedupe). The duplicate is then masked by the `isRestarting` guard, which only logs at `Debug`.

Call `parseConfig` in `addBff()` and leave only `buildBff` — which genuinely needs the mounted stack — in `start()`. That turns "app half-up, `/bff` bricked" into a throw on the line the developer wrote.

packages/agent/src/embedded-bff.ts
```diff
@@ -0,0 +1,128 @@
...
+      // Off unless asked for: the document is not filtered per caller, so mounting a BFF must not
+      // publish the name of every exposed collection and field on an already-open port.
+      BFF_OPENAPI_ENABLED: String(embedOptions.openapiEnabled ?? false),
+    });
+
+    this.bff = await buildBff({
+      config,
+      dispatcher,
+      basePath: BFF_PREFIX,
+      // Dropped rather than logged: the default sink reports gauges at Info, which would put a
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**

The comment justifies dropping **gauges**, and that reasoning is sound. The same object also drops every **counter**, and those counters are the only channel the schema cache has — `SchemaCache` takes no logger at all, only `Metrics`.

So in embedded mode: a failed schema refresh with a warm cache increments `schema_cache_refresh_error`, goes nowhere, and the BFF serves a schema that no longer matches the agent — a collection the customer added 404s, a field they removed is still advertised — indefinitely, with **zero lines in the host's logs**. The action-endpoint resolver's two `catch`/`increment`/`return undefined` paths are silenced the same way.

Standalone loses none of this: `createReadModel` defaults to `safeMetrics(metrics ?? createConsoleMetrics(logger))`, which routes every counter through the logger. This makes embedded strictly blinder than standalone on exactly the paths that matter, and `BffEmbedOptions` exposes neither `metrics` nor `logger`, so a host has no way to opt in.

Keeping the stated intent — drop the gauges, keep the errors — is a two-line change:

```ts
metrics: {
  increment: name => this.options.logger('Warn', formatLog(`metric ${name}`)),
  gauge: () => undefined,
},
```

.github/workflows/build.yml
```diff
@@ -177,7 +177,7 @@ jobs:
           key: ${{ runner.os }}-build-${{ github.sha }}
           fail-on-cache-miss: true
       - name: Run BFF integration tests
-        run: yarn workspace @forestadmin/agent-bff test --testPathPattern='search-agent.integration'
+        run: yarn workspace @forestadmin/agent test --testPathPattern='bff/embedded-bff.e2e'
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**

**Applies to:** `packages/agent-bff/test/data/search-agent.integration.test.ts` (deleted by this PR) — anchored on the job that used to gate it.

After this PR **no end-to-end coverage of the HTTP transport survives anywhere in the repo**, and that transport is the standalone deployment's only route to the agent — the configuration shipping today. The deleted suite was the only test booting a real agent on a real port and driving `createHttpTransport` through a socket; the replacement uses the in-process dispatcher exclusively. Every remaining `createHttpTransport` test mocks `@forestadmin/agent-client` away, and agent-bff can no longer boot an agent in its own suite at all, having dropped those devDeps.

The job is still named "BFF Integration Tests", so a reader assumes it gates the standalone path. It no longer does.

What that leaves unguarded on the HTTP path specifically: `HttpRequester` header and auth-token construction, query-string encoding, JSON:API deserialization over the wire, the socket timeout, HTTP status → BFF error mapping, and `AGENT_URL`/prefix joining.

Separately, thirteen assertions became five, and five behaviours now have **no case in any package**: relation crossing via the `author.name:` query syntax *without* `searchExtended` (whose deleted case commented that it was "pinned here so it cannot change unnoticed"), the `422 relation_field_not_supported` contrast for the same path in a filter, search ∩ filter intersecting rather than replacing, whitespace-only search listing everything, and the `count` search refusal.

The ticket sanctions moving the suite (step 7), but it says the deleted file's unique value *moves* — an equivalent that swaps the transport is not an equivalent for the transport it left behind. `packages/agent/test/bff/` is now the natural home for both: it already has the agent and supertest, so pointing a `createHttpTransport` BFF at `mountOnStandaloneServer` costs one describe block and restores the gate the job name still claims.

packages/agent/src/embedded-bff.ts
```diff
@@ -0,0 +1,128 @@
...
+  /** Drop what the BFF read from the SaaS. Called on a restart, which means the schema moved. */
+  invalidate(): void {
+    this.bff?.invalidate();
+  }
+
+  /**
+   * Stop answering. The host application keeps whatever middleware it registered, so without this
+   * a stopped agent would go on serving BFF data through a dispatcher pointing at a dead stack.
+   */
+  stop(): void {
```
Tonours, line 161: Claude Opus 5 (claude-opus-5): **Should fix**

`stop()` sets `bff = null` and nothing else, so `embedOptions` stays set and `handle` answers `503 bff_not_started` for the rest of the process lifetime. Three consequences:

**Boot and shutdown become indistinguishable.** A k8s or ALB probe on `/bff/health` gets the same `503 bff_not_started` while the agent is booting and after a deliberate `stop()`. During boot the right action is to wait; after stop it is to take the pod out of rotation. `EmbeddedBff` knows which state it is in — one flag separates `bff_not_started` from `bff_stopped`.

**The type misleads.** "not started" for a BFF that started, served traffic, and was stopped.

**`/bff` is permanently occupied.** `RootMiddleware.set` supports removal and `setBffCallback` already accepts `null`; neither is used on stop. So a host that stops the agent and mounts its own `/bff` handler, or a test harness cycling agents on one Express app, can never get the path back — its own routes stay shadowed by a 503. `setBffCallback(null)` in `stop()` is a one-liner and makes the middleware genuinely fall through.

Minor, same body: it omits `message`, unlike every other BFF error, so a client rendering `error.message` shows blank.

While here: the `if (!this.embedOptions) next?.()` branch just above is unreachable — `configure()` runs two lines before `setBffCallback(bff.handle)` — so its comment describes a fall-through that does not exist, while the state that *should* fall through (stopped) 503s instead.

packages/agent/src/embedded-bff.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    if (!this.bff) {
+      res.statusCode = 503;
+      res.setHeader('Content-Type', 'application/json');
+      res.end(JSON.stringify({ error: { type: 'bff_not_started', status: 503 } }));
+
+      return;
+    }
+
+    req.url = stripBffPrefix(req.url ?? BFF_PREFIX);
```
Tonours, line 190: Claude Opus 5 (claude-opus-5): **Should fix**

`req.url` is mutated in place and never restored, so on a Koa host every access log and metric of the customer's own middleware records the stripped path.

Express hosts are mostly fine — Express sets `req.originalUrl` at app entry and morgan reads `req.originalUrl || req.url`. Koa has no `originalUrl`, and `ctx.url` / `ctx.path` read `req.url` live: `expressToKoa` hands `ctx.req` straight through and resolves on `finish`, so any host middleware registered before the agent's that logs after `await next()` sees `/health` or `/agent/v1/books/list`. A `/bff/health` probe appears in their log as `/health` — colliding with paths the host may itself serve. Same on the raw-http and standalone paths, and for any `res.on('finish')` handler the host registered on any mount target.

The repo already has the pattern, one file over in this package: `fastify-adapter.ts:127-132` saves `originalUrl` and restores it in the `next` callback. There is no fall-through to restore in here, but the mutation still escapes to everything holding a reference to `req`. Restore it after `callback` returns, or set `req.originalUrl ??= req.url` before stripping.

packages/agent/src/agent.ts
```diff
@@ -184,6 +201,8 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
 
       this.setMcpCallback(mcpHttpCallback ?? null, mcpIsMcpRoute);
       await this.remount(router);
+      // A restart means the customizations changed, so the schema the BFF read is stale.
+      this.embeddedBff?.invalidate();
```
Tonours, line 224: Claude Opus 5 (claude-opus-5): **Violates conventions** — `skills/conventions/testing.md#Cover error and edge paths, not only the happy path` and `skills/conventions/agent.md#A module's public API is its curated barrel`

Four new reachable paths have no case, and one public type is missing from the barrel.

**`invalidate()` on restart** — this line. `restart()` is public and documented as the runtime-remount entry point; no test calls it on an agent with an embedded BFF. The one line whose job is to stop the BFF serving a schema that no longer matches the customizations is unguarded: delete it and every test passes, so the regression ships as a stale schema answered to third-party UIs.

**The missing-package path** — untested, though the sibling feature proves it costs nothing: `test/agent-workflow-executor-missing-dep.test.ts` mocks the module to throw and asserts `start()` rejects with the actionable message. The BFF has no equivalent, so a refactor of `importPackage` can turn its message into a generic startup failure with nothing failing.

**`buildBff`'s `dispatcher` branch** — `dispatcher` is a new public option on `buildBff`, and no test in `packages/agent-bff/test/` passes one. `test/agent/in-process-transport.test.ts` covers the transport unit, never the wiring that selects it. So the entire embedded path is unverified inside the package that owns the code.

**`healthy: dispatcher !== undefined || config.hasAllRequired`** — the deliberate new "embedded is always healthy" semantic, asserted nowhere in agent-bff; flipping the operands keeps that suite green.

**Barrel** — `BffEmbedOptions` is the parameter type of a public method on the exported `Agent`, and `index.ts` exports only `AgentOptions` and `WorkflowExecutorEmbedOptions`. The sibling feature's embed-options type is in the barrel; this one is not, so a consumer naming the options has to import from `@forestadmin/agent/dist/types` — the bypass the barrel exists to prevent.

packages/agent-bff/src/http/health-route.ts
```diff
@@ -19,7 +32,7 @@ export default function createHealthRoute({ config, version }: HealthRouteOption
       return;
     }
 
-    ctx.status = config.hasAllRequired ? 200 : 503;
-    ctx.body = { status: config.hasAllRequired ? 'ok' : 'degraded', version };
+    ctx.status = healthy ? 200 : 503;
+    ctx.body = { status: healthy ? 'ok' : 'degraded', version, features };
```
Tonours: Claude Opus 5 (claude-opus-5): **Preferential**

`/health` is unauthenticated — it sits second in the middleware array, long before `createAuthModeMiddleware` — and now returns `features: { oauth, ai, cors, openapi }`, where `oauth` and `ai` are `Boolean(config.tokenEncryptionKey)`. So an anonymous caller learns which capabilities the deployment configured.

No secret leaks and the values are useful, so this may well be the right trade. What changes it is the mount: embedded, this lands on `/bff/health` of the **customer's own public application**, where standalone it lived on a separately deployed port they could firewall. Worth deciding deliberately rather than inheriting — a `features` block gated behind the same credential the rest of the surface uses would keep the operational value without the anonymous reconnaissance.

What is not a judgment call: `test/http/bff-http-server.test.ts` has a case named *should never expose config presence or secret values in the response body*, and it stays green because it only greps for the secret literals. The invariant its name asserts is now false. Either the disclosure goes, or that test's name and body do.

Related in the same file's suite: the `/health` case was weakened from `toEqual` to `toMatchObject({ status, version })` and renamed nothing, so its *"with 200 ok and the version only"* is also false now that `features` is in the body — and the four flags have no assertion at all on the `BFFHttpServer` path (`build-bff.test.ts` does pin them for the `buildBff` path).

packages/agent/src/embedded-bff.ts
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * Serialize the BFF's structured log context onto the message: the agent's logger only accepts an
+ * Error as its third argument, so the context would be dropped otherwise. Never throws — logging
+ * must not break a request.
+ */
+function formatLog(message: string, context?: Record<string, unknown>): string {
+  if (!context || Object.keys(context).length === 0) return `[bff] ${message}`;
+
+  try {
+    return `[bff] ${message} ${JSON.stringify(context)}`;
```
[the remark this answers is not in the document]
nbouliol: Fixed: `formatLog` now passes a replacer that unfolds an `Error` into `{ name, message, stack }`. `docs-routes.ts:96` is the caller that hit this — it passed the raw `Error` and logged `{"error":{}}`. Covered in `agent-bff-lifecycle.test.ts`.

packages/agent/src/agent.ts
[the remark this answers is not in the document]
nbouliol: Fixed. Added `collidesWithBff()`, which mirrors `normalizeMountPath` (trim, leading slash, collapsed/stripped trailing slashes) and then tests containment via `isBffRoute(`${normalized}/mcp`)`, so `bff`, `/bff/`, `//bff//` and `/bff/ai` are all rejected now. Cases pinned in `bff-routes.test.ts` and `agent-bff.test.ts`.

packages/agent-bff/src/build-bff.ts
```diff
@@ -455,23 +481,52 @@ export default async function buildBff({
...
     agentMiddlewares.length > 0 ? [agentScoped(createErrorMiddleware({ logger }))] : [];
 
   const middlewares = [
     createVersionHeaderMiddleware(version),
-    createHealthRoute({ config, version }),
+    createHealthRoute({
+      version,
+      // Embedded, everything required is inherited from the agent, so there is no gap to report: a
+      // 503 here would let a load balancer restart a process that serves api-key traffic fine.
+      healthy: dispatcher !== undefined || config.hasAllRequired,
```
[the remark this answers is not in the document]
nbouliol: Fixed as suggested: `healthy: (dispatcher !== undefined && Boolean(config.forestAuthSecret)) || config.hasAllRequired`. Without the secret `buildAgentMiddlewares` returns an empty edge, so ok was a lie. Both halves asserted in `build-bff-dispatcher.test.ts`.

packages/agent/src/agent.ts
nbouliol: Fixed, and you are right on both counts. `collidesWithBff()` in `bff-routes.ts` normalizes the way `normalizeMountPath` does, then tests containment with `isBffRoute(`${normalized}/mcp`)` — so all four of your rows now throw, and the error names the basePath the caller actually passed instead of a hardcoded `/bff`. The normalization is mirrored rather than imported because `mcp-server` keeps it internal and exporting it would widen this PR into a third package. The comment is rewritten with the direction the right way round: the BFF registers at builder time, wins the Map first-match, and the MCP surface is the one that goes dark. Cases in `bff-routes.test.ts` and `agent-bff.test.ts`.

packages/agent/src/types.ts
nbouliol: Fixed. The doc now says what the key gates — the OAuth login/refresh flow and the AI relay — and says plainly that it is not an access control: the data routes accept a `bff_access` bearer either way, because that token is verified against `authSecret` alone. Also corrected the same sentence in the PR description.

.github/workflows/build.yml
nbouliol: Fixed: `--testPathIgnorePatterns='llm.integration|embedded-bff.e2e'`. Also updated the dedicated job comment and name, since it now gates both transports rather than only the embedded one.

packages/agent/src/embedded-bff.ts
nbouliol: Fixed, with the cause kept: the message now carries the load failure verbatim, and `cause` is set by assignment rather than through the constructor options bag — the repo targets ES2020, so `new Error(msg, { cause })` does not typecheck even though the declared engines guarantee a runtime that reads it. Three cases in `agent-bff-missing-dep.test.ts`.

packages/agent/src/embedded-bff.ts
nbouliol: Fixed. `parseConfig` moved into a new `EmbeddedBff.prepare()`, which `start()` calls first — before `buildRouterAndSendSchema()`, before `subscribeToServerEvents()`, before `mount()`. So a mistyped key throws with nothing mounted and nothing subscribed, instead of leaving a live `/forest` next to a bricked `/bff`; and since the failure is now ahead of the subscription, a retry cannot duplicate it either. `buildBff` stays in `start(dispatcher)`, which genuinely needs the mounted stack. Asserted in `agent-bff-lifecycle.test.ts`. The general "second `start()` on an agent with no BFF" guard is pre-existing `Agent.start()` behaviour on every failure path, so I left it out of this PR.

packages/agent/src/embedded-bff.ts
nbouliol: Fixed as suggested. Counters go to the host logs at `Warn`; gauges stay dropped. Worth noting all three counters in the package (`schema_cache_refresh_error`, `action_endpoint_error`, `action_endpoint_miss`) are failure signals, so routing every one of them is right rather than noisy. Both halves asserted in `agent-bff-lifecycle.test.ts`.

.github/workflows/build.yml
nbouliol: Fixed. `embedded-bff.e2e.test.ts` now runs its data contract twice through one shared `itServesTheDataContract()`: once over the in-process dispatcher, once over a `buildBff({ AGENT_URL })` pointed at the express app listening on a real port — so `createHttpTransport`, header and token construction, query encoding, JSON:API over the wire and the status mapping are gated again. All five behaviours you listed as orphaned are back in that contract, so each is now asserted on both transports (13 cases x 2, plus the health/`/forest`/`/bffalo`/stopped cases — 30 in the suite). Job renamed to say it gates both.

packages/agent/src/embedded-bff.ts
nbouliol: Half fixed. A stopped BFF now answers `bff_stopped` with a `message`, and boot answers `bff_not_started` — a probe can tell shutdown from boot, which was the substance of the finding. The unreachable `!embedOptions` branch is gone: `embedOptions` moved to the constructor, so the state cannot exist.

I did not call `setBffCallback(null)`. It contradicts the other half of your finding — with the handler removed there is no way to answer `bff_stopped` at all, and a stopped agent falling through to the host 404 is the "wrong url" reading this 503 exists to avoid. `agent.stop()` is process shutdown, not a handover, so a host reclaiming `/bff` afterwards is not a flow we support. Happy to revisit if you have a concrete case where it is.

packages/agent/src/embedded-bff.ts
nbouliol: Partly fixed. `req.originalUrl ??= req.url` before the rewrite, so anything reading `originalUrl` reports the url the client asked for.

On the Koa half: `ctx.originalUrl` is not missing — `koa/lib/application.js:215` sets `context.originalUrl = request.originalUrl = req.url` in `createContext`, at host entry, before any middleware, so it stays correct. What is genuinely live is `ctx.url` / `ctx.path`, and neither restoring after `callback()` nor on `finish` fixes those: Koa reads `req.url` lazily while routing, and a host `finish` listener registered before the agent middleware fires before any restore of ours. The only thing that would work is handing the BFF a `Proxy` over `req` with `url` overridden, and proxying a live `IncomingMessage` stream through the body parser is more risk than the log fidelity is worth. Asserted the `originalUrl` behaviour in `agent-bff-lifecycle.test.ts`; say the word if you want the proxy anyway.

packages/agent/src/agent.ts
nbouliol: All five fixed.

- `invalidate()` on restart: `agent-bff-lifecycle.test.ts` asserts `restart()` invalidates once and does not rebuild.
- Missing package: `agent-bff-missing-dep.test.ts`, mirroring the workflow-executor one, plus a case on the preserved `cause`.
- `buildBff`'s `dispatcher` branch: `build-bff-dispatcher.test.ts` drives a real list through the dispatcher (path, bearer, timeout, records) and asserts the no-dispatcher/no-`AGENT_URL` config falls back to the 501 stub.
- `healthy`: both operands asserted in the same file.
- Barrel: `BffEmbedOptions` exported from `index.ts` next to `WorkflowExecutorEmbedOptions`.

packages/agent-bff/src/http/health-route.ts
nbouliol: Test names fixed, disclosure kept.

The `/health` case is renamed and now asserts the full body including all four flags, plus a second case pinning them on a partial config — the `BFFHttpServer` path had none. The invariant test is renamed to what it actually guarantees ("secret values or the config itself"), which is the claim that survives.

On the trade: I am keeping `features`. It names capabilities, never values, and it is what makes an embedded `/bff/health` useful to a load balancer or to support — which is the deployment where the caller cannot reach a second port to ask. Gating it behind the same credential as the rest would leave a probe with nothing to read. Worth reopening if you would rather have it behind auth.

nbouliol:
Pushed 271c47bc7. Eleven inline findings answered individually — nine fixed, two partly, and the reasoning is in each thread. On the two points in the review body itself:

**Step 13, the retry half.** Addressed for the BFF: `parseConfig` now runs in `EmbeddedBff.prepare()`, called at the top of `start()` before `buildRouterAndSendSchema()`, `subscribeToServerEvents()` and `mount()`. The likeliest failure — a mistyped option — therefore happens before either the subscription or the restart listener exists, so a retry cannot duplicate them. The general "no guard against a second `start()`" is pre-existing `Agent.start()` behaviour on every failure path in the class, unrelated to `addBff()`, so I did not widen this PR into it.

**Step 17, the `Promise.race` caveat.** Documented on `agentTimeoutMs` in `types.ts`: it bounds how long the BFF waits and cancels nothing, an action cut at the deadline keeps running agent-side and still commits, so a client retrying on a timeout can double the mutation.

**`engines: ">=22.12.0"` under yarn 1.** Correct that it is a hard install failure, not a warning, and that it ships as a minor. Keeping the field — step 15 asked for it by name and `addBff()` pulls in a package that requires it — and I will add the release note, since a Node 20 consumer failing to install on a minor deserves a heads-up rather than a surprise.

Both suites green: agent-bff 82 suites / 1368 tests, agent 87 suites / 1587 tests, lint clean on both.

Tonours (approved):
All fourteen findings verified on 271c47b: `collidesWithBff()` normalizes the way mcp-server does, so 'bff', '/bff/' and '/bff/ai' all throw, and the comment has the direction right; the `tokenEncryptionKey` doc no longer claims an access control it does not provide; the unit job ignores `embedded-bff.e2e`; the load failure keeps its cause; `prepare()` validates before mount and before the SSE subscription; counters reach the host logs while gauges stay dropped; the e2e runs its data contract over both transports, HTTP included; `bff_stopped` is told apart from `bff_not_started`; `originalUrl` is claimed before the rewrite; the five test gaps and the barrel export are closed. Keeping the handler mounted on stop and keeping `features` on /health are both argued positions I accept.

One note, not blocking: the build workflow has not run on this head — only Macroscope and qlty report. Worth a re-run before merge.

Tonours (changes requested):
Delta since my last review (271c47b → 1e3ce47): rebase plus five commits — 5c8313d, a2c694d, 55875e0, ca5550f, 1e3ce47. One must-fix; the rest of the delta verified good (details at the end).

## Must fix

**The origin deny breaks same-origin mutating browser calls.** `createCorsMiddleware` is installed unconditionally (`build-bff.ts:526`) and now refuses every request whose `Origin` is not in the allow-list (`cors-middleware.ts:55-60`, from 55875e0). Browsers send `Origin` on every same-origin POST/PUT/PATCH/DELETE, so a deployment whose own origin is not listed — the default, `allowedOrigins` starts empty — answers `403 origin_not_allowed` to every list/count/action call from a same-origin UI. On main the same request was served (omit-header + log, #1868), and the README promises that contract: *"Empty ⇒ no cross-origin browser access"* (agent-bff/README.md:97) — same-origin implied working. This is also macroscopeapp's open thread on `cors-middleware.ts:55`; no test covers empty-list + `Origin` today.

Fix either way:
- treat a same-origin `Origin` as allowed (`origin === ctx.origin`, or Host comparison), or
- make the new contract explicit: the deployment's own origin must be listed, with a test for empty-list + same-origin POST and the README sentence rewritten.

## Verified good in this delta

- **5c8313d** (drop `engines` from the agent): right call — `engines` is a hard install failure under yarn 1 and this ships as a minor; the requirement stays on agent-bff, which is the optional peer.
- **ca5550f** (whitelist startup warning): both branches tested; never fails boot; names what still protects the route.
- **55875e0** health rename `features` → `configured`: honest naming, tests updated. Note the README in #1877 still documents the old key — flagged there.
- **a2c694d / 1e3ce47**: rename reconciliation and log prefixing, no behavior change.
- Rebase drift on the three rewritten commits is upstream absorption (mountPath, RootHandler, IN_PROCESS_AGENT_URL), nothing new snuck in.

CI: only `LLM Integration Tests (ai-proxy)` fails, on all seven PRs of the stack alike — an Anthropic-side `thinking.type.disabled` 400 on `claude-fable-5-1`, unrelated to this code.

packages/agent-bff/src/http/bff-http-server.ts
```diff
@@ -67,7 +67,18 @@ export default class BFFHttpServer {
...
 
     app.use(createVersionHeaderMiddleware(version));
-    app.use(createHealthRoute({ config, version }));
+    app.use(
+      createHealthRoute({
+        version,
+        healthy: config.hasAllRequired,
+        features: {
+          oauth: Boolean(config.tokenEncryptionKey),
+          ai: Boolean(config.tokenEncryptionKey),
```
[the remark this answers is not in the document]
nbouliol: Not changing this. The block is no longer named `features` — it is `configured`, and `HealthConfigured`'s own doc says why: "Which optional surfaces this deployment was CONFIGURED to serve. Deliberately not a statement that they work." `oauth: true` means an encryption key is set, not that a route answers, and that is the claim the name makes.

It also cannot be derived from installed routes on this path. `buildHandler` runs only when a host assembles its own stack and passes `middlewares` — it has no way to know what those mount. The real standalone deployment does not take this path at all: `runCli` builds the handler with `buildBff` and passes it as `callback`, so its `/health` is `build-bff`'s, where the flags *are* derived from the mounted chain (`oauth.middlewares.length > 0`, `agentMiddlewares.length > 0`).

packages/agent/src/agent.ts
```diff
@@ -362,6 +392,50 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
...
+  addBff(options: BffEmbedOptions = {}): this {
+    if (this.embeddedBff) {
+      throw new Error('addBff can only be called once.');
+    }
+
+    if (collidesWithBff(this.mcpBasePath)) {
+      throw new Error(bffMcpCollision(this.mcpBasePath as string));
+    }
+
+    const bff = new EmbeddedBff(this.options, options);
```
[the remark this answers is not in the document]
nbouliol: This is a known and deliberate decision, documented in `ca5550fbb` rather than left implicit — see the doc comment on `warnIfExemptFromIpWhitelist` in `embedded-bff.ts`, and the `Warn` it logs at every boot where the whitelist is enabled.

Propagating the caller's ip is the thing we specifically do not want: it would make embedded mode *stricter* than standalone, where the whitelist only ever sees the BFF's own host, and it would refuse the browsers a third-party UI is made of — which is the entire population of BFF callers. The whitelist is an operator control over who reaches the agent's admin surface, not over who reaches an application built on the BFF.

What the finding is right about is that it must not be silent, and that is what the startup warning is for. Access is still gated: a resolved API key or a valid OAuth session is required on every route.

packages/agent-bff/src/cors/cors-middleware.ts
```diff
@@ -35,6 +46,19 @@ export default function createCorsMiddleware({
       return;
     }
 
+    // Refuse rather than run and let the browser discard the answer. Omitting the header only stops
+    // the caller from READING the response: the request still executed, so a list was read and an
+    // action was executed for an origin the allow-list names as unwelcome. It also stops being
+    // theoretical once a host sits in front of this app — a permissive `cors()` of its own answers
+    // the preflight with `*`, and the real request arrives here regardless.
+    // A caller with no `Origin` at all is untouched: that is every server-to-server api-key call.
+    if (origin && !allowed) {
```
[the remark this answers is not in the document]
nbouliol: Fixed. The allow-list now applies to cross-origin callers only; the application serving the BFF is exempt. This mattered more than the wording suggests — every BFF data route is a POST, so a host serving its own UI next to a `/bff` mount was refused on *every* request under the default empty allow-list.

Matched on host rather than on the full origin, deliberately: a TLS-terminating proxy leaves `ctx.protocol` at `http` while the browser reports `https`, so comparing the scheme would have worked in development and silently failed in production. Nothing is weakened — a caller able to forge `Origin` can simply omit it, which the allow-list lets through by design, and a genuine cross-site request never carries this host as its origin. Four cases in `cors-middleware.test.ts`, including the same hostname on another port still being refused.

packages/agent-bff/src/http/bff-http-server.ts
```diff
@@ -67,7 +67,18 @@ export default class BFFHttpServer {
     const app = new Koa();
 
     app.use(createVersionHeaderMiddleware(version));
-    app.use(createHealthRoute({ config, version }));
+    app.use(
+      createHealthRoute({
+        version,
+        healthy: config.hasAllRequired,
```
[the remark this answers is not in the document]
nbouliol: Fixed, at the source rather than in this file: `hasAllRequired` no longer counts the encryption key, so both deployment modes agree.

Confirming your read with the repo's own evidence: `env-config.test.ts` already had a case named *"should not be part of REQUIRED_KEYS (it gates OAuth, not boot)"*, and the very next one pinned `hasAllRequired` to `false` without it. `warnMissingConfig` never named the key either — it only reports `presence`, which covers `REQUIRED_KEYS` — so the 503 arrived with no explanation of what was missing. Which optional surfaces are on is what `configured` reports. Pinned in `env-config.test.ts` and `bff-http-server.test.ts`.

packages/agent/src/embedded-bff.ts
```diff
@@ -0,0 +1,165 @@
...
+    this.bff = await buildBff({
+      config: this.config as BFFConfig,
+      dispatcher,
+      basePath: BFF_PREFIX,
+      // Counters are the schema cache's and the action-endpoint resolver's only channel — they take
+      // no logger — and every one of them reports a failure, so they go to the host's logs. Gauges
+      // do not: they are periodic cache sizes, and the default console sink reports them at Info,
+      // which would flood a host that only asked for a BFF.
+      metrics: {
+        increment: name => this.options.logger('Warn', formatLog(`metric ${name}`)),
```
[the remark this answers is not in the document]
nbouliol: Fixed as suggested — `increment: (name, tags) => logger('Warn', formatLog(`metric ${name}`, tags))`. `formatLog` already serialized a context, so the tags land as JSON on the line. Asserted in `agent-bff-lifecycle.test.ts` with the rendering/collection/action triple.

packages/agent/src/agent.ts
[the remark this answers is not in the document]
nbouliol: Fixed: `addBff()` now throws when the agent is already started, rather than registering a BFF nothing will start. Worth noting the diagnosis is even more complete than stated — beyond `EmbeddedBff` never starting, `getInProcessDispatcher()` pushes its hook onto `onFirstStart`, which `mount()` has already drained, so the dispatcher would have no handler either. Starting it late would have to redo both, for a call the documented builder chain never makes.

A `start()` that *failed* still accepts it — the flag is set only on success. Both cases in `agent-bff-lifecycle.test.ts`.

packages/agent-bff/src/cors/cors-middleware.ts
```diff
@@ -35,6 +71,20 @@ export default function createCorsMiddleware({
...
     }
 
+    // Refuse rather than run and let the browser discard the answer. Omitting the header only stops
+    // the caller from READING the response: the request still executed, so a list was read and an
+    // action was executed for an origin the allow-list names as unwelcome. It also stops being
+    // theoretical once a host sits in front of this app — a permissive `cors()` of its own answers
+    // the preflight with `*`, and the real request arrives here regardless.
+    // A caller with no `Origin` at all is untouched: that is every server-to-server api-key call.
+    // Nor is the host's own application, which the allow-list has nothing to say about.
+    if (origin && !allowed && !isSameOrigin(origin, ctx.host)) {
```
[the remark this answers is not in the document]
nbouliol: Fixed, and the diagnosis is exact — confirmed against Node: `new URL('https://x.example.com').host` is `x.example.com` while `ctx.host` stays `x.example.com:443`, so the exemption 403ed the very case it exists for.

Both spellings are accepted now, with the default port derived from the origin's scheme rather than by comparing full origins — matching the scheme is what breaks behind a TLS-terminating proxy, which is why the host comparison was chosen in the first place. A non-default port still has to match exactly, so `:8443` against `Host: x.example.com` is still cross-origin. Two cases added for the `:443` and `:80` spellings, alongside the one pinning that a different port is refused.

packages/agent/src/agent.ts
```diff
@@ -362,6 +397,57 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
...
+   */
+  addBff(options: BffEmbedOptions = {}): this {
+    if (this.embeddedBff) {
+      throw new Error('addBff can only be called once.');
+    }
+
+    // Refused rather than accepted and left dark: after start() the dispatcher hook has no mount to
+    // attach to and nothing calls the BFF's own start(), so `/bff` would answer 503 for the rest of
+    // the process while every other route works. A throw on the line the developer wrote instead.
+    if (this.started) {
```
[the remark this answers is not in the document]
nbouliol: Fixed. You are right that `started` was the wrong seam: `mount()` drains the `onFirstStart` hooks partway through `start()`, so a call landing while it is still in flight is just as late as one landing after it, and `start()` being async makes that reachable with a plain unawaited `agent.start()`.

The flag is now set on the first line of `start()` rather than its last, and cleared in the `catch` so a startup that failed leaves the agent configurable again — both of those are pinned in `agent-bff-lifecycle.test.ts`, including the in-flight case. Message reworded to "already starting".

nbouliol:
Pushed 55bc6fa71. Answering the `CHANGES_REQUESTED` review, and thanks for catching this one before it shipped.

**The must-fix is done, taking your first option.** Same-origin is exempt from the allow-list, which applies to cross-origin callers only. Your framing was the useful part: on main the same request was served, and `agent-bff/README.md:97` promises *"Empty ⇒ no cross-origin browser access"* — with the exemption in place that sentence is exactly true again, so it needed no rewrite. Every BFF data route being a POST is what made this total rather than partial: a same-origin UI was 403ing on every list, count and action.

Matched on host rather than on `origin === ctx.origin`. Comparing the full origin looked cleaner but breaks behind a TLS-terminating proxy, where `ctx.protocol` reads `http` while the browser reports `https` — it would have passed in development and failed in production. Nothing is weakened: a caller able to forge `Origin` can simply omit it, which the allow-list lets through by design, and a genuine cross-site request never carries this host as its origin.

Macroscope then found two real holes in that first attempt, both now fixed in the same commit:
- `new URL()` drops a default port while `ctx.host` keeps it, so `Host: app.example.com:443` against `Origin: https://app.example.com` 403ed — the exact case the exemption is for. Both spellings accepted, default port derived from the origin scheme; a non-default port still has to match exactly.
- the `addBff()`-after-`start()` guard keyed on a flag set at the *end* of `start()`, but `mount()` drains the `onFirstStart` hooks partway through, so an unawaited `agent.start()` followed by `addBff()` slipped past. Flag now set on the first line, cleared when startup fails.

Also fixed a doc claim my earlier `hasAllRequired` change falsified: the README said `/health` reports `degraded` without an encryption key. It reports `configured.oauth: false` and stays `ok`, since that key gates OAuth and not boot.

**On your non-blocking note:** the build workflow has now run on the head — everything green except `LLM Integration Tests (ai-proxy)`, the stack-wide Anthropic-side failure you already identified.

Local: agent-bff 85 suites / 1440 tests, agent 88 / 1599, lint clean on both.

packages/agent-bff/src/cors/cors-middleware.ts
```diff
@@ -2,11 +2,57 @@ import type { Middleware } from 'koa';
...
+    return false;
+  }
+
+  // Both spellings are accepted because the two sides normalize differently: `new URL()` drops a
+  // port that is the default for the scheme, while `ctx.host` is the raw `Host` header, which a
+  // proxy is free to spell out. Without this, `Host: app.example.com:443` against
+  // `Origin: https://app.example.com` reads as cross-origin.
+  const defaultPort = url.protocol === 'https:' ? '443' : '80';
+
+  return host === url.host || host === `${url.host}:${defaultPort}`;
```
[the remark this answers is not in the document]
nbouliol: Fixed. Confirmed against Node: `new URL('https://APP.EXAMPLE.COM').host` is `app.example.com`, so a raw `Host: APP.EXAMPLE.COM` did not match and the exemption refused a same-origin call.

This is the same shape as the default-port finding, on the other half of what `new URL()` normalizes — the comment now names both, so the next person touching this comparison sees why it absorbs two differences rather than one. A non-default port still has to match exactly. Case pinned in `cors-middleware.test.ts`.

packages/agent/src/agent.ts
Tonours: Verified at 5a6593a9: `collidesWithBff` mirrors the mount-path normalization (trim, leading slash, collapsed and stripped trailing slashes) and tests containment through `isBffRoute(`${normalized}/mcp`)` — so `bff`, `/bff/` and `/bff/ai` all collide while `/bffalo` still doesn't. Resolving.

packages/agent/src/types.ts
Tonours: Verified: the doc now states what the key gates — the OAuth login/refresh flow and the AI relay — and says plainly it is not an access control: the `bff_access` bearer path is protected by `authSecret`, key or no key. Resolving.

.github/workflows/build.yml
Tonours: Verified: the unit job now runs with `--testPathIgnorePatterns='llm.integration|embedded-bff.e2e'`, and the dedicated job gates both transports (its name says so, and it is green at head). Resolving.

packages/agent/src/embedded-bff.ts
Tonours: Verified: the load failure travels verbatim in the message, and `cause` is set by assignment — the ES2020 lib has no options bag on Error. Resolving.

packages/agent/src/embedded-bff.ts
Tonours: Verified: `parseConfig` moved into `EmbeddedBff.prepare()`, which `start()` calls ahead of `buildRouterAndSendSchema()`, `subscribeToServerEvents()` and `mount()`; the lifecycle test pins the rejection before anything is mounted or subscribed. Resolving.

packages/agent/src/embedded-bff.ts
Tonours: Verified, and hardened further in aaaaf589: counters go to the host logs at `Warn` with their tags (rendering, collection, action) now forwarded, gauges stay dropped. The test asserts the tags arrive with the metric name. Resolving.

.github/workflows/build.yml
Tonours: Verified: `embedded-bff.e2e.test.ts` runs the data contract over both the in-process dispatcher and a `buildBff({ AGENT_URL })` HTTP transport through one shared `itServesTheDataContract()`, and the dedicated CI job ("embedded and http transports") is green at head. Resolving.

packages/agent/src/embedded-bff.ts
Tonours: Verified: `handle()` answers `503 bff_stopped` with a message after `stop()`, and `bff_not_started` before boot — a probe can tell shutdown from never-started. Resolving.

packages/agent/src/embedded-bff.ts
Tonours: Verified: `req.originalUrl ??= req.url` before the rewrite, so host middleware reading `originalUrl` sees the untouched url. The emitted-prefix work in #1873 (`resolveEmittedBase`) now builds on exactly that invariant, e2e-asserted through a `/api` host mount. Resolving.

packages/agent/src/agent.ts
Tonours: All five verified at head: restart invalidates once without rebuilding, the missing-dep test exists, the lifecycle guards are in place, and the public surface stays read-only through `handle`. aaaaf589 additionally made `addBff()` after startup throw instead of registering a dark BFF — hardened to mid-flight calls and failed starts, both branches tested. Resolving.

packages/agent-bff/src/http/health-route.ts
Tonours: Verified, then superseded by the rename this stack landed: the block is `configured` (55875e0), tests assert the full body on complete and partial configs, and aaaaf589 keeps `/health` `ok` without the encryption key — `configured.oauth: false` is what says the surface is off. Resolving.

Tonours (approved):
Verified at 5a6593a9. The must-fix is closed properly: the same-origin exemption matches on host, absorbing both normalizations `new URL()` and the raw `Host` header disagree on (default port spelled out or not, case) — a non-default port is still another origin and refused, and the whole class is pinned by unit tests on both sides plus the empty-allow-list same-origin POST case. The three open bot findings are closed too: `/health` stays `ok` without the encryption key (it gates OAuth, not boot — `configured.oauth` says what is off, and the env-table promise was resynced), metric tags reach the host logs, and `addBff()` past startup throws — hardened to mid-flight calls and failed starts, both tested. Approving.

Left open on purpose: the macroscopeapp thread on the scheme-mismatch in `isSameOrigin` — host-only matching is the documented tradeoff (TLS-terminating proxies; a forgeable `Origin` buys nothing over omitting it under bearer auth), but it is the author's call to answer.

Tonours (approved):
Verified at 5a6593a9. The must-fix is closed properly: the same-origin exemption matches on host, absorbing both normalizations `new URL()` and the raw `Host` header disagree on (default port spelled out or not, case) — a non-default port is still another origin and refused, and the class is pinned by unit tests on both sides plus the empty-allow-list same-origin POST case. The three open bot findings are closed too: `/health` stays `ok` without the encryption key (it gates OAuth, not boot — `configured.oauth` says what is off, env-table promise resynced), metric tags reach the host logs, and `addBff()` past startup throws — hardened to mid-flight calls and failed starts, both tested. Approving.

Left open on purpose: the macroscopeapp thread on the scheme mismatch in `isSameOrigin` — host-only matching is the documented tradeoff (TLS-terminating proxies; a forgeable `Origin` buys nothing over omitting it under bearer auth), but it is the author's call to answer.

packages/agent/src/agent.ts
```diff
@@ -112,7 +139,10 @@ export default class Agent<S extends TSchema = TSchema> extends FrameworkMounter
       // Boot after mount(): the embedded executor reaches the agent over HTTP, and the
       // standalone server's host/port (used to derive that URL) are only known once mounted.
       await this.embeddedExecutor?.start(this.standaloneServerHost, this.standaloneServerPort);
+      // Same reason, without the socket: the dispatcher injects into the stack mount() just built.
+      await this.embeddedBff?.start(this.getInProcessDispatcher());
     } catch (error) {
+      this.startupBegun = false;
```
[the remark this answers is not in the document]
nbouliol: Fixed in e05ddeb54.

The chain holds: `mount()` sets `mounted = true`, a later failure (the embedded executor, say) hits the `catch`, and `startupBegun` was cleared unconditionally — so `addBff()` passed its guard on an agent the host is already serving, registered a BFF whose `start()` nothing calls, and `/bff` answered 503 for the rest of the process. `restart()` only calls `invalidate()`, never `start()`, so your last sentence is right too.

Took your diff as written — it is the same `mounted` distinction the audit-trail close two lines below already makes, so the two now agree.

Covered by `agent-bff-lifecycle.test.ts`: a post-mount failure (executor rejecting) must leave `addBff()` refused. Verified it fails without the guard.

packages/agent/src/embedded-bff.ts
```diff
@@ -0,0 +1,209 @@
...
+      metrics: {
+        // Tags carried through: `action_endpoint_error` and `action_endpoint_miss` name the
+        // rendering, collection and action that failed, which is the whole of what makes the line
+        // actionable — the metric name alone says only that something, somewhere, did not resolve.
+        increment: (name, tags) => this.options.logger('Warn', formatLog(`metric ${name}`, tags)),
+        gauge: () => undefined,
+      },
+      logger: (level, message, context) => this.options.logger(level, formatLog(message, context)),
+    });
+    this.stopped = false;
```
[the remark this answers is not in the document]
nbouliol: Fixed in e05ddeb54, with one correction to the diff.

The race is real: the result now goes to a local and is dropped when `this.stopped` is already set.

Your diff kept `this.stopped = false` after the assignment, which breaks a legitimate `start()` following a `stop()` — the flag is still `true` from the previous shutdown, so the new guard would bail on it and `/bff` would never come back. Moved it to the top of `start()` instead, before the `await`: a start clears the previous shutdown, while a stop landing *during* the await is still seen by the guard.

Two tests in `agent-bff-lifecycle.test.ts`: a `stop()` released mid-`buildBff()` must answer 503 `bff_stopped` and never call the BFF callback, and a `start()` after a `stop()` must serve again. Verified the first fails without the guard, and that the second is what rules out your version.

packages/agent-bff/src/cors/cors-middleware.ts
```diff
@@ -2,11 +2,59 @@ import type { Middleware } from 'koa';
...
+
+  // `url.host` is already normalized by `new URL()` — lowercased, and stripped of a port that is
+  // the default for the scheme — while `ctx.host` is the raw `Host` header, which a client or proxy
+  // spells however it likes. Both differences have to be absorbed here, or the exemption refuses
+  // the very requests it exists for: `Host: APP.EXAMPLE.COM` or `Host: app.example.com:443` against
+  // `Origin: https://app.example.com`. A non-default port still has to match exactly.
+  const normalizedHost = host.toLowerCase();
+  const defaultPort = url.protocol === 'https:' ? '443' : '80';
+
+  return normalizedHost === url.host || normalizedHost === `${url.host}:${defaultPort}`;
```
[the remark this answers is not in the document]
nbouliol: Not changing this — the scheme is left out deliberately, and the reason is in the comment above `isSameOrigin`.

Comparing it would break the common deployment rather than protect it: behind a TLS-terminating proxy the browser reports `https` while `ctx.protocol` stays `http`, so a scheme check passes in development and refuses every same-origin request in production. `ctx.protocol` only tells the truth when `app.proxy` is set and `X-Forwarded-Proto` survives the hop, neither of which this middleware can assume about a host application it is mounted into.

On the exposure itself: for a browser to send `Origin: http://app.example.com` to the HTTPS BFF, a page has to be served over plaintext on that same host — which needs either an attacker already on the network path or a listener on that host, and HSTS removes it for any site that sets it. That attacker does not gain a credential: the BFF is bearer-only, no cookies, so the forged-origin request carries no session unless the caller already holds the token. And the allow-list is not what stands between a cross-site page and this BFF — there the browser sets `Origin` to the attacking page, which the exemption never matches.

What is genuinely true, and worth stating rather than hiding: on a plaintext-reachable host, a `http://` page can read BFF responses for requests it can already authenticate. Anyone who cannot accept that should name their origins in `allowedOrigins` and serve the UI cross-origin, which skips the exemption entirely.

Happy to revisit if you can show a path that does not require plaintext on the same host.

PR #1877, opened 2026-09-01
docs(agent): document the embedded BFF and the mount order it needs

nbouliol:
Stacked on #1876, last of the stack.

## What

- **`agent-bff/README.md`** gains an "Embedded in an agent" section: what `addBff()` inherits and why it cannot be overridden, and a table of the four things that differ from the standalone deployment (configuration source, no `AGENT_URL`, no `HTTP_PORT`, `openapiEnabled` default inverted, `/health` always 200). The `/health` section now documents `features` as the chain it is — `tokenEncryptionKey` → `oauth` → `ai` — rather than four independent switches.
- **`agent/README.md`**, which was empty, now covers the two optional in-process components: `addBff()` and `addWorkflowExecutor()`.
- **`packages/_example`** wires `addBff()` alongside `mountAiMcpServer()`, on the seven frameworks it already mounts on. It is the only place where the ordering constraint below is visible in real code.

## The ordering constraint, verified rather than assumed

The review of PRD-1076 predicted that an `express.json()` registered before the mount would empty the BFF's request bodies. Measured, it is worse and better at once: the request fails with **`500 stream.not.readable`**. Worse because a `POST` cannot work at all; better because it is loud — no silently dropped filter, projection or search.

Both directions are now pinned in the e2e suite, so the contract stops being folklore:

- agent mounted before the host's parser → 200, one record
- host's parser first → 500 `stream.not.readable`

## Tests

agent 85 suites / 1535 tests.

## Production guidance

Recorded in the BFF README: embedded for a single deployment, which is most of them; standalone when several agents share one BFF, or when the BFF and the agent have to scale separately.

Fixes PRD-1076

> [!NOTE]
> ### Add embedded BFF support to `Agent` and document mount order
> - `Agent.addBff()` registers one embedded BFF before startup; requests route through an in-process `AgentDispatcher` instead of HTTP, and the BFF starts after the agent stack mounts, stops before the workflow executor, and is invalidated on restart
> - `Agent.mountAiMcpServer()` now rejects MCP base paths that normalize into the `/bff` route space, reporting the supplied `basePath` in the error
> - CORS middleware in `createCorsMiddleware` returns `403 origin_not_allowed` for disallowed cross-origin non-OPTIONS requests instead of omitting headers; same-host and no-`Origin` requests are unaffected
> - Health route exposes a `configured` object with OAuth, AI, CORS, and OpenAPI flags; `parseConfig` `hasAllRequired` no longer requires the token-encryption key, so missing encryption reports `configured.oauth: false` while health stays `ok`
> - Risk: `createCorsMiddleware` in [cors-middleware.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1877/files#diff-20b9196c8106f4ddff192abd6df8553352e300d26d8fea63a3d84b88eed67f90) now 403s previously-pass-through disallowed origins; `parseConfig` in [env-config.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1877/files#diff-c69256c4bc98cec6990650c1cae02b21e9431f72715351ebdf12346d2a3ed5cb) changes `hasAllRequired` semantics for callers relying on encryption-key presence gating health
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 264c74e.</sup>
> 
>

### DIFF
--- a/packages/_example/.env.example
+++ b/packages/_example/.env.example
@@ -16,6 +16,13 @@ FOREST_AUTH_SECRET=
 # MCP OAuth client allowlist (comma-separated domains, e.g. dust.tt); unset allows any registered client
 # FOREST_MCP_ALLOWED_OAUTH_CLIENTS=
 
+# Embedded BFF, served at /bff on every port above.
+# Browser origins allowed to call it: comma-separated, exact origins, no wildcard. Empty means no
+# browser can reach it — put your frontend's origin here, e.g. http://localhost:4200
+BFF_ALLOWED_ORIGINS=
+# Enables the BFF's OAuth flow, and with it the AI relay (openssl rand -hex 32)
+BFF_TOKEN_ENCRYPTION_KEY=
+
 # Production
 # FOREST_ENV_SECRET=
 # FOREST_AUTH_SECRET=
--- a/packages/_example/.env.example
+++ b/packages/_example/.env.example
@@ -20,7 +20,8 @@ FOREST_AUTH_SECRET=
 # Browser origins allowed to call it: comma-separated, exact origins, no wildcard. Empty means no
 # browser can reach it — put your frontend's origin here, e.g. http://localhost:4200
 BFF_ALLOWED_ORIGINS=
-# Enables the BFF's OAuth flow, and with it the AI relay (openssl rand -hex 32)
+# Enables the BFF's OAuth flow, and with it the AI relay. Base64, decoding to exactly 32 bytes
+# (openssl rand -base64 32) — a hex key is rejected at startup
 BFF_TOKEN_ENCRYPTION_KEY=
 
 # Production
--- a/packages/_example/package.json
+++ b/packages/_example/package.json
@@ -6,6 +6,7 @@
   "dependencies": {
     "@faker-js/faker": "^7.6.0",
     "@forestadmin/agent": "*",
+    "@forestadmin/agent-bff": "1.23.4",
     "@forestadmin/datasource-dummy": "*",
     "@forestadmin/datasource-mongo": "*",
     "@forestadmin/datasource-mongoose": "*",
--- a/packages/_example/package.json
+++ b/packages/_example/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "@faker-js/faker": "^7.6.0",
     "@forestadmin/agent": "*",
-    "@forestadmin/agent-bff": "1.23.4",
+    "@forestadmin/agent-bff": "*",
     "@forestadmin/datasource-dummy": "*",
     "@forestadmin/datasource-mongo": "*",
     "@forestadmin/datasource-mongoose": "*",
--- a/packages/_example/src/forest/agent.ts
+++ b/packages/_example/src/forest/agent.ts
@@ -46,6 +46,11 @@ export default function makeAgent() {
         .filter(Boolean)
     : undefined;
 
+  const bffAllowedOrigins = (process.env.BFF_ALLOWED_ORIGINS ?? '')
+    .split(',')
+    .map(origin => origin.trim())
+    .filter(Boolean);
+
   return createAgent<Schema>(envOptions)
     .addDataSource(createSqlDataSource({ dialect: 'sqlite', storage: './assets/db.sqlite' }))
 
@@ -95,6 +100,14 @@ export default function makeAgent() {
     })
     .mountAiMcpServer({
       ...(allowedOAuthClients && { allowedOAuthClients }),
+    })
+
+    // Serves the BFF at /bff on every port this agent is mounted on, in-process. Without
+    // `allowedOrigins` no browser can call it, which for a backend-for-frontend is a mistake the
+    // BFF warns about at startup.
+    .addBff({
+      allowedOrigins: bffAllowedOrigins,
+      tokenEncryptionKey: process.env.BFF_TOKEN_ENCRYPTION_KEY,
     })
 
     .customizeCollection('card', customizeCard)
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -326,5 +326,12 @@ called second throws on the spot, at the builder call and not from `start()`, ra
 the MCP server quietly claim `/bff/oauth` and `/bff/mcp`. Any `basePath` landing inside `/bff`
 counts (`bff`, `/bff/`, `/bff/ai`), so mount the MCP server elsewhere.
 
+**One security control does not apply to the embedded mode.** If the environment has Forest's IP
+whitelist enabled, requests served under `/bff` are not subject to it: they reach the agent
+in-process, over what the whitelist treats as a trusted loopback caller. This is deliberate —
+applying it would filter the browsers a third-party UI is made of, and the standalone deployment
+never filtered the end user either, only the BFF's own host. A resolved API key or a valid OAuth
+session is still required, and the agent still logs a warning at startup naming the exemption.
+
 When to prefer which: embedded for a single deployment, which is most of them. Standalone when
 several agents share one BFF, or when the BFF and the agent have to scale separately.
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -256,8 +256,8 @@ The body still never discloses which config *keys* are present or missing — th
 internal config surface to an unauthenticated probe. It reports what is served, not how it was
 configured. Missing keys are logged once at startup (`Warn`) for operators. Every response the BFF
 itself produces carries the `X-Forest-Bff-Version` header, read from `package.json`. The one
-exception is the embedded `503 bff_not_started` below: the agent writes it before there is a BFF to
-read a version from.
+exception is the embedded `503 bff_not_started` / `bff_stopped` below: the agent writes those
+itself, with no BFF to read a version from.
 
 ## Embedded in an agent
 
@@ -316,13 +316,15 @@ NestJS needs no special handling as long as you mount before `listen()`: `NestFa
 does not install its body parser, `init()` does, and `listen()` is what triggers `init()` — so the
 middleware `mountOnNestJs()` registers is already ahead of it.
 
-**`/bff/*` answers `503 bff_not_started`** between `addBff()` and the end of `start()`, and again
-after `stop()` — the agent claims the prefix as soon as it is mounted, and a 404 there would read as
-"wrong url" rather than "not ready yet". A host restarting an agent under traffic will see it.
+**`/bff/*` answers `503` while the BFF is not serving**, and the two ends of the lifecycle are told
+apart: `bff_not_started` between `addBff()` and the end of `start()`, `bff_stopped` after `stop()`.
+The agent claims the prefix as soon as it is mounted, and a 404 there would read as "wrong url"
+rather than "not available". A probe should wait on the first and drain on the second.
 
-**`addBff()` and `mountAiMcpServer({ basePath: '/bff' })` are mutually exclusive** and throw at
-startup rather than letting the MCP server quietly claim `/bff/oauth` and `/bff/mcp`. Mount the MCP
-server elsewhere.
+**`addBff()` and `mountAiMcpServer({ basePath: '/bff' })` are mutually exclusive** — whichever is
+called second throws on the spot, at the builder call and not from `start()`, rather than letting
+the MCP server quietly claim `/bff/oauth` and `/bff/mcp`. Any `basePath` landing inside `/bff`
+counts (`bff`, `/bff/`, `/bff/ai`), so mount the MCP server elsewhere.
 
 When to prefer which: embedded for a single deployment, which is most of them. Standalone when
 several agents share one BFF, or when the BFF and the agent have to scale separately.
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -33,6 +33,9 @@ dropped collection reachable.
 
 ## Usage
 
+Two ways to run it: embedded in a Forest agent (`agent.addBff()`, see
+[Embedded in an agent](#embedded-in-an-agent)) or standalone, described here.
+
 Packaged / production — run the bin:
 
 ```bash
@@ -230,11 +233,69 @@ with `session_expired`. Horizontal scaling requires both a shared session store 
 
 ```jsonc
 // 200 — all required config present
-{ "status": "ok", "version": "<package version>" }
+{
+  "status": "ok",
+  "version": "<package version>",
+  "features": { "oauth": true, "ai": true, "cors": true, "openapi": true }
+}
 // 503 — one or more required keys missing
-{ "status": "degraded", "version": "<package version>" }
+{ "status": "degraded", "version": "<package version>", "features": { /* … */ } }
 ```
 
-The body never discloses which config keys are present or missing — that would leak the internal
-config surface to an unauthenticated probe. Missing keys are logged once at startup (`Warn`) for
-operators. Every response carries the `X-Forest-Bff-Version` header, read from `package.json`.
+`features` says what this deployment actually serves. They form a chain, not four independent
+switches:
+
+| Feature | Switched on by |
+| --- | --- |
+| `oauth` | `BFF_TOKEN_ENCRYPTION_KEY` (`tokenEncryptionKey` when embedded) |
+| `ai` | `oauth` — the relay needs a session, and only the OAuth flow creates one |
+| `cors` | a non-empty `BFF_ALLOWED_ORIGINS` (`allowedOrigins`) |
+| `openapi` | `BFF_OPENAPI_ENABLED` (`openapiEnabled`), and a mounted agent edge |
+
+The body still never discloses which config *keys* are present or missing — that would leak the
+internal config surface to an unauthenticated probe. It reports what is served, not how it was
+configured. Missing keys are logged once at startup (`Warn`) for operators. Every response carries
+the `X-Forest-Bff-Version` header, read from `package.json`.
+
+## Embedded in an agent
+
+The same BFF runs inside a Forest agent, with no second deployment and no second port:
+
+```ts
+createAgent(options)
+  .addDataSource(/* … */)
+  .addBff({ allowedOrigins: ['https://my-app.com'] })
+  .start();
+```
+
+It answers under `/bff` on the agent's own port — `/bff/agent/v1/{collection}/list`,
+`/bff/health`, `/bff/oauth/*` — so the REST contract is the one documented above and only the base
+url changes. A client generated from the OpenAPI document stays portable: the document's `servers`
+entry carries the prefix.
+
+What differs from the standalone deployment:
+
+| | Standalone | Embedded |
+| --- | --- | --- |
+| Configuration | environment variables | `addBff()` options; the secrets, the Forest urls and the logger are inherited from the agent and cannot be overridden |
+| `AGENT_URL` | required, an http(s) url | gone — the BFF reaches the agent in the same process, without a socket |
+| `HTTP_PORT` | its own listener | gone — the agent's port serves it |
+| `openapiEnabled` | `true` | `false`. The document is not filtered per caller, so adding a BFF must not silently publish every collection and field name on an already-open port |
+| `/health` | 503 until every required key is set | always 200: everything required is inherited, so there is no gap to report. Read `features` |
+
+**Registration order matters on Express and Connect-style hosts.** Mount the agent *before* any
+body parser of your own:
+
+```ts
+const app = express();
+agent.mountOnExpress(app);  // first
+app.use(express.json());    // then yours
+```
+
+A body parser that runs first has already consumed the request stream, and nothing downstream can
+put it back: every BFF `POST` then answers `500 stream.not.readable`. The same applies to a
+permissive `cors()` registered ahead of the mount — it answers the preflight itself, and the BFF's
+strict allow-list never gets a say.
+
+When to prefer which: embedded for a single deployment, which is most of them. Standalone when
+several agents share one BFF, or when the BFF and the agent have to scale separately.
--- a/packages/agent-bff/README.md
+++ b/packages/agent-bff/README.md
@@ -162,9 +162,12 @@ Two layers, both driven by exact-origin matching (case-insensitive scheme/host, 
 normalized away, no trailing slash, no wildcard, no subdomain matching):
 
 - **Layer 1 (transport)** — the only layer that sets `Access-Control-Allow-Origin`. An origin in
-  `BFF_ALLOWED_ORIGINS` is echoed back exactly; anything else gets no CORS headers (the browser
-  blocks). Applies to `POST /oauth/token` too. Preflight `OPTIONS` from an allow-listed origin gets
-  the allowed methods + headers; credentials are never enabled.
+  `BFF_ALLOWED_ORIGINS` is echoed back exactly; any other `Origin` is refused outright with
+  `403 origin_not_allowed`, rather than served without the header and left for the browser to
+  discard — omitting the header only stops the caller from *reading* the answer, and the request
+  would already have run. A request with no `Origin` at all is untouched, which is every
+  server-to-server api-key call. Applies to `POST /oauth/token` too. Preflight `OPTIONS` from an
+  allow-listed origin gets the allowed methods + headers; credentials are never enabled.
 - **Layer 2 (per-key authorization, Mode 2 only)** — when the resolved key has a non-empty
   `allowedOrigins`, an `Origin` sent by the client must be in that list, else
   `403 origin_not_allowed`. A request with no `Origin` at all — a server-side client such as curl,
@@ -236,16 +239,17 @@ with `session_expired`. Horizontal scaling requires both a shared session store 
 {
   "status": "ok",
   "version": "<package version>",
-  "features": { "oauth": true, "ai": true, "cors": true, "openapi": true }
+  "configured": { "oauth": true, "ai": true, "cors": true, "openapi": true }
 }
 // 503 — one or more required keys missing
-{ "status": "degraded", "version": "<package version>", "features": { /* … */ } }
+{ "status": "degraded", "version": "<package version>", "configured": { /* … */ } }
 ```
 
-`features` says what this deployment actually serves. They form a chain, not four independent
-switches:
+`configured` says which optional surfaces this deployment was set up to serve — deliberately not
+that they work: `oauth` is `true` as soon as a key is set, whether or not the Forest server ever
+answers. They form a chain, not four independent switches:
 
-| Feature | Switched on by |
+| Surface | Switched on by |
 | --- | --- |
 | `oauth` | `BFF_TOKEN_ENCRYPTION_KEY` **and** `FOREST_SERVER_URL`, `FOREST_ENV_SECRET`, `FOREST_APP_URL`, `FOREST_AUTH_SECRET` — the routes need a Forest server to talk to as much as a key. Embedded, only `tokenEncryptionKey` is yours to set: the other four are inherited |
 | `ai` | `oauth` — the relay needs a session, and only the OAuth flow creates one |
@@ -253,8 +257,8 @@ switches:
 | `openapi` | `BFF_OPENAPI_ENABLED` (`openapiEnabled`), and a mounted agent edge |
 
 The body still never discloses which config *keys* are present or missing — that would leak the
-internal config surface to an unauthenticated probe. It reports what is served, not how it was
-configured. Missing keys are logged once at startup (`Warn`) for operators. Every response the BFF
+internal config surface to an unauthenticated probe. It names the surfaces, never the values behind
+them. Missing keys are logged once at startup (`Warn`) for operators. Every response the BFF
 itself produces carries the `X-Forest-Bff-Version` header, read from `package.json`. The one
 exception is the embedded `503 bff_not_started` / `bff_stopped` below: the agent writes those
 itself, with no BFF to read a version from.
@@ -278,9 +282,10 @@ It answers under `/bff` on whatever port the agent is mounted on — `/bff/agent
 `/bff/health`, `/bff/oauth/*` — so the REST contract is the one documented above and only the base
 url changes.
 
-The prefix is fixed at `/bff` and is resolved against the url the host hands the agent, so **mount
-the agent on the root application, not on a sub-router**: an agent mounted on an express router at
-`/api` answers at `/api/bff`, while the OpenAPI document and the docs page still say `/bff`.
+A host serving the agent under a sub-path of its own is supported: `app.use('/api', mounted)` puts
+the BFF at `/api/bff`, and the emitted prefix is derived per request from `originalUrl`
+(`base-path.ts`), so the `servers` entry and the docs page carry `/api/bff` rather than the
+configured `/bff`.
 
 The *served* OpenAPI document carries the prefix in its `servers` entry, so a client generated from
 `GET /bff/agent/openapi.json` is correct as-is. The *exported* one does not: `forest-bff openapi`
@@ -295,7 +300,7 @@ What differs from the standalone deployment:
 | `AGENT_URL` | required, an http(s) url | gone — the BFF reaches the agent in the same process, without a socket |
 | `HTTP_PORT` | its own listener | gone — the agent's port serves it |
 | `openapiEnabled` | `true` | `false`. The document is not filtered per caller, so adding a BFF must not silently publish every collection and field name on an already-open port |
-| `/health` | 503 until every required key is set | always 200: an in-process dispatcher short-circuits the readiness check, since a 503 here would let a load balancer restart a process that serves api-key traffic fine. It stays 200 with no `tokenEncryptionKey`, which is *not* inherited — read `features` to know what is on |
+| `/health` | 503 until every required key is set | always 200: an in-process dispatcher short-circuits the readiness check, since a 503 here would let a load balancer restart a process that serves api-key traffic fine. It stays 200 with no `tokenEncryptionKey`, which is *not* inherited — read `configured` to know what is on |
 
 **Registration order matters on Express and Connect-style hosts.** Mount the agent *before* any
 body parser of your own:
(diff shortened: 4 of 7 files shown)


packages/agent/README.md
```diff
@@ -0,0 +1,48 @@
...
+```ts
+createAgent(options)
+  .addDataSource(/* … */)
+  .addBff({ allowedOrigins: ['https://my-app.com'] })
+  .start();
+```
+
+Serves the REST BFF at `/bff` on the agent's own port, on every mount target. It reaches the agent
+in the same process, so there is no second port, no agent url to configure, and no secrets to keep
+in sync — `authSecret`, `envSecret`, the Forest urls and the logger are inherited.
```
Tonours: Claude Opus 5 (claude-opus-5): **Must fix**

This quick-start opens no HTTP listener, and the sentence two lines below tells the reader it does.

`start()` calls `mount()`, which only runs the `onFirstStart` / `onEachStart` hooks. The single `createServer(...).listen(...)` in the package is inside `mountOnStandaloneServer`; the other four targets register onto a server the host provides. With no `mountOn*` call, no socket is ever opened.

What the reader observes: the process starts, `EmbeddedBff.start()` logs `Embedded BFF mounted on /bff` — actively confirming the false expectation — and `curl localhost:PORT/bff/health` answers *connection refused*. Then the prose says *"Serves the REST BFF at `/bff` on the agent's own port"*, describing a port the snippet never creates.

Add `.mountOnStandaloneServer(3351)` before `.start()`, or an explicit `mountOnExpress(app)`.

Same in `agent-bff/README.md`'s embedded quick-start; and its registration-order sample mounts but never calls `agent.start()`, so that one yields `503 bff_not_started` on `/bff/*` and `{"error":"Agent is not started"}` on `/forest/*`.

Worth fixing at the source too: `addBff`'s own JSDoc in `agent.ts` carries the same mountless snippet, which is where both READMEs got it. Fixing only the READMEs leaves the copy a reader sees in their editor wrong.

packages/agent-bff/README.md
```diff
@@ -206,11 +209,69 @@ with `session_expired`. Horizontal scaling requires both a shared session store
...
+  .start();
 ```
 
-The body never discloses which config keys are present or missing — that would leak the internal
-config surface to an unauthenticated probe. Missing keys are logged once at startup (`Warn`) for
-operators. Every response carries the `X-Forest-Bff-Version` header, read from `package.json`.
+It answers under `/bff` on the agent's own port — `/bff/agent/v1/{collection}/list`,
+`/bff/health`, `/bff/oauth/*` — so the REST contract is the one documented above and only the base
+url changes. A client generated from the OpenAPI document stays portable: the document's `servers`
+entry carries the prefix.
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**

> A client generated from the OpenAPI document stays portable: the document's `servers` entry carries the prefix.

True of the **served** document, false of the **exported** one — and embedded, the exported one is the only document a customer can normally get.

The served route carries it: `embedded-bff.ts` passes `basePath: '/bff'` → `createOpenApiRoutes({ basePath })` → `servers: [{ url: basePath || '/' }]`. The CLI export never passes it: `cli-dispatch.ts` calls `generateOpenApiDocument(version, { hasAiQueryRoute })` and `buildUnfoldedDocument(…, { version, hasAiQueryRoute })` with no `basePath`, and there is no `--base-path` in `USAGE` — the CLI form *cannot* carry a prefix, so it always emits `servers: [{ url: '/' }]`.

What closes the trap: embedded defaults `openapiEnabled` to `false`, and `openapi` is additionally gated on a mounted agent edge, so in the default embedded configuration the served document is not reachable at all — while this same README recommends `forest-bff openapi` as the form to generate a client from. The reader follows that recommendation, generates against base `/`, and every call goes to `/agent/v1/{collection}/list` instead of `/bff/agent/v1/…`. The embedded BFF only claims `/bff*`, so all of them fall through to the host and 404, with nothing for the client to branch on.

Either scope the sentence to the served document and say the export assumes the origin root, or add `--base-path` to the export and thread it through. Note this is the same divergence I raised on #1873 from the other side; fixing it there fixes it here.

Tonours (commented):
**Spec (PRD-1076)**: steps 18, 20 and 21 are delivered. Step 19 is delivered on its documentation half and half of its test half — and the half that got the test is the less dangerous one. Step 17's documentation requirement is not met at all.

Step 19 asked to *"Document the host-registration contract **and test it**"*, naming two hazards. The body-parser hazard is done properly: documented with the measured failure mode, and the new test asserts both directions with an exact status **and** an exact error type, on isolated agents and per-instance schema paths. The `cors()` hazard is documented well (`agent-bff/README.md:273-274`) and has no test — and it is the security-relevant one, since a permissive host `cors()` silently neutralises the BFF's strict origin allow-list.

**The ticket's prediction was wrong and this PR is right to correct it.** Step 19 predicted an emptied request body; the measured behaviour is `500 stream.not.readable`, and I traced the chain to confirm it: `express.json()` leaves `req.readable === false`, `@koa/bodyparser` reaches `raw-body`, which throws `createError(500, 'stream is not readable', { type: 'stream.not.readable' })`, and `isSerializableError` duck-types on `status`/`type`/`message` so the type is echoed verbatim rather than flattened to `internal_error`. Replacing a guess with a measurement is the right call and worth saying.

**Also verified accurate**, so it does not get re-checked: the inherited secrets/urls/logger really cannot be overridden (`BffEmbedOptions` exposes none of them); `openapiEnabled` really defaults off embedded and on standalone; the `features` block's four flags match their stated sources, including the `openapi` conjunction with a mounted agent edge; `/bff` really is registered on all five mount targets plus the standalone callback; and the `/health` disclosure note draws the right distinction — it says which capabilities are configured without naming config keys, which is what the code does.

Seven findings inline.

packages/agent/README.md
```diff
@@ -0,0 +1,48 @@
...
+with it the AI relay), `allowedOrigins` enables browser access, `openapiEnabled` serves the docs
+(off by default when embedded). `GET /bff/health` reports which of them are on.
+
+**Mount the agent before any body parser of your own** — a parser that runs first consumes the
+request stream, and every BFF `POST` then answers `500 stream.not.readable`.
+
+See [`@forestadmin/agent-bff`](../agent-bff/README.md) for the routes, the auth modes and the
+differences with the standalone deployment.
+
+### A workflow executor — `agent.addWorkflowExecutor()`
```
Tonours, line 66: Claude Opus 5 (claude-opus-5): **Should fix**

PRD-1076 step 17 asked, in these words, to *"document that the `Promise.race` cancels nothing: an action cut at 10s keeps running agent-side, and a retry can double the mutation"*. Nothing in this PR says it — I grepped `agent/README.md`, `agent-bff/README.md` and `mcp-server/README.md` at this SHA for `cancels nothing`, `keeps running`, `double the mutation` and `not cancel`: no hit. The only "abandoned, not cancelled" note in the repo is about `UploadStorage.download`, a different path.

The mechanism is real: `InProcessDispatcher.injectWithTimeout` races `inject()` against a `setTimeout` and, on timeout, leaves the injection unobserved — its own comment says so — because `light-my-request` exposes no abort. The BFF routes agent calls through it with `agentTimeoutMs` as the bound, an option this PR documents on `BffEmbedOptions`.

So the sequence a host can walk into: they set `agentTimeoutMs` low because the docs offer it, a `POST /bff/agent/v1/{collection}/actions/{action}/execute` exceeds it, the BFF returns an error — and the action finishes agent-side anyway. The client retries and the mutation applies twice. A timeout that is not a cancellation is exactly the kind of thing a host cannot infer and must be told.

This is the natural place for it, next to the mount-order warning.

packages/_example/package.json
```diff
@@ -6,6 +6,7 @@
   "dependencies": {
     "@faker-js/faker": "^7.6.0",
     "@forestadmin/agent": "*",
+    "@forestadmin/agent-bff": "1.23.4",
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**

This is the only internal dependency in the file pinned to a version — the seven `@forestadmin/*` siblings around it all use `"*"`. It resolves today only because `packages/agent-bff/package.json` happens to be at `1.23.4`.

It breaks on the next agent-bff release. `_example` is `"private": true`, so `@qiwi/multi-semantic-release` does not rewrite its ranges — the mechanism that keeps every other exact pin in this repo current. When agent-bff goes to `1.23.5` the workspace version stops satisfying `1.23.4`, yarn resolves the published copy from the registry instead of linking the workspace, and `yarn install --frozen-lockfile` fails on a stale lockfile. Locally the example would then run against the published BFF rather than the branch's sources — which defeats the point of the example added in step 20, since it exists to make the body-parser ordering visible against *this* code.

`"*"`, like its six neighbours.

packages/agent/test/bff/embedded-bff.e2e.test.ts
```diff
@@ -240,4 +240,75 @@ describe('embedded BFF', () => {
...
+    function listBooks(hostApp: express.Express) {
+      return supertest(hostApp)
+        .post('/bff/agent/v1/books/list')
+        .set('Authorization', `Bearer ${sessionToken()}`)
+        .set('X-Forest-Timezone', 'Europe/Paris')
+        .send({ projection: ['id', 'title'], search: 'foundation' });
+    }
+
+    it(
+      'should serve normally when the agent is mounted before the host body parser',
```
Tonours, line 631: Claude Opus 5 (claude-opus-5): **Should fix**

Step 19 named two host-registration hazards and asked for both to be documented **and tested**. This block does the body-parser one properly — both orders, exact status, exact `error.type`, isolated agents and per-instance schema paths, and a positive control that genuinely discriminates (a 3-book fixture with `search: 'foundation'` can only return one row if the body was parsed). Nothing to fix in what is here.

The missing half is the `cors()` one, and it is the more dangerous of the two. The docs describe it correctly — *"a permissive `cors()` registered ahead of the mount — it answers the preflight itself, and the BFF's strict allow-list never gets a say"* — but no test exercises it: `git grep -i 'preflight|OPTIONS|cors'` over this file returns only the `features` assertion.

The consequence of getting it wrong is a security control silently disabled, not a 500: the BFF's exact-origin allow-list is the thing standing between a browser on an arbitrary origin and the customer's data, and a host `cors()` registered first answers the preflight with its own policy. That is worth a case in the same shape as the one above — `app.use(cors())` before the mount, an `OPTIONS` on a `/bff/agent/*` route from a non-allow-listed origin, and an assertion on which `Access-Control-Allow-Origin` comes back.

packages/agent/README.md
```diff
@@ -0,0 +1,48 @@
...
+  .addBff({ allowedOrigins: ['https://my-app.com'] })
+  .start();
+```
+
+Serves the REST BFF at `/bff` on the agent's own port, on every mount target. It reaches the agent
+in the same process, so there is no second port, no agent url to configure, and no secrets to keep
+in sync — `authSecret`, `envSecret`, the Forest urls and the logger are inherited.
+
+Everything `addBff()` takes is a feature it switches on: `tokenEncryptionKey` enables OAuth (and
+with it the AI relay), `allowedOrigins` enables browser access, `openapiEnabled` serves the docs
```
Tonours, line 38: Claude Opus 5 (claude-opus-5): **Preferential**

Two imprecisions in this sentence, and the second one hides a real misconfiguration.

**"on the agent's own port"** is true only for `mountOnStandaloneServer`. On Express, Fastify, Koa and NestJS the agent registers into the host app and the listener is the host's. The neighbouring "there is no second port" is the accurate way to say it; "the agent's own port" reads as if the agent opened one. `on whatever port the agent is mounted on` would cover both.

**"at `/bff`"** holds only when the agent is mounted on the root app. `mountOnExpress` accepts, in its own words, the express app *or router*, and `isBffRoute` matches the url the router has already stripped. So an agent mounted on a router at `/api` serves the BFF at `/api/bff` — while `basePath` is hardcoded to `/bff`, so the OpenAPI `servers` entry and the docs page still emit `/bff` and a generated client points at a 404.

That case is exactly what PRD-1076 step 4 asked to handle by deriving the emitted base from `req.originalUrl` with `basePath` as the fallback, which is the half of step 4 that has not landed (I raised it on #1873). Worth knowing that these three sit on one thread: the derivation is missing, so a sub-router mount is silently wrong, and this sentence plus the portability promise are what keep it invisible. Until the derivation lands, saying that the prefix is fixed at `/bff` and that mounting the agent on a sub-router is not supported would at least make it loud.

packages/agent-bff/README.md
```diff
@@ -206,11 +209,69 @@ with `session_expired`. Horizontal scaling requires both a shared session store
...
+| --- | --- |
+| `oauth` | `BFF_TOKEN_ENCRYPTION_KEY` (`tokenEncryptionKey` when embedded) |
+| `ai` | `oauth` — the relay needs a session, and only the OAuth flow creates one |
+| `cors` | a non-empty `BFF_ALLOWED_ORIGINS` (`allowedOrigins`) |
+| `openapi` | `BFF_OPENAPI_ENABLED` (`openapiEnabled`), and a mounted agent edge |
+
+The body still never discloses which config *keys* are present or missing — that would leak the
+internal config surface to an unauthenticated probe. It reports what is served, not how it was
+configured. Missing keys are logged once at startup (`Warn`) for operators. Every response carries
+the `X-Forest-Bff-Version` header, read from `package.json`.
```
Tonours: Claude Opus 5 (claude-opus-5): **Preferential**

> Every response carries the `X-Forest-Bff-Version` header, read from `package.json`.

Not every response, as of the sibling PR in this stack. `EmbeddedBff.handle` answers the pre-start and post-`stop()` case by writing to `res` directly — `res.statusCode = 503`, `res.setHeader('Content-Type', …)`, `res.end(…)` — which bypasses the Koa chain entirely, so `createVersionHeaderMiddleware` never runs and that response carries no version header.

Two related gaps in the same embedded section, both being things a host will actually meet:

- **`bff_not_started` is undocumented.** It is what `/bff/*` answers between `addBff()` and the end of `start()`, and again after `stop()`. A host restarting an agent under traffic sees it, and its type does not appear anywhere in the error-contract documentation.
- **The MCP collision guard is in neither README.** `addBff()` combined with `mountAiMcpServer({ basePath: '/bff' })` throws at startup. That is a hard boot failure a host can hit, and it belongs next to the mount-order warning rather than being discovered from a stack trace.

One more, smaller: the table's justification for the embedded `/health` always answering 200 — "everything required is inherited, so there is no gap to report" — is not why. `tokenEncryptionKey` is part of `hasAllRequired` and is *not* inherited; it is an optional `addBff()` option. The 200 comes from the dispatcher short-circuit in `build-bff.ts`. The operational advice that follows (read `features`) is right either way, so only the reason needs correcting.

packages/agent-bff/README.md
```diff
@@ -206,11 +209,69 @@ with `session_expired`. Horizontal scaling requires both a shared session store
...
+
+## Embedded in an agent
+
+The same BFF runs inside a Forest agent, with no second deployment and no second port:
+
+```ts
+createAgent(options)
+  .addDataSource(/* … */)
+  .addBff({ allowedOrigins: ['https://my-app.com'] })
+  .start();
```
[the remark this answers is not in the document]
nbouliol: Fixed — both quick-starts now chain `.mountOnStandaloneServer(3351)` before `.start()`, and each says explicitly that a `mountOn*` call is what opens the socket. Same fix applied to the `addBff()` JSDoc, which is where the READMEs got the snippet.

packages/_example/src/forest/agent.ts
```diff
@@ -97,6 +102,14 @@ export default function makeAgent() {
       ...(allowedOAuthClients && { allowedOAuthClients }),
     })
 
+    // Serves the BFF at /bff on every port this agent is mounted on, in-process. Without
+    // `allowedOrigins` no browser can call it, which for a backend-for-frontend is a mistake the
+    // BFF warns about at startup.
+    .addBff({
```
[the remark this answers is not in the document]
nbouliol: Not changing this — the premise does not hold. `NestFactory.create()` does not install the body parser; `NestApplication.init()` does (`nest-application.js:99-102`), and `listen()` is what triggers `init()`. `mountOnNestJs()` calls `nestJs.use(...)`, which goes straight to the express instance, and the example mounts before `listen()` (`_example/src/index.ts`), so the agent middleware is ahead of Nest's parser. Measured it: a middleware registered the same way sees `req.readable === true` and `req.body === undefined` on a JSON POST. I documented the ordering rule in the BFF README so the constraint (mount before `listen()`) is written down.

packages/agent-bff/README.md
```diff
@@ -206,11 +209,69 @@ with `session_expired`. Horizontal scaling requires both a shared session store
...
-{ "status": "degraded", "version": "<package version>" }
+{ "status": "degraded", "version": "<package version>", "features": { /* … */ } }
+```
+
+`features` says what this deployment actually serves. They form a chain, not four independent
+switches:
+
+| Feature | Switched on by |
+| --- | --- |
+| `oauth` | `BFF_TOKEN_ENCRYPTION_KEY` (`tokenEncryptionKey` when embedded) |
```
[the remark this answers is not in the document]
nbouliol: Fixed — the `oauth` row now lists all five: `resolveOAuthConfig` (`build-bff.ts:143-151`) requires `FOREST_SERVER_URL`, `FOREST_ENV_SECRET`, `FOREST_APP_URL`, `FOREST_AUTH_SECRET` and `BFF_TOKEN_ENCRYPTION_KEY`. The row also says that embedded, only `tokenEncryptionKey` is yours to set — the other four are inherited from the agent.

packages/_example/src/forest/agent.ts
```diff
@@ -46,6 +46,11 @@ export default function makeAgent() {
         .filter(Boolean)
     : undefined;
 
+  const bffAllowedOrigins = (process.env.BFF_ALLOWED_ORIGINS ?? '')
```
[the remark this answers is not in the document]
nbouliol: Fixed — `.env.example` gains `BFF_ALLOWED_ORIGINS` (with the frontend-origin guidance and a `http://localhost:4200` example) and `BFF_TOKEN_ENCRYPTION_KEY`.

packages/agent/README.md
nbouliol: Fixed in all three places. Both quick-starts now chain `.mountOnStandaloneServer(3351)` before `.start()` and state that `start()` only builds the router; the registration-order sample ends with `await agent.start()`; and the `addBff()` JSDoc `@example` carries the mount too, so the copy shown in an editor matches. The "on the agent's own port" wording is gone as well — see the reply on the other thread.

packages/agent-bff/README.md
nbouliol: Documented rather than fixed in code, since `--base-path` belongs to the CLI in #1873. The README now separates the two: the *served* document carries the prefix in `servers` (`openapi-routes.ts:36` passes `basePath`), while `forest-bff openapi` (`cli-dispatch.ts:99,104`) has no way to be told about one and always emits `servers: [{ "url": "/" }]` — so generating a client from the export against an embedded BFF means setting the base url to `/bff` yourself.

packages/agent/README.md
nbouliol: Documented — the agent README now has an `agentTimeoutMs` bounds the wait, not the work paragraph: the timeout is not a cancellation, the in-process request runs to completion, an action cut at the timeout still applies its mutation, and a client retry applies it twice. Guidance added to size the timeout above the slowest action and to make actions idempotent if they are retried. Verified the chain: `resolveTransport` (`build-bff.ts:282-284`) hands `agentTimeoutMs` to the in-process transport, which reaches `InProcessDispatcher.injectWithTimeout`.

packages/_example/package.json
nbouliol: Fixed — `"*"`, like the seven siblings.

packages/agent/test/bff/embedded-bff.e2e.test.ts
nbouliol: Added — two cases in the same shape as the body-parser pair, with a permissive host CORS handler (allow `*`, answer any preflight) and an `OPTIONS /bff/agent/v1/books/list` from a non-allow-listed origin. Mounted first: 204 with no `Access-Control-Allow-Origin`, the BFF allow-list refusing. Host handler first: 204 with `Access-Control-Allow-Origin: *`. The pair discriminates, so the hazard is pinned rather than described.

packages/agent/README.md
nbouliol: Fixed both. "on the agent's own port" is now "on whatever port the agent is mounted on", with the mount targets listed. And both READMEs now say the `/bff` prefix is fixed and resolved against the url the host hands the agent, so an agent mounted on an express router at `/api` answers at `/api/bff` while the document still advertises `/bff` — mount on the root application, not on a sub-router. Left as a documented limitation here; the `req.originalUrl` derivation is #1873's.

packages/agent-bff/README.md
nbouliol: All four fixed. (1) The version-header sentence now says "every response the BFF itself produces", with the embedded `503 bff_not_started` named as the exception — `EmbeddedBff.handle` writes it on `res` directly and never enters the Koa chain. (2) `bff_not_started` is now documented: what answers it, and that it spans `addBff()` → end of `start()` and again after `stop()`. (3) The `addBff()` / `mountAiMcpServer({ basePath: "/bff" })` collision is documented next to the mount-order warning in both READMEs. (4) The `/health` row now gives the real reason — the in-process dispatcher short-circuits the readiness check (`build-bff.ts:522`) — and notes it stays 200 with no `tokenEncryptionKey`, which is not inherited.

Tonours (commented):
My eight findings are all addressed on 7b79c5a: both quick-starts and the `addBff()` JSDoc mount before `start()`, the registration-order sample starts the agent, the served/exported OpenAPI split is documented, `agentTimeoutMs bounds the wait, not the work` is there, the example dependency is `"*"`, the CORS preflight pair is tested in the same shape as the body-parser one, the fixed `/bff` prefix and the sub-router limitation are stated, and the four items on the agent-bff README are in.

Holding the approval on one thing that is still open — Macroscope's last comment, which got no reply:

`packages/_example/.env.example:23` tells the reader to generate `BFF_TOKEN_ENCRYPTION_KEY` with `openssl rand -hex 32`. That is 64 hex characters; `parseEncryptionKey` (`env-config.ts:72-86`) accepts the charset but decodes it to 48 bytes, not 32, so the agent throws `Invalid configuration: BFF_TOKEN_ENCRYPTION_KEY must be base64-encoded and exactly 32 bytes (AES-256).` at startup. Anyone following the example file cannot boot it.

`openssl rand -base64 32` is the right command. The line right below (`FOREST_EXECUTOR_ENCRYPTION_KEY`, line 34) is correct as `-hex 32` — that value is not validated — so only line 23 changes. Both READMEs already say "base64-encoded 32-byte", so this is the example file alone.

Approving as soon as that line is fixed.

packages/_example/.env.example
```diff
@@ -16,6 +16,13 @@ FOREST_AUTH_SECRET=
 # MCP OAuth client allowlist (comma-separated domains, e.g. dust.tt); unset allows any registered client
 # FOREST_MCP_ALLOWED_OAUTH_CLIENTS=
 
+# Embedded BFF, served at /bff on every port above.
+# Browser origins allowed to call it: comma-separated, exact origins, no wildcard. Empty means no
+# browser can reach it — put your frontend's origin here, e.g. http://localhost:4200
+BFF_ALLOWED_ORIGINS=
+# Enables the BFF's OAuth flow, and with it the AI relay (openssl rand -hex 32)
```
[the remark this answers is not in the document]
nbouliol: Fixed — `-base64 32`. Confirmed the arithmetic: `openssl rand -hex 32` is 64 hex characters, `Buffer.from(…, "base64")` turns those into 48 bytes, and `isValidEncryptionKey` (`env-config.ts:72-74`) wants exactly 32, so the agent threw at startup. `openssl rand -base64 32` gives 44 characters decoding to 32. The line now also states what the parser wants, not just how to generate it. `FOREST_EXECUTOR_ENCRYPTION_KEY` below stays `-hex 32` — that value is not validated.

packages/agent-bff/README.md
```diff
@@ -206,11 +209,96 @@ with `session_expired`. Horizontal scaling requires both a shared session store
...
+permissive `cors()` registered ahead of the mount — it answers the preflight itself, and the BFF's
+strict allow-list never gets a say.
+
+NestJS needs no special handling as long as you mount before `listen()`: `NestFactory.create()`
+does not install its body parser, `init()` does, and `listen()` is what triggers `init()` — so the
+middleware `mountOnNestJs()` registers is already ahead of it.
+
+**`/bff/*` answers `503 bff_not_started`** between `addBff()` and the end of `start()`, and again
+after `stop()` — the agent claims the prefix as soon as it is mounted, and a 404 there would read as
+"wrong url" rather than "not ready yet". A host restarting an agent under traffic will see it.
```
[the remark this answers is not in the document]
nbouliol: Fixed — the paragraph now names both: `bff_not_started` between `addBff()` and the end of `start()`, `bff_stopped` after `stop()` (`embedded-bff.ts:121`), with the reason they are distinct (a probe waits on the first, drains on the second). The version-header exception two sections up lists both types as well.

packages/agent-bff/README.md
```diff
@@ -206,11 +209,96 @@ with `session_expired`. Horizontal scaling requires both a shared session store
...
+
+**`addBff()` and `mountAiMcpServer({ basePath: '/bff' })` are mutually exclusive** and throw at
+startup rather than letting the MCP server quietly claim `/bff/oauth` and `/bff/mcp`. Mount the MCP
+server elsewhere.
```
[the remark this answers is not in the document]
nbouliol: Fixed in both READMEs — "whichever is called second throws on the spot, at the builder call and not from `start()`". Both guards are synchronous: `mountAiMcpServer` (`agent.ts:343`) and `addBff` (`agent.ts:428`). Also widened the wording to match `collidesWithBff`, which normalizes the path — `bff`, `/bff/` and `/bff/ai` collide too, not only an exact `/bff`.

nbouliol:
Fixed on 49a7010. `.env.example:23` is now `openssl rand -base64 32`, with the line stating what `parseEncryptionKey` wants (base64, decoding to exactly 32 bytes) rather than only how to generate it. `FOREST_EXECUTOR_ENCRYPTION_KEY` stays `-hex 32`, as you noted.

Two more from Macroscope in the same commit: the lifecycle paragraph now distinguishes `bff_not_started` from `bff_stopped`, and the MCP collision is documented as throwing from the builder call rather than from `start()` (also widened to any `basePath` landing inside `/bff`, matching `collidesWithBff`).

agent: 63/63 on `test/bff`, eslint `src test` 0 errors.

Tonours (changes requested):
Delta since my last review (7b79c5a → 0c6f3df): rebase onto #1876 plus three doc commits and 8e84f1f. Three statements no longer match the code they describe — ironic given 2340409's own subject.

## Should fix

**1. `/health` still documents `features`; the code now serializes `configured`** — agent-bff/README.md:215 and :218 show `"features": { … }`, and :221 says *``features`` says what this deployment actually serves*. 55875e0 (#1876) renamed the key to `configured` and inverted the claim on purpose (*"Deliberately not a statement that they work"*, health-route.ts:6-10). A reader following this README branches on a key that no longer exists. Same for the closing sentence at :230 (*"It reports what is served, not how it was configured"*) — that is now precisely backwards. This is macroscopeapp's open thread; sample, table intro and closing sentence all need the rename.

**2. The sub-router paragraph describes the pre-90001a0 behavior** — agent/README.md:46-48: *"an agent mounted on an express router at `/api` answers at `/api/bff` while the OpenAPI document still advertises `/bff`"*. Since 90001a0 (#1873) the document advertises `/api/bff`: the e2e suite added in 55875e0 asserts `servers: [{ url: '/api/bff' }]` and the prefixed docs-page urls for exactly this `app.use('/api', mounted)` shape. Either drop the warning or reword it around whatever residual reason remains to prefer a root mount.

**3. The cors paragraph describes the pre-deny behavior** — agent/README.md:41-44: a permissive host `cors()` *"answers the preflight with its own policy, and the BFF's exact-origin allow-list never gets a say."* Since 55875e0 the allow-list does get a say: the request that follows the hijacked preflight is refused with `403 origin_not_allowed` (e2e-asserted). Reword to say the host preflight is lost but the followed request is still refused — and keep it in sync with whatever #1876 decides for same-origin `Origin` (see my review there: same-origin POSTs currently 403 unless the own origin is listed).

## Verified good

8e84f1f (supertest import name) and 0c6f3df (IP-whitelist exemption paragraph) check out against the code; the mountless quick-starts now chain `mountOnStandaloneServer` before `start()` as reviewed.

packages/agent-bff/README.md
```diff
@@ -206,11 +209,105 @@ with `session_expired`. Horizontal scaling requires both a shared session store
 
 ```jsonc
 // 200 — all required config present
-{ "status": "ok", "version": "<package version>" }
+{
+  "status": "ok",
+  "version": "<package version>",
+  "features": { "oauth": true, "ai": true, "cors": true, "openapi": true }
```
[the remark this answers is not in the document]
nbouliol: Fixed — the sample, the table intro and the table header all say `configured` now, matching `createHealthRoute` (`health-route.ts:22,40`). Carried the meaning over too, since the rename was not cosmetic: the block says which surfaces were set up, deliberately not that they work. The closing sentence said "what is served, not how it was configured", which was backwards — it now says the body names the surfaces, never the values behind them. The agent README's one-liner points at `configured` as well.

nbouliol:
All three fixed on 8e07e4f.

**1. `/health` `features` → `configured`** — sample, table intro and header, plus the agent README one-liner. Carried the meaning over rather than just the key: the block says which surfaces were *set up*, deliberately not that they work. The closing sentence ("what is served, not how it was configured") was backwards and now says the body names the surfaces, never the values behind them.

**2. Sub-router** — warning dropped. `app.use("/api", mounted)` is now documented as supported, with the emitted prefix derived per request from `originalUrl` (`base-path.ts`), so `servers` and the docs page carry `/api/bff`. Confirmed against the e2e assertions at `embedded-bff.e2e.test.ts:492-495`.

**3. CORS** — reworded to what the code does: the host preflight is lost and unrecoverable, but the request that follows still reaches the app and is refused with `403 origin_not_allowed`. Also fixed the Layer 1 bullet two sections up, which still said "anything else gets no CORS headers (the browser blocks)" — same stale claim, and it was not in this PR's diff so nothing else was going to catch it. Both mention that a request with no `Origin` passes untouched, which is what keeps server-to-server api-key calls working.

On the same-origin question you raised on #1876: I have not documented anything about it here, since it is that PR's call. If it changes, the two paragraphs above are the ones to revisit.

agent: `test/bff` 69/69, eslint `src test` 0 errors.

packages/agent/README.md
Tonours: Verified: both quick-starts chain `.mountOnStandaloneServer(3351)` before `.start()`, and the text now says a `mountOn*` call is what opens a socket — `start()` only builds the router, so a chain without one serves nothing, BFF included. Resolving.

packages/agent-bff/README.md
Tonours: Verified: the README separates the two — the *served* document carries the prefix in `servers`, the *exported* one (`forest-bff openapi`) does not, with `--base-path` ownership left to #1873 where it belongs. Resolving.

packages/agent/README.md
Tonours: Verified: the `agentTimeoutMs` paragraph states the timeout bounds the wait and cancels nothing — an action cut at the deadline still commits, and a retry can double the mutation. The step-17 ask, documented. Resolving.

packages/_example/package.json
Tonours: Verified: `"*"`, like the siblings. Resolving.

packages/agent/test/bff/embedded-bff.e2e.test.ts
Tonours: Verified: the permissive-host-CORS pair exists in `embedded-bff.e2e.test.ts` — a hijacked preflight the host answers permissively, then the followed request asserted refused with `403 origin_not_allowed` for a forbidden origin and served for a listed one. Resolving.

packages/agent/README.md
Tonours: Verified: "on whatever port the agent is mounted on" with the four mount targets listed, and both READMEs state the `/bff` prefix is fixed and resolved against the url the host hands the agent. Resolving.

packages/agent-bff/README.md
Tonours: Verified: "every response the BFF itself produces", with the embedded `503 bff_not_started` / `bff_stopped` named as the exception — matching `EmbeddedBff.handle` writing that response itself, with no BFF to read a version from. Resolving.

Tonours (approved):
Verified at 8e07e4f: all three paragraphs now match the code. `/health` documents `configured` with the set-up-not-working semantics and the closing sentence no longer says the opposite; the sub-router warning became a supported-shape statement matching `resolveEmittedBase` (e2e-asserted `servers: /api/bff`); the cors paragraph states the 403 deny and keeps the no-Origin case that server-to-server calls rely on. The env-table degraded claim was resynced in 55bc6fa. Approving.

One non-blocking note for the next rebase: once aaaaf589's same-origin host exemption lands under this branch, Layer 1's "any other `Origin` is refused outright" wants the exemption named — same-origin callers are served without being listed.

Tonours (approved):
Verified at 8e07e4f: all three paragraphs now match the code — `/health` documents `configured` with the set-up-not-working semantics, the sub-router warning became a supported-shape statement matching `resolveEmittedBase` (e2e-asserted `servers: /api/bff`), and the cors paragraph states the 403 deny while keeping the no-Origin case server-to-server calls rely on. Env-table degraded claim resynced in 55bc6fa. Approving.

One non-blocking note for the next rebase: once aaaaf589's same-origin host exemption lands under this branch, Layer 1's "any other `Origin` is refused outright" wants the exemption named — same-origin callers are served without being listed.

packages/agent/src/agent.ts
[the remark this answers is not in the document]
nbouliol: Correct hazard, and the diagnosis is precise — verified against the registry rather than reasoned about:

- published `1.28.1` (which `^1.28.0` could resolve): `buildBff({ config, logger, basePath })` — no `dispatcher` parameter, and neither `IN_PROCESS_AGENT_URL` nor `AgentDispatcher` is exported.
- published `1.29.0`: `buildBff({ config, logger, basePath, dispatcher, metrics })`, with both symbols exported.

So an install that resolved 1.28.0 or 1.28.1 would indeed leave the embedded BFF with no agent transport.

It never reached npm, though. #1876 changes `agent-bff/src`, so its merge released `agent-bff@1.29.0` and `agent@1.100.0` in the same run, and `multi-semantic-release --deps.bump=override` rewrote the range on the way out. `npm view @forestadmin/agent@1.100.0 peerDependencies` shows `"@forestadmin/agent-bff": "1.29.0"` — exact, not a range.

The caret only ever existed to keep this long-lived branch linking the workspace while `agent-bff` released four times underneath it; it was always going to be overwritten at release. What you caught is that it also stated a compatibility that was false while it sat in the source.

Now moot on this PR: rebased onto `main`, which carries the bot's exact `1.29.0` in both `devDependencies` and `peerDependencies`. No caret remains anywhere under `packages/`, `yarn workspaces info` reports no mismatch repo-wide, and `agent` still links the workspace copy. agent 88 suites / 1624 tests, agent-bff 91 / 1754, lint clean.

PR #1879, opened 2026-09-04
feat: add option to skip rbac

arnaud-moncel:
### General

### Security

> [!NOTE]
> ### Add `skipRelationReadPermissions` option to bypass relation-path read checks in `AuthorizationService`
> - Adds the optional `skipRelationReadPermissions` boolean to `AgentOptions` (default `false`). When `true`, the agent skips read-permission checks on fields reached through relation paths — projections, filters, sorts, searches, and chart groupings — while authorization on the queried collection still runs.
> - `AuthorizationService` gains the constructor parameter and early-returns in `redactProjection`, `assertCanReadQueryFields`, and `assertCanReadUsages` when the option is enabled. The chart route in [chart.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1879/files#diff-09b20a04c73a95c8683f474f2818d52a6875552b1f2ce6c66690677ecdd5373b) also skips browse-permission checks on the counted collection for Count leaderboards.
> - The capabilities endpoint advertises `checksRelationReadPermissions` (the inverse of the option) so internal consumers know whether relation-path checks are active.
> - `OptionsValidator.withDefaults` logs a warning at startup when the option is enabled.
> - Behavioral Change: enabling `skipRelationReadPermissions` means clients can read, filter, sort, and search through relations they lack browse/read permission on. Root collection authorization (list, get-one, export) is unaffected and still enforced.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized e3684fd.</sup>
> 
>

### DIFF
--- a/packages/agent/src/routes/access/chart.ts
+++ b/packages/agent/src/routes/access/chart.ts
@@ -258,8 +258,8 @@ export default class ChartRoute extends CollectionRoute {
       ]);
 
       // A count exposes the cardinality of the relation, which `/relationships/<name>/count` puts
-      // behind `browse`.
-      if (!aggregation.field) {
+      // behind `browse`. Added with the related-read checks, so it goes away with them.
+      if (!aggregation.field && !this.options.skipRelationReadPermissions) {
         await this.services.authorization.assertCanBrowse(context, field.foreignCollection);
       }
 
--- a/packages/agent/src/routes/capabilities.ts
+++ b/packages/agent/src/routes/capabilities.ts
@@ -46,6 +46,7 @@ export default class Capabilities extends BaseRoute {
         // `x-forest-correlation-id` is emitted on every response regardless — the frontend must
         // gate the History tab on this flag rather than inferring the feature from that header.
         canUseAuditTrail: this.options.auditTrail !== null,
+        checksRelationReadPermissions: !this.options.skipRelationReadPermissions,
       },
       collections:
         collections?.map(collection => {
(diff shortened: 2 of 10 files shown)


matthv:
`packages/agent/src/routes/access/chart.ts:262-263` — the leaderboard chart's count-only path (`assertCanBrowse` on `field.foreignCollection`, no `aggregation.field`) isn't gated by `skipRelationReadPermissions`, unlike `assertCanReadAggregatedFields` right above it. So a Count leaderboard over a relation still 403s with the flag on.

The option's JSDoc ("The route's own browse/read/export check on the collection being queried is unaffected") doesn't quite cover this either — the surviving check here is on the *related* collection, not the one being queried.

Intentional (mirroring the pre-#1840 behavior for this one shape) or worth gating too?

matthv:
A couple more things from a review pass:

- Nothing in `related-read-permissions.test.ts` proves the primary collection's `browse`/`read`/`export` check still holds with `skipRelationReadPermissions: true` — that boundary is the whole security claim of the option. `buildServices`'s `canOnCollection` mock always allows `cards`, so a denied-primary-collection case is never exercised. Worth a test asserting `List`/`CsvRelated` still reject when the *root* collection is denied, flag on.
- The new default (`skipRelationReadPermissions: false`) isn't asserted in `http-driver-options.test.ts`'s `'should add default values when they are missing'`, unlike `maxRecordsForApproval`/`auditTrail` when they landed.

Minor: no `Warn` log on boot when the option is on (existing pattern for `permissionsCacheDurationInSeconds`/`skipSchemaUpdate`) — would help an auditor spot the weakened posture from the boot log.

matthv (approved):
LGTM

PR #1880, opened 2026-09-04
feat(workflow-executor): evaluate the operators a list view filter offers, relative dates included

Scra3:
fixes PRD-1147, fixes PRD-1130

## Why

A deterministic Decision on a Date field could only say `is`, `is not`, `is present`, `is blank`. Forest exposes `before`/`after` and a relative family for dates, none of which were in the contract, so the first condition anyone writes on a date could not be written. Meanwhile the contract carried two operators (`>=`, `<=`) that the editor can never build, since the front registry does not define them.

## What

**Operator set = what a list view filter offers.** `CONDITION_OPERATORS` goes from 12 to 32: the two orphans leave, twenty-two arrive.

| Family | Operators |
|---|---|
| Absolute dates | `before`, `after` |
| Relative dates | `past`, `future`, `today`, `yesterday`, `previous_x_days`, `previous_x_days_to_date`, `before_x_hours_ago`, `after_x_hours_ago` |
| Calendar periods | `previous_week`, `previous_month`, `previous_quarter`, `previous_year`, and each one's `_to_date` variant |
| Strings | `starts_with`, `ends_with`, `i_contains` |
| Arrays | `includes_all` |

`not_in` stays in the contract and gains an implementation: `memberOf` is tri-state, so `not_in` is met only when the candidate is definitively absent from the list, never on a value it cannot compare.

**Relative dates read an injected `Clock`**, `{ now, timezone }`, never the machine's. The evaluator stays pure. The step executor builds one clock per step so every row of a Decision sees the same instant.

**Timezone is the project's, carried on the run.** `AvailableStepExecution.timezone` is new; the mapper falls back to `UTC` when the project has none or the orchestrator predates the field. The machine's zone is never used: the fleet runs several executor instances and the same run must route the same on each.

**Semantics mirror `datasource-toolkit`'s time transforms**, which is what the list filter runs on: day windows exclude their end and, like the toolkit's `GreaterThan` / `GreaterThanOrEqual` split, include their start for a calendar date only, `previous_x_days` excludes today, `previous_x_days_to_date` includes it up to the instant. A calendar date (Dateonly) is read at midnight in the project zone, otherwise a Honolulu record's own today is counted as yesterday. `i_contains` folds case and keeps accents, measured on Postgres: `'É' ILIKE '%é%'` holds, `'é' ILIKE '%e%'` does not. A time of day is read as milliseconds since midnight rather than text, so `08:30` and `08:30:00` are one instant and `08:30:00.10` does not order below `08:30:00.1`; the upper bound is `24:00:00`, the top of Postgres's `time` range, which it returns for the end of a day — past that instant is not a time.

**The trace records `evaluatedAt` and `timezone`.** A relative condition makes routing depend on the clock; without the instant, a check on "previous 7 days" cannot be explained a day later.

## Decisions taken with product (on PRD-1147)

- Dates are evaluated at the instant the condition is evaluated.
- Timezone: project's, fallback UTC. Not the user's (automated runs have none), not the host's.
- A retry may route differently. Accepted.
- All operators ship together.

## Compatibility

Dropping `greater_than_or_equal` and `less_than_or_equal` makes a published workflow that still carries one of them fail at parse on this executor (and be refused by the orchestrator's parser). The editor never offered them, and the deterministic Decision has only been open on executor 1.27.0 since 2 September, so no such workflow should exist; a definition that did not come from the builder is the only way to have one.

The reverse direction is what the front gates. This release is 1.28.0 — the executor is published at 1.27.2 with nothing unreleased, and this is a `feat` — and `MIN_EXECUTOR_VERSIONS.deterministicConditionListFilterOperators` on the front is set to it, so the new operators are offered only where they can be evaluated. That floor is separate from the Decision's own, which stays on 1.27.0: measured on the fleet, 13 environments run 1.27.x and none run anything newer, so those 13 are the whole population that has the step, Qonto App Prod and Qonto CBS Production among them. An operator such a runtime does not know is refused while the run is mapped, which reports the whole run malformed and unmappable on every later poll — not a step error.

## Paired with

- forestadmin-server: the contract list, the value rules per operator, and sending `project.timezone` on the run.
- forestadmin: the editor's allowed list and the run view.

Until the orchestrator sends the timezone, relative dates resolve in UTC.

## Tests

1772 passing, 7 skipped. Every new operator is pinned with a fixed clock in `Europe/Paris`, the calendar-date case in `Pacific/Honolulu`, and the condition executor proves the context timezone reaches the evaluator with fake timers straddling midnight.

> [!NOTE]
> ### Add list-view filter operators and timezone-aware date evaluation to conditions
> - Removes `greater_than_or_equal`, `less_than_or_equal`, and `not_in` from `CONDITION_OPERATORS` in [step-definition.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1880/files#diff-66dcc667cb1ebecf3964eecfb28391b1cb81e16c853ad79d4f9017172521aea3); adds `includes_all`, `not_contains`, `starts_with`, `ends_with`, `i_contains`, `before`, `after`, `past`, `future`, `today`, `yesterday`, `previous_x_days`, `previous_x_days_to_date`, `before_x_hours_ago`, and `after_x_hours_ago`
> - Introduces an injected `Clock` in [deterministic-condition-evaluator.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1880/files#diff-554011f1797a9f5d8b27048dda98e6f799a04b7c8b3b2a89ca4b978f347ecd12) so relative date and time operators evaluate against a captured instant and project timezone; `ConditionStepExecutor` persists the evaluation timestamp and timezone in execution params
> - Adds a required `timezone` field to `ExecutionContext` and `AvailableStepExecutionSchema`, validated as an IANA zone via Luxon; `toAvailableStepExecution` in [run-to-available-step-mapper.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1880/files#diff-d8e6d94d08cd810252fadde0e0f846eca9f05c1b39b880ed4bb1d2357ca64abf) forwards the run timezone and falls back to `UTC` for missing or invalid values
> - Adds `luxon` as a runtime dependency in [package.json](https://github.com/ForestAdmin/agent-nodejs/pull/1880/files#diff-39959c5d53211b31724a713f0f4021345ce98a0dd752aadcde43c1729bdff151) for IANA zone validation and timezone-aware date math
> - Behavioral Change: `ConditionStepDefinitionSchema` rejects the three removed operators at the schema boundary; `AvailableStepExecutionSchema` rejects available-step values that omit `timezone` or contain an invalid IANA zone; `ExecutionContext.timezone` is now required, so all callers and test fixtures must supply it; `greater_than` and `less_than` now use strict comparison instead of inclusive
>
> 
> #### Changes since #1880 opened
>
> - Added support for 8 new calendar-period operators to the `workflow-executor` package's deterministic condition evaluation system [0148a2a]
> - Updated documentation to reflect the expanded operator set and new calendar-period functionality [0148a2a]
> - Refactored `memberOf` utility function in `deterministic-condition-evaluator` to return tri-state results (true/false/null) instead of boolean, added `not_in` operator to the `EVALUATORS` map that evaluates to true only when candidate is definitively not a member of the expected list, and updated `in` operator to check for definite membership using tri-state result [fe1ff08]
> - Extended the `CONDITION_OPERATORS` constant in `step-definition` types to include `not_in` operator, increasing the total count from 31 to 32 operators [fe1ff08]
> - Added test coverage for `not_in` operator behavior and updated operator validation tests to accept `not_in` as a valid operator [fe1ff08]
> - Updated documentation in `CLAUDE.md` to reflect the addition of `not_in` operator and clarify tri-state membership semantics [fe1ff08]
> - Implemented time-of-day parsing infrastructure converting time strings to milliseconds since midnight [ceaaae3]
> - Updated time-of-day comparison operations in condition evaluator to use numeric millisecond values instead of string comparison [ceaaae3]
> - Added test coverage for fractional second handling and invalid time behavior in time-of-day comparisons [ceaaae3]
> - Modified the `toMillisOfDay` utility in `deterministic-condition-evaluator` to accept '24:00:00' and '24:00' as valid end-of-day time values [5c7c638]
> - Added test coverage for '24:00:00' end-of-day handling and updated existing test assertions in `deterministic-condition-evaluator` test suite [5c7c638]
> - Replaced millisecond-based time-of-day parsing with microsecond-based parsing in the `workflow-executor` package [bb75085]
> - Documented architectural decision for case-insensitive decision semantics [80ae746]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c86a5a5.</sup>
> 
>


packages/workflow-executor/src/types/validated/execution.ts
```diff
@@ -49,6 +49,8 @@ export const AvailableStepExecutionSchema = z
     stepDefinition: StepDefinitionSchema,
     previousSteps: z.array(StepSchema),
     user: StepUserSchema,
+    /** IANA zone the run's relative dates are read in: the project's, never the machine's. */
+    timezone: z.string().min(1),
```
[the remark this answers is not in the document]
Scra3: Both, now. The mapper already normalised an unknown zone to UTC (`IANAZone.isValidZone` guard in `run-to-available-step-mapper.ts`), so nothing invalid could reach the evaluator. The schema now refuses it too, in the spirit of the "defense against mapper bugs" note below it: `timezone: z.string().refine(IANAZone.isValidZone)`. The mapper is the only producer, so the refinement only ever fires on a mapper bug.

packages/workflow-executor/src/executors/deterministic-condition-evaluator.ts
```diff
@@ -108,6 +156,17 @@ function compare(actual: unknown, expected: unknown): number | null {
   const expectedTs = toTimestamp(expected);
   if (actualTs !== null && expectedTs !== null) return actualTs - expectedTs;
 
+  // Without this, "is greater than" on a Time column can never be met: the builder offers the
+  // operator, and the step would route to the fallback on every record without saying why.
+  const actualTime = toTimeOfDay(actual);
```
hercemer42: **Applies to:** `deterministic-condition-evaluator.ts:111-125` (`scalarEqual`, not in this diff). This line is where the time-of-day rule is applied, and it is the only place.
Claude Fable 5.1 (claude-fable-5-1): **Should fix** — A Decision "is 08:30" on a Time column never matches a record stored as `08:30:00`, and "is not 08:30" matches it, while "is greater than" and "is less than" both treat the pair as one instant.
`compare` pads and orders a time of day at `:161-168`, but `scalarEqual` (`:111-125`), which backs `equal`, `not_equal`, `in` and `includes_all`, still ends on `typeof actual === typeof expected ? false : null`, so `'08:30:00'` and `'08:30'` are two different strings; a Postgres or MySQL Time column arrives as `HH:mm:ss` and the builder's `<input type="time">` writes `HH:mm`.
Parse both sides to milliseconds since midnight in one helper used by `scalarEqual` and `compare`, which also settles the fractional-precision ordering already raised on this PR.

1. Time column `openingTime` stored `08:30:00` in Postgres, read by a Get Data step.
2. Decision: option A `openingTime` is `08:30` (the time widget writes `08:30`), option B `openingTime` is not `08:30`, fallback C.
3. Run on that record: A is `met: false`, B is `met: true`; the list view filter "is 08:30" matches the same row.
</details>

hercemer42 (commented):
**Spec (PRD-1147, PRD-1130)**: conforms — the 23-operator set, the injected clock, the project-zone/UTC rule and the `evaluatedAt`/`timezone` trace are in place and pinned; "today included up to the instant" for `previous_x_days_to_date` holds for datetime columns, and a Dateonly column excludes today exactly as the list filter does (deliberate, documented in the package CLAUDE.md).

Claude Fable 5.1 (claude-fable-5-1): **Should fix** — The next engineer touching either half of the relative-date rule (the front anchoring the list filter on the browser's zone, the orchestrator sending `project.timezone`) finds no decision record outside this package's CLAUDE.md and PRD-1147.
The diff fixes two cross-repo decisions that are hard to reverse once workflows are published on them: relative dates evaluate in the project's zone with a UTC fallback, never the viewer's or the host's (`run-to-available-step-mapper.ts:182-190`), and `i_contains` follows Postgres `ILIKE` across dialects (`deterministic-condition-evaluator.ts:279-283`); no ADR in `docs/adr/**` covers either (the org-wide search on `timezone` returns nothing).
Record them with `/adr`, one per decision.

packages/workflow-executor/src/executors/deterministic-condition-evaluator.ts
Scra3: Right on both counts, and fixed in ceaaae35f — with the parse you suggest rather than the padding I had.

`toMillisOfDay` parses a time of day to milliseconds since midnight, and both `scalarEqual` and `compare` use it, so `equal`, `not_equal`, `in`, `includes_all`, `greater_than` and `less_than` all read `08:30` and `08:30:00` as one instant. That also settles the fractional-precision ordering raised by the bot on the same lines: `08:30:00.1` and `08:30:00.10` now compare equal instead of ordering by text length. A time nobody can be at (`25:00`, `08:70`) parses to nothing rather than to a number.

Six assertions cover it: the two formats through equality and membership, the fraction, the out-of-range hours and minutes, and a non-time value.

Worth noting the starting point, since your reproduction understates it: before this PR `greater_than` and `less_than` on a Time column were never met at all, in any format — `compare` had no branch for a time of day, so the step took the fallback on every record with nothing in the trace.

packages/workflow-executor/src/executors/deterministic-condition-evaluator.ts
```diff
@@ -108,6 +156,17 @@ function compare(actual: unknown, expected: unknown): number | null {
...
+  const expectedTime = toTimeOfDay(expected);
```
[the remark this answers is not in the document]
Scra3: Fixed in ceaaae35f, by parsing rather than by normalising the text. `toMillisOfDay` reads a time of day into milliseconds since midnight, so `08:30:00.1` and `08:30:00.10` are the same number and neither `less_than` nor `greater_than` matches; the same parse backs `scalarEqual`, which the human review on these lines pointed out was the other half of the problem. A regex over the text would have left equality and ordering on two different rules.

packages/workflow-executor/src/executors/deterministic-condition-evaluator.ts
hercemer42: Closed: equality and ordering now read a time of day through the same parse.
Predicate: two strings for one time of day, differing in seconds or fraction, on which equal and compare disagree.
Ran evaluateOperator at a1a8fc6 on five same-instant pairs through equal, not_equal, in, not_in, includes_all, greater_than and less_than, and all agree.
Pinned by the times-of-day tests at lines 184 to 199 of the evaluator test, which fail on the pre-fix evaluator.

hercemer42:
Reopened: no ADR records the project-zone rule or the Postgres ILIKE contract yet.
Predicate: docs/adr on the branch holds no decision covering either, and the finding has no reply.
Checked by listing docs/adr at a1a8fc6 and reading the issue comments since the review.
Nothing to pin, the fix is an ADR, so this stays open until one lands or the finding is contested.

packages/workflow-executor/src/executors/deterministic-condition-evaluator.ts
```diff
@@ -68,6 +110,23 @@ function compareIntegers(actual: unknown, expected: unknown): number | null {
...
+// Milliseconds since midnight rather than text: comparing the strings would order "08:30:00.10"
+// before "08:30:00.1", which is the same instant written twice.
+function toMillisOfDay(value: unknown): number | null {
+  if (typeof value !== 'string') return null;
+
+  const parts = TIME_OF_DAY.exec(value);
+  if (!parts) return null;
+
+  const [, hours, minutes, seconds = '0', fraction = '.0'] = parts;
+  if (Number(hours) > 23 || Number(minutes) > 59 || Number(seconds) > 59) return null;
```
hercemer42: Claude Fable 5.1 (claude-fable-5-1): **Should fix** — A Time column holding `24:00:00`, which Postgres accepts and returns for the end of a day, now satisfies no operator: "is greater than 08:00" is not met and "is not 24:00" is met.
`toMillisOfDay` returns `null` for `hours > 23` at `:122`, so `scalarEqual` and `compare` read the value as not a time; at `b9d1855` the text ordering still placed it above `23:59:59`.
Accept exactly `24:00:00`, with zero minutes, seconds and fraction, as 86 400 000 ms and keep the refusal for anything else past 23.

1. Postgres `time` column `closesAt` holding `24:00:00` (`select '24:00'::time` stores and returns it as such).
2. Decision: option A `closesAt` is greater than `08:00`, option B `closesAt` is `24:00`, fallback C.
3. Run on that record: A and B are `met: false` and the run takes C; the list view filter "is greater than 08:00" matches the same row.
</details>

Scra3: Confirmed and fixed in 5c7c63827.

Checked the premise before the code: 24:00:00 is the top of Postgres's `time` range, which it accepts and returns for the end of a day, and the list view filter orders it above every other time because the comparison happens in SQL. So the divergence you describe is real, and it is the one this PR set out not to have.

Worth noting the existing test hid it: `ev('equal', '24:00:00', '24:00:00')` passed only because `scalarEqual` short-circuits on identical strings before reaching `toMillisOfDay`. Ordering never got that shortcut.

Fixed with a bound on the instant rather than a special case on the hour: the minute and second ranges are still checked, the milliseconds are computed, and anything past 86 400 000 is refused. 24:00:00 and 24:00 both read as the end of the day, 24:00:00.001 does not, and 25:00:00, 08:70:00 and 08:30:70 stay refused exactly as before. One bound instead of a branch, and no separate rule for equality and ordering.

Four assertions pin it, and the old `24:00:00` line moved out of "a time nobody can be at" — it is a time something can be at. Verified the new test fails without the change and passes with it; 1772 pass, 7 skipped on the package.

Scra3:
One half fixed, the other contested.

**The Postgres ILIKE contract — you were right, and my earlier reasoning was wrong.** Recorded in 80ae746bd as `docs/adr/2026-09-09-a-case-insensitive-decision-follows-postgres-s-ilike-whatever-the-data.md`, the first ADR in this repo, so the helper created `docs/adr/` and the index.

I had ruled it out of the gate on the grounds that mirroring the list view filter left no alternative to choose. That was the error: there is no single list-filter behaviour to mirror. `'É' ILIKE '%é%'` holds on Postgres and MySQL's and SQLite's default collations differ, which the code comment says outright — so picking Postgres *is* a choice, with a named cost (a Decision and the customer's own filter disagreeing on the same record on those datasources) and two real alternatives I rejected: following each datasource's collation, which makes routing depend on which database a run happens to read, and folding accents too, which is one rule everywhere but unlike every filter the customer sees.

**The project-zone rule is already recorded** — in forestadmin-server, not here: `docs/adr/2026-09-08-relative-dates-in-a-deterministic-decision-are-read-in-the-project-s-t.md` on that PR's branch (#8481). It states the project's zone, UTC when the project has none, never the caller's or the executor host's, and the accepted divergence from the list filter's browser-anchored `today`.

Deliberately one record rather than one per repo. It is a single cross-repo decision — the orchestrator resolves the zone and puts it on the run, the executor reads it — and it lives in the repo that owns the run contract. Two copies would be two things to keep in sync, and the convention is one decision, one record. Your predicate only listed `docs/adr` in this repo, which is why it came back empty; the search that finds it has to cross repos.

Happy to add a pointer here if you would rather each repo carry a stub, but I would rather not duplicate the decision itself.

packages/workflow-executor/src/executors/deterministic-condition-evaluator.ts
hercemer42: Closed: 24:00:00 is read as the end of the day and everything past it is refused.
Predicate: a time Postgres stores that no operator meets, or a time nobody can be at that one does.
Ran evaluateOperator at 80ae746 on 24:00:00, 24:00, 24:00:00.001, 25:00:00, 08:70:00 and 08:30:70 through every scalar operator, and the micros change keeps 08:30:00.0004 distinct from 08:30:00.
Pinned by the tests at lines 204 to 225 of the evaluator test, and the 24:00:00 ordering fails on a1a8fc6.

hercemer42:
Closed: each decision now has an accepted ADR, one record per decision as the finding asked.
Predicate: a decision this diff fixes that no docs/adr file in any Forest repo records.
Read the ILIKE record at 80ae746 here and the project zone record at f39dca9 on forestadmin-server PR 8481, which lands with that PR.
Nothing to pin for a record, and no stub needed here since the zone ADR names workflow-executor in affected_components.

hercemer42 (approved):

PR #1883, opened 2026-09-07
fix(security): replace excel4node with write-excel-file to remove vulnerability

EnkiP:
### General

### Security

> [!NOTE]
> ### Replace `excel4node` with `write-excel-file` in xlsx export plugin
> - Replaces the `excel4node` dependency with `write-excel-file` to remove a security vulnerability in `plugin-export-advanced`
> - Rewrites the xlsx renderer to map records to `write-excel-file` cell descriptors, encoding booleans, numbers, and dates as typed cells
> - `addExportAdvanced` now awaits the renderer output before returning the file result
> - Risk: `toCell` in [xlsx.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1883/files#diff-0d084ebae2a26293a5c1582477d1e675670966b18aafc786041bd35e678809ac) now infers types, converting numeric strings to numbers and lowercase boolean strings to booleans, changing how some data is represented in exported spreadsheets
>
> 
> #### Changes since #1883 opened
>
> - Added validation in the `toCell` utility within `plugin-export-advanced` to convert non-finite numeric values (NaN, Infinity, -Infinity) to string cells instead of number cells, and to return null for invalid Date instances [dba86e3]
> - Changed the date format for XLSX date cells from 'yyyy-mm-dd hh:mm:ss' to 'm/d/yy' using a new `XLSX_DATE_FORMAT` constant [dba86e3]
> - Configured the `writeXlsxFile` function call in `plugin-export-advanced` to specify the worksheet name as 'Export' [dba86e3]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 5a6cf6e.</sup>
> 
>

### DIFF
--- a/packages/plugin-export-advanced/src/renderers/xlsx.ts
+++ b/packages/plugin-export-advanced/src/renderers/xlsx.ts
@@ -4,7 +4,7 @@ import writeXlsxFile from 'write-excel-file/node';
 
 import getFieldValue from '../utils/get-field-value';
 
-function toCell(value: unknown): Cell {
+export function toCell(value: unknown): Cell {
   if (value === null || value === undefined) return null;
   if (typeof value === 'boolean') return { type: Boolean, value };
   if (typeof value === 'number') return { type: Number, value };
--- a/packages/plugin-export-advanced/src/renderers/xlsx.ts
+++ b/packages/plugin-export-advanced/src/renderers/xlsx.ts
@@ -1,61 +1,40 @@
-import type { Readable } from 'stream';
+import type { Cell } from 'write-excel-file/node';
 
-import xl from 'excel4node';
+import writeXlsxFile from 'write-excel-file/node';
 
 import getFieldValue from '../utils/get-field-value';
 
-function getExcel4NodeTypeFromValue(value: unknown): string {
+function toCell(value: unknown): Cell {
   if (value === null || value === undefined) return null;
-  if (typeof value === 'boolean') return 'bool';
-  if (typeof value === 'number') return 'number';
+  if (typeof value === 'boolean') return { type: Boolean, value };
+  if (typeof value === 'number') return { type: Number, value };
+  if (value instanceof Date) return { type: Date, value, format: 'yyyy-mm-dd hh:mm:ss' };
 
   if (typeof value === 'string') {
-    if (value === 'true' || value === 'false') return 'bool';
-    if (!Number.isNaN(Number(value))) return 'number';
-    if (!Number.isNaN(Date.parse(value))) return 'date';
+    if (value === 'true' || value === 'false') return { type: Boolean, value: value === 'true' };
 
-    return 'string';
-  }
-
-  if (value instanceof Date) return 'date';
-
-  return 'string';
-}
-
-function castValue(value) {
-  switch (getExcel4NodeTypeFromValue(value)) {
-    case 'bool':
-      return Boolean(value);
-    case 'number':
-      return Number(value);
-    case 'string':
-      return String(value);
-    case 'date':
-      return new Date(value);
-    default:
-      return String(value);
-  }
-}
-
-export default function render(records: Record<string, unknown>[], projection: string[]): Readable {
-  const wb = new xl.Workbook();
-  const ws = wb.addWorksheet('Export');
-
-  for (const [index, name] of projection.entries()) {
-    ws.cell(1, index + 1).string(name);
-  }
-
-  for (const [row, record] of records.entries()) {
-    for (const [col, name] of projection.entries()) {
-      const value = getFieldValue(record, name);
-      const castedValue = castValue(value);
-      const excel4NodeCellFunction = getExcel4NodeTypeFromValue(castedValue);
-
-      if (value !== null && value !== undefined) {
-        ws.cell(2 + row, 1 + col)[excel4NodeCellFunction](castedValue);
-      }
+    if (value.trim() !== '' && !Number.isNaN(Number(value))) {
+      return { type: Number, value: Number(value) };
     }
+
+    const timestamp = Date.parse(value);
+
+    if (!Number.isNaN(timestamp)) {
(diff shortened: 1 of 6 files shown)


packages/plugin-export-advanced/src/renderers/xlsx.ts
```diff
@@ -1,61 +1,40 @@
...
-  }
+export default function render(
+  records: Record<string, unknown>[],
+  projection: string[],
+): Promise<Buffer> {
+  const header: Cell[] = projection.map(name => ({ type: String, value: name }));
+  const rows = records.map(record => projection.map(name => toCell(getFieldValue(record, name))));
 
-  return wb.writeToBuffer();
+  return writeXlsxFile([header, ...rows]).toBuffer();
```
Tonours: **Must fix**

The old code named the sheet `Export` (`wb.addWorksheet('Export')`). This line passes no `sheetOptions`, so `write-excel-file` falls back to `"Sheet".concat(sheetIndex + 1)` (`generateXlsxFileContents.js:261`, checked against the published 4.1.1 tarball) and the workbook ships a tab named `Sheet1`.

Anything pointing at `Export!A1` now gets `#REF!`: a formula, a Power Query connection, a customer script. Nothing on our side logs it.

```ts
return writeXlsxFile([header, ...rows], { sheet: 'Export' }).toBuffer();
```

packages/plugin-export-advanced/src/renderers/xlsx.ts
```diff
@@ -1,61 +1,40 @@
...
 
 import getFieldValue from '../utils/get-field-value';
 
-function getExcel4NodeTypeFromValue(value: unknown): string {
+export function toCell(value: unknown): Cell {
   if (value === null || value === undefined) return null;
-  if (typeof value === 'boolean') return 'bool';
-  if (typeof value === 'number') return 'number';
+  if (typeof value === 'boolean') return { type: Boolean, value };
+  if (typeof value === 'number') return { type: Number, value };
```
Tonours: **Must fix**

`NaN` goes straight into a numeric cell here, and an `Invalid Date` does the same on line 11 (`convertDateToSerialNumber` yields `NaN`). `write-excel-file` writes a bare `String(value)` for `type: Number`, so the file carries `<v>NaN</v>` and Excel offers to repair it.

This is new. `excel4node` threw: `numberSetter` does `if (parseFloat(val) !== val) throw` (`lib/cell/index.js:65`), and `NaN !== NaN`. The error used to surface. Now the export succeeds and hands back a broken workbook.

```ts
if (typeof value === 'number') {
  return Number.isFinite(value) ? { type: Number, value } : { type: String, value: String(value) };
}
if (value instanceof Date) {
  return Number.isFinite(value.getTime()) ? { type: Date, value, format: XLSX_DATE_FORMAT } : null;
}
```

packages/plugin-export-advanced/src/renderers/xlsx.ts
```diff
@@ -1,61 +1,40 @@
...
+  if (value instanceof Date) return { type: Date, value, format: 'yyyy-mm-dd hh:mm:ss' };
```
Tonours: **Should fix**

`excel4node` defaulted to `m/d/yy` (`lib/workbook/workbook.js:47`). This sets `yyyy-mm-dd hh:mm:ss` on every date column: component order flips, and a time shows up on date-only fields. Every admin who exports sees it.

Fine as a deliberate change, but then put it in the PR description. Otherwise keep `m/d/yy`.

packages/plugin-export-advanced/src/renderers/xlsx.ts
```diff
@@ -1,61 +1,40 @@
...
-    if (value === 'true' || value === 'false') return 'bool';
-    if (!Number.isNaN(Number(value))) return 'number';
-    if (!Number.isNaN(Date.parse(value))) return 'date';
+    if (value === 'true' || value === 'false') return { type: Boolean, value: value === 'true' };
 
-    return 'string';
-  }
-
-  if (value instanceof Date) return 'date';
+    if (value.trim() !== '' && !Number.isNaN(Number(value))) {
```
Tonours: **Should fix**

`Number.isNaN(Infinity)` is `false`, so `'Infinity'` and `'1e309'` pass this test and land as `<v>Infinity</v>`. XSD spells infinity `INF`, so that is not a valid double and Excel offers to repair the file.

Pre-existing, `parseFloat(Infinity) === Infinity` got through `excel4node` too. But the PR rewrites this line and the fix is one identifier:

```ts
if (value.trim() !== '' && Number.isFinite(Number(value))) {
```

Keep the `trim()` guard, `Number('')` is `0`.

packages/plugin-export-advanced/src/renderers/xlsx.ts
```diff
@@ -1,61 +1,40 @@
...
-    default:
-      return String(value);
-  }
-}
-
-export default function render(records: Record<string, unknown>[], projection: string[]): Readable {
-  const wb = new xl.Workbook();
-  const ws = wb.addWorksheet('Export');
+    if (!Number.isNaN(timestamp)) {
+      return { type: Date, value: new Date(timestamp), format: 'yyyy-mm-dd hh:mm:ss' };
```
Tonours: **Violates conventions**

`'yyyy-mm-dd hh:mm:ss'` sits here and on line 11. `CLAUDE.md:85` forbids magic strings.

Concretely: whoever changes the export date format next has to find both sites. Miss one and the sheet ships two different formats, real `Date` columns in one and `Date.parse`-coerced strings in the other.

```ts
const XLSX_DATE_FORMAT = 'yyyy-mm-dd hh:mm:ss';
```

File-local is enough, this package has no `constants.ts`.

Tonours (commented):
excel4node gone, security goal met. 5 findings in xlsx.ts, two are changes the description misses.

Tonours (approved):

PR #1885, opened 2026-09-07
feat(agent-bff): audit reads and action executions

nbouliol:
`agent-bff` wrote no activity log at all — zero occurrences of `activity` / `ActivityLog` / `activityLogsService` in its source. A user who fetched data or triggered an action through the BFF left no audit trail, and the agent it proxies to writes none either (`routes/access/audit-trail.ts` only reads the trail). Only `mcp-server` and `workflow-executor` were writing logs.

Fixes PRD-1150

Depends on ForestAdmin/forestadmin-server#8483, which supplies the credential and the `BFF` source value. Implemented against that contract and mocked in tests, so this branch is reviewable now but must not ship before it.

## Audited surface

Strict parity with mcp-server: three routes.

| Route | `action` |
| -- | -- |
| `list` | `search` if the body carries one, else `filter`, else `index` |
| `relations/:rel/list` | `listRelatedData`, parent id as the record, label naming the relation and its refinements |
| `actions/:name/execute` | `action`, record ids, label naming the action |

`count`, `relations/:rel/count` and `actions/:name/form` write nothing, deliberately. mcp-server has no standalone count tool — count is folded inside the list tool's single log — and does not audit `get-action-form` either. The accepted consequence is that a filtered `count` called on its own stays unaudited, which is an information oracle the MCP surface does not expose; auditing it would double the audit volume, since a table page-load fires `list` and `count` as two separate requests and every pagination click replays both.

## How it is wired

New `src/activity-log/`: the service (its own instance, because `ForestAdminClientOptions` has no `headers` field and the source must travel as one), the creator holding the action→type map and the fail policy, the wrapper, the drainer, and a composition root so the route middlewares take a writer and never see the service or the token plumbing.

`src/auth/forest-server-token-middleware.ts` lands a **lazy** `ctx.state.resolveForestServerToken`, memoised per request. API-key mode returns the token that came with the resolve response; OAuth mode calls `ensureFreshServerAccess`. Lazy because `/health`, permissions, context, the OpenAPI document and the docs route audit nothing and must not pay a session lookup — and permissions is hit on every page load.

**Fail policy, verbatim from mcp-server:** a write whose pending log cannot be created is blocked (503 `audit_unavailable`, new, mirroring `permissions_unavailable`); a read proceeds with a warning; an authorization refusal propagates even for a read (403 `audit_not_authorized`, new). Both statuses were already declared on these routes and `Retry-After` was already wired for 503, so only the OpenAPI descriptions needed extending.

## Points worth a reviewer's attention

**Approval needed an explicit special case.** mcp-server treats `approvalRequested` as a success, so its entry ends `completed`. The BFF throws `actionRequiresApproval`, which verbatim wrapping would record as `failed` — the same business event recorded differently depending on the channel, making action-failure statistics unusable. The log is marked `completed` before the rethrow.

**`activityLogs` is a required option** on both route middlewares rather than reached from `ctx.state`, so a wiring mistake cannot silently disable auditing. That is why the existing construction sites in the test suite now pass an explicit passthrough. The *token* resolver is what stays off their dependency lists.

**Execute wraps the whole sequence**, `loadAction` and `setFields` included, matching `execute-action.ts`. So an unknown action, or an agent down at form load, produces a `failed` entry for an attempt that never touched data. That is the intent — capture the attempt.

**`installShutdownHandlers` replaces the previously installed pair** instead of adding one. `runCli` is called many times in a single test file and would otherwise accumulate signal listeners; one process runs one BFF, so replacement is also the right production semantics.

**An API-key write with no token answers 503, not 401.** There is no session to have expired — the resolve response simply predates the server change.

## Known follow-ups, not addressed here

- The pending log is awaited before the operation, as mcp-server does, which adds a blocking round-trip to 100% of audited reads. Unmeasurable today: the `Metrics` port has `increment` and `gauge` but no duration.
- The session store is in-memory, so behind more than one instance an OAuth read can land on an instance without the session. It then proceeds unaudited rather than 401-ing, which also avoids up to 15 minutes of 401s after each deploy while stateless access tokens outlive the process.
- The action→type policy now exists in three places in the monorepo (mcp-server's map, workflow-executor's inline literals, and this). PRD-644 covers extracting a shared core.
- **Merge order:** PRD-1076 moves this whole middleware chain from `cli-core.ts` into `build-bff.ts` and is already in review. Land it first and rebase this.

## Verification

`tsc --noEmit` clean; `yarn workspace @forestadmin/agent-bff test` → 91 suites, 1698 tests passing; targeted eslint clean.

> [!NOTE]
> ### Add `ActivityLogWriter` to audit BFF reads and action executions
> - Adds an activity-log pipeline that creates a pending log, runs the audited operation, and asynchronously schedules the completed or failed status transition.
> - Wraps list, relation-list, and action-execute handlers in `withActivityLog`; action-form loads remain unaudited.
> - Read-side audit failures let the request proceed unaudited. Write-side (action) audit failures block the operation with a 503 `audit_unavailable` or 403 `audit_not_authorized` response.
> - `BFFHttpServer.stop` now drains in-flight activity logs within a 10-second default deadline, triggered by `Agent.stop` or CLI SIGTERM/SIGINT handlers.
> - Behavioral Change: `DataRoutesMiddlewareOptions` and `ActionRoutesMiddlewareOptions` now require an `ActivityLogWriter`; `ApiKeyAuthenticator` requires an `invalidate` method and may return a `saasAccessToken`.
>
> 
> #### Changes since #1885 opened
>
> - Introduced auditing for unexposed action execution attempts in `agent-bff` [b37ccb3]
> - Implemented credential-aware API key invalidation with retry windowing in `agent-bff` [b37ccb3]
> - Added bounded activity log drain with timeout on `EmbeddedBff` shutdown [b37ccb3]
> - Expanded retry logic for activity log status transitions [a77f003]
> - Added test coverage for activity log status transition retry scenarios [a77f003]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8d74201.</sup>
> 
>

### DIFF
--- a/packages/agent-bff/src/activity-log/activity-logs-creator.ts
+++ b/packages/agent-bff/src/activity-log/activity-logs-creator.ts
@@ -10,6 +10,7 @@ import type { Context } from 'koa';
 
 import { HttpError, NotFoundError } from '@forestadmin/forestadmin-client';
 
+import { invalidateApiKeyIdentity } from '../api-key/api-key-middleware';
 import {
   resolveForestServerToken,
   resolveRenderingId,
@@ -32,8 +33,8 @@ export type BffActivityLogAction = Extract<
  * created is blocked (no unaudited side effect), while a read proceeds with a warning (an audit
  * store outage must not take down the read surface).
  *
- * One case is arbitrated by the cause instead of the action type: an authorization refusal
- * (401/403) propagates for reads too — the read itself is not authorized either.
+ * One case is arbitrated by the cause instead of the action type: an authorization refusal (403)
+ * propagates for reads too — the read itself is not authorized either.
  */
 const ACTION_TO_TYPE: Record<BffActivityLogAction, ActivityLogType> = {
   index: 'read',
@@ -81,8 +82,48 @@ function describeCause(error: unknown): string {
   return error instanceof Error ? `${error.name}: ${error.message}` : String(error);
 }
 
+const FORBIDDEN = 403;
+const UNAUTHORIZED = 401;
+
+/**
+ * A 403 only. A 401 is not the caller being refused: the bearer the BFF audits with is minted by
+ * the Forest server and cached, so a 401 says that token expired — answering the caller with
+ * `audit_not_authorized` would refuse a read the fail-open policy lets through.
+ */
 function isAuthorizationRefusal(error: unknown): boolean {
-  return error instanceof HttpError && (error.status === 401 || error.status === 403);
+  return error instanceof HttpError && error.status === FORBIDDEN;
+}
+
+function isExpiredAuditCredential(error: unknown): boolean {
+  return error instanceof HttpError && error.status === UNAUTHORIZED;
+}
+
+/**
+ * What locates the failure for support, and nothing else: the credential, the record ids and the
+ * label are the payload this must never carry.
+ */
+function auditIdentifiers(
+  ctx: Context,
+  context?: ActivityLogContext,
+): Record<string, string | number> {
+  const renderingId = resolveRenderingId(ctx);
+
+  return {
+    ...(renderingId === undefined ? {} : { renderingId }),
+    ...(context?.collectionName === undefined ? {} : { collectionName: context.collectionName }),
+  };
+}
+
+function isPresent(value: string | undefined | null): boolean {
+  return value !== null && value !== undefined;
+}
+
+/**
+ * The status transition reads both the id and the index, so an answer carrying only an id strands
+ * the entry `pending`: the fail-closed policy has to engage here, not asynchronously afterwards.
+ */
+function isTransitionable(activityLog: ActivityLogResponse): boolean {
+  return isPresent(activityLog?.id) && isPresent(activityLog?.attributes?.index);
 }
 
 async function resolveCredentials(ctx: Context): Promise<AuditCredentials> {
@@ -111,6 +152,7 @@ export default async function createPendingActivityLog({
     credentials = await resolveCredentials(ctx);
   } catch (error) {
     logger('Error', `Activity log for '${action}' has no credentials to be created with`, {
+      ...auditIdentifiers(ctx, context),
       cause: describeCause(error),
     });
 
@@ -136,23 +178,26 @@ export default async function createPendingActivityLog({
     });
   } catch (error) {
     logger('Error', `Activity log for '${action}' could not be created`, {
+      ...auditIdentifiers(ctx, context),
       cause: describeCause(error),
     });
 
     if (isAuthorizationRefusal(error)) throw auditNotAuthorized();
+    if (isExpiredAuditCredential(error)) invalidateApiKeyIdentity(ctx);
     if (type === 'write') throw auditUnavailable(AUDIT_RETRY_AFTER_SECONDS);
 
     return null;
   }
 
-  if (activityLog?.id === null || activityLog?.id === undefined) {
-    if (type === 'write') throw auditUnavailable(AUDIT_RETRY_AFTER_SECONDS);
-
+  if (!isTransitionable(activityLog)) {
     logger(
       'Error',
       `Activity log for '${action}' could not be created: the server answered with no activity ` +
-        'log id, so the audit store dropped the write',
+        'log id or index, so the audit store dropped the write',
+      auditIdentifiers(ctx, context),
     );
+
+    if (type === 'write') throw auditUnavailable(AUDIT_RETRY_AFTER_SECONDS);
 
     return null;
   }
@@ -210,6 +255,8 @@ export function markActivityLog(options: MarkActivityLogOptions): void {
     .track(() => updateStatus(options))
     .catch(error => {
       logger('Error', `Failed to mark the activity log as '${status}'`, {
+        activityLogId: options.pending.activityLog.id,
+        index: options.pending.activityLog.attributes?.index,
         cause: describeCause(error),
       });
     });
--- a/packages/agent-bff/src/activity-log/activity-logs-creator.ts
+++ b/packages/agent-bff/src/activity-log/activity-logs-creator.ts
@@ -1,0 +1,216 @@
+import type ActivityLogDrainer from './activity-log-drainer';
+import type { ActivityLogsWriter } from './activity-logs-service';
+import type { Logger } from '../ports/logger-port';
+import type {
+  ActivityLogAction,
+  ActivityLogResponse,
+  ActivityLogType,
+} from '@forestadmin/forestadmin-client';
+import type { Context } from 'koa';
+
+import { HttpError, NotFoundError } from '@forestadmin/forestadmin-client';
+
+import {
+  resolveForestServerToken,
+  resolveRenderingId,
+} from '../auth/forest-server-token-middleware';
+import { sessionExpired } from '../http/bff-http-error';
+import {
+  AUDIT_RETRY_AFTER_SECONDS,
+  auditNotAuthorized,
+  auditUnavailable,
+} from '../http/bff-local-errors';
+
+/** The actions the BFF writes: its data routes read, and its action route writes. */
+export type BffActivityLogAction = Extract<
+  ActivityLogAction,
+  'index' | 'search' | 'filter' | 'listRelatedData' | 'action'
+>;
+
+/**
+ * Fail policy for the audit trail, keyed by action type: a write whose activity log cannot be
+ * created is blocked (no unaudited side effect), while a read proceeds with a warning (an audit
+ * store outage must not take down the read surface).
+ *
+ * One case is arbitrated by the cause instead of the action type: an authorization refusal
+ * (401/403) propagates for reads too — the read itself is not authorized either.
+ */
+const ACTION_TO_TYPE: Record<BffActivityLogAction, ActivityLogType> = {
+  index: 'read',
+  search: 'read',
+  filter: 'read',
+  listRelatedData: 'read',
+  action: 'write',
+};
+
+const MAX_STATUS_ATTEMPTS = 5;
+const STATUS_RETRY_DELAY_MS = 500;
+
+const NO_RENDERING_MESSAGE = 'This request carries no usable rendering';
+
+export interface ActivityLogContext {
+  collectionName?: string;
+  recordId?: string | number;
+  recordIds?: string[] | number[];
+  label?: string;
+}
+
+/**
+ * The token that created the log is kept for the status transition: the transition is fired after
+ * the response, when the session it came from may already be unreachable.
+ */
+export interface PendingActivityLog {
+  activityLog: ActivityLogResponse;
+  forestServerToken: string;
+}
+
+export interface CreatePendingActivityLogOptions {
+  ctx: Context;
+  service: ActivityLogsWriter;
+  action: BffActivityLogAction;
+  context?: ActivityLogContext;
+  logger: Logger;
+}
+
+interface AuditCredentials {
+  forestServerToken: string;
+  renderingId: string;
+}
+
+function describeCause(error: unknown): string {
+  return error instanceof Error ? `${error.name}: ${error.message}` : String(error);
+}
+
+function isAuthorizationRefusal(error: unknown): boolean {
+  return error instanceof HttpError && (error.status === 401 || error.status === 403);
+}
+
+async function resolveCredentials(ctx: Context): Promise<AuditCredentials> {
+  const renderingId = resolveRenderingId(ctx);
+
+  if (renderingId === undefined) throw sessionExpired(NO_RENDERING_MESSAGE);
+
+  return {
+    forestServerToken: await resolveForestServerToken(ctx),
+    renderingId: String(renderingId),
+  };
+}
+
+export default async function createPendingActivityLog({
+  ctx,
+  service,
+  action,
+  context,
+  logger,
+}: CreatePendingActivityLogOptions): Promise<PendingActivityLog | null> {
+  const type = ACTION_TO_TYPE[action];
+
+  let credentials: AuditCredentials;
+
+  try {
+    credentials = await resolveCredentials(ctx);
+  } catch (error) {
+    logger('Error', `Activity log for '${action}' has no credentials to be created with`, {
+      cause: describeCause(error),
+    });
+
+    if (type === 'write') throw error;
+
+    return null;
+  }
+
+  const { forestServerToken, renderingId } = credentials;
+
+  let activityLog: ActivityLogResponse;
+
+  try {
+    activityLog = await service.createMcpActivityLog({
+      forestServerToken,
+      renderingId,
+      action,
+      type,
+      collectionName: context?.collectionName,
+      recordId: context?.recordId,
+      recordIds: context?.recordIds,
+      label: context?.label,
+    });
+  } catch (error) {
+    logger('Error', `Activity log for '${action}' could not be created`, {
+      cause: describeCause(error),
+    });
+
+    if (isAuthorizationRefusal(error)) throw auditNotAuthorized();
+    if (type === 'write') throw auditUnavailable(AUDIT_RETRY_AFTER_SECONDS);
+
+    return null;
+  }
+
+  if (activityLog?.id === null || activityLog?.id === undefined) {
+    if (type === 'write') throw auditUnavailable(AUDIT_RETRY_AFTER_SECONDS);
+
+    logger(
+      'Error',
+      `Activity log for '${action}' could not be created: the server answered with no activity ` +
+        'log id, so the audit store dropped the write',
+    );
+
+    return null;
+  }
+
+  return { activityLog, forestServerToken };
+}
+
+export interface MarkActivityLogOptions {
+  service: ActivityLogsWriter;
+  drainer: ActivityLogDrainer;
+  pending: PendingActivityLog;
+  status: 'completed' | 'failed';
+  logger: Logger;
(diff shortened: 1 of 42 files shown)


packages/agent-bff/src/http/bff-http-server.ts
```diff
@@ -116,21 +126,50 @@ export default class BFFHttpServer {
   }
 
   async stop(): Promise<void> {
+    await this.closeConnections();
+    await this.options.drainActivityLogs?.();
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**: a SIGTERM during an audit store slowdown holds the process about 50s, so the orchestrator kills it mid drain.

`bff-http-server.ts:130` awaits the drain after the bounded close, and `activity-log-drainer.ts:27` loops with no deadline, up to 50s per stuck transition.

A way out: share `shutdownTimeoutMs` with the drain and log what was still in flight when it expired.
</details>

1. Run an audited `list` against a Forest server whose activity log endpoint answers slowly.
2. Send SIGTERM while the status transition is in flight.
3. The process stays up past the 10s connection deadline until the drain returns.
</details>

packages/agent-bff/src/activity-log/activity-logs-creator.ts
```diff
@@ -0,0 +1,263 @@
...
+  logger,
+}: CreatePendingActivityLogOptions): Promise<PendingActivityLog | null> {
+  const type = ACTION_TO_TYPE[action];
+
+  let credentials: AuditCredentials;
+
+  try {
+    credentials = await resolveCredentials(ctx);
+  } catch (error) {
+    logger('Error', `Activity log for '${action}' has no credentials to be created with`, {
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**: every list request logs at Error level on a Forest server that mints no audit token, a deployment this code calls supported.

`activity-logs-creator.ts:154` logs at `Error` and `with-activity-log.ts:37` logs at `Warn` for the same event, which `api-key-client.ts:20` calls a supported degradation.

A way out: log a structurally absent credential once at `Warn`, as PRD-1150 states for the read path.
</details>

1. Resolve an API key against a Forest server that sends no `saasAccessToken`.
2. Call `POST /agent/v1/collections/:name/list`.
3. The response is 200 and the logs carry one `Error` line plus one `Warn` line, on every request.
</details>

packages/agent-bff/src/activity-log/activity-logs-creator.ts
```diff
@@ -0,0 +1,263 @@
...
+    });
+  } catch (error) {
+    logger('Error', `Activity log for '${action}' could not be created`, {
+      ...auditIdentifiers(ctx, context),
+      cause: describeCause(error),
+    });
+
+    if (isAuthorizationRefusal(error)) throw auditNotAuthorized();
+    if (isExpiredAuditCredential(error)) invalidateApiKeyIdentity(ctx);
+    if (type === 'write') throw auditUnavailable(AUDIT_RETRY_AFTER_SECONDS);
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**: a Forest server without the activity log route answers every action execute with 503 and a retry hint that can never succeed.

`activity-logs-creator.ts:187` maps every failure that is neither 403 nor 401 to `auditUnavailable(AUDIT_RETRY_AFTER_SECONDS)`, a 404 on that route included.

A way out: route a 404 or a 501 from `createMcpActivityLog` through the no `retryAfter` path of `forest-server-token-middleware.ts:46`.
</details>

1. Point the BFF at a Forest server that does not expose `/api/activity-logs-requests/mcp`.
2. Call `POST /agent/v1/collections/:name/actions/:action/execute`.
3. The answer is 503 `audit_unavailable` with `Retry-After: 5`, and every retry repeats it.
</details>

packages/agent-bff/src/activity-log/activity-logs-creator.ts
```diff
@@ -0,0 +1,263 @@
...
+function isPresent(value: string | undefined | null): boolean {
+  return value !== null && value !== undefined;
+}
+
+/**
+ * The status transition reads both the id and the index, so an answer carrying only an id strands
+ * the entry `pending`: the fail-closed policy has to engage here, not asynchronously afterwards.
+ */
+function isTransitionable(activityLog: ActivityLogResponse): boolean {
+  return isPresent(activityLog?.id) && isPresent(activityLog?.attributes?.index);
```
Tonours, line 142: Claude Opus 5 (claude-opus-5): **Should fix**: an activity log answered with an empty id lets the action run and strands its entry pending for good.

`activity-logs-creator.ts:117` rejects only `null` and `undefined`, so an `id` of `''` passes the fail closed guard at `:126`.

A way out: require a non empty value in `isPresent`.
</details>

1. Have the audit endpoint answer a create with `{ id: '', attributes: { index: '' } }`.
2. Call an action execute.
3. The action runs, and its entry stays `pending` once the 5 retries are exhausted.
</details>

packages/agent-bff/src/activity-log/activity-logs-creator.ts
```diff
@@ -0,0 +1,263 @@
...
+      label: context?.label,
+    });
+  } catch (error) {
+    logger('Error', `Activity log for '${action}' could not be created`, {
+      ...auditIdentifiers(ctx, context),
+      cause: describeCause(error),
+    });
+
+    if (isAuthorizationRefusal(error)) throw auditNotAuthorized();
+    if (isExpiredAuditCredential(error)) invalidateApiKeyIdentity(ctx);
```
Tonours, line 240: Claude Opus 5 (claude-opus-5): **Should fix**: an audit endpoint that keeps answering 401 defeats the API key resolve cache, so every list costs two Forest server round trips.

`activity-logs-creator.ts:186` invalidates the identity on any audit 401, and the fail open read then re-resolves the key on the next request.

A way out: invalidate at most once per key per TTL window in `resolve-cache.ts`.
</details>

1. Resolve an API key whose `saasAccessToken` the audit endpoint rejects with 401.
2. Call `POST /agent/v1/collections/:name/list` twice.
3. Both calls hit `/liana/v1/bff-api-keys/resolve`, instead of the second being served from cache.
</details>

packages/agent-bff/src/cli-core.ts
```diff
@@ -6,16 +6,57 @@ import { parseConfig } from './config/env-config';
...
+export function installShutdownHandlers(server: BFFHttpServer, logger: Logger): void {
+  for (const { signal, handler } of installedShutdownHandlers) {
+    process.removeListener(signal, handler);
+  }
+
+  installedShutdownHandlers = SHUTDOWN_SIGNALS.map(signal => {
+    const handler = () => {
+      logger('Info', 'Stopping the Forest BFF', { signal });
+
+      server.stop().catch(error => {
```
Tonours, line 30: Claude Opus 5 (claude-opus-5): **Violates conventions**: the only signal an operator gets from a failed shutdown has no test, so a regression into an unhandled rejection ships green.

`cli-core.ts:30` logs the `stop()` rejection, and `test/cli-shutdown.test.ts:10` stubs a resolving `stop` only, against `testing.md#Cover error and edge paths, not only the happy path`.

A way out: a case where the stubbed `stop` rejects, asserting the `Error` line and its `cause`.
</details>
<details><summary>Verification</summary>

`test/http/bff-http-server.test.ts:326` pins that `stop()` rejects on a close error and skips the drain, so the branch is reachable.
</details>

Tonours (commented):
**Spec (PRD-1150)**: conforms. The three audited routes, the action selection, the approval special case, the two new errors and the lazy two-mode resolver all match the ticket, and `count` / `relation count` / `form` stay unaudited as it asks.

packages/agent-bff/src/http/bff-http-server.ts
Tonours: Closed: `stop` now spends one deadline across the connection close and the drain, so it cannot outlast `shutdownTimeoutMs`.

Predicate tried: a stop whose wall time exceeds `shutdownTimeoutMs` while a status transition stalls against the audit store.

`bff-http-server.ts` fixes the deadline before `closeConnections` and hands the drain what is left, while `drain` breaks on a non positive residual.

Regression pinned by three cases: the drainer returning on an expired deadline, on a zero deadline, and the server handing over the residual.
</details>

packages/agent-bff/src/activity-log/activity-logs-creator.ts
Tonours: Closed: a deployment that mints no audit credential is now reported once, at `Warn`, and the read path states nothing of its own.

Predicate tried: a request whose audit credential is structurally absent producing an `Error` line, or two lines for one event.

`forest-server-token-middleware.ts:46` throws an `audit_unavailable` with no `retryAfter`, which `isUnretryableAuditFailure` routes to `Warn`, and `with-activity-log.ts` no longer logs.

Regression pinned by a read asserting one `Warn` call and by a retryable resolution failure asserting the `Error` line is kept.
</details>

packages/agent-bff/src/activity-log/activity-logs-creator.ts
Tonours: Closed: a 404 or a 501 from the audit route now answers without a retry hint, and a real outage keeps one.

Predicate tried: a write refused with a `Retry-After` no retry can ever satisfy, because the Forest server serves no activity log route.

`activity-logs-creator.ts:243` routes an absent endpoint through `auditUnavailable` with no `retryAfter`, and the 500 case still carries `AUDIT_RETRY_AFTER_SECONDS`.

Regression pinned by three cases: a 404 and a 501 asserting `retryAfter` undefined, and a 500 asserting the hint survives.
</details>

packages/agent-bff/src/activity-log/activity-logs-creator.ts
Tonours: Closed: an empty id or index now fails the guard exactly like a missing one, so the action is blocked rather than stranded.

Predicate tried: any value that reaches `isTransitionable` truthy while naming no document, so the transition can never land.

`activity-logs-creator.ts:133` rejects the empty string too, and `isTransitionable` is the only caller, reading just the id and the index.

Regression pinned by three cases: an empty id and an empty index blocking an action, and a read served unaudited instead.
</details>

packages/agent-bff/src/activity-log/activity-logs-creator.ts
Tonours: Closed: a key is now forgotten at most once per positive TTL window, so a repeated audit 401 costs one extra resolve per window.

Predicate tried: a sequence of audit 401 answers that makes two Forest server round trips the steady state of every request.

`resolve-cache.ts:113` returns early while `now` is under the recorded window, and `cache.invalidate` has exactly one caller chain, the audit 401 path.

Regression pinned by three cases: a second invalidation inside the window ignored, another key still forgotten, and invalidation allowed again at the boundary.
</details>

packages/agent-bff/src/cli-core.ts
Tonours: Closed: the failed shutdown branch now has its own case, so a regression into an unhandled rejection no longer ships green.

Rule read as cited: every branch a caller can hit, thrown errors included, needs a case of its own.

The new case in `cli-shutdown.test.ts` stubs a rejecting `stop` and asserts the exact line `cli-core.ts:31` emits, cause included.

The added case is itself the pin, and the agent-bff job passes on this head.
</details>

Tonours (approved):
All six findings from the previous round are verified closed on this head, each with its regression pinned by a new case: the shared shutdown deadline, the single `Warn` for a deployment that mints no audit credential, the absent audit route answering without a retry hint, the empty id or index failing the guard, and the throttled key invalidation. The untested failed-shutdown branch now has its own case.

One unrelated job is red, `LLM Integration Tests (ai-proxy)`, on a package this branch does not touch. `Linting & Testing (agent-bff)` and the BFF integration tests pass.

packages/agent-bff/src/http/bff-http-server.ts
Tonours: Closed again at `63e1941`, which unrefs the status retry timer without touching the deadline `stop` spends across the close and the drain.

Predicate run against the fixed code: a `stop` with a transition that never settles took 301 ms for a 300 ms `shutdownTimeoutMs`.

The later commit only unrefs the retry wait, so the drain still breaks on a non positive residual.

Regression stays pinned by the drainer and server cases, and the six agent-bff suites pass on this head.
</details>

packages/agent-bff/src/activity-log/activity-logs-creator.ts
Tonours: Closed again at `63e1941`, which bounds the invalidation window map without losing the one forget per window property.

Predicate run against the fixed code: a hundred audit 401 answers over fifty seconds forgot the key once.

The new eviction drops a window only when `maxEntries` distinct keys hold live ones, so the guard degrades at ten thousand keys rather than leaking.

Regression stays pinned by the window cases plus two new eviction cases, and `resolve-cache.test.ts` passes on this head.
</details>

packages/agent-bff/src/activity-log/activity-logs-creator.ts
```diff
@@ -0,0 +1,216 @@
...
+  try {
+    await service.updateActivityLogStatus({
+      forestServerToken: pending.forestServerToken,
+      activityLog: pending.activityLog,
+      status,
+    });
+  } catch (error) {
+    // The document may not exist yet when the transition lands, and only then is a retry worth
+    // anything: a network failure loses the transition permanently.
+    if (error instanceof NotFoundError && attempt < MAX_STATUS_ATTEMPTS) {
```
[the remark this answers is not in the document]
Tonours: Fixed in a77f0032. `updateStatus` now retries every failure a later attempt can land — a `NotFoundError`, status 0 (unreachable), 408, 429, any 5xx, and a transport failure the client rethrows with no status at all — under the same bound of 5 attempts, 500 ms apart, on the unref'd timer. A refusal or a malformed request (403, 401, 400) still fails at once instead of burning four pointless round trips.

Covered by three cases in `activity-logs-creator.test.ts`: a 500 retried and landing on the second attempt, an `ECONNREFUSED`-style error with no status retried, a 503 exhausting the 5 attempts and reporting the entry, plus the 403 case asserting a single call.

packages/agent/src/embedded-bff.ts
```diff
@@ -157,10 +157,17 @@ export default class EmbeddedBff {
...
+   * nothing else holds them and a shutdown would leave the entries `pending`.
    */
-  stop(): void {
+  async stop(): Promise<void> {
+    const { bff } = this;
+
     this.bff = null;
     this.stopped = true;
+
+    await bff?.drainActivityLogs?.();
```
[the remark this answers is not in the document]
Tonours: Already answered in b37ccb3, which landed after this comment. Two changes close the race:

- `activity-log-writer.ts:44` tracks the audited **request** in the drainer, not only the status transition it fires afterwards.
- `activity-log-drainer.ts:39` loops instead of settling one snapshot: it re-reads `inFlight` after each pass, so a transition registered while the drain was awaiting the request it belongs to is picked up by the next iteration, and the loop only exits when the set is empty or the shared deadline expires.

`embedded-bff.ts:182` passes `shutdownTimeoutMs`, so that loop is bounded and what it leaves unfinished is logged by name.

packages/agent-bff/src/api-key/api-key-authenticator.ts
```diff
@@ -91,5 +103,13 @@ export default function createApiKeyAuthenticator({
 
       return authenticated;
     },
+
+    invalidate(rawKey) {
+      const parsed = parseApiKey(rawKey);
+
+      if (!parsed) return;
+
+      cache.invalidate(hashApiKey(parsed.keyId, parsed.secret));
```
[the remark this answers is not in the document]
Tonours: Already answered in b37ccb3, pushed after this review ran on 8d74201. `invalidate` is now credential-aware: `api-key-authenticator.ts:110` forwards the refused token to `resolve-cache.ts:130`, which keys the suppression window on it.

The scenario described plays out as: T1 refused → window opens, entry dropped, T2 resolved → T2 refused → `open.credential !== credential` and `retried` is false, so the entry is dropped again and a third resolution happens. Only a *repeat* refusal of the same token, or a third distinct one inside the window, is suppressed — that residual is the deliberate bound documented on `ResolveCache.invalidate`: a server refusing every token it mints would otherwise cost two round trips on every request forever. Pinned by the window/eviction cases in `resolve-cache.test.ts`.

Tonours (approved):
Approving on `a77f0032`. All review threads are resolved, CI is green (31 checks), and the dependency ForestAdmin/forestadmin-server#8483 is approved on its side — it still merges first.

The last round closed three threads:

- Status transitions dropped on a transient failure: fixed. `updateStatus` now retries a 404, status 0, 408, 429, any 5xx and a transport failure carrying no status; a 403 / 401 / 400 still fails at once. Same bound of 5 attempts, 500 ms apart, on the unref'd timer.
- Drain race on the embedded stop path: already answered by b37ccb3, which tracks the audited request itself and loops the drain instead of settling one snapshot.
- API-key invalidation suppressed per key: already answered by b37ccb3, which keys the suppression window on the refused credential.

Full disclosure on the record: the retry commit `a77f0032` is mine, so that part of the diff carries a reviewer's code and has not been read by a third party. Worth a glance from @nbouliol before merge.

PR #1887, opened 2026-09-08
fix(agent-bff): publish recordKey on context fields the agent renames

Tonours:
`GET /agent/v1/context` now publishes a `recordKey` on every field whose technical name differs from the key the record actually carries.

```json
{ "field": "created_at", "recordKey": "createdAt", "type": "Date" }
```

Read a record value under `recordKey ?? field`, keep sending `field` on the request side. The key is absent when both names match, so nothing changes for a camelCase agent.

## Why

`agent-client` deserializes the agent JSON:API with `keyForAttribute: 'camelCase'` (`http-requester.ts:31`), so a `created_at` column is projected under that name and returned as `createdAt`. The context published the technical name only. On any snake_case agent (agent-ruby, forest-rails) a consumer had no way to join the schema it read to the records it got back.

Two alternatives were rejected:

- Renaming `field` to the camelCase form. `field` is what a consumer sends back in a filter, a sort or a projection, and the BFF passes it to the agent verbatim. Renaming it means translating the whole request path.
- Returning records under their technical names. The published OpenAPI already documents record keys as camelCase, and the transform is not reversible: `first_name` and `firstName` collapse to the same key.

## How

`recordKey()` already existed in `openapi/record-schemas.ts`, where it renames fields in the generated record schemas. It moved to `data/record-key.ts`, the layer that describes agent-client deserialization rather than presentation, and both callers share it. `build-context.ts` sets the key only when it differs from `field`, and `ContextFieldSchema` carries it as optional.

Known limitation: sub-fields of a composite `type` carry no `recordKey`, although the same transform applies to them. Covering them means rewriting the type tree, which is passed through verbatim today. The gap is stated in the `ContextResponse` description rather than left for a consumer to discover.

## How to test

```bash
```

1657 tests pass, lint reports no error, build is clean.

Against a real snake_case agent (agent-ruby 1.39 on ActiveRecord), a projection on the technical name comes back under the record key, and the context now carries both:

```
POST /agent/v1/Category/list   projection: ["id","name","created_at"]
  { "createdAt": "2026-08-20T08:03:40.617Z", "id": "5", "name": "Finance" }
```

### General

### Security

### DIFF
--- a/packages/agent-bff/src/context/build-context.ts
+++ b/packages/agent-bff/src/context/build-context.ts
@@ -7,7 +7,7 @@ import type {
   ForestSchemaField,
 } from '@forestadmin/forestadmin-client';
 
-import recordKey from '../data/record-key';
+import recordKey, { groupByRecordKey } from '../data/record-key';
 
 export interface ContextActionField {
   field: string;
@@ -66,39 +66,33 @@ function toArray<T>(value: T[] | null | undefined): T[] {
 
 type FieldWithWireEnums = ForestSchemaField & { enums?: string[] };
 
-/** The deserializer writes the JSON:API resource identifier under this key, always. */
+/**
+ * The deserializer writes the JSON:API resource identifier under this key, always, over whatever
+ * attribute landed there. The resource id is a string whatever the column type says, and a
+ * composite key reaches it packed, so it is never the value of the field that camelizes to `id` —
+ * a primary key named `Id` included.
+ */
 const RESOURCE_ID_KEY = 'id';
 
 /**
  * The record keys this collection cannot promise, because more than one thing lands on them.
  *
  * Two fields whose technical names differ only by casing collapse onto one key (`first_name` and
- * `firstName` both reach the response as `firstName`), and a field named `Id` or `ID` collapses onto
- * the resource identifier, whose value the deserializer writes over the projected attribute — so
- * reading that field under `id` yields the record's id rather than the field's value. Publishing a
- * `recordKey` in either case would point a consumer at a value that is not the field's, which is
- * worse than publishing nothing: `recordKey` is absent, the caller falls back to `field`, and the
- * ambiguity stays visible instead of being papered over.
+ * `firstName` both reach the response as `firstName`), and `id` always belongs to the resource
+ * identifier. Publishing a `recordKey` in either case would point a consumer at a value that is not
+ * the field's, which is worse than publishing nothing: `recordKey` is absent, the caller falls back
+ * to `field`, and the ambiguity stays visible instead of being papered over.
+ *
+ * Not covered: the deserializer also writes `meta` from the resource meta, so a field named `Meta`
+ * on an agent that emits one is shadowed the same way. agent-nodejs emits no per-resource meta, and
+ * reserving the key would drop a working `recordKey` on every agent that emits none.
  */
 function ambiguousRecordKeys(fields: FieldWithWireEnums[]): Set<string> {
-  const claimants = new Map<string, number>();
+  const ambiguous = new Set<string>([RESOURCE_ID_KEY]);
 
-  for (const field of fields) {
-    const key = recordKey(field.field);
-    claimants.set(key, (claimants.get(key) ?? 0) + 1);
+  for (const [key, group] of groupByRecordKey(fields, field => field.field)) {
+    if (group.length > 1) ambiguous.add(key);
   }
-
-  const ambiguous = new Set<string>();
-
-  for (const [key, count] of claimants) {
-    if (count > 1) ambiguous.add(key);
-  }
-
-  const idClaimedByAnotherField = fields.some(
-    field => !field.isPrimaryKey && recordKey(field.field) === RESOURCE_ID_KEY,
-  );
-
-  if (idClaimedByAnotherField) ambiguous.add(RESOURCE_ID_KEY);
 
   return ambiguous;
 }
--- a/packages/agent-bff/src/context/build-context.ts
+++ b/packages/agent-bff/src/context/build-context.ts
@@ -66,6 +66,43 @@ function toArray<T>(value: T[] | null | undefined): T[] {
 
 type FieldWithWireEnums = ForestSchemaField & { enums?: string[] };
 
+/** The deserializer writes the JSON:API resource identifier under this key, always. */
+const RESOURCE_ID_KEY = 'id';
+
+/**
+ * The record keys this collection cannot promise, because more than one thing lands on them.
+ *
+ * Two fields whose technical names differ only by casing collapse onto one key (`first_name` and
+ * `firstName` both reach the response as `firstName`), and a field named `Id` or `ID` collapses onto
+ * the resource identifier, whose value the deserializer writes over the projected attribute — so
+ * reading that field under `id` yields the record's id rather than the field's value. Publishing a
+ * `recordKey` in either case would point a consumer at a value that is not the field's, which is
+ * worse than publishing nothing: `recordKey` is absent, the caller falls back to `field`, and the
+ * ambiguity stays visible instead of being papered over.
+ */
+function ambiguousRecordKeys(fields: FieldWithWireEnums[]): Set<string> {
+  const claimants = new Map<string, number>();
+
+  for (const field of fields) {
+    const key = recordKey(field.field);
+    claimants.set(key, (claimants.get(key) ?? 0) + 1);
+  }
+
+  const ambiguous = new Set<string>();
+
+  for (const [key, count] of claimants) {
+    if (count > 1) ambiguous.add(key);
+  }
+
+  const idClaimedByAnotherField = fields.some(
+    field => !field.isPrimaryKey && recordKey(field.field) === RESOURCE_ID_KEY,
+  );
+
+  if (idClaimedByAnotherField) ambiguous.add(RESOURCE_ID_KEY);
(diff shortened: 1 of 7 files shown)


packages/agent-bff/src/openapi/schemas.ts
```diff
@@ -412,6 +413,12 @@ export const ContextResponseSchema = z
       'collection this document does not expose. Cross a target against `collections[]` before ' +
       'following it — unlike the per-collection route documents, nothing here is dropped for ' +
       'pointing outside the served set. ' +
+      '`field` is the technical name the agent declares, and the one to send back in a filter, ' +
+      'a sort or a projection. It is NOT always the key the record carries: a snake_case agent ' +
+      '(Rails, Django) declares `created_at` and the response returns `createdAt`. When the two ' +
```
nbouliol: remove django since we dont support it anymore

Tonours: Done in 9de8b9d5b — description now reads `(Rails)` only.

packages/agent-bff/src/context/build-context.ts
```diff
@@ -79,6 +82,9 @@ function toContextValidations(validations: unknown[] | null | undefined): Contex
 function toContextField(field: FieldWithWireEnums): ContextField {
   const serialized: ContextField = { field: field.field, type: field.type };
 
+  const key = recordKey(field.field);
+  if (key !== field.field) serialized.recordKey = key;
```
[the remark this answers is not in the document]
Tonours: Confirmed, and it is worse than described: the attribute value is not just shadowed, it is gone. Reproduced against the real deserializer:

```
attributes sent : { Id: 'ATTRIBUTE-VALUE', created_at: '...', name: 'n' }
record returned : { id: '42', createdAt: '...', name: 'n' }
```

`Id` camelizes to `id`, and the deserializer writes the resource identifier there, so publishing `recordKey: "id"` points a consumer at the record's id rather than the field's value. Fixed in 8ae93e2e1.

I did not take the suggested diff, because gating on `isPrimaryKey` would drop `recordKey` for every non-key field, which is the entire point of the change: `created_at` -> `createdAt` on a snake_case agent is not a primary key. It would have removed the feature rather than the defect.

The fix instead omits `recordKey` whenever the key is ambiguous, which covers this case and one the review did not mention:

- a field whose camelized name collides with the resource identifier and is not the primary key (`Id`, `ID`)
- two fields differing only by casing, which collapse onto one key (`first_name` and `firstName`)

A primary key named `Id` still publishes `recordKey: "id"`, since there the resource identifier genuinely is that field's value.

`recordKey`'s presence is now a promise rather than a hint, and the `ContextResponse` description says so. Three tests pin it: the collision, the non-key `Id`, and the primary-key `Id` that must keep its key.

packages/agent-bff/src/context/build-context.ts
```diff
@@ -63,6 +66,43 @@ function toArray<T>(value: T[] | null | undefined): T[] {
...
+  }
+
+  const ambiguous = new Set<string>();
+
+  for (const [key, count] of claimants) {
+    if (count > 1) ambiguous.add(key);
+  }
+
+  const idClaimedByAnotherField = fields.some(
+    field => !field.isPrimaryKey && recordKey(field.field) === RESOURCE_ID_KEY,
```
nbouliol: The `!isPrimaryKey` exemption publishes `recordKey: "id"` for a PK named `Id`, but `record.id` is the JSON:API resource id — always a string whatever `type` says, and the packed `a|b` form for a composite key. Drop the exemption?

packages/agent-bff/src/context/build-context.ts
```diff
@@ -76,9 +116,15 @@ function toContextValidations(validations: unknown[] | null | undefined): Contex
...
 
-function toContextField(field: FieldWithWireEnums): ContextField {
+function toContextField(
+  field: FieldWithWireEnums,
+  ambiguousKeys: ReadonlySet<string>,
+): ContextField {
   const serialized: ContextField = { field: field.field, type: field.type };
 
+  const key = recordKey(field.field);
+  if (key !== field.field && !ambiguousKeys.has(key)) serialized.recordKey = key;
```
nbouliol, line 120: A non-PK field named `id` exits on `key === field.field`, so the caller falls back to `field` and reads the resource id instead of the field — what about letting the ambiguity check run before the equality one?

packages/agent-bff/src/context/build-context.ts
```diff
@@ -63,6 +66,43 @@ function toArray<T>(value: T[] | null | undefined): T[] {
...
+  for (const field of fields) {
+    const key = recordKey(field.field);
+    claimants.set(key, (claimants.get(key) ?? 0) + 1);
+  }
+
+  const ambiguous = new Set<string>();
+
+  for (const [key, count] of claimants) {
+    if (count > 1) ambiguous.add(key);
+  }
```
nbouliol: Third site deriving which record keys collapse (`record-schemas.ts:54-62` and `:104-112` group the same way) — what about a shared helper next to `record-key.ts`?

packages/agent-bff/src/context/build-context.ts
```diff
@@ -63,6 +66,43 @@ function toArray<T>(value: T[] | null | undefined): T[] {
 
 type FieldWithWireEnums = ForestSchemaField & { enums?: string[] };
 
+/** The deserializer writes the JSON:API resource identifier under this key, always. */
```
nbouliol: The deserializer also writes `meta` from the resource meta, so a field named `Meta` still gets a `recordKey` — narrow the comment, or cover `meta` too?

packages/agent-bff/src/context/build-context.ts
```diff
@@ -63,6 +66,43 @@ function toArray<T>(value: T[] | null | undefined): T[] {
...
+ * reading that field under `id` yields the record's id rather than the field's value. Publishing a
+ * `recordKey` in either case would point a consumer at a value that is not the field's, which is
+ * worse than publishing nothing: `recordKey` is absent, the caller falls back to `field`, and the
+ * ambiguity stays visible instead of being papered over.
+ */
+function ambiguousRecordKeys(fields: FieldWithWireEnums[]): Set<string> {
+  const claimants = new Map<string, number>();
+
+  for (const field of fields) {
+    const key = recordKey(field.field);
```
nbouliol: `recordKey(field.field)` is computed here, again at :98, and once more at :125 — a single pass into a `Map` would do.

packages/agent-bff/src/context/build-context.ts
Tonours: Right, and the OpenAPI record schema already said so — `ID_SCHEMA` documents `id` as always a string, packed for a composite key, disagreeing with the typed `__forest.primaryKey`. Publishing `recordKey: "id"` for a PK named `Id` contradicted that in the same package.

Exemption dropped in 38efd288b: `id` is now reserved unconditionally, so no field ever publishes it. A field literally named `id` never published one anyway (`key === field`), so this only removes the `Id`/`ID` case. Test flipped.

packages/agent-bff/src/context/build-context.ts
Tonours: Checked, and the reorder is a no-op: both branches omit `recordKey`, since a non-PK field named `id` has `key === field.field` and the ambiguity set now also holds `id`. `recordKey` absent means "fall back to `field`" — there is no way in the current contract to say "this field is not readable at all", which is what that case would need. That would take a separate marker; out of scope here, and the `ContextResponse` description already states that a field colliding with the resource identifier is not readable by name.

packages/agent-bff/src/context/build-context.ts
Tonours: Done in 38efd288b. `groupByRecordKey(items, name)` lives next to `recordKey` in `data/record-key.ts` and the three sites use it — `record-schemas.ts:54` and `:96`, plus `ambiguousRecordKeys`, which now reads a group length instead of counting by hand.

packages/agent-bff/src/context/build-context.ts
Tonours: Confirmed in the deserializer: `deserializer-utils.js` does `if ("meta" in from) { dest.meta = keyForAttribute(from.meta) }`, so resource meta shadows a field named `Meta` exactly like the resource id shadows `Id`.

I narrowed the comment rather than reserving the key. agent-nodejs emits no per-resource meta (`json-api-serializer` is registered without one, and the search metadata is top-level), so reserving `meta` would drop a working `recordKey` on every agent that emits none, to protect a case that needs both a resource meta and a field named `Meta`. The comment now names the gap.

packages/agent-bff/src/context/build-context.ts
Tonours: Two of the three are merged by the shared helper — `ambiguousRecordKeys` is one pass now instead of a count pass plus an `id` scan. `toContextField` still calls it once per field, which is one `recordKey` per field overall, so a `Map` threaded through the signature would buy nothing but a parameter.

nbouliol (approved):
Both round-one blockers are gone: `id` is reserved unconditionally in `ambiguousRecordKeys`, so no field publishes it, and `groupByRecordKey` now backs all three collapse sites instead of three hand-rolled copies. Verified the deserializer mirror is the same `inflected` pair the agent-client uses, not a transcription.

PR #1889, opened 2026-09-09
fix(workflow-executor): read an xToOne linkage from the raw JSON:API body

Scra3:
## Problem

A Load related record step on a **forest-rails smart `belongs_to`** finds the relation and loads nothing — the next step then fails with *"that step didn't load any record"*.

Reported on a Qonto workflow (`CardClaim__Claim.card` → `CardLifecycle__Card`), after #1884 and the server-side counterpart already made the editor and the engine agree the field *is* a relation.

## Root cause

`getSingleRelatedData` read the relation off the **deserialized** parent record. Both halves of the chain are individually correct, and they lose the value between them:

- forest-rails serializes a smart `belongs_to` as `has_one(… include_data: true)`, so `relationships.card.data = { type, id }` **is** on the wire;
- but its by-id route builds `include:` from real ActiveRecord associations only, so the Card never lands in `included`;
- and `jsonapi-serializer` **deletes a relationship key entirely** when its linkage has no matching `included` entry (`deserializer-utils.js`, `findIncluded` → `null` → key never assigned).

So the adapter saw `undefined`, returned `null`, and the step reported zero candidates. Real relations were never affected: they *are* in `include:`, hence in `included`, hence kept by the deserializer.

## Fix

Read the parent's **raw JSON:API body** (`getOne(…, { skipDeserialization: true })` — what `resolvePolymorphicType` already does, 200 lines below, for this exact reason), take the id, then load the target by id:

```ts
const raw = node?.relationships?.[relation]?.data?.id ?? node?.attributes?.[relation];
```

One path now covers both shapes: the linkage (Rails `belongs_to` DSL, forest-express references, v2 agents) and the scalar attribute (forest-rails `field ... reference:`, whose value *is* the related id). Two consequences worth noting:

- raw keys are the agent's own field names, so the `toCamelCase(relation)` lookup — and the inflection guessing it required (#1787) — is gone from this path;
- the target is always re-read by id, so `values` come from a normal `getRecord` (full projection, field names restored) instead of whatever a single sub-field projection had packed into the linkage.

A 404 on the parent still maps to `RecordNotFoundError`.

## Tests

The old tests mocked `getOne` with **already-deserialized** objects, so the deserializer — the thing that breaks — was never in the picture. They asserted a belief about the wire format, not the wire format. Rewritten with real JSON:API bodies:

- `follows a linkage that has no included entry` — the exact reported shape, red before this change;
- the four inflection tests collapse into one `it.each` over raw relation names, and the five empty cases into one `it.each` (`data: null`, linkage without id, neither relationship nor attribute, `null` attribute, `''` attribute) → `null` without touching the target;
- added: 404 on the parent → `RecordNotFoundError`, non-404 → `AgentPortError`.

`83 passed` on the file (10 red without the src change), `1772 passed` on the package.

## Review follow-up (2nd commit)

`/validator` on the first commit found three defects in the seam the rewrite opened; all three are fixed in `aca2746`:

- **The parent lost its "200 + empty body ⇒ not found" mapping.** The old path inherited it from `getRecord` (which documents the agents that answer a missing composite-key record that way). Losing it reported a source record that could not be read as *"no related record to load"* — the exact conflation this PR exists to remove, reintroduced two lines below the fix. Restored, with the case added to the table.
- **An unreadable target aborted the run.** The linkage branch used to build the result from the parent's own body, so the target's readability never came into play; it now always reads the target. A v2 by-id route intersects the caller's scope while the relation projected on the parent does not, so a valid linkage to a scoped-out (or dangling, or soft-deleted) record is reachable — and it used to complete. It now returns no record rather than failing the run with *"The record no longer exists."*, which was false for the scope case. The parent is deliberately the opposite: it is the step's own source record, so failing to read it stays an error.
- **A full record was fetched to be discarded.** `fetchXToOneCandidate` passes `fields` only when the related schema has a `referenceField`; otherwise it reads `recordId` alone. With nothing to project, the linkage id is the whole answer and the second read is skipped.

`88 passed` on the file, `1777 passed` on the package.

### General

### Security

- [x] Consider the security impact of the changes made — read-only path, no new data leaves the executor

> [!NOTE]
> ### Fix `AgentClientAgentPort.getSingleRelatedData` to read xToOne linkage from raw JSON:API body
> - Replaces the deserialized parent-record lookup with a direct agent-client request using `skipDeserialization` and a single relation projection, reading raw relationship linkage or attribute values from the parent record.
> - Handles null, empty, missing, and id-less relations; splits packed IDs only for composite keys; maps parent 404s and empty bodies to `RecordNotFoundError`; avoids reading the target when no fields are requested.
> - Reworks the test suite to use raw JSON:API-shaped parent responses, covering linkages without included records, scalar smart-field attributes, attribute-held records, composite/single-key IDs, no-field target avoidance, and separate parent vs. target not-found behavior.
> - Behavioral Change: target `RecordNotFoundError` is now converted to `null` instead of propagating; parent 404s and empty bodies now map to `RecordNotFoundError` in `getSingleRelatedData`.
>
> 
> #### Changes since #1889 opened
>
> - Modified `AgentClientAgentPort.getSingleRelatedData` port method to always read xToOne relation targets from the database [62b0a23]
> - Added test fixture and parameterized test cases in `AgentClientAgentPort` to verify reading xToOne relationship linkage from JSON:API responses [0927e41]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 024927d.</sup>
> 
>

### DIFF
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -293,15 +293,16 @@ export default class AgentClientAgentPort implements AgentPort {
       const recordId =
         relatedSchema.primaryKeyFields.length > 1 ? String(raw).split('|') : [String(raw)];
 
-      // Nothing to project: the caller wants the linkage id alone (it reads `values` only for the
-      // reference field it asked for), so reading the target would fetch a whole record to discard.
-      if (!fields?.length) {
-        return { collectionName: relatedSchema.collectionName, recordId, values: {} };
-      }
+      // The target is read even when the caller projected nothing, on its first PK alone. Skipping
+      // the read to save the round trip also skipped the readability check below, so a relation
+      // whose collection happens to carry no reference field would load a record the caller cannot
+      // open, while the same relation with one reported no record — the outcome turning on a
+      // schema detail rather than on what the caller can see.
+      const projection = fields?.length ? fields : [relatedSchema.primaryKeyFields[0]];
 
       try {
         return await this.getRecord(
-          { collection: relatedSchema.collectionName, id: recordId, fields },
+          { collection: relatedSchema.collectionName, id: recordId, fields: projection },
           user,
         );
       } catch (error) {
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -285,8 +285,11 @@ export default class AgentClientAgentPort implements AgentPort {
 
       if (raw == null || raw === '') return null;
 
-      // Only a composite key is pipe-packed. A single-key value can legitimately hold a pipe —
-      // a smart field's value is written by the client — so splitting it would tear it in two.
+      // Only a composite key is pipe-packed, so a single-key value is never split — a smart field's
+      // value is written by the client and splitting it would tear it in two. Such a value cannot
+      // hold a pipe either way: `serializeRecordId` refuses one rather than emit an id the agent
+      // would unpack into the wrong number of key parts. That surfaces as an AgentPortError naming
+      // the value, which is cryptic but loud; naming the relation too would need a domain error.
       const recordId =
         relatedSchema.primaryKeyFields.length > 1 ? String(raw).split('|') : [String(raw)];
 
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -260,7 +260,12 @@ export default class AgentClientAgentPort implements AgentPort {
         throw error;
       }
 
-      const node = body?.data;
+      // Some agents answer a missing composite-key record with a 200 + empty body instead of 404,
+      // like getRecord above. Reading that as an unset relation would report "nothing to load" for
+      // a parent that was never read — the exact confusion this method exists to remove.
+      if (!body?.data) throw new RecordNotFoundError(collection, id);
+
+      const node = body.data;
       // Raw keys are the agent's own field names, so no inflection to second-guess. A linkage
       // covers the Rails `belongs_to` DSL, forest-express references and v2 agents; the attribute
       // covers a forest-rails `field ... reference:`, whose value IS the related id.
@@ -270,16 +275,32 @@ export default class AgentClientAgentPort implements AgentPort {
 
       if (raw == null || raw === '') return null;
 
-      return this.getRecord(
-        {
-          collection: relatedSchema.collectionName,
-          // Only a composite key is pipe-packed. A single-key value can legitimately hold a pipe —
-          // a smart field's value is written by the client — so splitting it would tear it in two.
-          id: relatedSchema.primaryKeyFields.length > 1 ? String(raw).split('|') : [String(raw)],
-          ...(fields?.length ? { fields } : {}),
-        },
-        user,
-      );
+      // Only a composite key is pipe-packed. A single-key value can legitimately hold a pipe —
+      // a smart field's value is written by the client — so splitting it would tear it in two.
+      const recordId =
+        relatedSchema.primaryKeyFields.length > 1 ? String(raw).split('|') : [String(raw)];
+
+      // Nothing to project: the caller wants the linkage id alone (it reads `values` only for the
+      // reference field it asked for), so reading the target would fetch a whole record to discard.
+      if (!fields?.length) {
+        return { collectionName: relatedSchema.collectionName, recordId, values: {} };
+      }
+
+      try {
+        return await this.getRecord(
+          { collection: relatedSchema.collectionName, id: recordId, fields },
+          user,
+        );
+      } catch (error) {
+        // A target the caller cannot read — scoped out, segmented out, or a dangling link — is not
+        // a broken run: the by-id route intersects the caller's scope while the relation projected
+        // on the parent does not, so a valid linkage to an invisible record is reachable. "No
+        // related record" is the true answer from this caller's vantage. The parent above is the
+        // opposite case: it is the step's own source record, so failing to read it stays an error.
+        if (error instanceof RecordNotFoundError) return null;
+
+        throw error;
+      }
     });
   }
 
--- a/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
+++ b/packages/workflow-executor/src/adapters/agent-client-agent-port.ts
@@ -58,6 +58,13 @@ function mapActionExecutionError(action: string, cause: unknown): unknown {
   return cause;
 }
 
+type RawRecordBody = {
+  data?: {
+    attributes?: Record<string, unknown>;
+    relationships?: Record<string, { data?: { id?: string } | null }>;
+  };
+};
+
 // Transcribes inflected's `underscore` + `camelize(_, false)` — the inflection agent-client's
 // jsonapi-serializer deserializer applies to every response key. Divergence = silent lookup miss.
 function toCamelCase(name: string): string {
@@ -220,10 +227,11 @@ export default class AgentClientAgentPort implements AgentPort {
     });
   }
 
-  // xToOne relations have no /relationships/<relation> route on the agent. We read the
-  // parent record with a `<relation>@@@<field>` projection and unpack the relation linkage
-  // jsonapi-serializer emits as a nested object on the parent (with the related PK packed
-  // under "id" when composite).
+  // xToOne relations have no /relationships/<relation> route on the agent, so the relation is read
+  // off the parent record — from the RAW JSON:API body. jsonapi-serializer drops a relationship key
+  // whose linkage has no matching `included` entry, and forest-rails only compounds real
+  // ActiveRecord associations, so a Rails smart belongs_to reached us as "no relation at all".
+  // resolvePolymorphicType reads raw for the same reason.
   async getSingleRelatedData(
     { collection, id, relation, relatedSchema, fields }: GetSingleRelatedDataQuery,
     user: StepUser,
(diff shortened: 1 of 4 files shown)


packages/workflow-executor/src/adapters/agent-client-agent-port.ts
```diff
@@ -232,52 +240,46 @@ export default class AgentClientAgentPort implements AgentPort {
...
+      } catch (error) {
+        if (HttpRequester.is404Error(error)) throw new RecordNotFoundError(collection, id);
 
-      const linkage = raw as Record<string, unknown> | null | undefined;
-      const packedId = linkage?.id as string | undefined;
+        throw error;
+      }
 
-      if (!linkage || !packedId) return null;
+      const node = body?.data;
```
Scra3: Claude Opus 5 (claude-opus-5): **Must fix** — a source record that no longer exists is reported to the operator as "no related record to load", and the run continues past it.

The parent used to be read through `getRecord`, which maps **two** not-found cases: the 404, and a 200 with an empty body — "Some agents answer a missing composite-key record with a 200 + empty body instead of 404" (`agent-client-agent-port.ts:160-163`, covered by its own test at `test/adapters/agent-client-agent-port.test.ts:233`). The raw read re-implements only the 404 arm.

So a deleted composite-key parent on such an agent answers `200 {}` → `body.data` is `undefined` → `raw` is `undefined` → line 271 returns `null`. In `FullyAutomated` the step calls `persistSkip` and the run continues; in AI-assisted the operator gets "No X to load" pre-checked. Indistinguishable from an empty relation — the exact conflation this PR exists to remove, reintroduced two lines below the fix.

`if (!body?.data) throw new RecordNotFoundError(collection, id);` before reading the linkage, plus the case in the `it.each` table — which currently enumerates five "no relation" shapes, none of them a `data`-less body.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
```diff
@@ -232,52 +240,46 @@ export default class AgentClientAgentPort implements AgentPort {
...
 
-      const restored = restoreFieldNames(linkage, [projectedField]);
+      if (raw == null || raw === '') return null;
 
-      return {
-        collectionName: relatedSchema.collectionName,
-        recordId: packedId.split('|'),
-        values: restored,
-      };
+      return this.getRecord(
```
Scra3: Claude Opus 5 (claude-opus-5): **Should fix** — a workflow that completed before this PR now fails mid-step, and blames the wrong thing.

The linkage branch used to build `RecordData` from the parent's own body, so the target's readability never came into play. It now always reads the target by id. A v2 agent's by-id route intersects the caller's scope into the filter and 404s when nothing matches (`packages/agent/src/routes/access/get.ts:19-40`), while a relation projected on the parent is **not** scope-filtered — so a valid linkage plus a target the user's scope excludes is reachable, as is a dangling FK or a soft-deleted target.

`RecordNotFoundError` is a `WorkflowExecutorError`, so `callAgent` rethrows it untouched and neither `fetchXToOneCandidate` nor `handleFirstCall` catches it: the run errors out with *"The record no longer exists. It may have been deleted."* — false for the scope case, and naming only the target collection, so nothing says the id came from following this relation.

Not leaking a scoped-out record is right; aborting the run to do it is not. Catching `RecordNotFoundError` from the target read and returning `null` restores the pre-PR outcome. Either way the choice deserves a test — both error tests today reject only the *first* `getOne`.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
```diff
@@ -232,52 +240,46 @@ export default class AgentClientAgentPort implements AgentPort {
...
+        {
+          collection: relatedSchema.collectionName,
+          // Only a composite key is pipe-packed. A single-key value can legitimately hold a pipe —
+          // a smart field's value is written by the client — so splitting it would tear it in two.
+          id: relatedSchema.primaryKeyFields.length > 1 ? String(raw).split('|') : [String(raw)],
+          ...(fields?.length ? { fields } : {}),
```
Scra3: Claude Opus 5 (claude-opus-5): **Should fix** — every xToOne step whose target has no `referenceField` pays a second round trip and a full record payload for values nobody reads.

`fetchXToOneCandidate` passes `fields` only when the related schema has a `referenceField` (`load-related-record-step-executor.ts:471-489`); otherwise it uses `candidate.recordId` alone and writes `referenceFieldValue: null`, never touching `candidate.values`. With `fields` empty this line still calls `getRecord` with no projection, so the agent returns the entire related record — every attribute, its own relationships — and all of it is discarded. The port already holds the only thing the caller wants: the linkage id.

It also inherits the failure surface above for data nobody reads. Returning `{ collectionName, recordId, values: {} }` when `fields` is empty skips both.

Scra3 (commented):
### Spec (no ticket)

No Linear ticket is linked to this PR — no `linear-code[bot]` linkback, nothing in the body, the branch name, the title or the commit trailers — so the functional check was limited to the PR description. Against that description the change does what it claims: the premise is verified in `node_modules/jsonapi-serializer/lib/deserializer-utils.js:120-124` (a linkage with no matching `included` entry has its key dropped outright), the `??` ordering is sound, and the composite split is strictly better than the unconditional one it replaces.

The three findings below are all in the seam the rewrite opened: what the method does when the body is not the shape it expects, and what the new second read costs.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
Scra3: **Closed.** Predicate: *a 200 whose body carries no `data` node makes the method answer "no related record" instead of `RecordNotFoundError`.*

What closes the class is a guard placed before any relation is read — `if (!body?.data) throw new RecordNotFoundError(collection, id)` — so the relation resolution is unreachable without a truthy `data` node, whatever the body is: `{}`, `{data:null}`, `null`, or a body that is not an object at all.

Checked by running the head test file against the pre-fix source in a scratch worktree at `0b6c1eb`: `treats a 200 with an empty parent body ({})` and `({"data":null})` both fail there and pass at `87ebe7a`. Regression pinned.

Boundary worth stating: `{"data":{}}` — a data node present but empty — still resolves to no relation. That is outside this failure mode (an agent signalling a missing record answers with an empty body, and JSON:API requires a resource object to carry `type` and `id`), so it is not treated as a live input here.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
Scra3: **Closed.** Predicate: *a linkage or attribute resolving to a target the caller cannot read makes the step throw instead of reporting no related record.*

The target read is now wrapped, and `RecordNotFoundError` — the only error either not-found arm of `getRecord` produces, the 404 and the 200-with-empty-body — returns `null`; everything else is rethrown, so a real failure still surfaces. The parent read above is deliberately the opposite, and the comment says why: it is the step's own source record.

Checked two ways. The head test `returns null when the target is unreadable, instead of failing the step` fails against the pre-fix source at `0b6c1eb` and passes at `87ebe7a`, and `rethrows a non-404 on the target read` pins the other arm. Then the behaviour change (throw → `null`) was swept at its consumers: `getSingleRelatedData` is reached from `agent-with-log.ts:76` (a pass-through wrapper) and `load-related-record-step-executor.ts:477` (`fetchXToOneCandidate`), which already treats `null` as "no candidate". No other call site. Regression pinned.

One consequence to keep in view, since it is now silent by design: a scoped-out target and a genuinely absent relation are indistinguishable to the operator, and this adapter has no logger to separate them.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
Scra3: **Closed.** Predicate: *with no field projected, the method issues a second request and fetches a full record that is discarded.*

An early return before the target read makes that read unreachable unless `fields?.length`, so the linkage id is answered on its own. Pinned by `returns the linkage id without reading the target when no field is projected`, which asserts a single `getOne`; it fails against the pre-fix source at `0b6c1eb` and passes at `87ebe7a`.

The premise turned out stronger than the finding assumed: forest-rails' by-id route forces `unoptimized: true` in `serialize_model`, so it ignores the `fields` projection and returns the whole record regardless — measured against a live v1 agent, not inferred. The skipped read was a full record on that stack, not three fields.

Re-derived after `87ebe7a`, which moved how `raw` is computed one commit later: the early return depends on `recordId`, so its premise moved. It holds — the return is independent of how `raw` was derived, and for a record-valued attribute `recordId` is now a real id where it would have been `["[object Object]"]`.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
```diff
@@ -232,52 +240,77 @@ export default class AgentClientAgentPort implements AgentPort {
...
+        linkageId ??
+        (attribute && typeof attribute === 'object'
+          ? (attribute as { id?: unknown }).id
+          : attribute);
+
+      if (raw == null || raw === '') return null;
+
+      // Only a composite key is pipe-packed. A single-key value can legitimately hold a pipe —
+      // a smart field's value is written by the client — so splitting it would tear it in two.
+      const recordId =
```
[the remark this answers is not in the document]
Scra3: Confirmed on the observation, declining the fix.

**The throw is real.** `serializeRecordId` refuses a `|` inside a key part, and the id is passed as a one-element array, so `acme|corp` on a single-key target does raise.

**Passing it as a scalar would make it worse.** That bypasses the guard and puts `acme|corp` on the wire. The agent unpacks a packed id by splitting on the separator and requires exactly as many parts as the collection has primary keys — `IdUtils.unpackId` in `packages/agent/src/utils/id.ts:37-47` — so two values against one key is a `ValidationError` there, and on a laxer stack it is a lookup that can match the wrong record. A single-key value containing a pipe is not addressable in Forest's by-id URL scheme at all, which is why the serializer refuses it early rather than emit it: *"Throws rather than silently corrupting the key"*, in its own comment. Moving the failure from our process to theirs is not a fix.

**What was actually wrong here was mine, and it is fixed in `024927d8d`.** The test named `keeps a pipe intact when the target key is single` asserted that the pipe survives to the target read. It cannot. It passed only because `@forestadmin/agent-client` is mocked in that file, so the real serializer never ran — the test asserted the opposite of the truth. It is renamed to `does not split a single-key value on the pipe` and scoped to what it legitimately pins: that a single-key value is not split, which is the behaviour the composite branch changed. The throw itself is already covered where it lives, `packages/agent-client/test/record-id.test.ts:38`. The source comment claimed such a value *can legitimately hold a pipe* — also false, also corrected.

So: no behaviour change. The value surfaces an `AgentPortError` naming it, which is loud but cryptic — it does not say the id came from following a relation. If that turns out to matter in the field, the fix is a domain error naming the relation and the parent collection, not a looser id.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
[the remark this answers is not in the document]
Scra3: Fixed in `aca274601`, and thank you — this one was real and I had reintroduced it two lines below the fix it was about.

The guard is back before any relation is read: `if (!body?.data) throw new RecordNotFoundError(collection, id)`, so a 200 with an empty body can no longer read as an unset relation. Pinned by `it.each([{}, { data: null }])` in the adapter suite; both cases fail against the pre-fix source at `0b6c1eb` and pass at head, checked in a scratch worktree.

An independent conventions audit reached the same finding from the test-coverage side, which is a good sign for this check.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
```diff
@@ -232,52 +240,80 @@ export default class AgentClientAgentPort implements AgentPort {
...
+      // value is written by the client and splitting it would tear it in two. Such a value cannot
+      // hold a pipe either way: `serializeRecordId` refuses one rather than emit an id the agent
+      // would unpack into the wrong number of key parts. That surfaces as an AgentPortError naming
+      // the value, which is cryptic but loud; naming the relation too would need a domain error.
+      const recordId =
+        relatedSchema.primaryKeyFields.length > 1 ? String(raw).split('|') : [String(raw)];
+
+      // Nothing to project: the caller wants the linkage id alone (it reads `values` only for the
+      // reference field it asked for), so reading the target would fetch a whole record to discard.
+      if (!fields?.length) {
```
EnkiP: Claude Opus 5 (claude-opus-5): **Should fix**: a workflow loads a related record the operator cannot open, but only when the related collection has no reference field.

Line 298 skips the target read, and `load-related-record-step-executor.ts:482` sends `fields` only when `referenceField` is set.

A way out: project `relatedSchema.primaryKeyFields.slice(0, 1)` when `fields` is empty.
</details>

1. Take a BelongsTo whose related collection has no reference field.
2. Point the linkage at a record the run's role cannot read.
3. Run the workflow.
4. The step loads that record instead of reporting none.
5. Add a reference field to that collection and rerun.
6. The step now reports no related record.
</details>

EnkiP: **Applies to:** `packages/workflow-executor/src/ports/agent-port.ts:23-26` (not in this diff)

Claude Opus 5 (claude-opus-5): **Should fix**: the port contract still tells the next reader the relation is unpacked from the parent record.

`agent-port.ts:25` says the port "unpacks the relation linkage embedded on the parent", which line 298 is the last path resembling.

A way out: restate that comment around the raw body and the `getRecord` target read.
</details>

1. Read `agent-port.ts:23-26` for how the port reads an xToOne relation.
2. Compare it with `getSingleRelatedData` at this sha.
3. The mechanism it describes is gone.
</details>

EnkiP (commented):
**Spec (no ticket)**: no Linear ticket is linked on the PR, in the branch, or in the commits, and a topical search found none, so the functional check was limited to the PR description. Against that description the change does what it claims: the raw-body read removes the deserializer drop and the inflection guess, and the three error semantics the second commit added are each implemented and tested.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
Scra3: Fixed in `62b0a23b9` — you are right, and it is worth naming what went wrong: the inconsistency is one I introduced, in answer to one of my own review findings.

That finding asked me to skip the target read when the caller projects nothing, to avoid fetching a whole record to discard. Doing so also skipped the readability check the *previous* finding had just added, so the outcome ended up turning on whether the related collection happens to carry a reference field rather than on what the caller can see. Two findings pulling opposite ways; correctness wins, and the round trip comes back.

Implemented as you suggested: the read now goes out on `relatedSchema.primaryKeyFields[0]` when `fields` is empty, so there is one code path and the target's readability always decides. Pinned by two tests — the target is read with `{ fields: ['id'] }`, and an unreadable target with no projection returns `null` (that second one fails on the previous commit).

The saving I was protecting was partly illusory anyway: forest-rails' by-id route forces `unoptimized: true` in `serialize_model` and ignores the projection, so on a v1 stack it returns the full record whatever we ask for. Measured against a live agent while validating this PR, not inferred.

Scra3: Fixed in `62b0a23b9`. The comment described the mechanism the first commit removed, so it was pointing the next reader at the one thing that is no longer true.

It now states what the port actually does — reads the parent's raw JSON:API body with the `<relation>@@@<field>` projection, takes the id from the relationship linkage or the reference attribute, then reads the target by id — and why the raw body is not an implementation whim: deserializing drops a relationship key whose linkage has no matching `included` entry, which is exactly what an agent emits for a smart relation.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
Scra3: Reopening this one's conclusion rather than the thread: the optimisation it closed is **withdrawn** in `62b0a23b9`.

A later review (@EnkiP) showed that skipping the target read also skipped the readability check the finding above had just introduced, so a relation whose collection carries no reference field loaded a record the caller cannot open, while the same relation with one reported no record. The saving was not worth an outcome that turns on a schema detail. The target is read again, projecting its first primary key when the caller asked for nothing.

Leaving the thread resolved — the defect it described was real and is gone — but the fix that closed it no longer stands, and that should not be discoverable only by reading the diff.

packages/workflow-executor/src/adapters/agent-client-agent-port.ts
Scra3: **Closed.** The target read is now unconditional, so readability decides the outcome for every relation.

Predicate: a relation whose collection carries no reference field loads a record the caller cannot open.
Checked by running the head tests against 024927d8d in a worktree, where both new tests fail.
Regression pinned by those two tests.

</details>

Scra3: **Closed.** The port comment now describes the raw body read that the implementation performs.

Predicate: the contract describes a mechanism the implementation no longer has.
Checked by reading agent-port.ts and getSingleRelatedData at head side by side.
Nothing pins prose, so no regression test exists and none can.

</details>

EnkiP (approved):

PR #1891, opened 2026-09-10
feat(agent-bff): log one line per request

nbouliol:
Fixes PRD-1133

## Problem

The BFF logged nothing per request, in either deployment mode. Everything it logged was assembly-time or exceptional: configuration gaps, a cache invalidation, an unhandled edge error.

Found while debugging the embedded BFF against a real agent. Calling `/bff/agent/openapi.json` and `/bff/health` produced **no output at all** — correct for the code as written, and indistinguishable from "my mount is broken":

- neither route reaches the agent, so the agent's own request logger never fires
- and the BFF said nothing itself

A data call that *does* reach the agent logs one line, but from the agent and with no marker — indistinguishable from Forest UI traffic. The standalone docker image had the same blind spot.

## What this does

Adds an access-log middleware at the head of the chain assembled by `buildBff`, so it lands in both deployment modes at once.

- `[status] METHOD path - Xms` — the shape the agent's own `Logger` route emits, so both layers read the same way. Embedded, it comes out `[BFF]`-prefixed for free through the existing logger adapter.
- `Info` under 400, `Warn` from 400, `Error` from 500.
- At the head of the chain, so it also covers what never reaches the agent or the agent-scoped error middleware: `/health`, `/docs`, `/oauth/*`, a 401 from the auth edge, a rejected origin, `openapi_disabled`, a 404 on an unclaimed path.
- No request body, no `Authorization` header, no `X-Forest-Bff-Key`.

## Decisions worth reviewing

**Message only, no structured fields.** The standalone sink (`console-logger`) can flatten a context into JSON fields, and a log pipeline would rather query `status` than regex a string. Chosen against: it keeps the format identical to the agent's, and it keeps the change inside one package — passing a context would make `formatLog` in `packages/agent` re-serialize it behind the message, so every embedded line would carry the same values twice.

**Status read off the exception, then rethrown.** The middleware sits above every error handler the BFF has, so at the moment it logs, `ctx.status` is still Koa's default 404 and the line would name a status nobody was ever answered. Copying the agent's `Logger` verbatim was the other option — it catches without rethrowing, which here would have turned a 500 into a bare 404 on the public surface.

**No cause on the line, and a guarded logger.** The line carries a status and nothing else. Passing a `cause` through the port's third argument is what `formatLog` in `packages/agent` re-serializes behind the message, so an embedded deployment would carry the same values twice on every line that has one — and silencing it means a marker shared between two separately released packages. The detail is not lost where it matters: `createErrorMiddleware` already logs `cause` for the whole `/agent` surface. Outside it — a body-parser rejection on an OAuth POST is the only route today — the stack reaches stderr through Koa's handler, as before this middleware existed. The real fix is that `createErrorMiddleware` is `agentScoped` and should not be, which changes OAuth's error contract and deserves its own ticket.

The `logger` call itself is wrapped: it comes from the host, nothing in the port forbids it from throwing, and this middleware runs on every route. Unguarded it would answer 500 to a request the route served, and on the error path it would replace the exception being reported with its own.

**No query string, ever.** Nothing routed today needs it: OAuth's sensitive exchange is a POST body, and the data routes read their filters and `search` from the body too. Logging it would be safe *today* and would leak the first time a route accepts something sensitive there, with nobody revisiting this line.

**Path named as the caller asked for it.** Embedded, `req.url` is rewritten before the BFF sees it, so the BFF's own view is `/agent/openapi.json` for a client that called `/bff/agent/openapi.json` — an operator grepping `/bff/` would still find nothing. Uses the existing `emittedBaseOf` helper, which also prefers the prefix a host actually stripped over the configured one (`/api/bff/health`).

**Two lines per data call, embedded.** One from the BFF, one from the agent, not correlated. Kept deliberately: the paths and timings differ, and the delta is the BFF's own cost. Silencing either side would break the other deployment mode. End-to-end correlation is left out — it would change the log line of every agent, BFF or not.

**Not configurable.** No `BFF_LOG_LEVEL`. The standalone `createConsoleLogger()` is built with a hardcoded `Info` minimum, so gating this behind `Debug` would mean never emitting it there. Embedded deployments already have the knob: `loggerLevel: 'Warn'` on the agent drops the 2xx lines.

## Deliberate consequences

The standalone output changes: one line per request where there were none. Anyone with a log pipeline behind the docker image will see the volume change — including `/health`, which a liveness probe hits continuously. The alternative was an embedded-only access log, which leaves the standalone undebuggable and the two modes divergent for no good reason.

Routine 4xx now logs at `Warn`: an expired token (401), a stale bookmark's origin (403), a bot scan (404). That is the agent's mapping, kept for symmetry.

## Known gap

`BFFHttpServer` in its `AssembledOptions` mode builds its own Koa app and gets no access log. It is exported but unused in this repo — `runCli` goes through the prebuilt callback.

## Tests

18 new tests. The unit suite covers the behaviour — level mapping, status derivation from `status`/`statusCode`/neither, rethrow, base-path resolution, query string exclusion. The integration suite covers the **position**, through the assembled `buildBff` callback: `/health`, a CORS rejection, a 401 from the auth edge, a 404 on an unclaimed path, and the mounted path. The split is on purpose — position is what regresses silently when someone reorders the middleware array.

`agent-bff`: 93 suites / 1774 tests pass. `agent`: 88 suites / 1624 tests pass. Lint clean.

> [!NOTE]
> Fixes PRD-1133
>
> ## Problem
>
> The BFF logged nothing per request, in either deployment mode. Everything it logged was assembly-time or exceptional: configuration gaps, a cache invalidation, an unhandled edge error.
>
> Found while debugging the embedded BFF against a real agent. Calling `/bff/agent/openapi.json` and `/bff/health` produced **no output at all** — correct for the code as written, and indistinguishable from "my mount is broken":
>
> - neither route reaches the agent, so the agent's own request logger never fires
> - and the BFF said nothing itself
>
> A data call that *does* reach the agent logs one line, but from the agent and with no marker — indistinguishable from Forest UI traffic. The standalone docker image had the same blind spot.
>
> ## What this does
>
> Adds an access-log middleware at the head of the chain assembled by `buildBff`, so it lands in both deployment modes at once.
>
> - `[status] METHOD path - Xms` — the shape the agent's own `Logger` route emits, so both layers read the same way. Embedded, it comes out `[BFF]`-prefixed for free through the existing logger adapter.
> - `Info` under 400, `Warn` from 400, `Error` from 500.
> - At the head of the chain, so it also covers what never reaches the agent or the agent-scoped error middleware: `/health`, `/docs`, `/oauth/*`, a 401 from the auth edge, a rejected origin, `openapi_disabled`, a 404 on an unclaimed path.
> - No request body, no `Authorization` header, no `X-Forest-Bff-Key`.
>
> ## Decisions worth reviewing
>
> **Message only, no structured fields.** The standalone sink (`console-logger`) can flatten a context into JSON fields, and a log pipeline would rather query `status` than regex a string. Chosen against: it keeps the format identical to the agent's, and it keeps the change inside one package — passing a context would make `formatLog` in `packages/agent` re-serialize it behind the message, so every embedded line would carry the same values twice.
>
> **Status read off the exception, then rethrown.** The middleware sits above every error handler the BFF has, so at the moment it logs, `ctx.status` is still Koa's default 404 and the line would name a status nobody was ever answered. Copying the agent's `Logger` verbatim was the other option — it catches without rethrowing, which here would have turned a 500 into a bare 404 on the public surface.
>
> **No cause on the line, and a guarded logger.** The line carries a status and nothing else. Passing a `cause` through the port's third argument is what `formatLog` in `packages/agent` re-serializes behind the message, so an embedded deployment would carry the same values twice on every line that has one — and silencing it means a marker shared between two separately released packages. The detail is not lost where it matters: `createErrorMiddleware` already logs `cause` for the whole `/agent` surface. Outside it — a body-parser rejection on an OAuth POST is the only route today — the stack reaches stderr through Koa's handler, as before this middleware existed. The real fix is that `createErrorMiddleware` is `agentScoped` and should not be, which changes OAuth's error contract and deserves its own ticket.
>
> The `logger` call itself is wrapped: it comes from the host, nothing in the port forbids it from throwing, and this middleware runs on every route. Unguarded it would answer 500 to a request the route served, and on the error path it would replace the exception being reported with its own.
>
> **No query string, ever.** Nothing routed today needs it: OAuth's sensitive exchange is a POST body, and the data routes read their filters and `search` from the body too. Logging it would be safe *today* and would leak the first time a route accepts something sensitive there, with nobody revisiting this line.
>
> **Path named as the caller asked for it.** Embedded, `req.url` is rewritten before the BFF sees it, so the BFF's own view is `/agent/openapi.json` for a client that called `/bff/agent/openapi.json` — an operator grepping `/bff/` would still find nothing. Uses the existing `emittedBaseOf` helper, which also prefers the prefix a host actually stripped over the configured one (`/api/bff/health`).
>
> **Two lines per data call, embedded.** One from the BFF, one from the agent, not correlated. Kept deliberately: the paths and timings differ, and the delta is the BFF's own cost. Silencing either side would break the other deployment mode. End-to-end correlation is left out — it would change the log line of every agent, BFF or not.
>
> **Not configurable.** No `BFF_LOG_LEVEL`. The standalone `createConsoleLogger()` is built with a hardcoded `Info` minimum, so gating this behind `Debug` would mean never emitting it there. Embedded deployments already have the knob: `loggerLevel: 'Warn'` on the agent drops the 2xx lines.
>
> ## Deliberate consequences
>
> The standalone output changes: one line per request where there were none. Anyone with a log pipeline behind the docker image will see the volume change — including `/health`, which a liveness probe hits continuously. The alternative was an embedded-only access log, which leaves the standalone undebuggable and the two modes divergent for no good reason.
>
> Routine 4xx now logs at `Warn`: an expired token (401), a stale bookmark's origin (403), a bot scan (404). That is the agent's mapping, kept for symmetry.
>
> ## Known gap
>
> `BFFHttpServer` in its `AssembledOptions` mode builds its own Koa app and gets no access log. It is exported but unused in this repo — `runCli` goes through the prebuilt callback.
>
> ## Tests
>
> 18 new tests. The unit suite covers the behaviour — level mapping, status derivation from `status`/`statusCode`/neither, rethrow, base-path resolution, query string exclusion. The integration suite covers the **position**, through the assembled `buildBff` callback: `/health`, a CORS rejection, a 401 from the auth edge, a 404 on an unclaimed path, and the mounted path. The split is on purpose — position is what regresses silently when someone reorders the middleware array.
>
> `agent-bff`: 93 suites / 1774 tests pass. `agent`: 88 suites / 1624 tests pass. Lint clean.
>
> 
> #### Changes since #1891 opened
>
> - Wrapped logger invocations in try/catch blocks within the `createAccessLogMiddleware` Koa middleware [6e036b8]
> - Added test coverage for exception handling during access logging [6e036b8]
> 
>

### DIFF
--- a/packages/agent-bff/src/http/access-log-middleware.ts
+++ b/packages/agent-bff/src/http/access-log-middleware.ts
@@ -43,6 +43,12 @@ function statusOf(error: unknown): number {
  *
  * The query string is deliberately absent. Nothing routed today needs it, and a line that carries it
  * would leak whatever a future route accepts there without anyone revisiting this decision.
+ *
+ * Carries no cause either, so that a host logger which cannot take structured fields still reads as
+ * one line per request. The detail is not lost where it matters: `createErrorMiddleware` already
+ * reports it for everything under `/agent`. Outside it — a body parser rejection on an OAuth POST is
+ * the only route there today — the stack reaches stderr through Koa's own handler, as it did before
+ * this middleware existed.
  */
 export default function createAccessLogMiddleware({
   logger,
@@ -54,7 +60,14 @@ export default function createAccessLogMiddleware({
     const write = (status: number) => {
       const path = `${emittedBaseOf(ctx, basePath)}${ctx.path}`;
 
-      logger(levelOf(status), `[${status}] ${ctx.method} ${path} - ${Date.now() - startedAt}ms`);
+      try {
+        logger(levelOf(status), `[${status}] ${ctx.method} ${path} - ${Date.now() - startedAt}ms`);
+      } catch {
+        // The logger belongs to the host and nothing in the port forbids it from throwing. Left
+        // unguarded, a throw here would answer 500 to a request the route served, and on the error
+        // path it would replace the exception being reported with its own. There is nowhere to
+        // report this: the only sink available is the logger that just failed.
+      }
     };
 
     try {
--- a/packages/agent-bff/src/http/access-log-middleware.ts
+++ b/packages/agent-bff/src/http/access-log-middleware.ts
@@ -1,0 +1,70 @@
+import type { Logger, LoggerLevel } from '../ports/logger-port';
+import type { Middleware } from 'koa';
+
+import { emittedBaseOf } from '../base-path';
+
+export interface AccessLogMiddlewareOptions {
+  logger: Logger;
+  /** Prefix the host serves this BFF under, so the line names a url the caller can actually call. */
+  basePath: string;
+}
+
+const CLIENT_ERROR = 400;
+const SERVER_ERROR = 500;
+
+function levelOf(status: number): LoggerLevel {
+  if (status >= SERVER_ERROR) return 'Error';
+  if (status >= CLIENT_ERROR) return 'Warn';
+
+  return 'Info';
+}
+
+/**
+ * The status the caller will see, read off an exception rather than off the response: this
+ * middleware sits above every error handler the BFF has, so at the moment it logs, `ctx.status` is
+ * still the 404 Koa defaults to and would name a status nobody was ever answered.
+ */
+function statusOf(error: unknown): number {
+  if (typeof error !== 'object' || error === null) return SERVER_ERROR;
+
+  const { status, statusCode } = error as { status?: unknown; statusCode?: unknown };
+  const value = typeof status === 'number' ? status : statusCode;
+
+  return typeof value === 'number' ? value : SERVER_ERROR;
+}
+
+/**
+ * One line per request, in the shape the agent's own request logger emits (`[200] GET /path - 12ms`)
+ * so both layers read the same way in an embedded deployment.
+ *
+ * Belongs at the head of the chain: everything below it — `/health`, `/docs`, `/oauth/*`, a 401 from
+ * the auth edge, a rejected origin — either never reaches the agent or never reaches the
+ * agent-scoped error middleware, and is exactly the traffic that used to leave no trace at all.
+ *
+ * The query string is deliberately absent. Nothing routed today needs it, and a line that carries it
+ * would leak whatever a future route accepts there without anyone revisiting this decision.
+ */
+export default function createAccessLogMiddleware({
+  logger,
+  basePath,
+}: AccessLogMiddlewareOptions): Middleware {
+  return async function accessLog(ctx, next) {
+    const startedAt = Date.now();
+
+    const write = (status: number) => {
+      const path = `${emittedBaseOf(ctx, basePath)}${ctx.path}`;
+
+      logger(levelOf(status), `[${status}] ${ctx.method} ${path} - ${Date.now() - startedAt}ms`);
+    };
+
+    try {
+      await next();
+    } catch (error) {
+      write(statusOf(error));
+
+      throw error;
+    }
+
+    write(ctx.response.status);
+  };
+}
--- a/packages/agent-bff/src/build-bff.ts
+++ b/packages/agent-bff/src/build-bff.ts
@@ -32,6 +32,7 @@ import createCorsMiddleware from './cors/cors-middleware';
 import createPerKeyOriginMiddleware from './cors/per-key-origin';
 import createDataRoutesMiddleware from './data/data-routes-middleware';
 import createDocsRoutes from './docs/docs-routes';
+import createAccessLogMiddleware from './http/access-log-middleware';
 import { unauthorized, unsupportedMediaType } from './http/bff-http-error';
 import BODY_LIMIT, { AI_BODY_LIMIT } from './http/body-limit';
 import createErrorMiddleware from './http/error-middleware';
@@ -534,6 +535,7 @@ export default async function buildBff({
   const agentJsonOnlyGuard = hasAgentEdge ? [agentScoped(createJsonOnlyGuard())] : [];
 
   const middlewares = [
+    createAccessLogMiddleware({ logger, basePath: mountPath }),
     createVersionHeaderMiddleware(version),
     createHealthRoute({
       version,
--- a/packages/agent-bff/test/build-bff-access-log.test.ts
+++ b/packages/agent-bff/test/build-bff-access-log.test.ts
@@ -1,0 +1,94 @@
+import type { Logger, LoggerLevel } from '../src/ports/logger-port';
+
+import request from 'supertest';
+
+import buildBff from '../src/build-bff';
+import { restoreFetchAfterEach, stubEnvironmentIdFetch } from './helpers/fetch-stub';
+import { parseConfig } from '../src/config/env-config';
+
+const VALID_ENV = {
+  FOREST_AUTH_SECRET: 'auth-secret',
+  FOREST_ENV_SECRET: 'env-secret',
+  FOREST_SERVER_URL: 'https://api.forestadmin.com',
+  FOREST_APP_URL: 'https://app.forestadmin.com',
+  AGENT_URL: 'https://agent.example.com',
+  BFF_TOKEN_ENCRYPTION_KEY: Buffer.alloc(32).toString('base64'),
+  BFF_ALLOWED_ORIGINS: 'https://allowed.example.com',
+} satisfies NodeJS.ProcessEnv;
+
+type Line = [LoggerLevel, string];
+
+async function buildWithSpy(basePath?: string) {
+  const lines: Line[] = [];
+  const logger: Logger = (level, message) => lines.push([level, message]);
+  const { callback } = await buildBff({ config: parseConfig(VALID_ENV), logger, basePath });
+
+  lines.length = 0;
+
+  return { callback, lines };
+}
+
+function accessLines(lines: Line[]): Line[] {
+  return lines.filter(([, message]) => /^\[\d{3}\] /.test(message));
+}
+
+describe('buildBff access log', () => {
+  restoreFetchAfterEach();
+
+  beforeEach(() => {
+    stubEnvironmentIdFetch();
+  });
+
+  it('logs /health, which no other layer ever sees', async () => {
+    const { callback, lines } = await buildWithSpy();
+
+    await request(callback).get('/health');
+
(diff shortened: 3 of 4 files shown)


packages/agent-bff/src/http/access-log-middleware.ts
```diff
@@ -0,0 +1,70 @@
...
+  logger,
+  basePath,
+}: AccessLogMiddlewareOptions): Middleware {
+  return async function accessLog(ctx, next) {
+    const startedAt = Date.now();
+
+    const write = (status: number) => {
+      const path = `${emittedBaseOf(ctx, basePath)}${ctx.path}`;
+
+      logger(levelOf(status), `[${status}] ${ctx.method} ${path} - ${Date.now() - startedAt}ms`);
```
Tonours: Claude Opus 5 (claude-opus-5): **Should fix**: a host logger that throws turns a served 200 into a bare 500 on every route, and swallows the original exception on the error path.

`access-log-middleware.ts:57` logs unguarded, and `write` is called from both branches.

On the success path a throw there makes Koa answer 500 instead of the status the route computed.

On the error path it is worse: `write(statusOf(error))` at `access-log-middleware.ts:63` runs *before* `throw error` at `access-log-middleware.ts:65`, so a logger that throws replaces the original exception with its own and the cause is lost — on the middleware whose reason to exist is instrumenting that path.

Nothing in `ports/logger-port.ts` states the port must not throw, and the logger is host-supplied.

A way out: one try/catch around the `logger` call inside `write`, which covers both branches at a single site.
</details>

1. Embed the BFF with a `logger` that throws on its first call.
2. Call `GET /bff/health` — the answer is a bare 500, not the 200 the health route computed.
3. Make an agent-scoped route throw a `TypeError`, then read what propagates: the logger error, not the `TypeError`.
</details>

packages/agent-bff/src/http/access-log-middleware.ts
```diff
@@ -0,0 +1,70 @@
...
+    };
+
+    try {
+      await next();
+    } catch (error) {
+      write(statusOf(error));
```
Tonours, line 76: Claude Opus 5 (claude-opus-5): **Should fix**: an unhandled OAuth or docs error reaches the host log sink as a status line with no cause.

`access-log-middleware.ts:63` logs no cause, and `createErrorMiddleware` is `agentScoped` in `build-bff.ts`, so the stack only reaches stderr.

A way out: pass a `cause` field through the `Logger` port's third argument, as `error-middleware.ts:52` does.
</details>

1. Embed the BFF with a logger shipping to a structured sink.
2. Make the OAuth token route throw a `TypeError`.
3. The sink holds `[500] POST /oauth/token - 4ms` and no cause.
</details>

Tonours (commented):
**Spec (`PRD-1133`)**: conforms. The middleware is first in the chain `buildBff` assembles, and the line shape, the level mapping and the header and body exclusions all match the ticket.

packages/agent-bff/src/http/access-log-middleware.ts
nbouliol: Fixed in 6e036b8dd. One `try/catch` around the `logger` call inside `write`, so it covers both branches at a single site as suggested.

Silent on purpose, and commented as such: the only sink available to report a logging failure is the logger that just failed.

Two tests, both verified to fail without the guard — one asserting a served 200 survives a throwing logger, one asserting the original `TypeError` is what propagates on the error path rather than the logger's own error.

packages/agent-bff/src/http/access-log-middleware.ts
nbouliol: Your read of the code is right — outside `/agent` the stack only reaches stderr through Koa's own handler. Keeping it as is, deliberately, and I've recorded the reasoning in the file since review had to reconstruct it (6e036b8dd).

Passing a `cause` through the port's third argument is the one thing this middleware cannot do cheaply. `formatLog` in `packages/agent` re-serializes any non-empty context behind the message, so an embedded deployment would get `[BFF] [500] POST /oauth/token - 4ms {"cause":"..."}` on every logged line that carries one — and teaching it to stay quiet means a shared marker between two separately released packages. That is a real change to `packages/agent`, for a path whose only route today is a body-parser rejection on an OAuth POST.

Worth noting the detail is not lost where it matters: `createErrorMiddleware` already logs `cause` for everything under `/agent`, which is the whole data and action surface.

The honest fix is the one this scope refuses: `createErrorMiddleware` is `agentScoped` and should not be — a global error handler would give `/oauth/*` and `/docs` a structured body and a logged cause. It changes OAuth's error shape, which has its own contract, so it deserves its own ticket rather than riding along here.

packages/agent-bff/src/http/access-log-middleware.ts
Tonours: Closed, a throwing host logger can no longer change the status or replace the exception.

The predicate was a host logger whose throw changes the response status or replaces the propagating exception.

The only logger call now sits inside the try at `access-log-middleware.ts:63`, and `emittedBaseOf` above it is total.

Both new tests pass at `6e036b8dd` and both fail against the pre-fix source when run locally.
</details>

packages/agent-bff/src/http/access-log-middleware.ts
Tonours: Accepted, the finding asked for a fix whose cost on embedded deployments I had not priced.

The predicate was an unhandled error outside `/agent` reaching the host sink as a status line with no cause.

`formatLog` at `packages/agent/src/embedded-bff.ts:13` appends any non-empty context after the message, and line 99 wires the port through it.

Nothing is pinned and nothing is owed, since the observation now lives at `access-log-middleware.ts:47` rather than in code.
</details>

Tonours (approved):
Both findings from the earlier review are settled: the logger guard is in with two tests that fail on the pre-fix source, and the missing cause is accepted as argued, with the reasoning now recorded in the file.

The architectural fix you name, taking `createErrorMiddleware` out of `agentScoped`, is worth its own ticket. Not a blocker here.

PR #1893, opened 2026-09-10
feat(agent-bff): synthesize capabilities for legacy lianas

Tonours:
Top of a three-PR stack: #1894 then #1895 merge first, and this one carries the last three commits. Review those three; the parents' diffs belong to their own PRs.

## What

The BFF serves constrained reads — filter, sort, projection — in front of a v1 liana (`forest-express-sequelize`, `forest-express-mongoose`, `forest-rails`). It decides from the liana the schema announces. `BFF_LEGACY_CAPABILITIES` is gone.

## Why

v1 lianas never served `POST /forest/_internal/capabilities`. Every constrained read died on that 404 and the collection was unusable. The first shape of this feature asked the operator to declare the generation with a flag, for information the BFF already receives in the schema's `meta.liana`.

## How

`forestadmin-client` stops dropping the schema's `meta`, so the liana name reaches the BFF. On a 404 from the capabilities route, the fetcher matches `meta.liana` against the three legacy names and synthesizes from the apimap: the operator set that liana implements per column type, plus the fields the schema marks filterable and sortable. Any other liana keeps throwing.

`includes_all` is not published. Measured on a scalar column it answers 500 with a leaking SQL fragment in Express and 422 in Rails, so advertising it would advertise a crash.

The third commit is a refactor the synthesis needs: three implementations converted a PascalCase condition tree to the snake_case the wire expects, each with its own idea of what to do with an operand-less leaf. They agreed on all 43 operators when measured, which is luck, not a contract. The walk now lives in one exported place.

## Scope and safety

No agent code changes, v1 or v2. Nothing in the wire format is new to either generation: v2 already parses snake_case operators and accepts `value: null` on operand-less leaves.

`SchemaFetcher` now returns `{ collections, meta }` instead of a bare array. The in-process dispatcher added by #1876 is the one other consumer and its test follows the new shape.

## How to test

`yarn workspace @forestadmin/agent-bff test` — 1822 tests, 95 suites. Also measured on five running agents: `forest-express-sequelize`, `forest-express-mongoose`, `forest-rails`, a v2 node agent and a v2 ruby agent, with the capabilities route absent on the three v1 ones. Constrained reads answer 200 on all five. A cold OpenAPI unfold is pinned on a 269-collection schema, the largest measured on the fleet.

### General

### Security


packages/agent-bff/src/read-model/synthesize-capabilities.ts
```diff
@@ -0,0 +1,147 @@
...
+      return field.relationship === 'BelongsTo' ? [{ name: field.field, type: MANY_TO_ONE }] : [];
+    }
+
+    const entry: CapabilitiesResult['fields'][number] = {
+      name: field.field,
+      type: field.type,
+      operators: operatorsFor(field, collection.name, logger),
+    };
+
+    return [field.isSortable === false ? { ...entry, sortable: false } : entry];
```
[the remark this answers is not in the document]
Tonours: Measured rather than argued: `GET /liana/forest-schema` on a live forest-rails 9.21.0 environment serves camelCase field keys — `isFilterable`, `isSortable`, `isPrimaryKey` — and 6 of its fields carry `isFilterable: false`. The SaaS normalises the liana's snake_case apimap before serving it, and `ForestSchemaField` in forestadmin-client types it that way, so reading `isFilterable` is correct for what the BFF actually receives.

The finding did surface a real defect next to it, though: the fixture in this PR was a raw apimap in snake_case, which made every filterability assertion over it vacuous. It is now normalised to the shape the SaaS serves, and two tests assert the denials are honoured — one of which caught a genuine miss on to-one relations, fixed in 8760afb80.

packages/forestadmin-client/src/schema/index.ts
```diff
@@ -45,6 +49,10 @@ export default class SchemaService {
     return this.forestAdminServerInterface.getSchema(toHttpOptions(this.options));
   }
 
+  async getSchemaWithMeta(): Promise<ForestSchemaWithMeta> {
+    return this.forestAdminServerInterface.getSchemaWithMeta(toHttpOptions(this.options));
```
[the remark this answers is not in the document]
Tonours: Fixed in fb639f594. The guard now mirrors `WorkflowsService.resolveTransportMethod` in the same package (`src/workflows/index.ts:31`), which already resolves these optional transport members and throws "The configured Forest server transport does not support <name>." — same message shape here, plus a test asserting it on a transport with `getSchemaWithMeta: undefined`.

Worth noting the scope: every member of `ForestAdminServerInterface` is optional and the package compiles without `strictNullChecks`, so `getSchema`, `postSchema` and `checkSchemaHash` a few lines above have the same unguarded call. Only the method this PR adds is fixed here; the siblings are pre-existing and belong to their own change.

packages/agent-bff/src/openapi/unfolded-document.ts
```diff
@@ -38,7 +38,12 @@ export default async function buildUnfoldedDocument(
   const unfolding = await collectUnfolding({
     readModel,
     store: source.store,
-    capabilitiesFetcher: createAgentCapabilitiesFetcher({ transport: source.transport, token }),
+    capabilitiesFetcher: createAgentCapabilitiesFetcher({
```
nbouliol, line 41: **Applies to:** packages/agent-bff/src/openapi/unfolded-paths.ts:263 (not in this diff)
Claude Opus 5 (claude-opus-5): **Should fix**: the OpenAPI document lists a field in the sort enum that the BFF then answers 422 field_not_sortable on.

`unfolded-paths.ts:263` builds the sort enum from `projectable`, which `collect-unfolding.ts:136` fills without the new `sortable` flag.

A way out: carry `sortable` into `CollectionFields` and filter that enum on `sortable !== false`.
</details>

1. Run the BFF over a forest-rails agent whose apimap marks a field isSortable false.
2. Fetch the unfolded OpenAPI document.
3. The SortClause enum for that collection lists the field.
4. Send a list request sorting on it.
5. The BFF answers 422 field_not_sortable.
</details>

packages/agent-bff/src/read-model/synthesize-capabilities.ts
```diff
@@ -0,0 +1,154 @@
...
+ */
+const OPERATORS_BY_COLUMN_TYPE = new Map<string, readonly Operator[]>(
+  Object.entries(allowedOperatorsForColumnType),
+);
+
+function primitiveNameOf(type: FieldType): string | undefined {
+  const normalized = normalizeFieldType(type);
+
+  if (typeof normalized === 'string') return normalized;
+  if (Array.isArray(normalized) && typeof normalized[0] === 'string') return normalized[0];
```
nbouliol: Claude Opus 5 (claude-opus-5): **Should fix**: an array column advertises the element type's scalar operators, so filtering one answers 503 agent_unavailable to the user.

`primitiveNameOf` unwraps `['Number']` to `Number`, so line 112 publishes the scalar table including `greater_than` on an array column.

A way out: return `undefined` from `primitiveNameOf` for an array, taking the existing unknown-type path.
</details>

1. Point the BFF at a forest-rails agent publishing a Postgres array column as filterable.
2. Read that collection's capabilities.
3. They list greater_than.
4. Filter on that column with greater_than.
5. The agent raises and the BFF answers 503.
</details>

packages/agent-client/src/filter-wire-format.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function isBranch(node: unknown): node is { aggregator?: string; conditions: unknown[] } {
+  if (typeof node !== 'object' || node === null) return false;
+
+  const { aggregator, conditions } = node as { aggregator?: unknown; conditions?: unknown };
+
+  // The aggregator is lowercased below, so a non-string one would throw a TypeError from inside the
+  // walk instead of being reported as the malformed filter it is.
+  return Array.isArray(conditions) && (aggregator === undefined || typeof aggregator === 'string');
```
nbouliol, line 22: Claude Opus 5 (claude-opus-5): **Violates conventions**: a caller sending a non-string aggregator gets 503 agent_unavailable where a 400 was available.

`capabilities-validator.ts:31` accepts any `conditions` array as a branch, so this guard refuses the node and line 69 forwards it verbatim.

A way out: reject a non-string `aggregator` in that `isBranch`, and cover both branches per `testing.md#Cover error and edge paths`.
</details>

1. POST /agent/v1/Article/list with filter {"aggregator": 5, "conditions": [{"field": "title", "operator": "StartsWith"}]}.
2. The BFF validates it green.
3. It forwards aggregator 5 with operator StartsWith.
4. The agent rejects the filter.
5. The BFF answers 503 agent_unavailable.
</details>

nbouliol (commented):
**Spec**: no Linear ticket is linked on this PR, so the functional check was limited to the PR description. The change matches it.

Claude Opus 5 (claude-opus-5): **Should fix**: no ADR records that the BFF decides an agent's capabilities from the liana name it self-reports.

`agent-capabilities-fetcher.ts:67` reads `meta.liana` from a cached schema and treats three self-reported names as a capability set.

A way out: run `/adr` and record the decision against the capability negotiation it replaced.
</details>

1. Read docs/adr in agent-nodejs.
2. It holds one ADR, about case-insensitive decisions.
3. Search the org ADR corpus for capabilities, schema, liana, operators and filter.
4. None covers this decision.
5. Read the diff, which adds no ADR file.
</details>

packages/agent-bff/src/openapi/unfolded-document.ts
Tonours: Fixed in b255c7728. ProjectableField now carries the sortable denial through collect-unfolding.ts, and fieldRefs builds the sort enum from the fields that keep it: SortableFields_<key> when something is denied, the projectable enum itself otherwise, so a v2 document gains no extra component. A collection where every field denies the sort falls back to the shared free-form SortClause.

Covered in openapi-unfolded.test.ts (the denied field is absent from the sort enum but still projectable, plus the all-denied fallback) and in collect-unfolding.test.ts (the flag survives collection).

packages/agent-bff/src/read-model/synthesize-capabilities.ts
Tonours: Fixed in b255c7728. primitiveNameOf no longer unwraps an array, so an array column takes the existing unknown-type path: no operator published, one Warn logged. That also matches v2 — allowedOperatorsForColumnType is keyed by primitive names only, so the agent publishes nothing for such a column either.

Covered for both spellings the apimap uses, NumberList and ["Number"].

packages/agent-client/src/filter-wire-format.ts
Tonours: Fixed in b255c7728, though not where the comment pointed.

Rejecting the node in capabilities-validator's isBranch would have made it worse: collectLeaves would then read the branch as neither leaf nor branch, skip it, and every leaf underneath would go unvalidated before the filter was forwarded anyway. The guard belongs to the structural pass, so it went into agent-query's assertFilterNode next to the stray-key and depth checks, where a rejection is a 400 invalid_request before any agent call.

It also covers the sibling case with the same failure mode: "xor" is a string and passes every shape check, reaches the agent and comes back 503 just like 5 does.

One deliberate narrowing: the comparison is case-insensitive although the document enumerates And and Or only. toWireFilter lowercases the aggregator on the way out, so "and" already reaches the agent as the "and" it parses — rejecting it would break a caller nothing else refuses today. Say the word if you would rather hold the runtime to the document exactly.

Covered on both branches: 5, "xor" and null rejected with the 400; And, Or, and, or accepted; a branch with no aggregator still accepted, as the document says.

packages/agent-bff/src/openapi/unfolded-document.ts
nbouliol: **Closed**: the sort enum now drops a field the apimap denies, so the document stops offering a sort the runtime refuses.

Predicate: a name in `SortClause_<key>` whose capability field carries `sortable: false`.

`unfolded-paths.ts:250` builds the enum from the filtered list, and `record-schemas.ts:121` groups by name alone.

Pinned by `openapi-unfolded.test.ts`, which asserts `SortableFields_users` enumerates `id` alone.
</details>

packages/agent-bff/src/read-model/synthesize-capabilities.ts
nbouliol: **Closed**: an array column publishes no operator now, so nothing advertises a filter the liana raises on.

Predicate: a field whose apimap type is an array or a List name publishing any operator.

`primitiveNameOf` returns a name only for a plain string type, so every array takes the unknown-type path.

Pinned by `synthesize-capabilities.test.ts`, which asserts an empty operator list for both spellings.
</details>

packages/agent-client/src/filter-wire-format.ts
nbouliol: **Closed**: a branch aggregator that is not And or Or is now a 400, and keep the case-insensitive comparison.

Predicate: a branch the BFF forwards whose aggregator is present and not one the agent parses.

`assertAggregator` sits in `assertFilterNode`, which all four request parsers reach, and your `collectLeaves` point holds.

Pinned by `agent-query.test.ts`, which rejects 5, xor and null and accepts And, Or, and and or.
</details>

nbouliol:
**Reopened**: no ADR records deciding an agent's capabilities from the liana name it self-reports.

Predicate: an accepted ADR covering the decision to read `meta.liana` as a capability set.

`docs/adr` at `1d9c219` holds one file, about case-insensitive decisions, and the head adds none.

No reply addressed the finding, so the rationale still lives only in file comments.
</details>

nbouliol:
**Closed**: an accepted ADR now records deciding a v1 agent's capabilities from the liana name the schema carries.

Predicate: an accepted ADR covering the decision to read `meta.liana` as a capability set.

The ADR added in `2ac05bf` is `status: accepted` and names the self-reported name as an accepted trade-off.

Nothing pins a decision record, so the ADR file itself is where the absence is accounted for.
</details>

nbouliol (approved):
Four findings raised, four closed: the sort enum no longer advertises a sort the runtime refuses, an array column publishes no operator, a non-enumerated aggregator is a 400 before any agent call, and the liana-name decision is recorded in an accepted ADR. Three carry a regression test that fails on the pre-fix code.

Two nits left deliberately unaddressed, neither worth holding this up: `schemas.ts:521` still describes `details` as `{ field }` on `unknown_field` and `field_not_filterable` only, though `field_not_sortable` carries one too, and `SchemaCache.meta` has no production caller since `getPayload()` replaced it.

`LLM Integration Tests (ai-proxy)` is red but failed identically before the fixes and this PR touches no ai-proxy file. Merge order still stands: #1894, then #1895.

PR #1894, opened 2026-09-10
fix(agent-bff): serve legacy lianas, and stop leaking and overpromising

Tonours:
> **Read this first.** This PR was approved on 11 Sep, when it was the 4xx body fix alone — 6 files. #1893 and #1895 have since been squash-merged into this branch, which now carries 67 files. **The approval predates roughly 90% of the diff**, and nothing is stacked on this any more: it is the whole change.

## What

Four things, all of them about the BFF telling the truth to its caller.

**1. An agent 4xx no longer leaks the agent's body** (`agent-error-mapper.ts`, `agent-client/domains/action.ts`). Whatever the agent happened to print was forwarded verbatim — a framework stack, an HTML error page, a raw SQL fragment. The mapper now forwards only what it recognises, keeps the status and drops the rest.

**2. A legacy liana gets synthesized capabilities** (#1893). `forest-express-sequelize`, `forest-express-mongoose` and `forest_liana` never served `POST /forest/_internal/capabilities`, so every constrained read died on that 404. The BFF now matches `meta.liana` against those three names and synthesizes the answer from the apimap the liana already pushed. Any other liana keeps throwing, so a v2 agent is never silently downgraded.

**3. A collection whose schema declares no primary key gets one derived from its `id` field** (#1895). `forest_liana` only publishes `isPrimaryKey` from 9.17.6 (2026-06-04); below that line every collection is keyless and a plain list answers `500 mapping_error`. Of the 125 `forest-rails` environments active in the last 120 days, the twenty oldest run between 2.14.6 and 8.3.2, several in production.

**4. The OpenAPI document stops promising what the runtime refuses.** Sort, filter and projection are each capped when the collection's capabilities were read and name no usable field, and left unconstrained when they could not be read at all — an unknown field set is not an empty one.

## Decisions worth a look

Two guesses are made, and both are published as guesses rather than hidden:

- **Arity.** A keyless collection whose real key is composite packs `tenant|42`. Splitting that against one derived key would 500 the whole list, so a derived key is flagged and passed through whole.
- **Name.** 12 of the 269 collections in the measured schema declare no `id` field at all, so `__forest.primaryKey` carries the opaque id under a name that is not a column. `ForestRecordMeta` states this as the one exception to its per-column promise, and the context flags a derived key `isPrimaryKeyDerived` so a client reads identities with it rather than filtering by it.

Both are recorded in `docs/adr/`.

## Scope and safety

- Every fallback is keyed on the **shape of the schema**, never on a version string, so a collection that declares a key keeps it and every v2 agent is untouched. The fallbacks only fire where the alternative is a 500.
- The 4xx change narrows what reaches the client. No status code moves.
- No wire shape changes: `parentId` stays the same union, only its description hedges.

## How to test

```
```

1867 tests, 95 suites, green. Lint clean. CI is green except `LLM Integration Tests (ai-proxy)`, which is `continue-on-error: true` (`build.yml:127`), reaches the real OpenAI/Anthropic APIs and touches no file in this diff.

### General

### Security

### DIFF
--- a/packages/agent-client/src/domains/action.ts
+++ b/packages/agent-client/src/domains/action.ts
@@ -85,7 +85,7 @@ function toActionError(error: unknown): unknown {
   if (error.status === 400 || error.status === 422) {
     return new ActionFormValidationError(
       detail ?? 'The action form values were rejected.',
-      typeof body.html === 'string' ? body.html : undefined,
+      detail !== undefined && typeof body.html === 'string' ? body.html : undefined,
       detail === undefined ? error : undefined,
     );
   }
--- a/packages/agent-client/src/domains/action.ts
+++ b/packages/agent-client/src/domains/action.ts
@@ -40,15 +40,28 @@ type ActionErrorBody = {
   data?: { roleIdsAllowedToApprove?: number[] };
 };
 
+function firstNonEmptyString(candidates: unknown[]): string | undefined {
+  for (const candidate of candidates) {
+    if (typeof candidate === 'string' && candidate.trim() !== '') return candidate.trim();
+  }
+
+  return undefined;
+}
+
 // Only a recognised error shape yields a detail. A raw body or responseText is whatever the agent's
 // host emitted -- an HTML error page, a proxy notice, a stack trace -- and it would reach the client
 // verbatim as the action error message, so it is left out and callers use their own wording.
 function extractDetail(error: AgentHttpError): string | undefined {
   const body = (error.body ?? {}) as ActionErrorBody;
   const first = body.errors?.[0];
-  const message = first?.detail || first?.message || first?.title || body.error || body.message;
 
-  return message?.trim() || undefined;
+  return firstNonEmptyString([
+    first?.detail,
+    first?.message,
+    first?.title,
+    body.error,
+    body.message,
+  ]);
 }
 
 // Translate the transport AgentHttpError into a semantic action error so callers route on meaning.
@@ -72,7 +85,8 @@ function toActionError(error: unknown): unknown {
   if (error.status === 400 || error.status === 422) {
     return new ActionFormValidationError(
       detail ?? 'The action form values were rejected.',
-      body.html,
+      typeof body.html === 'string' ? body.html : undefined,
+      detail === undefined ? error : undefined,
     );
   }
 
--- a/packages/agent-client/src/domains/action.ts
+++ b/packages/agent-client/src/domains/action.ts
@@ -40,17 +40,13 @@ type ActionErrorBody = {
   data?: { roleIdsAllowedToApprove?: number[] };
 };
 
+// Only a recognised error shape yields a detail. A raw body or responseText is whatever the agent's
+// host emitted -- an HTML error page, a proxy notice, a stack trace -- and it would reach the client
+// verbatim as the action error message, so it is left out and callers use their own wording.
 function extractDetail(error: AgentHttpError): string | undefined {
   const body = (error.body ?? {}) as ActionErrorBody;
   const first = body.errors?.[0];
-  const message =
-    first?.detail ||
-    first?.message ||
-    first?.title ||
-    body.error ||
-    body.message ||
-    (typeof error.body === 'string' ? error.body : undefined) ||
-    error.responseText;
+  const message = first?.detail || first?.message || first?.title || body.error || body.message;
 
   return message?.trim() || undefined;
 }
--- a/packages/agent-bff/src/action/action-routes-middleware.ts
+++ b/packages/agent-bff/src/action/action-routes-middleware.ts
@@ -155,6 +155,15 @@ async function handleExecute({
     }
 
     if (error instanceof ActionFormValidationError) {
+      const { unstructuredCause } = error;
+
+      if (unstructuredCause) {
+        logger('Warn', 'Agent action 4xx carried no structured error; client message is generic', {
+          status: unstructuredCause.status,
+          cause: unstructuredCause.responseText ?? unstructuredCause.body,
+        });
+      }
+
       const html = sanitizeActionHtml(error.html, logger);
 
       throw actionError(error.message, html === null ? undefined : { html });
--- a/packages/agent-bff/src/http/agent-error-mapper.ts
+++ b/packages/agent-bff/src/http/agent-error-mapper.ts
@@ -134,15 +134,19 @@ function parseJsonApiFromMessage(error: unknown): AgentJsonApiError | undefined 
   }
 }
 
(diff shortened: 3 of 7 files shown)


nbouliol (commented):
**Spec**: no Linear ticket is linked to this pull request, so the functional check was limited to the PR description.

**Applies to:** `packages/agent-bff/src/action/action-routes-middleware.ts:160`
Claude Opus 5 (claude-opus-5): **Should fix**: a third-party UI executing an action still receives the agent's raw 4xx body, such as a proxy HTML page, as the error message.

`action-routes-middleware.ts:160` throws the message `agent-client` extracted from the raw body, skipping the recognition check `mapFlatBody` now applies.

A way out: apply the same recognised-shape check to `ActionFormValidationError.message` before it reaches `actionError`.
</details>

1. Put a proxy in front of the agent that answers an action execute with a 400 HTML page.
2. Call `POST /agent/v1/:collection/actions/:action/execute` through the BFF.
3. Read `error.message` in the response body, which carries the proxy HTML.
</details>

nbouliol:
Claude Opus 5 (claude-opus-5): **Closed**: `extractDetail` dropped its raw body and `responseText` fallbacks, so no unstructured body reaches the action error message.

The predicate is an action 4xx whose body is not a recognised error shape whose content reaches the client as `error.message`.

No input satisfies it now: `action.ts:49` keeps only recognised keys, and `workflow-executor`, the one other consumer, discards the message.

The regression is pinned by two tests in `agent-client/test/domains/action.test.ts` that return the raw body on the pre-fix code.
</details>

packages/agent-client/src/domains/action.ts
```diff
@@ -40,17 +40,13 @@ type ActionErrorBody = {
...
   const first = body.errors?.[0];
-  const message =
-    first?.detail ||
-    first?.message ||
-    first?.title ||
-    body.error ||
-    body.message ||
-    (typeof error.body === 'string' ? error.body : undefined) ||
-    error.responseText;
+  const message = first?.detail || first?.message || first?.title || body.error || body.message;
```
nbouliol: Claude Opus 5 (claude-opus-5): **Should fix**: an operator debugging a failed action has no record of the agent's 4xx body, because nothing logs it or keeps it.

`action.ts:49` drops the body and `errors.ts:22` keeps no `cause`, so the BFF, workflow-executor and mcp-server all lose it.

A way out: carry the `AgentHttpError` as `cause`, which `ApprovalRequestCreationError` already does and `parseAgentError` already walks.
</details>

1. Put a proxy in front of the agent that answers an action execute with a 400 HTML page.
2. Run the action through the BFF.
3. Search the BFF logs for that HTML, which appears nowhere.
</details>

packages/agent-client/src/domains/action.ts
```diff
@@ -40,17 +40,13 @@ type ActionErrorBody = {
...
 
   return message?.trim() || undefined;
```
nbouliol: Claude Opus 5 (claude-opus-5): **Should fix**: a gateway 400 whose body nests an object under `error` reaches the client as a 502 `network_error` instead.

`action.ts:51` calls `trim` on whatever `body.error` held, so a non-string throws a TypeError out of `execute`.

A way out: guard with `typeof`, the way `agent-error-mapper.ts:143` already does on the sibling path.
</details>

1. Put a gateway in front of the agent that answers a 400 with `{"error":{"code":"invalid_form"}}`.
2. Execute an action through the BFF.
3. Read the response, which is a 502 `network_error`.
</details>

nbouliol (commented):
**Spec**: no Linear ticket is linked to this pull request, so the functional check was limited to the PR description.

packages/agent-client/src/domains/action.ts
nbouliol: Claude Opus 5 (claude-opus-5): **Closed**: an unrecognised action body now reaches the BFF logs as `cause`, so the failure is diagnosable again.

The predicate is an unstructured action 4xx whose body no process retains or logs.

No input satisfies it now: `errors.ts:25` keeps it as `unstructuredCause` and `action-routes-middleware.ts:161` logs it.

Pinned by a middleware test asserting the exact warn payload, and a client test, both failing on the pre-fix code.
</details>

packages/agent-client/src/domains/action.ts
nbouliol: Claude Opus 5 (claude-opus-5): **Closed**: `firstNonEmptyString` type-checks each candidate, so no body shape can throw out of `execute` any more.

The predicate is a candidate field whose value is not a string reaching `trim` and throwing.

No input satisfies it now: `action.ts:45` calls `trim` only behind a `typeof candidate === 'string'` guard.

Pinned by a client test sending an object under `error`, which threw a TypeError on the pre-fix code.
</details>

nbouliol (approved):
All review findings are closed, each with a regression test that fails on the pre-fix code. The red `LLM Integration Tests (ai-proxy)` check is unrelated to this diff: it calls the live OpenAI API and `gpt-live-1` now answers 404 on the chat-completions endpoint.

packages/agent-bff/src/data/pack-id.ts
```diff
@@ -7,11 +7,22 @@ export const PACKED_ID_SEPARATOR = '|';
 // The only column type unpacked to a number, mirroring the agent's `IdUtils.unpackId`.
 const NUMBER_COLUMN_TYPE = 'Number';
 
+function toNumberIfWhole(value: string): string | number {
+  const numeric = Number(value);
+
+  return Number.isNaN(numeric) ? value : numeric;
```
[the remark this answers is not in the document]
Tonours: Real defect, fixed in 9b05807 — but with a stricter guard than the suggestion.

You are right on the principle: the derived branch exists precisely to carry the agent id opaque, so a cast that loses anything defeats it. `Number.isSafeInteger` alone still lets through values that round-trip to different characters than `record.id` holds — `1e3` becomes `1000`, `042` becomes `42`. Both are safe integers, and both name an id the record does not carry.

So the test is the round trip itself:

```ts
return Number.isSafeInteger(numeric) && String(numeric) === value ? numeric : value;
```

`isSafeInteger` covers your two cases (`9007199254740993` and the non-finite ones, which also stops the `null` serialization), and the string comparison covers the spelling ones. Renamed `toNumberIfLossless`, since "whole" no longer describes what it checks.

Unchanged: a genuinely numeric id still types (`42` -> `42`), and the declared-key path is untouched — it keeps its own cast and its mapping errors.

Tests: five cases, one per way the cast can lose (past the safe range, non-finite, exponent form, leading zero, fractional). All five fail on the previous implementation; 1873 pass, lint 0 errors.

PR #1895, opened 2026-09-10
fix(agent-bff): derive a primary key when the schema declares none

Tonours:
Stacked on #1894, which merges first. Review only the last commit; the parent's diff belongs to that PR.

## What

A collection whose schema declares no primary key gets one derived from its `id` field.

## Why

`forest_liana` publishes `isPrimaryKey` only from 9.17.6 (2026-06-04). Below that line every collection is keyless, so `unpackPrimaryKey` rejects every record the agent returns and a plain list answers `500 mapping_error`. Of the 125 `forest-rails` environments active in the last 120 days, the twenty oldest run between 2.14.6 and 8.3.2, several of them production.

## How

`buildPrimaryKeys` falls back to a field literally named `id`, typed as the schema declares it, `String` when no such field exists. Keyed on the shape of the schema rather than the liana version: a collection that declares a key keeps it, and every v2 agent declares one, so this only fires where the alternative is a 500.

## How to test

`yarn workspace @forestadmin/agent-bff test`. Three unit tests cover the declared-key, derived-key and no-id-field cases.

### General

### Security


packages/agent-bff/src/read-model/read-model.ts
```diff
@@ -131,13 +131,33 @@ export default class ReadModel {
...
+   */
   private buildPrimaryKeys(collection: ForestSchemaCollection): void {
     const keys: PrimaryKeyField[] = [];
 
     for (const field of collection.fields ?? []) {
       if (field.isPrimaryKey) keys.push({ name: field.field, type: field.type });
     }
 
+    if (keys.length === 0) {
+      const declaredId = (collection.fields ?? []).find(field => field.field === 'id');
```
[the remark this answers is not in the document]
Tonours: Not applying this one.

`Number(value)` for a `Number` column is not a fallback quirk, it is the contract the whole stack already follows: the agent itself casts the same way in `packages/agent/src/utils/id.ts` (`schemaField.columnType === 'Number' ? Number(value) : value`), and `unpackPrimaryKey` mirrors it deliberately for every declared key.

Forcing `String` only on the derived key would make the same column type produce a string here and a number on the declared-key path, for no gain: the precision loss above `Number.MAX_SAFE_INTEGER` is identical on the declared path this PR does not touch. If we want lossless ids, the fix belongs in `IdUtils.unpackId` and `pack-id.ts` together, not in the fallback branch.

Note the fallback already defaults to `String` when the schema declares no `id` field at all, which is the case where nothing tells us the type.

packages/agent-bff/src/read-model/read-model.ts
```diff
@@ -131,13 +131,33 @@ export default class ReadModel {
...
+
+      keys.push({ name: 'id', type: declaredId?.type ?? 'String' });
```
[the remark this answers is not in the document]
Tonours: Already handled — `buildContext` does not read `isPrimaryKey` from the schema alone.

`toContextCollection` builds `derivedPrimaryKeys` from `readModel.getPrimaryKeys(collection.name)` (build-context.ts:174) and `toContextField` flags the field when either source says key (build-context.ts:133), so a schema with no `isPrimaryKey` anywhere still publishes `id` as the primary key.

It was untested though, so I added a regression test covering exactly that shape: `test/context/build-context.test.ts`, "when the schema declares no primary key at all".

nbouliol (commented):
**Spec (no ticket)**: no Linear ticket is linked to this PR, so the functional check was limited to the PR description, which the diff matches.

packages/agent-bff/src/read-model/read-model.ts
nbouliol: Claude Opus 5 (claude-opus-5[1m]): **Must fix**: A collection with no `id` field now answers 200 with a primary key naming a column it does not have.

`read-model.ts:158` pushes a key named `id` with no field behind it, so `pack-id.ts:20` and `unfolded-paths.ts:367` can no longer refuse it.
A way out: derive only when `collection.fields` declares a field named `id`, leaving the rest keyless.
Also moves: `build-context.ts:133`, which publishes no `isPrimaryKey` for that shape, and `openapi-unfolded.test.ts:478`.
</details>

1. Serve a schema whose collection declares no `isPrimaryKey` and no field named `id`.
2. Call `POST /agent/v1/<collection>/list`.
3. The response carries `__forest.primaryKey.id`, a column the collection does not have.
4. Call `GET /agent/v1/context` and no field of that collection carries `isPrimaryKey`.
</details>

Tonours: Real point on the contract, but I am not applying the fix as proposed — deriving only when an `id` field is declared puts those collections back on the 500 this PR exists to remove.

The schema this was measured against (`forest_liana 9.15.8`, 269 collections, not one `isPrimaryKey`) has 12 collections with no `id` field, and they are not exotic: `Api__Person` carries `birthdate`, `birth_city`, `birth_zipcode`, a `person_address` relation. Leaving it keyless means `unpackPrimaryKey` throws on every record and a plain list of a collection a user browses answers `500 mapping_error`. A key whose name is wrong beats a list that does not answer.

What the invented name actually costs: every route takes the record id packed and opaque, so nothing in the request path breaks. It only misses for a consumer filtering ON the key column — which a keyless collection could not do either way, since nothing in the schema names that column.

So I fixed the lie rather than the behaviour:

- `ListResponse` now states the exception — against a schema with no key and no `id` field, `primaryKey` carries the opaque id under the name `id`, which is then NOT a column, and a consumer must filter on a field the context lists.
- `buildPrimaryKeys` documents the trade instead of leaving it implicit.

On the context divergence you flagged: that one is deliberate and stays. `buildContext` marks `isPrimaryKey` on a field that exists; for this shape there is no field to carry it, so the context stays silent rather than inventing one too.

nbouliol: Reopened on the contract only: `ForestRecordMeta` still promises typed key columns with no exception, so the documented remedy stops one schema short.

Predicate: a keyless collection with no `id` field puts a non-column name in `primaryKey`, still satisfied at `read-model.ts:166`.
Your behaviour argument holds: every route takes the id opaque, and `recordSchema` always publishes `id`.
Checked by reading `schemas.ts:276` at the head sha, where the `ForestRecordMeta` description carries no exception.
</details>

Tonours: Fair — the remedy was one schema short. `ForestRecordMeta` is where `primaryKey` is defined, so a consumer reading the component never saw the caveat I had only put on `ListResponse`.

Fixed in 61a17d6:

- `ForestRecordMeta` now carries the exception itself: against a schema declaring no key and no `id` field, it holds the opaque agent id under the name `id`, which is NOT a column, and the context marks no field as the key. Filter on a field the context lists, never on a key name read back from there.
- `ListResponse` points at `ForestRecordMeta` instead of restating it.
- The `buildPrimaryKeys` comment now names `ForestRecordMeta` as the place that documents the trade.

Behaviour unchanged, as agreed.

packages/agent-bff/src/read-model/read-model.ts
```diff
@@ -131,13 +131,41 @@ export default class ReadModel {
...
```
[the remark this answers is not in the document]
Tonours: Real defect, fixed in 8a22964.

You are right that the fallback was guessing arity it does not have: the schema publishes no key, so a collection whose real one is composite packs `tenant|42`, and splitting that against one declared key answers 500 on every list — the exact failure this PR exists to remove, just moved.

Taking the second of your two options, since the first is not available: nothing in the published schema names the columns of an unpublished composite key, so there is no authoritative shape to obtain.

- `PrimaryKeyField` gains `derived?: true`, set only on the fallback key.
- `unpackPrimaryKey` takes a derived key whole: no split, no arity check, no throw. `tenant|42` comes back as `{ id: 'tenant|42' }`, the opaque agent id under a name the contract already warns is not necessarily a column.
- A numeric id stays typed (`{ id: 42 }`) when the declared `id` field says `Number`; a non-numeric one stays a string instead of throwing.
- Declared keys are untouched — they keep splitting, the arity check and the mapping errors.
- `ForestRecordMeta` now says the derived key is never split per column.

Tests: three cases on `unpackPrimaryKey` (packed composite kept whole, numeric id still typed, non-numeric id not throwing) plus the derived flag on the read-model. 1763 pass, lint 0 errors, tsc clean.

packages/agent-bff/src/read-model/read-model.ts
nbouliol: Accepted: the trade is now documented on `ForestRecordMeta` itself, and the derived key no longer splits, so the contract matches the behaviour.

Predicate: the invented name still reaches a client, and your argument that it beats a 500 on a plain list stands.
Checked `schemas.ts:286` and `read-model.ts:171` at the head sha, and `pack-id.test.ts` pins the composite case.
Still unmoved: `openapi-unfolded.test.ts:487` stays green on a `primaryKeys: []` no `ReadModel` produces.
</details>

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -251,13 +269,13 @@ function fieldRefs(deps: Deps, plan: Pick<CollectionPlan, 'key' | 'collection'>)
     projectable,
     filter: filterSchema(deps, plan),
     sort:
-      fields.projectable.length === 0
+      sortableNames.length === 0
```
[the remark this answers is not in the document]
Tonours: Right, fixed in 4523e44 — but not with the suggested condition, which lands on the same lie.

`fieldsEnum` already returns a plain `{ type: 'string' }` for an empty list, so switching the guard to `fields.projectable.length === 0` would still hand `SortClause_<key>.field` an unrestricted string. It would only move the unrestricted field from the shared component to a per-collection one.

What the document has to say is that no clause is valid, so the cap goes on the array instead: when every known field denies sorting, the list body's `sort` gets `maxItems: 0`. That matches the runtime exactly — an empty sort passes, any clause answers 422 field_not_sortable — and a generated client refuses the clause at compile time rather than at request time.

One distinction the suggested condition would have lost: an empty `projectable` is an UNKNOWN field set, not a denial. Capping there would forbid a sort the agent accepts, so sort stays unconstrained in that case and only a known set with every field denied is capped. Three tests pin the three states (capped, uncapped with a sortable field, uncapped when the field set is unknown).

packages/agent-bff/src/read-model/agent-capabilities-fetcher.ts
```diff
@@ -28,11 +75,33 @@ export default function createAgentCapabilitiesFetcher({
...
+    try {
+      return await fetch(collection);
+    } catch (error) {
+      if (!(error instanceof AgentHttpError) || error.status !== 404) throw error;
+
+      const synthesized = await synthesizeForLegacyLiana(collection, { transport, store, logger });
+
+      if (!synthesized) throw error;
+
+      return synthesized;
```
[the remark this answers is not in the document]
Tonours: Correct, and it caught an inconsistency in my own previous fix too. Both are in 59ef8a1.

The gap is real but it is not in `agent-capabilities-fetcher`: returning the synthesis is right, since a legacy collection genuinely has fewer filterable fields. It is `filterLeaves` that conflates two different empties.

`collectFields` already distinguishes them, and the document was throwing the distinction away:

- `capabilities_unavailable` — the agent could not be asked, so the field set is UNKNOWN. The free-form leaf is correct: the collection still accepts whatever it really exposes, and the runtime is the one that rejects.
- `no_fields`, or `degraded: null` with nothing filterable — capabilities were READ and name nothing. There is no valid leaf, so the free-form one offered a filter the validator answers 422 on whatever field it carries.

So the leaf is now dropped when the field set is known, leaving the tree its branch and empty-object alternatives, and the description says so instead of staying silent.

The same split applies to the sort cap I pushed in 4523e44: I had keyed it on `projectable.length === 0`, which lumped `no_fields` in with `capabilities_unavailable` and left sort unconstrained on a collection whose capabilities name no field at all. Both now read `degraded`, through one `isFieldSetKnown` helper, so filter and sort cannot drift apart again.

One existing test moved: a filterable field whose operator list is empty used to assert a free-form `FilterLeaf_E`. Its stated intent — never emit an enum no value satisfies — is better served by emitting no leaf at all.

Tests: six cases across the two axes (leaf dropped when known-and-empty, leaf kept when capabilities are unavailable, sort capped on `no_fields`, uncapped on `capabilities_unavailable`, plus the two sortable states). 1853 pass, lint 0 errors, tsc clean.

packages/agent-bff/src/context/build-context.ts
```diff
@@ -126,7 +127,10 @@ function toContextField(
   const polymorphicTargets = toArray(field.polymorphicReferencedModels);
   if (polymorphicTargets.length > 0) serialized.polymorphicTargets = [...polymorphicTargets];
 
-  if (field.isPrimaryKey) serialized.isPrimaryKey = true;
+  // The read-model derives a key when the schema declares none, and the BFF builds record
+  // identifiers from it. Publishing only the schema's flag would leave a client unable to name the
+  // key the BFF is actually using.
+  if (field.isPrimaryKey || derivedPrimaryKeys.has(field.field)) serialized.isPrimaryKey = true;
```
nbouliol: Claude Opus 5 (claude-opus-5[1m]): **Should fix**: A client filtering on the key the context lists gets no rows, because a derived key looks declared.

`build-context.ts:133` flags a derived key the same as a declared one, and `schemas.ts:291` sends the client to filter on it.
A way out: carry `derived` onto `ContextField`, or publish no `isPrimaryKey` for a derived key.
Also moves: `openapi/schemas.ts:291`, whose guidance assumes the context lists only verified keys.
</details>

1. Serve a keyless schema whose collection declares an `id` column that is not the real key.
2. Read `GET /agent/v1/context` and see `id` carry `isPrimaryKey`.
3. List the collection and read `__forest.primaryKey.id`, which holds the packed agent id.
4. Filter `id Equal` that value and the agent returns 200 with no rows.
</details>

packages/agent-bff/src/data/pack-id.ts
```diff
@@ -21,6 +32,14 @@ export default function unpackPrimaryKey(
     throw mappingError('Cannot build primary key: the collection exposes no key metadata');
   }
 
+  const [first] = primaryKeys;
+
+  if (first.derived) {
```
nbouliol, line 37: **Applies to:** `packages/agent-bff/src/openapi/unfolded-paths.ts:425` (not in this diff)
Claude Opus 5 (claude-opus-5[1m]): **Should fix**: The unfolded document names the guessed key as a real column, contradicting `ForestRecordMeta` inside the same document.

`pack-id.ts:37` honours `derived`, while `parentIdSchema` branches on arity only and asserts a `(name, type)` pair.
A way out: emit a hedged description when `primaryKeys[0].derived`, the flag the anchored line reads.
Also moves: `test/openapi/openapi-unfolded.test.ts`, whose opaque-parent case runs on a `primaryKeys: []` nothing produces.
</details>

1. Serve a schema from a liana older than 9.17.6, so no collection declares a key.
2. Run `forest-bff openapi` with unfolding configured.
3. Read a relation list path and its `parentId` description names the column `id`.
4. Read `ForestRecordMeta` in the same document, which says that `id` is not a column.
</details>

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -285,7 +336,19 @@ function requestProperties(deps: Deps, plan: Pick<CollectionPlan, 'key' | 'colle
     list: {
       filter: refs.filter,
       projection: { type: 'array', items: refs.projectable },
```
nbouliol: Claude Opus 5 (claude-opus-5[1m]): **Should fix**: On a `no_fields` collection the document closes sort and filter but still offers any projection, which answers 422 on every call.

`unfolded-paths.ts:338` leaves the projection items a bare string, while `:46` says projection is rejected like the other two.
A way out: cap projection the way `anySortable` caps sort, keeping `capabilities_unavailable` uncapped.
</details>

1. Serve a collection whose capabilities read succeeds and names no field.
2. Generate the unfolded document and read its `ListRequest`.
3. Sort is capped at zero and the filter has no leaf, but projection accepts any string.
4. Send `projection: ["id"]` and the BFF answers 422 unknown_field.
</details>

packages/agent-bff/src/openapi/unfolded-paths.ts
```diff
@@ -158,9 +168,15 @@ function filterLeaves(pool: ComponentPool, plan: Pick<CollectionPlan, 'key' | 'c
...
-  // No known filterable field: the field stays free-form and every operator stays allowed, because a
-  // collection whose capabilities could not be read still accepts whatever it really exposes.
+  // No filterable field, and the two reasons are not the same document. Capabilities that could not
+  // be read leave the field set UNKNOWN, so the leaf stays free-form: the collection still accepts
+  // whatever it really exposes, and the runtime is the one that rejects. Capabilities that were read
+  // and name nothing filterable leave no valid leaf at all — advertising a free-form one would offer
+  // a filter the validator answers 422 on for every field, so the tree keeps only its branch and
+  // empty-object alternatives.
   if (groups.length === 0) {
+    if (isFieldSetKnown(fields)) return [];
```
nbouliol: Claude Opus 5 (claude-opus-5[1m]): **Should fix**: A collection the agent filters fine is documented as filterable by nothing, once one operator of each field has no mapping.

`unfolded-paths.ts:178` reads an empty `filterable` as an answer, but `collect-unfolding.ts` also empties it by dropping fields for operator skew.
A way out: carry a third `DegradedReason` for a field dropped as undocumentable, and treat it as unknown here.
</details>

1. Have the agent report, on every field, one operator the BFF cannot map to a canonical name.
2. Each field is dropped from `filterable` while `degraded` stays null.
3. Generate the document and the filter tree carries no leaf at all.
4. It states no field is filterable, on a collection the agent filters.
</details>

packages/agent-bff/src/read-model/read-model.ts
```diff
@@ -131,13 +131,46 @@ export default class ReadModel {
...
     const keys: PrimaryKeyField[] = [];
 
     for (const field of collection.fields ?? []) {
       if (field.isPrimaryKey) keys.push({ name: field.field, type: field.type });
     }
 
+    if (keys.length === 0) {
+      const declaredId = (collection.fields ?? []).find(field => field.field === 'id');
+
+      keys.push({ name: 'id', type: declaredId?.type ?? 'String', derived: true });
```
nbouliol, line 171: Claude Opus 5 (claude-opus-5[1m]): **Should fix**: The decision to invent a primary key named `id` is recorded only in a code comment, not in `docs/adr/`.

`read-model.ts:135` carries a 25-line rationale with a weighed alternative, and no ADR in `docs/adr/` records it.
A way out: run `/adr`, as the sibling decision on the same surface already did in `docs/adr/2026-09-11-the-liana-name...`.
</details>

1. Read `ForestRecordMeta` and meet a primary key that names no column.
2. Search `docs/adr/` for the decision behind it.
3. Only the liana-capabilities ADR is there, on a different subject.
4. The reasoning is reachable only by opening `read-model.ts`.
</details>

packages/agent-bff/test/context/build-context.test.ts
```diff
@@ -449,6 +449,28 @@ describe('buildContext', () => {
     });
   });
 
+  describe('when the schema declares no primary key at all', () => {
```
nbouliol, line 452: Claude Opus 5 (claude-opus-5[1m]): **Violates conventions**: The keyless collection with no `id` field has no test, though `ForestRecordMeta` promises clients the context marks no key.

`testing.md#cover-error-and-edge-paths-not-only-the-happy-path`: the block at `build-context.test.ts:452` covers only the declared-`id` branch.
A way out: add a case whose fields carry no `id`, asserting no field carries `isPrimaryKey`.
</details>

packages/agent-bff/test/openapi/openapi-unfolded.test.ts
```diff
@@ -783,22 +792,187 @@ describe('a collection whose key ends in an index-like suffix', () => {
...
 
 describe('an unfolding carrying a filterable field with no operator', () => {
-  it('should leave it out rather than emit an enum no value satisfies', () => {
+  it('should register no leaf rather than emit an enum no value satisfies', () => {
     // An empty enum forbids every value, so the leaf would be unsatisfiable. `collectFilterableFields`
-    // cannot produce one, but the generator takes hand-constructible plain data.
+    // cannot produce one, but the generator takes hand-constructible plain data. The field set is
+    // known here, so dropping the only field leaves no valid leaf — not a free-form one, which would
+    // offer a filter every field answers 422 on.
     const empty = unfoldedDocument({
```
nbouliol, line 800: Claude Opus 5 (claude-opus-5[1m]): **Violates conventions**: Inverting the new `isFieldSetKnown` prose leaves every test green, so the document can contradict its own schema unnoticed.

`testing.md#cover-error-and-edge-paths-not-only-the-happy-path`: neither new description string appears anywhere under `packages/agent-bff/test`.
A way out: assert both strings with `toContain`, as this file already does in about twenty places.
</details>

nbouliol (commented):
**Spec (no ticket)**: no Linear ticket is linked to this PR, so the functional check was limited to the PR description, which the diff matches.

packages/agent-bff/src/context/build-context.ts
Tonours: Real, and fixed in 643e4fd — not by dropping `isPrimaryKey`, which would put the context back on the earlier finding that a client cannot name the key the BFF builds identifiers from.

The context now publishes the guess as a guess: a derived key carries `isPrimaryKeyDerived: true` alongside `isPrimaryKey`, and the `ContextField` schema says what that costs — the key is what `__forest.primaryKey` carries, nothing confirms it is the column the records are really keyed on, so a filter on it can answer 200 with no row. Read identities with it, do not filter by identity.

A declared key stays unflagged, so the two are distinguishable rather than merged. `ForestRecordMeta` points at the flag instead of only covering the no-`id`-field case.

Tests: derived flagged, declared not flagged, and the no-`id`-field shape marking no field at all.

packages/agent-bff/src/data/pack-id.ts
Tonours: Right, and fixed in 643e4fd. The document was contradicting itself inside one file: `ForestRecordMeta` warns that this `id` may not be a column while `parentId` stated `(id, String)` the way a verified key is stated.

`parentIdSchema` now reads `derived` the way `pack-id.ts` already does. On a derived key the description says the schema does not publish the primary key, so the value is the opaque agent id and NOT necessarily the value of a column of that name. A declared key keeps naming its column and type.

The shape is untouched — the wire takes the id opaque either way — so this is description-only and no generated client narrows.

Two tests pin both branches on the relation request.

packages/agent-bff/src/openapi/unfolded-paths.ts
Tonours: Correct, same lie one axis over. Fixed in 643e4fd.

`projection` now takes the same cap as `sort`, keyed on the same split: a known-empty field set (`no_fields`) gets `maxItems: 0` with a description saying the collection exposes no field, while `capabilities_unavailable` stays uncapped because the field set is UNKNOWN and the runtime is the one that rejects.

That closes the third axis: on a `no_fields` collection the document now states that sort, filter and projection all take nothing, which is exactly what the runtime answers.

Two tests: capped on `no_fields`, uncapped on `capabilities_unavailable`.

packages/agent-bff/src/openapi/unfolded-paths.ts
Tonours: Real, and the sharpest of the batch — an empty `filterable` had two causes and the document was reading both as an answer. Fixed in 643e4fd.

I took your third-state suggestion but not on `DegradedReason`: that enum drives `hasDegradedCollection`, `isFullyDegraded` and the degraded notes, so a third member would un-cache a document and warn a reader about a collection that is perfectly fine. The state is orthogonal to `degraded`, so it is its own flag.

- `collectFilterableFields` now reports whether it dropped a field for operator skew; `CollectionFields` carries `undocumentableFilter?: true`.
- `isFilterSetKnown` = field set known AND nothing dropped. `filterLeaves` and the filter tree prose read it instead of `isFieldSetKnown`.
- A skewed collection therefore keeps the free-form leaf and no longer states that no field is filterable.

The distinction is kept where it belongs: a field with zero operators is still an answer (genuinely not filterable), only an unmappable operator makes the set incomplete.

Four tests: the collector sets the flag on skew and leaves it off when every field maps, and the document keeps the leaf and drops the false prose.

packages/agent-bff/src/read-model/read-model.ts
Tonours: Fair — the rationale outgrew the comment it was living in. Recorded in 643e4fd as `docs/adr/2026-09-11-a-schema-declaring-no-primary-key-gets-one-derived-from-its-id-field.md`, indexed in `CLAUDE.md`.

It carries what the code comment could not: the measured numbers, the two things that are guessed (arity and name) and how each is published rather than hidden, the bounded cost, and the two rejected alternatives — deriving only when an `id` field is declared, and reading the key from capabilities the v1 lianas do not serve.

packages/agent-bff/test/context/build-context.test.ts
Tonours: Right, the block only covered the declared-`id` branch. Added in 643e4fd.

`build-context.test.ts` now covers the shape with no `id` field at all: no field carries `isPrimaryKey`, and none carries `isPrimaryKeyDerived` either. Two more cases landed with the derived-key fix — the derived key flagged, and a declared key left unflagged — so all three branches of that condition are pinned.

packages/agent-bff/test/openapi/openapi-unfolded.test.ts
Tonours: Right, both strings were unpinned prose. Added in 643e4fd.

`openapi-unfolded.test.ts` now asserts, with `toContain`, the sort cap description, the new projection cap description and the filter tree prose on a `no_fields` collection, plus the negative case: on `capabilities_unavailable` neither array carries a description and the tree does not claim nothing is filterable.

Inverting `isFieldSetKnown` now fails those tests instead of only contradicting the schema in prose.

packages/agent-bff/src/context/build-context.ts
nbouliol: Closed: a derived key now carries `isPrimaryKeyDerived` alongside `isPrimaryKey`, so a guess and a declaration are distinguishable.

Predicate: a consumer cannot tell a derived key from a declared one, no longer satisfiable at `build-context.ts:134`.
The flag is set only on the branch where the field carries no schema flag, and `ContextFieldSchema` publishes it with its caveat.
Pinned by three cases in `build-context.test.ts` covering derived, declared and the no-`id` shape.
</details>

packages/agent-bff/src/data/pack-id.ts
nbouliol: Closed: `parentIdSchema` now reads `derived` and hedges instead of naming a column the schema never published.

Predicate: the unfolded document asserts a `(name, type)` pair for a guessed key, no longer satisfiable at `unfolded-paths.ts:469`.
A declared key still names its column, so only the guessed branch hedges.
Pinned by two cases at `openapi-unfolded.test.ts:1149`, one asserting the hedge and one asserting `(id, String)` is absent.
</details>

packages/agent-bff/src/openapi/unfolded-paths.ts
nbouliol: Closed: projection now takes the same cap as sort, so a known-empty field set is documented as taking nothing.

Predicate: a known-empty field set leaves projection uncapped, no longer satisfiable at `unfolded-paths.ts:353`.
`anyProjectable` mirrors `anySortable`, so `capabilities_unavailable` stays uncapped where the set is unknown.
Pinned by the cap description asserted in `openapi-unfolded.test.ts`, on both the capped and uncapped cases.
</details>

packages/agent-bff/src/openapi/unfolded-paths.ts
nbouliol: Closed: an empty `filterable` is read as an answer only when no field was dropped as undocumentable.

Predicate: operator skew empties `filterable` and the document states nothing is filterable, no longer satisfiable at `unfolded-paths.ts:188`.
`collectFilterableFields` reports the drop, `CollectionFields.undocumentableFilter` carries it, and `isFilterSetKnown` gates both the leaf and the prose.
Pinned at `collect-unfolding.test.ts:206`, which asserts the flag on skew and its absence when every operator maps.
</details>

packages/agent-bff/src/read-model/read-model.ts
nbouliol: Closed: the decision is now recorded in `docs/adr/2026-09-11-a-schema-declaring-no-primary-key-gets-one-derived-from-its-id-field.md`.

Predicate: the rationale exists only as a code comment, no longer satisfiable since the ADR is indexed in `CLAUDE.md`.
It records both guesses, the measured numbers, the bounded cost and the two rejected alternatives.
No regression test applies to a recorded decision, so `pin_filed` names the ADR itself rather than a ticket.
</details>

packages/agent-bff/test/context/build-context.test.ts
nbouliol: Closed: the no-`id` shape is now covered, so all three branches of that condition are held against the rule.

Rule `testing.md#Cover error and edge paths, not only the happy path` re-read, and the uncovered branch is the one now at `build-context.test.ts:503`.
It asserts no field carries `isPrimaryKey` and none carries `isPrimaryKeyDerived`.
The declared and derived branches are covered by the two cases above it.
</details>

packages/agent-bff/test/openapi/openapi-unfolded.test.ts
nbouliol: Closed: the three description strings are now asserted, so inverting the prose fails a test instead of only the document.

Rule `testing.md#Cover error and edge paths, not only the happy path` re-read, and each new string now matches once under `packages/agent-bff/test`.
The sort cap, the projection cap and the filter tree prose are all asserted with `toContain`.
A negative case pins that `capabilities_unavailable` claims none of them.
</details>

nbouliol (approved):
Approving on the review side: the seven findings from the validator run at `59ef8a1a` are all closed and verified at `3797d1ee`, each with the property that closes its class and a regression pinned where a test applies.

Two things this approval does not clear:

- #1894 is still open, and this PR is based on its branch, so it cannot merge first.
- `LLM Integration Tests (ai-proxy)` is red. Unrelated to this diff, but it is red.

PR #1896, opened 2026-09-11
feat(agent-client): allow overriding the default request timezone

EnkiP:
fixes PRD-1182

## What

`HttpRequester` hardcoded `timezone=Europe/Paris` on every agent request (`http-requester.ts:81` and `:120`). This adds an optional `timezone` to the requester options and to `createRemoteAgentClient`, defaulting to `Europe/Paris`.

Needed by the PRD-1177 automation poller, which must read segments in the **project** timezone so a relative-date filter (`was in previous X days`) resolves the same way on every executor instance.

## Backward compatibility

Both call sites already spread `query` *after* the timezone, so the precedence chain is preserved exactly:

```
query.timezone  >  options.timezone  >  'Europe/Paris'
```

Every existing caller passes no option and keeps sending `Europe/Paris`. Verified by reverting the two source lines to the old literal and confirming that exactly the new behaviour tests fail while the pre-existing ones still pass.

The `httpRequester` injection escape hatch is untouched.

## Per-query timezone: already exists, left alone

`Collection.action()` threads `actionContext.timezone` into `FieldFormStates`, which reaches the `/hooks/load` and `/hooks/change` calls through the existing `query.timezone` precedence. That path needs nothing from this change, so the new option is correctly requester-level and `src/types.ts` was not touched.

## Not covered by this change

Two subclasses fully override `query()`/`stream()` and still hardcode `Europe/Paris` in their own dispatch:

- `mcp-server/src/in-process-http-requester.ts:52`
- `agent-bff/src/agent/in-process-transport.ts:90`

They never reach the base implementation, so nothing regressed — but they will not honour the option. Both are in-process transports that also hardcode their `url`; wiring a timezone through them is a separate change. **Neither is on the workflow-executor's path**: `agent-client-agent-port.createClient` calls `createRemoteAgentClient` with no injected requester, so the poller gets the fixed behaviour.

## Release

No changeset — the repo generates versions from conventional commits via `@qiwi/multi-semantic-release`. This `feat` produces a minor bump from the current `1.15.2`.

## Tests

```
cd packages/agent-client && jest --config jest.config.ts --coverage=false --runInBand
→ 20 suites, 414 passed
```

New cases on **both** `query()` and `stream()`: the default when no option is given, the option overriding the default, and an explicit `query.timezone` beating the option. Each asserts the actual serialized query.

Dependents, since `HttpRequester` is exported and subclassed: agent-bff 1774 passed, mcp-server 1015 passed, agent-testing 37 passed / 1 skipped. `tsc --noEmit` exits 0; ESLint reports 0 errors.

> [!NOTE]
> ### Allow overriding default request timezone in `HttpRequester` and `createRemoteAgentClient`
> - `HttpRequester` now accepts an optional `timezone` in constructor options, defaulting to `Europe/Paris` when unset. The `query` and `stream` methods use this resolved timezone instead of a hard-coded value, while an explicit query-level timezone still takes precedence.
> - `createRemoteAgentClient` forwards a new optional `timezone` parameter to newly constructed `HttpRequester` instances. When an `httpRequester` is injected, the factory timezone is ignored and the injected instance keeps its own configuration.
> - Behavioral Change: requests that previously always sent `Europe/Paris` now send the configured timezone (or `Europe/Paris` if unset); callers injecting a custom `HttpRequester` should verify their timezone configuration.
>
> 
> #### Changes since #1896 opened
>
> - Changed access modifier of `HttpRequester.timezone` getter from private to protected in `agent-client` package [dc66eec]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dcc53a7.</sup>
> 
>


packages/agent-client/test/index.test.ts
```diff
@@ -94,6 +104,75 @@ describe('createRemoteAgentClient', () => {
     expect(mockMakeCreateApprovalRequest).not.toHaveBeenCalled();
   });
 
+  it('should send the default timezone when none is provided', async () => {
```
Scra3, line 107: suggestion: the default and the explicit-override cases are already proven directly in `http-requester.test.ts`, so keeping only the injected-requester test here leaves the one behaviour the factory actually owns.

Scra3 (commented):
Complexity-only pass. Four lines of source and a getter, nothing to cut there. One note on the test layer.

packages/agent-client/src/http-requester.ts
```diff
@@ -18,6 +18,10 @@ function parseJson(text: string | undefined): unknown {
 export default class HttpRequester {
   protected readonly deserializer: Deserializer;
 
+  private get timezone(): string {
```
Scra3: Claude Opus 5 (claude-opus-5[1m]): **Preferential**

Making this `protected` rather than `private` would let the two subclasses that override `query()`/`stream()` inherit the option instead of keeping their own literal — `mcp-server/src/in-process-http-requester.ts:52` and `agent-bff/src/agent/in-process-transport.ts:90` both hardcode `'Europe/Paris'` and, as written, cannot do otherwise because the getter is invisible to them.

Nothing misbehaves today: both call `super(token, { url })` with no timezone, so the values coincide, and the BFF resolves a per-request timezone into `query` where the spread order makes it win anyway. The divergence only appears the day either constructor starts accepting one, which is exactly when nobody will be looking at this file. One keyword now instead.

Scra3 (commented):
**Spec (`PRD-1182`)**: conforms. Optional `timezone` on both `HttpRequester` and `createRemoteAgentClient`, default `Europe/Paris`, backward compatible — I verified the precedence `query.timezone > options.timezone > default` holds at both call sites, and that no existing caller passes the option.

No convention violations and no architecture violations. One note below.

packages/agent-client/src/http-requester.ts
EnkiP: Done in `dc66eece7` — `protected`, with a one-line comment saying why.

Verified the premise first: both subclasses do `super(bearerToken, { url })` with no timezone, so the values coincide today and nothing was misbehaving. Agreed it is worth closing before either constructor grows a timezone. `deserializer` directly above is already `protected`, so this matches the file's own convention.

I left the two literals in `in-process-http-requester.ts:52` and `in-process-transport.ts:90` alone — switching them to `this.timezone` is now possible but belongs with whatever change actually gives those transports a timezone. `tsc --noEmit` clean on agent-client, mcp-server and agent-bff; 414 agent-client tests pass.

packages/agent-client/test/index.test.ts
EnkiP: I'd rather keep them — I don't think they're the same test at a different address.

`http-requester.test.ts` asserts on `mockRequest.query` arguments against a mocked superagent. These three go through `nock` and a real round-trip, then read `timezone` back off the actual URL query string. So they cover the wiring the factory owns — `createRemoteAgentClient` → `HttpRequester` → superagent → the wire — which is exactly where this PR's new parameter can go wrong.

Concretely: drop `timezone: params.timezone` from the `new HttpRequester(...)` call in `index.ts` and every requester-level test stays green while these fail. That's the regression I want caught, and it is the one the option actually introduces.

Happy to be overruled if you'd still rather trim — it's two tests.

hercemer42 (approved):

PR #1897, opened 2026-09-11
feat(workflow-executor): support Amazon Bedrock as a self-hosted AI provider

Scra3:
## What

`AI_PROVIDER` accepted `openai` and `anthropic` only. Customers who need their LLM calls to stay inside their AWS account had no path — Bedrock speaks the Converse API behind SigV4, which neither `ChatOpenAI` nor `ChatAnthropic` can reach.

```bash
AI_PROVIDER=bedrock
AI_MODEL=eu.anthropic.claude-sonnet-5
AWS_REGION=eu-west-1
# no AI_API_KEY
```

**Verified against a real AWS account** in `eu-west-1`: the integration suite passes 8/8, with all three advertised Claude lines driven through a forced tool call.

## Scope, and why the proxy still refuses it

Self-hosted executor only: `AiClient` / `createBaseChatModel`, the CLI env config, and the embedded `ai` option of `addWorkflowExecutor()`.

The front-facing `Router` / `ProviderDispatcher` keeps refusing `bedrock`, and now does so at construction rather than on the first end-user query. The reason is not that nothing feeds it: **the server-side AI provider choice is being retired** — the SaaS will serve OpenAI only, and customisation will go through the embedded executor — so wiring Bedrock into the dispatcher would build on a surface we are removing. Recorded in `docs/adr/2026-09-15-bedrock-is-executor-only-and-its-models-pass-an-allowlist.md`, because the asymmetry reads as an inconsistency and the natural instinct is to "fix" it.

## Supported models: an allowlist, not a denylist

**Claude sonnet, haiku and opus only** — as a bare id, behind any cross-region inference-profile prefix (`us.`, `eu.`, `apac.`, `jp.`, `au.`, `global.`), or as an inference-profile ARN, which is reduced to its id. Application inference profiles carry an opaque id and are refused.

This inverts what `openai` and `anthropic` use, on purpose: Bedrock resells hundreds of models we do not own, and the executor forces a tool call on **every** AI step. `amazon.titan-text-express-v1` passed a permissive filter without being able to honour one, which surfaces as a failure mid-run instead of a refusal at startup.

Ids are unwrapped before matching, then still run through the Anthropic rules — that is what keeps the streaming-only and retired Claude releases refused here too, including through Bedrock's context-window suffixes (`-v1:0:200k`).

## Credentials

The standard AWS chain (IAM role, `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`, shared profile, assumed roles and IRSA). Two consequences, both checked at boot rather than mid-workflow:

- `AI_API_KEY` alongside `bedrock` is **refused**, not silently ignored — `ChatBedrockConverse` drops it and falls back to the ambient role, so an operator would ship a key believing it authenticates.
- A missing `AWS_REGION` / `AWS_DEFAULT_REGION` is **refused**. `ChatBedrockConverse` demands a region at construction and reads `AWS_DEFAULT_REGION` only, so the AWS SDK below it never gets to resolve one from the shared profile.

## Two LangChain behaviours worked around

**1. `tool_choice` is gated by a stale hardcoded list.** `@langchain/aws` infers support from model-family substrings (`claude-3`, `claude-4`, `mistral-large`) and throws *client-side* for anything else. The executor binds every AI step with `tool_choice: 'any'`, so this is on the critical path. Measured against the unpatched default:

```
eu.anthropic.claude-sonnet-5 → Model does not currently support 'tool_choice'
amazon.nova-pro-v1:0         → Model does not currently support 'tool_choice'
```

We pass `supportsToolChoiceValues` explicitly and let Bedrock be the authority: a model that genuinely cannot do it answers with a `ValidationException`, and its id leaves the allowlist.

**2. Region.** LangChain reads `AWS_DEFAULT_REGION` only, while ECS/EKS/Lambda set `AWS_REGION`.

## Errors an operator can act on

A model the provider refuses (`ValidationException`, `ResourceNotFoundException`, `AccessDeniedException`) used to reach the workflow author as *"An unexpected error occurred"*, and on guidance and load-related-record steps it did not reach them at all — `withAiAssist` re-tagged it and the step degraded to manual, permanently and silently.

It now surfaces as `AiModelUnusableError`, naming the model and the error class, classified `configuration`, and passing through the degrade path because a refused model is a permanent fault rather than the transient outage that path exists for. The provider's own sentence stays in the technical message: an `AccessDeniedException` spells out the caller's IAM role ARN and AWS account id, which has no place in the Forest UI or the audit trail.

## Tests

- **Unit** — id unwrapping through every profile prefix, ARN and version suffix; model construction without an `apiKey`; the region fallbacks and refusal; the per-provider CLI rules; the embedded `ai` option; the Router boundary; the degrade-path interaction.
- **`bedrock-tool-choice.test.ts`** — the `supportsToolChoiceValues` contract against the real LangChain class, no mock and no AWS, with a negative control that goes green if the override is deleted.
- **`bedrock.integration.test.ts`** — the real Converse API. It **fails in CI when the secrets are missing** rather than skipping, because a silent skip reads as "Bedrock is verified". Locally it skips.

It asserts each advertised line has a model the account can drive, then sweeps `ListFoundationModels` ∪ `ListInferenceProfiles`. Entitlement and AWS's 30-day parking of a Legacy model are account states and are excluded; **throttling is not** — it is our own serial walk, so it fails the run rather than excusing it.

The three CI secrets (`BEDROCK_AWS_REGION`, `BEDROCK_AWS_ACCESS_KEY_ID`, `BEDROCK_AWS_SECRET_ACCESS_KEY`) are set.

## Notes for the reviewer

- `@langchain/aws` is pinned to **1.4.3**, the newest release still accepting `@langchain/core ^1.0.0`; 1.4.4+ require `^1.2.8` and the repo pins `1.1.48`.
- It pulls ~107 `@aws-sdk/*` packages into the Docker image, including `client-kendra` and `client-bedrock-agent-runtime` we never use — hard dependencies of the package, not prunable. `docker/deps/yarn.lock` is regenerated; the `@forestadmin` closure is unchanged.
- `yarn start:with-executor:with-bedrock` in `packages/_example` runs the agent and executor against Bedrock locally.
- Docs are a separate PR: ForestAdmin/docs#33.
- A validator pass ran on this PR; its three findings are resolved in the threads.

### General

### Security

No new credential material is handled: Bedrock auth is delegated to the AWS SDK's own chain. Refusing `AI_API_KEY` stops an operator believing a key is in use when it is not, and the provider's error message is kept out of the user-facing channel because it names the caller's IAM identity and AWS account.

> [!NOTE]
> ### Add Amazon Bedrock as a self-hosted AI provider in `workflow-executor`
> - Adds `bedrock` to the AI provider union and constructs Bedrock chat models via `ChatBedrockConverse` with zero retries, no API key, and AWS credential-chain authentication. Region resolution prefers explicit config, then `AWS_REGION`, then `AWS_DEFAULT_REGION`.
> - Restricts Bedrock models to an allowlist of Claude sonnet, haiku, and opus family identifiers (including cross-region inference-profile prefixes and ARNs). Non-allowlisted models produce `AIModelNotAllowlistedError`.
> - Makes Bedrock executor-only: `Router.validateConfigurations` and `ProviderDispatcher` reject Bedrock configurations on the server-side AI proxy. The embedded executor and CLI accept `AI_PROVIDER=bedrock` without `AI_API_KEY` and require a region.
> - Wraps `ResourceNotFoundException` and `AccessDeniedException` from Bedrock as `AiModelUnusableError`, preventing permanent config errors from degrading to manual steps. Sensitive provider messages are kept separate from user-facing text.
> - CI runs a Bedrock integration suite (30-min timeout) gated on AWS secrets; the unit-test matrix excludes `bedrock.integration`.
> - Risk: `createBaseChatModel` Bedrock branch omits explicit credentials entirely — any deployment relying on static key auth for Bedrock will fail. `isModelSupportingTools(model, 'bedrock')` now applies the allowlist; callers passing a Bedrock model without the provider argument will get OpenAI rules and incorrect results. `claude-3-sonnet-20240229` is newly rejected for Anthropic.
>
> 
> #### Changes since #1897 opened
>
> - Modified `isAiConfigurationError` util in `workflow-executor` to classify Bedrock `ValidationException` errors as configuration errors when the error message matches `/model identifier/i` regex pattern [25fe21f]
> - Changed `resolveBedrockRegion` util in `ai-proxy` to use logical-OR operator instead of nullish coalescing when selecting the AWS region [25fe21f]
> - Added test coverage for Bedrock provider in `ai-proxy` verifying `apiKey` parameter validation, model identifier validation errors, and region resolution with empty `AWS_REGION` environment variable [25fe21f]
> - Added test coverage in `workflow-executor` for Bedrock `ValidationException` classification as configuration error when message references model identifier [25fe21f]
> - Reformatted test assertions and import statements across multiple test files in `ai-proxy` without changing test behavior [25fe21f]
> - Updated inline comments in `bedrock.integration.test.ts` within the `ai-proxy` package [037198e]
> - Removed duplicate '-live' pattern entry from `OPENAI_UNSUPPORTED_PATTERNS` constant in `ai-proxy` package [0a5e52c]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 032024e.</sup>
> 
>


packages/agent/src/embedded-workflow-executor.ts
```diff
@@ -64,13 +64,20 @@ export default class EmbeddedWorkflowExecutor {
...
           'Omit `ai` entirely to use Forest’s AI server.',
       );
     }
 
+    if (ai && ai.provider !== 'bedrock' && !ai.apiKey) {
+      throw new Error(
+        `addWorkflowExecutor: \`ai\` requires \`apiKey\` for provider '${ai.provider}'. ` +
+          'Only `bedrock` reads its credentials from the AWS credential chain.',
+      );
+    }
```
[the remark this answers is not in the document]
Scra3: Fixed in d1d3855 — valid, and it was the one real asymmetry left between the two entry points.

Worth recording *why* it cannot self-recover: `ChatBedrockConverse` resolves its region from `AWS_DEFAULT_REGION` only. It never reads `region` from the AWS profile, so a deployment that sets the region the ordinary way (profile or task definition) still has none as far as the client is concerned. There is no later point at which this gets better — hence the boot-time refusal in both paths now.

Covered by three tests in `agent-workflow-executor.test.ts`: no region reachable throws, and each of `AWS_REGION` / `AWS_DEFAULT_REGION` is accepted on its own.

Scra3:
Addressed the two bot findings in d1d3855:

- **Macroscope, region validation (High)** — valid. The embedded `addWorkflowExecutor()` path now refuses a Bedrock config with no `region`, `AWS_REGION` or `AWS_DEFAULT_REGION`, matching what the standalone CLI already did. Replied in the thread with why it cannot recover later.
- **Qlty, `parseAiConfig` complexity 12** — fair, I had added a fourth branch to it. Extracted `parseBedrockConfig`, so the function stops growing a block per provider.

Coverage stayed at 100% on all modified files.

On the red checks, none is this PR:

- **`Linting & Testing` (9 packages)** — one flaky timeout in `mcp-server` (`does not serve /files`, 5000 ms, 1014/1015 passing) tripped the fail-fast matrix and cancelled the rest. `workflow-executor` shows `1787 passed` then `The operation was canceled`. `mcp-server` does not depend on `ai-proxy`, so this PR cannot reach it. Re-running.
- **`Validate Dockerfile build`** — the image builds and the smoke test passes (`smoke test passed for workflow-executor:pr`). Trivy fails it: two HIGH CVEs in `libpcre2-8-0` (CVE-2026-86145, CVE-2026-89161), fixed in `10.42-1+deb12u1`. The base image is pinned by digest (`node:22-bookworm-slim@sha256:813a748…`), so it cannot pick the fix up on its own — it needs a digest bump, which will hit `main` and every other PR too. Deliberately not bundled here; it wants its own PR.
- **`LLM Integration Tests (ai-proxy)`** — already red on `main` before this branch existed: OpenAI shipped `gpt-live-1`, a non-chat model the denylist does not catch. Not a required check.

Same SHA was green on Sept 11; the Trivy and flake failures both appeared on a re-run of unchanged code.

packages/ai-proxy/src/supported-models.ts
```diff
@@ -102,10 +102,55 @@ function isAnthropicModelSupported(model: string): boolean {
...
+  'us.',
+  'us-gov.',
+  'eu.',
+  'apac.',
+  'jp.',
+  'au.',
+  'global.',
+];
+
+const BEDROCK_SUPPORTED_ANTHROPIC_FAMILIES = ['claude-sonnet', 'claude-haiku', 'claude-opus'];
```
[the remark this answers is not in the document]
Scra3: Fixed in 17fcdc54 — correct, and the diagnosis is exact. Anthropic renamed its lines mid-Claude-4 (`claude-3-5-sonnet` became `claude-sonnet-4-6`) and I matched only the new spelling, which silently excluded every Claude 3.x id Bedrock still serves.

Added the legacy families alongside the new ones. One deliberate difference from your suggestion: I also added `claude-3-5-haiku` and `claude-3-7-sonnet`, because they belong to the same lines and the chaining handles them correctly — the family allowlist admits them, then `isAnthropicModelSupported` rejects the specific EOL ids. That ordering is now pinned by a test (`lets the EOL denylist fire after the family allowlist passes`), so the two filters cannot drift apart unnoticed.

packages/ai-proxy/src/supported-models.ts
```diff
@@ -102,10 +102,56 @@ function isAnthropicModelSupported(model: string): boolean {
...
+
+const BEDROCK_VERSION_SUFFIX = /-v\d+:\d+$/;
+
+function isBedrockModelSupported(model: string): boolean {
+  const profilePrefix = BEDROCK_INFERENCE_PROFILE_PREFIXES.find(prefix => model.startsWith(prefix));
+  const vendorId = profilePrefix ? model.slice(profilePrefix.length) : model;
+  const separatorIndex = vendorId.indexOf('.');
+
+  if (separatorIndex === -1) return false;
```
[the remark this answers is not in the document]
Scra3: Half fixed in 17fcdc54, and the other half is a real open question.

**Standard inference-profile ARNs now work.** `arn:aws:bedrock:<region>:<account>:inference-profile/<profile id>` carries the profile id in its tail, so the ARN form reduces to the id form; it is unwrapped and re-evaluated. Covered for both `aws` and `aws-us-gov` partitions.

**Application inference profile ARNs are still refused**, and I do not think I can fix that silently. Their id is opaque (`.../application-inference-profile/abcdef123`) — nothing in the string says which model it points at, so validating one means calling Bedrock at startup. The options are to allow them unvalidated, which puts a hole in the allowlist exactly where the allowlist is the only thing protecting the user from a model that cannot honour the forced tool call, or to resolve them with a `GetInferenceProfile` call, which adds an AWS API dependency to config validation.

This matters because application inference profiles are how AWS documents per-team cost allocation and quota tagging, so an enterprise following that guidance hits the refusal. Raising it with the team rather than picking one alone. The error message at least now says ARNs of that form are not accepted instead of claiming the model lacks tool support.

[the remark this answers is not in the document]
Scra3: Not yet — thanks, but I would rather not land that call in this PR.

`GetInferenceProfile` would resolve the opacity, but it carries two costs that belong in the team decision rather than in a follow-up commit:

- **An IAM permission.** Self-hosted deployments would need `bedrock:GetInferenceProfile` on top of `bedrock:InvokeModel`. Anyone whose policy is scoped to invoke-only would go from "my model is refused" to "my agent will not start", which is a worse failure for the same configuration.
- **A network dependency at config-validation time.** Today the allowlist is a pure function; adding an AWS round trip makes startup fail on a Bedrock blip, in a code path whose whole job is to fail fast on static configuration.

There is also a third option neither of us has put on the table: accept application profile ARNs, and let the first forced tool call be the thing that fails. That keeps startup pure and keeps the enterprise path open, at the price of moving the diagnosis from startup to first use — which is exactly what the allowlist exists to prevent, so I do not think it wins, but it should be rejected explicitly rather than by omission.

So: refusal stays for now, with an error message that names the ARN form instead of blaming the model. I will raise the three options with the team, and if we take the `GetInferenceProfile` route I will come back to you for it — the resolution itself is the easy part, the permission and the startup contract are not.

packages/ai-proxy/src/supported-models.ts
```diff
@@ -82,6 +83,7 @@ const ANTHROPIC_UNSUPPORTED_MODELS = [
   'claude-3-5-haiku-20241022', // EOL 2026-02-19
   'claude-3-5-haiku-latest', // Points to deprecated claude-3-5-haiku-20241022
   'claude-3-7-sonnet-20250219', // EOL 2026-02-19
+  'claude-3-sonnet-20240229', // EOL — Bedrock: "reached the end of its life"; caught by bedrock.integration
```
Scra3: Claude Opus 5 (claude-opus-5[1m]): `Must fix`

An existing customer running `AI_PROVIDER=anthropic` with `AI_MODEL=claude-3-sonnet-20240229` boots today and **stops booting** after this upgrade — the executor refuses at startup, not at the first AI step, so the whole process is down rather than one workflow degraded.

`ANTHROPIC_UNSUPPORTED_MODELS` is shared: `isBedrockModelSupported` delegates to `isAnthropicModelSupported`, so an entry added for a Bedrock EOL also binds the direct Anthropic provider. Verified against the built module — `isModelSupportingTools('claude-3-sonnet-20240229', 'anthropic')` now returns `false`, and `validateAiConfigurations` runs in the `AiClient` constructor, which `build-workflow-executor.ts` calls at startup.

The model is genuinely retired at Anthropic, so refusing it is defensible. Turning a runtime 404 into a boot failure on a minor bump is the part that deserves to be chosen rather than inherited from the Bedrock work — either scope the entry to bedrock, or say so in the release notes.

packages/ai-proxy/test/bedrock.integration.test.ts
```diff
@@ -0,0 +1,238 @@
...
+          if (response.tool_calls?.length) verified.push(model);
+          else failures.push({ model, error: 'no tool call in the response' });
+        } catch (error) {
+          const message = error instanceof Error ? error.message : String(error);
+          const name = (error as { name?: string }).name ?? '';
+
+          // Keyed on the AWS error identity, not on message text: a broken InvokeModel policy makes
+          // every model raise "not authorized", which would file the whole catalogue under
+          // unavailable and leave failures empty.
+          if (name === 'AccessDeniedException' || name === 'ThrottlingException') {
```
Scra3: Claude Opus 5 (claude-opus-5[1m]): `Should fix`

On a throttled run this suite reports green having verified two models out of forty, and the only trace is a `console.log` in a job nobody opens when it is green. The allowlist is the sole thing standing between a customer and a model that cannot honour the forced tool call, and this suite is its only empirical justification — so a silent degradation here is the failure the `unavailable` bucket was invented to expose.

`AccessDeniedException` belongs in that bucket: entitlement is per-account and a catalogue-wide denial is caught by the floor below. `ThrottlingException` is different, and the suite manufactures it: clients are built through `new AiClient(...)`, inheriting `createBaseChatModel`'s `maxRetries: 0`, where production goes through `ai-client-adapter.ts` with `maxRetries: 2` — so one throttle is terminal here and nowhere else. The catalogue is then walked serially with no backoff.

Cheapest fix that closes it: split `throttled` from `unavailable`, assert `throttled` is empty, and give the test client the `maxRetries: 2` production uses.

Scra3 (commented):
Claude Opus 5 (claude-opus-5[1m]): **Spec (no ticket)** — no Linear ticket governs this PR. Searched the `linear-code[bot]` linkback, the body, the branch name, the title and the commit trailers, then the workspace itself: `bedrock` returns zero issues. The two ancestors that established the `AI_PROVIDER` path (PRD-63 anthropic, PRD-134 openai) are Done and do not govern this one. The functional check was therefore limited to the PR description.

---

Claude Opus 5 (claude-opus-5[1m]): `Should fix`

A reader six months from now cannot tell whether the provider asymmetry was reasoned or accidental, and the natural instinct on finding it will be to "fix" the inconsistency.

This change makes two decisions that are hard to reverse, surprising without context, and carry a real trade-off — the `/adr` gate, all three met:

- **Bedrock is gated by an allowlist** (`supported-models.ts`) while `openai` and `anthropic` use denylists. The inversion is deliberate — we vouch for nothing on a catalogue we do not own — but nothing records that.
- **Bedrock is executor-only**: `Router` refuses it at construction, so the same `AiConfiguration` union is served by one surface and refused by the other.

Both were reached independently by the ADR lens and the architecture lens this run, and no accepted ADR covers either (13 incidental matches, healthy control query). Worth `/adr`.

packages/ai-proxy/src/supported-models.ts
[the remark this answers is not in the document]
Scra3: Withdrawing this one — I over-rated it, and the reviewer who raised it (me) judged the entry in isolation from the list it joins.

`ANTHROPIC_UNSUPPORTED_MODELS` contains nothing but direct-Anthropic ids, every one of them EOL: `claude-3-haiku-20240307`, `claude-3-5-haiku-20241022`, `claude-3-5-haiku-latest`, `claude-3-7-sonnet-20250219`. `claude-3-sonnet-20240229` is Anthropic's own id too — Bedrock reuses it behind its prefixes and suffixes — and Claude 3 Sonnet is retired at Anthropic as well. The entry is in the right list with the right neighbours.

The residual objection was that it turns a runtime 404 into a boot refusal on a minor bump. That is true, and it is equally true of every other entry here, all of which shipped long before this PR. It is the established behaviour of this denylist, not something this change introduces — so there is nothing here for this PR to decide.

What was genuinely wrong was the comment: it credited the EOL to Bedrock's error message, which reads as a Bedrock fact being applied to the Anthropic provider. Reworded to say the model is EOL at Anthropic, with Bedrock's wording as corroboration.

packages/ai-proxy/test/bedrock.integration.test.ts
Scra3: **closed** — the predicate no longer holds, checked by running it.

Predicate: *a run where a minority of models answer and the majority are throttled reports green*. Reproduced against the fixed code with a probe copy of this suite — the first model answering, the other nineteen raising `ThrottlingException` — and the sweep now **fails**, reporting `Verified 1/20` with the throttled list attached. On the pre-fix classification that same run was green: `failures` stayed empty and the `verified.length > 0` floor was satisfied by the single model that answered.

What closes the class is that throttling has its own bucket and its own assertion (`expect(throttled).toEqual([])`) instead of sharing `unavailable` with entitlement. The two are not the same fact: entitlement says this account may not use the model, throttling says we asked too fast — only the second means the run itself is unsound. The test client also takes `maxRetries: 2` now, matching `AiClientAdapter`, so the suite stops manufacturing the throttling it used to excuse.

Regression pinned by that assertion: it fires exactly when the condition recurs.

One thing the fix surfaced that the finding did not predict, from the first real run against AWS: a third account state. AWS parks a Legacy model an account has not called in 30 days, and it answers fine on an account that does. It is excluded like entitlement, but read from the message rather than the exception name — it shares `ResourceNotFoundException` with a genuine end-of-life, and that one has to stay a failure, since it is how a retired id earns its denylist entry.

packages/ai-proxy/src/supported-models.ts
Scra3: **reopened** — and I am withdrawing the withdrawal above, which was wrong.

Predicate: *a configuration that boots before this PR fails to boot after it*. It is still satisfied, by `AI_PROVIDER=anthropic` + `AI_MODEL=claude-3-sonnet-20240229`. Checked at the two shas rather than argued: the id is **absent** from `ANTHROPIC_UNSUPPORTED_MODELS` at the merge base `057f634e` and **present** at head, so this PR is what introduces the refusal. Validation runs in the `AiClient` constructor, so the failure is at startup.

My earlier reply argued the entry sits in the right list beside the right neighbours, and that boot-refusal is the established behaviour of every entry in it. Both of those remain true, and neither answers the predicate. The difference is that the other entries were added for models already dead when they were added, so no working configuration broke; this one is added in the same commit that makes it refuse. "The entry belongs here" and "no working configuration breaks" are two claims, and I let the first stand in for the second.

Not reopening to ask for a code change — scoping the entry to bedrock and release-noting it are both defensible. Reopening because the choice between them has not been made, and a resolved thread would record it as settled.

No regression test pins this either way: nothing asserts that a previously-bootable configuration still boots.

Scra3:
## Verification of the review's body finding — run `1897-20260915082549-Scra3-verify`

**closed** — the ADR asked for exists.

Predicate: *the PR embeds a decision meeting the `/adr` gate that no accepted ADR records*. It no longer holds: `docs/adr/2026-09-15-bedrock-is-executor-only-and-its-models-pass-an-allowlist.md`, written through the skill's helper so the `CLAUDE.md` index is its output rather than a hand edit. `validate-adrs` passes at 4 records, and the supersession analysis found no candidate — the repo's only other ADR is about `i_contains` collation semantics, with no term overlap.

Checked by reading it back against the gate rather than against the finding's wording: hard to reverse (a public contract about which providers work on which surface), surprising (an allowlist beside two denylists, accepted on one surface and refused on the other), real trade-off (both alternatives existed and one was chosen).

Worth recording that the ADR says something the review could not have: the decisive reason is that the **server-side AI provider choice is being retired** — the SaaS will serve OpenAI only, and customisation will go through the embedded executor. That turns the Router's refusal from an inconsistency into an alignment, and it came from the team, not from the diff. A version of this ADR written from the code alone would have said "nothing feeds the dispatcher", which is true, weak, and would not have stopped anyone wiring it up later.

Regression pinned: not applicable — an ADR is not code, and `validate-adrs` already guards its integrity in CI.

packages/ai-proxy/src/supported-models.ts
Scra3: **accepted** — the finding was wrong, and so was the reopening above. Third and final position on this thread, this time from a checked fact rather than an argued one.

Anthropic's deprecation page lists `claude-3-sonnet-20240229` as **Retired, 21 July 2025**, and states that requests to retired models fail. So the predicate I reopened on — *a configuration that boots before this PR fails to boot after it* — is satisfied only in the narrow sense that validation used to pass. There is no working configuration behind it: pre-PR the executor booted and then failed on **every** AI step with a 404 from Anthropic. The change converts a silent boot followed by guaranteed runtime failure into a refusal that names the model. That is strictly better, and there is nothing here to decide.

What went wrong twice is the same thing: I argued from the shape of the list instead of checking the published lifecycle, which was one fetch away. The withdrawal reached the right answer for a reason I had not verified; the reopening then correctly rejected that reasoning and landed on the wrong answer.

Recorded outside this thread, per the rule that a counter-argument living only in a resolved thread is invisible: the entry now reads `// EOL 2025-07-21` in `ANTHROPIC_UNSUPPORTED_MODELS`, matching the dated convention of its neighbours, in 178f3aa37. The date is the fact; Bedrock's wording was only where it surfaced.

packages/ai-proxy/src/validate-ai-configurations.ts
```diff
@@ -1,12 +1,26 @@
...
-import { AIModelNotSupportedError } from './errors';
+import { AIBadRequestError, AIModelNotAllowlistedError, AIModelNotSupportedError } from './errors';
 import isModelSupportingTools from './supported-models';
 
 export default function validateAiConfigurations(aiConfigurations: AiConfiguration[]): void {
   for (const config of aiConfigurations) {
+    // The CLI and the embedded option both refuse this, but a direct `new AiClient(...)` reaches
+    // neither, and ChatBedrockConverse drops the key silently and authenticates with the ambient
+    // role instead — wrong account, wrong bill, no error.
+    if (config.provider === 'bedrock' && (config as { apiKey?: string }).apiKey) {
```
EnkiP, line 11: Claude Fable 5.1 (claude-fable-5-1): **Violates conventions**: the `AiClient` guard refusing an `apiKey` on a bedrock configuration has no test, so a refactor dropping it passes CI.

`validate-ai-configurations.ts:11` throws for a bedrock config carrying `apiKey`, no test reaches it, and `testing.md#Cover error and edge paths` asks for one per branch.

A way out: one `it()` beside `ai-client.test.ts:54` passing `apiKey` on a bedrock configuration.
</details>

packages/ai-proxy/src/create-base-chat-model.ts
```diff
@@ -2,10 +2,32 @@ import type { AiConfiguration } from './provider';
...
+// LangChain infers tool_choice support from a hardcoded family list (claude-3/4, mistral-large)
+// and throws client-side for anything else, so every claude-5 id would be rejected before reaching
+// AWS. Let Bedrock be the authority instead: a model that truly cannot do it answers with a
+// ValidationException, and its id leaves the allowlist in supported-models.ts.
+const BEDROCK_TOOL_CHOICE_VALUES = ['auto', 'any', 'tool'] as const;
+
+// LangChain reads AWS_DEFAULT_REGION only, and its own error names just that one — unhelpful to
+// anyone who set AWS_REGION, the variable the AWS SDK and CLI treat as primary.
+function resolveBedrockRegion(region?: string): string {
+  const resolved = region ?? process.env.AWS_REGION ?? process.env.AWS_DEFAULT_REGION;
```
EnkiP: Claude Fable 5.1 (claude-fable-5-1): **Should fix**: an embedded agent with an empty `AWS_REGION` and a set `AWS_DEFAULT_REGION` boots, then every AI step fails as unavailable.

`create-base-chat-model.ts:19` resolves with `??` so the empty string wins, while `embedded-workflow-executor.ts:97` treats it as unset and the lazy `ai-client.ts:36` build throws inside a step.

A way out: `||` in `resolveBedrockRegion`, matching `cli-core.ts:181`.
</details>

1. Set `AWS_REGION=` empty and `AWS_DEFAULT_REGION=eu-west-3`, then embed the executor with a bedrock `ai` option and no `region`.
2. Call `agent.start()` and observe it succeeds.
3. Run a workflow with an AI step and observe it fails with the AI service unavailable message.
</details>

packages/workflow-executor/src/errors.ts
```diff
@@ -308,6 +308,37 @@ export class StepTimeoutError extends WorkflowExecutorError {
...
+// where the generic message sends them to look at the workflow rather than at AI_MODEL and
+// AWS_REGION. The provider's own sentence is the actionable part, so it is carried through — these
+// AWS messages name the model and the reason, never a credential.
+// ValidationException is deliberately absent: Bedrock also raises it for request-level faults the
+// model configuration is innocent of — an oversized prompt above all, which a large record can
+// produce on a correctly configured model. Classifying those as configuration would send them down
+// the no-degrade path below and fail the step permanently instead of falling back to manual. The
+// cost is that a typo in AI_MODEL (also a ValidationException) degrades to manual rather than
+// naming itself; the warn log carries Bedrock's sentence.
+const AI_CONFIGURATION_ERROR_NAMES = ['ResourceNotFoundException', 'AccessDeniedException'];
```
EnkiP, line 316: Claude Fable 5.1 (claude-fable-5-1): **Should fix**: a mistyped Bedrock model id that passes the allowlist fails condition, record and MCP steps with an unclassified unexpected error.

`errors.ts:322` omits `ValidationException`, which Bedrock returns for a bad id that `supported-models.ts:164` let through, so `base-step-executor.ts:94` logs an unexpected error with no `errorKind`.

A way out: classify `ValidationException` on the model identity sentence only, `model identifier is invalid`, keeping oversized prompts on the degrade path.
</details>

1. Start the executor with `AI_PROVIDER=bedrock`, `AI_MODEL=eu.anthropic.claude-sonnet-4-7-v1:0` and `AWS_REGION=eu-west-1`, which boots.
2. Run a workflow whose first step is a condition step.
3. The step fails with `Unexpected error during step execution` and no configuration classification.
</details>

EnkiP (commented):
**Spec (no ticket)**: the functional check was limited to the PR description. It conforms except one line: the description says a `ValidationException` surfaces as `AiModelUnusableError`, and since 032024e it degrades or lands in the generic error instead.

packages/ai-proxy/src/create-base-chat-model.ts
Scra3: Fixed in 25fe21f24 — real, and the reproduction is exact.

`resolveBedrockRegion` used `??` where `cli-core.ts:181` uses `||`, so the two entry points disagreed on what an empty string means. The embedded boot check reads `''` as unset and lets the deployment through; `??` then keeps it and throws at model construction, inside the first AI step. Reproduced before the fix:

```
AWS_REGION='' AWS_DEFAULT_REGION=eu-west-3
→ THROW: AIBadRequestError - Bedrock requires a region…
```

Now `||`, matching the standalone path, with a test asserting the fallback to `AWS_DEFAULT_REGION` when `AWS_REGION` is set but empty. An exported-but-unfilled variable is how a container ships a value it never got, so it has to read as absent rather than as a refusal.

packages/ai-proxy/src/validate-ai-configurations.ts
Scra3: Fixed in 25fe21f24 — correct, the guard had no test reaching it.

Added one beside the allowlist cases in `ai-client.test.ts`, asserting the thrown message. Worth stating why this guard earns a test rather than being deleted as redundant with the two entry points: a direct `new AiClient(...)` meets neither, and the failure is silent — `ChatBedrockConverse` drops the key and authenticates with the ambient role instead, so the call succeeds against an account nobody chose and bills it. A refactor dropping the guard would have shipped green.

packages/workflow-executor/src/errors.ts
Scra3: Fixed in 25fe21f24 — right, and this one caught a regression I introduced earlier today.

Sequence: Macroscope raised the opposite defect on the same line (`ValidationException` classified as configuration sends an oversized prompt down the no-degrade path). I removed the name outright, which fixed that and broke this one. Both findings are correct about their own case, and the middle ground you propose is the only thing that satisfies both — I had considered it and rejected it for a bad reason: I could not verify AWS's wording from here, so I preferred asserting nothing.

That reasoning was wrong, because the failure modes are not symmetric. A pattern that fails to match costs one classification and leaves today's behaviour; a pattern that matches too much stops a workflow a human could have finished. So a narrow match with "not configuration" as the default is strictly better than no match.

Implemented as `/model identifier/i`, checked only for `ValidationException`. It cannot match `Input is too long for requested model`, which is the case that has to keep degrading.

The part I could not do from a laptop, I moved to CI instead of guessing: `bedrock.integration.test.ts` now invokes a well-formed but nonexistent id against the real account and asserts the failure is classifiable — by exception name, or by that sentence. If AWS rewords it, CI turns red and prints the actual string, which is how we learn before a customer does.

On the premise itself: confirmed that `condition-step-executor.ts:281` and `mcp-step-executor.ts:318` call `invokeWithTool` directly, outside `withAiAssist`, so an unclassified error there does surface as "Unexpected error during step execution" with no `errorKind`.

packages/workflow-executor/src/executors/base-step-executor.ts
```diff
@@ -343,6 +345,14 @@ export default abstract class BaseStepExecutor<TStep extends StepDefinition = St
       // Detect the timeout via our own signal, not the thrown error's name: providers wrap an
       // aborted request differently (AbortError, TimeoutError, APIUserAbortError, …).
       if (timeoutS !== undefined && signal?.aborted) throw new AiInvokeTimeoutError(timeoutS);
+
+      if (isAiConfigurationError(err)) {
```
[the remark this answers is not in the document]
Scra3: Fixed, in two rounds — the first one was mine and it overshot.

Your finding is correct: `ValidationException` is Bedrock's catch-all, an oversized prompt raises it on a model that works, and classifying it as configuration made `AiModelUnusableError` bypass `withAiAssist` and fail the step permanently.

First fix (e879b44): removed the name from the list entirely. That closed this, and opened the mirror defect — a mistyped `AI_MODEL` also raises `ValidationException`, and condition and MCP steps call `invokeWithTool` outside the degrade path, so it reached operators as "Unexpected error during step execution" with no `errorKind`. A separate reviewer raised it on `errors.ts` and was right.

Second fix (25fe21f24): classify `ValidationException` again, but only when the message names the model identity (`/model identifier/i`), which cannot match `Input is too long for requested model`. Default stays "not configuration", so a reworded AWS message costs a classification rather than stopping a workflow.

The sentence is now pinned by an integration test against the live API, so a reword turns CI red instead of reaching a customer. Both cases have a unit test that fails without the change.

.github/workflows/build.yml
```diff
@@ -145,10 +147,15 @@ jobs:
           key: ${{ runner.os }}-build-${{ github.sha }}
           fail-on-cache-miss: true
       - name: Run LLM integration tests
-        run: yarn workspace @forestadmin/ai-proxy test --testPathPattern='llm.integration'
+        run: yarn workspace @forestadmin/ai-proxy test --testPathPattern='llm.integration|bedrock.integration'
```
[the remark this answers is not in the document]
Scra3: Correct on the facts, and it caught an overclaim of mine — corrected in 037198e78.

`continue-on-error: true` on `llm-integration-tests` and the absence of any downstream `needs` are both pre-existing on `main`; this PR only extended the test pattern and added the Bedrock env. But my own comment in `bedrock.integration.test.ts` claimed the CI guard turns a missing region into "a failure", which under `continue-on-error` is a red job nobody is gated on. That comment now says what is true.

I am not moving Bedrock into a gating job, and the reason is the one that put `continue-on-error` there in the first place: every suite in that job reaches a third party. The Bedrock sweep in particular walks the catalogue serially with `maxRetries: 0` and already fails the run on throttling — deliberately, so a mostly-throttled sweep cannot read as proof. Gating on it would redden `main` whenever AWS throttles or an entitlement lapses, which is a worse failure than the one being prevented.

Splitting the deterministic part (secrets present) from the flaky part (the sweep) into two jobs would give the gate without the flakiness, and I would support it — but as a change to how all three providers gate, not as a Bedrock-only exception landed here. Happy to open a ticket for it.

.github/workflows/build.yml
```diff
@@ -145,10 +147,15 @@ jobs:
...
         env:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          # The Bedrock suite fails the job when AWS_REGION is unset: a silent skip would let the
+          # provider rot untested. Set BEDROCK_AWS_* in the repository secrets.
+          AWS_REGION: ${{ secrets.BEDROCK_AWS_REGION }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.BEDROCK_AWS_ACCESS_KEY_ID }}
```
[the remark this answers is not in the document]
Scra3: Real, and worth a ticket — but it is not specific to this PR and I am not fixing it here.

The exposure is pre-existing: `OPENAI_API_KEY` and `ANTHROPIC_API_KEY` sit in the same `env` block on `main`, in the same `pull_request`-triggered job. This PR adds a third secret to an established pattern. Fixing it properly — OIDC with a short-lived role behind a protected environment — covers all three and changes how the repo handles CI credentials; doing it for AWS alone inside a Bedrock PR would leave the other two exposed and bury a repo-wide decision in an unrelated review.

Two things that bound the blast radius, neither of which makes the finding wrong:

- The trigger is `pull_request`, not `pull_request_target`, so fork PRs get no secrets at all. The reachable attacker is someone who already has write access to this repository.
- The IAM user is scoped to Bedrock, and narrowly: this run's own log shows it denied on `bedrock:InvokeModel` for the `global.*` profiles. It is not a general-purpose AWS credential.

Separately, and regardless of this finding: that key needs rotating for an unrelated reason — it was shared in clear while we were setting the CI up, before a dedicated one existed. That was already on the list.

hercemer42 (approved):

PR #1898, opened 2026-09-14
fix(agent-bff): order composite key segments by the record, not by the apimap

Tonours:
**Stacked on #1894** (`fix/agent-bff-4xx-body-leak`), which merges first. This PR targets that branch, so the diff shown here is this change alone.

## What changed

`__forest.primaryKey` paired the `|`-split segments of a record's packed id with the collection's primary keys **by index**. That index came from the apimap, which sorts its fields alphabetically (`generator-collection.ts`), while the agent packs in declaration order (`agent/src/utils/id.ts`). Neither order is contractual and they agree only by accident.

In front of `forest-express-sequelize` 9.3.8, with a `tenant_id` + `seq` table: the apimap publishes `seq` first, the agent packs `acme|42`, and every list of that collection answered

```
500 mapping_error: Cannot build primary key: invalid numeric value "acme" for "seq"
```

The record's own attributes now decide which segment belongs to which key. A segment equal to a key's attribute value belongs to that key, whatever order the schema published.

## Why the record gives the order and not the value

The record decides the ordering and nothing else. Every emitted value still comes from the packed id, cast exactly as before. Reading the value from the record instead looks simpler and is wrong:

- `jsonapi-serializer` overwrites the `id` attribute with the resource id (`deserializer-utils.js:90`). A key named `id` would read back `acme|42`, and a single `Number` `id` key would read `'42'` where the contract promises `42`.
- an attribute can be `null`, a boolean, a relation object written over the same key, or the JSON of a Buffer, and none of those fit `Record<string, string | number>`.
- a `Date` key serializes as ISO in the attributes but as `String(date)` in the packed id, and only the packed form round-trips through the agent's `unpackId`.

A key that matches nothing, `id` included, keeps its positional segment. That is the behavior that shipped before.

## Scope and safety

The matcher skips any key whose response key another field of the collection shares. The record then holds a single value under that key and nothing here can tell whose it is, so reading it could claim a sibling's segment. `ReadModel` flags those keys with `ambiguousRecordKey`, reusing `groupByRecordKey`, the collision detector the context builder already relies on.

What does **not** change: single-key collections, derived keys (#1895), the arity and numeric-cast errors, and every call that takes an id opaque (`parentId`, `recordIds`). A request whose record cannot answer falls back to the positional split, errors included.

Out of scope: a liana that publishes a single `isPrimaryKey` on a composite-key table. `forest-rails` flags only `tenant_id` and its JSON:API id carries no second component, and no BFF code recovers a value the agent never sent.

## How to test

Against a v1 agent with a composite-key collection whose fields do not sort in declaration order:

```
POST /agent/v1/<collection>/list   {"page": {"limit": 5, "offset": 0}}
```

Before: `500 mapping_error`. After: `200`, with `__forest.primaryKey` naming both columns.

Measured on the conformance bench in front of `forest-express-sequelize` 9.3.8: `pk-composite` goes from `fail 500` to `pass 200`, and that agent's column from 23 pass / 2 fail to 24 pass / 1 fail. The one failure left is `caps-route`, which is structural, since no v1 liana serves the capabilities route.

`packages/agent-bff`: 95 suites, 1888 tests, 0 failures. `tsc --noEmit` clean, `eslint src test` reports no error on the diff.

### General

### Security

https://claude.ai/code/session_01TrJS9jCkAfmyWQc1PdSEA2


packages/agent-bff/src/data/pack-id.ts
```diff
@@ -20,6 +21,73 @@ function toNumberIfLossless(value: string): string | number {
...
+
+    if (index === -1) return null;
+    claimed[index] = true;
+
+    return values[index];
+  });
+
+  const leftovers = values.filter((_, index) => !claimed[index]);
+
+  return matched.map(value => value ?? (leftovers.shift() as string));
```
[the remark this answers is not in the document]
Tonours: Fixed in 61233ec64, with a different rule than the suggested diff.

The failure mode is real: an unmatched key could be shifted off a segment the published order happened to get right. But returning `values` whole whenever two keys stay unmatched also throws away a match that is *proven* correct, so neither version dominates the other.

The rule now is: an unmatched key keeps its own segment, and moves only when a matched key claimed it. That is a strict improvement over pairing by position, because a key whose segment was claimed by a proven match was wrong under the positional pairing by definition. On the example given (`[createdAt, id, seq]`, packed `42|<date>|7`), `seq` matches its own segment, `createdAt` and `id` both keep theirs, and the outcome is identical to the positional one.

Two tests cover it: one where no match claims anything and every unmatched key stays put, one where a match takes another key's segment and only that key moves.

nbouliol (commented):
**Spec**: no Linear ticket is linked, so the functional check was limited to the PR description, which the diff satisfies.

**Applies to:** `packages/agent-bff/src/openapi/unfolded-paths.ts:463` (not in this diff)
Claude Fable 5.1 (claude-fable-5-1): **Should fix**: the OpenAPI document tells a client to build a composite parentId in apimap order, which the agent does not unpack.

`unfolded-paths.ts:460` describes the composite id as the read-model keys joined by `|` in that order, the apimap order the agent does not pack in.

A way out: drop the `in that order` claim and state the id is taken opaque from a record's `id`.
</details>

1. Generate the unfolded document for a collection declared `tenant_id`, `seq`, with a to-many relation.
2. Read the relation route's `parentId` description, which says the values of `seq, tenant_id` joined by `|` in that order.
3. Send `42|acme` as `parentId`, and the agent unpacks `tenant_id=42`, `seq=acme`, so the relation list fails.
</details>

packages/agent-bff/src/data/pack-id.ts
```diff
@@ -20,6 +21,85 @@ function toNumberIfLossless(value: string): string | number {
...
+    const index = wanted === null ? -1 : values.findIndex((v, i) => !claimed[i] && v === wanted);
+
+    if (index === -1) return null;
+    claimed[index] = true;
+
+    return values[index];
+  });
+
+  const placed = matched.map((value, index) => {
+    if (value !== null || claimed[index]) return value;
```
nbouliol: Claude Fable 5.1 (claude-fable-5-1): **Should fix**: a composite key with two unreadable columns can list as 200 with two columns swapped, where it used to 500.

`pack-id.ts:92` lets a matched key displace an unmatched one, which then takes any leftover: keys `id`, ambiguous `owner_id`, `sku` packed `ab|9|5` swap `id` and `owner_id`.

A way out: return `values` whole when more than one key is unmatched and a match displaced any of them.
Also moves: `pack-id.test.ts:280`, `pack-id.test.ts:294`, and the strict-improvement claim in the docblock at `pack-id.ts:65`.
</details>

1. Declare a collection with primary keys `sku`, `id`, `owner_id` in that order, plus a non-key column `ownerId`.
2. List it through the BFF, so the agent packs `ab|9|5` and the apimap publishes `id, owner_id, sku`.
3. The row answers 200 with `__forest.primaryKey.id` as 5 and `owner_id` as 9, swapped.
</details>

Tonours: Confirmed and fixed in a5f32c202. Your reproduction is exact: keys `id`, ambiguous `owner_id`, `sku` packed `ab|9|5` returned 200 with `id` and `owner_id` swapped, where the positional pairing used to raise its numeric-cast error on `id`. A silent wrong key is worse than the 500, so the 'strict improvement' claim in the docblock was false and is gone.

The guard is the one you suggested, with the condition read off the failing case: the reordering is dropped whenever more than one key stays unread AND a match displaced any of them. A single unread key is still placed, since the one segment left over is necessarily its own.

My first attempt keyed the guard on the number of DISPLACED keys rather than unread ones, and your case slipped through it: only `id` ends up displaced there, `owner_id` keeps its own slot and is wrong anyway. The test caught it.

Added `should keep the positional pairing, error included, when two displaced keys stay unread`, which is your triple verbatim. The three earlier cases are unaffected, each has one unread key or none displaced.

Tonours:
On the second finding, the `parentId` description in `unfolded-paths.ts:463`: fixed in 02c8a68af, even though that line is not in this PR's diff. It is the same defect as the one this PR is about, stated in prose instead of code, and leaving it would have kept telling clients to assemble an id the agent unpacks onto the wrong columns.

The description no longer claims an order a client can rely on. It names the keys, says the id is taken verbatim from a record's own `id`, and says to copy it from a listed record rather than assemble it.

Both findings are in. `packages/agent-bff`: 95 suites, 1891 tests, 0 failures, `tsc --noEmit` clean, `eslint src test` no error.

packages/agent-bff/src/data/pack-id.ts
nbouliol: Claude Fable 5.1 (claude-fable-5-1): **Reopened**: two unread keys can still list as 200 swapped when neither is displaced, where the base 500'd.

Predicate: more than one unread key, and a cast-valid output differing from the truth where positional pairing threw.
A port of `segmentsByKey` over 1230 orders: keys `id`, ambiguous `owner_id`, `seq`, `sku` declared `owner_id, id, sku, seq` answer `{ id: 5, owner_id: 9 }`.
Not pinned: `pack-id.test.ts:155` covers the displaced triple only, and the closing property is `unread > 1` alone, the finding's displacement route being too narrow.
</details>

nbouliol:
Claude Fable 5.1 (claude-fable-5-1): **Closed**: the composite parentId description no longer names an order a client could assemble.

Predicate: the document states a key order for a composite parentId that differs from the order the agent packs in.
Read `unfolded-paths.ts:460` at the head: it names the agent's own order and tells the client to copy the id from a listed record.
Not pinned: `openapi-unfolded.test.ts:484` matches a substring the pre-fix text also contained, so the missing assertion is filed as PRD-1221.
</details>

packages/agent-bff/src/data/pack-id.ts
Tonours: Reopen confirmed, fixed in cc29a2b0e. Your four-key case is the one my guard missed: `id` and ambiguous `owner_id` both sit on a published segment, neither is displaced, and both are wrong. Displacement was the wrong closing property.

The rule is now `unread > 1` alone, as you stated it: as soon as a second key cannot be read from the record, the whole reordering is dropped and the positional pairing stands, 500 included. That also deletes the keep-your-own-slot pass, since with at most one unread key the single leftover segment is necessarily its own. Net 16 lines removed.

Added your triple verbatim as `should keep the positional pairing when two unread keys each sit on a published segment`, alongside the three-key one from the first round.

On the parentId description not being pinned (PRD-1221): pinned in 5d702a509. `openapi-unfolded.test.ts` now asserts the new sentences and asserts the old `in that order` is absent, so a revert fails the test. Close the ticket if that covers it.

95 suites, 1892 tests, 0 failures. tsc and eslint clean.

nbouliol: Claude Fable 5.1 (claude-fable-5-1): **Closed**: a second unread key now drops the reordering, so no cast-valid pairing the record did not prove can reach a 200.

Predicate: more than one unread key, and a cast-valid output differing from the truth where positional pairing threw.
Executed a port of `pack-id.ts:97` over the same 1230 declaration orders of six columns, and no input satisfies it any more.
Pinned: `pack-id.test.ts:155` fails on the previous guard, which returned `{ id: 9, owner_id: '5', seq: 7, sku: 'ab' }` for that input.
</details>

nbouliol:
Claude Fable 5.1 (claude-fable-5-1): **Closed**: the composite parentId description is now pinned to its new wording, so PRD-1221 is done.

Predicate: the document states a key order for a composite parentId that differs from the order the agent packs in.
`unfolded-paths.ts:460` is unchanged since the last pass, and `openapi-unfolded.test.ts:485` now asserts the new sentences and the absence of `in that order`.
Pinned: those three assertions fail on the pre-fix wording, so the ticket filed for the gap is closed.
</details>

nbouliol (approved):
Approved. Both validator findings verified closed at 5d702a5 (guard `unread > 1`, brute-forced over 1230 declaration orders; parentId description pinned by test). Merge after #1894, which this PR is stacked on.

packages/agent-bff/src/data/pack-id.ts
```diff
@@ -20,6 +21,82 @@ function toNumberIfLossless(value: string): string | number {
...
+    const wanted = comparableValue(record, key);
+    const index = wanted === null ? -1 : values.findIndex((v, i) => !claimed[i] && v === wanted);
+
+    if (index === -1) return null;
+    claimed[index] = true;
+
+    return values[index];
+  });
+
+  if (matched.filter(value => value === null).length > 1) return values;
```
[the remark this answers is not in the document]
Tonours: Not reproducible: the premise that a sparse projection leaves no key attributes on the record does not hold on either agent generation.

Measured against forest-express-sequelize 9.3.8, the very agent this PR fixes, on a `tenant_id`/`seq` collection whose published order is the reverse of the packing order:

    POST /agent/v1/edge_composite_pk/list  {"projection": ["payload"]}
    -> 200, record keys: ['__forest', 'id', 'payload', 'seq', 'tenantId']
    -> __forest.primaryKey: {"seq": 1, "tenant_id": "acme"}

The liana returns the key columns whatever the projection asks for, so the matching pass still has what it needs and the pairing is correct. On a v2 agent the same holds by construction: `withPks` is applied after redaction (`agent/src/routes/access/list.ts:33`) and the serializer does not drop key attributes.

For an agent that did omit them, `segmentsByKey` returns the positional segments, which is exactly what `main` does today for every request. That path keeps its numeric-cast error, so a wrong pairing on a Number key still fails loudly rather than returning a bad key. No regression is introduced by this PR there.

Forcing the key columns into the agent projection would fix nothing measurable and would put attributes the caller did not ask for into the response, so I am leaving the code as is.

PR #1899, opened 2026-09-14
fix(workflow-executor): apply Debian security updates to the Docker image

Scra3:
## Why

The `Validate Dockerfile build` job has been failing on every build since 12 Sep on two HIGH CVEs in `libpcre2-8-0`:

```
Total: 2 (HIGH: 2, CRITICAL: 0)
libpcre2-8-0  CVE-2026-86145  HIGH  fixed  10.42-1 → 10.42-1+deb12u1
              CVE-2026-89161
```

This is not branch-specific. `main` and every open PR fail the same step, and a commit that passed on 11 Sep fails today with no code change: the advisories landed in between.

The image builds and the smoke test passes. Only the Trivy gate (step 8, `Scan OS packages (blocking)`) fails.

## Why not just bump the digest

That was my first instinct, and I checked it before writing this: the newest `node:22-bookworm-slim` **still ships the vulnerable `10.42-1`**. Upstream has not rebuilt.

```
newest node:22-bookworm-slim → libpcre2-8-0 10.42-1
pinned digest 813a7480…      → libpcre2-8-0 10.42-1
```

The digest pin is the right call for reproducibility, but it also guarantees Debian security updates never reach the image on their own. Applying them in the `base` stage addresses the class of failure rather than this instance of it, so the next advisory does not need a digest hunt.

## What

- The `runtime` stage runs `apt-get update && apt-get upgrade -y --with-new-pkgs` and drops the index afterwards. In `runtime` rather than `base` so a refresh rebuilds `COPY` layers, not `yarn install` and `lerna run build`.
- `APT_EPOCH` build arg, set by CI to `github.run_id-run_attempt`. It sits in the layer's cache key, so every run refetches the Debian index: a rebuild of a released tag, a re-run of a failed job and both arches of one release all resolve against the current archive. The gated build and the push share the value, so the push is a cache hit and ships the image the scan saw.
- `apt-get update -o APT::Update::Error-Mode=any`: an unreachable mirror fails the build instead of exiting zero on a half patched layer.

## Verified locally, with the CI's own scan settings

`--scanners vuln --pkg-types os --severity CRITICAL,HIGH --ignore-unfixed`:

| | Before | After |
|---|---|---|
| Trivy HIGH/CRITICAL | 2 | **0** |
| `libpcre2-8-0` | `10.42-1` | `10.42-1+deb12u1` |
| `docker/smoke-test.sh` | pass | pass |
| Image size | ~410 MB | ~435 MB |

Also checked the cache key: a second build with the same `APT_EPOCH` shows the apt layer as `CACHED`, a different value re-runs it.

## Trade off, stated plainly

The image is reproducible from the commit plus the build date, not from the commit alone. That is the point: the digest pin had frozen the package set and nobody noticed until CI went red. What stays fixed is the base digest, the lockfile-driven dependency tree, the SBOM on every push and the blocking scan before publish. The README no longer claims byte reproducibility.

Unblocks the Docker job on `main` and on every open PR, including #1897.

> [!NOTE]
> ### Apply Debian security updates to `workflow-executor` Docker image
> - Adds an `apt-get update && apt-get upgrade -y --with-new-pkgs` step in the runtime stage of [Dockerfile](https://github.com/ForestAdmin/agent-nodejs/pull/1899/files#diff-d7484fdcdb83aeab152a6c03a53f6a75ea7d34e6a5d6ce41374fe9f3cd1ff9b5) so published images install current Debian security and package updates rather than relying solely on the pinned base image.
> - Introduces an `APT_SECURITY_EPOCH` build arg composed of the current UTC ISO week plus a serial read from [apt-refresh](https://github.com/ForestAdmin/agent-nodejs/pull/1899/files#diff-4b48d6e956c153da3ec85a513b3d18099d7aba2e979c664f00c4d2357bf4928d); the epoch busts the Docker build cache to force APT layer refresh on week rollover or serial bump.
> - The epoch is computed in both the validate and publish jobs in [docker-publish.yml](https://github.com/ForestAdmin/agent-nodejs/pull/1899/files#diff-5b21991be47c2922383bdc0b6bf00b65af7db51b82d049dd8d6ad03e3d37ac98) and passed through the gated and push-by-digest builds so the published image matches the one that was validated.
> - Risk: base package contents now vary over time, so reproducibility of runtime layers depends on archive availability and the epoch value; bump the serial in `apt-refresh` to force a refresh.
>
> 
> #### Changes since #1899 opened
>
> - Replaced `APT_SECURITY_EPOCH` build argument with `APT_EPOCH` across the `docker-publish` workflow and `workflow-executor` Dockerfile, computing the epoch value from `github.run_id` and `github.run_attempt` instead of a weekly security epoch [d3f93d2]
> - Added `-o APT::Update::Error-Mode=any` flag to the `apt-get update` command in the `workflow-executor` Dockerfile runtime stage [d3f93d2]
> - Deleted the `apt-refresh` file from the `workflow-executor` docker directory [d3f93d2]
> - Updated the documented image size from approximately 410 MB to approximately 435 MB in the `workflow-executor` docker README and clarified that the dependency tree is fully pinned [d3f93d2]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2c4b08f.</sup>
> 
>


packages/workflow-executor/Dockerfile
```diff
@@ -11,6 +11,15 @@
 # Bump via Renovate/Dependabot.
 FROM node:22-bookworm-slim@sha256:813a7480f28fdadac1f7f5c824bcdad435b5bc1322a5968bbbdef8d058f9dff4 AS base
 
+# Debian ships security fixes faster than the upstream node image rebuilds, and the digest pin means
+# we would otherwise never receive them: the Trivy gate below went red on CVE-2026-86145 and
+# CVE-2026-89161 (libpcre2) while the newest node:22-bookworm-slim still carried the vulnerable
+# 10.42-1. Applying the security updates here keeps the pin reproducible and the image patched
+# without chasing a digest per advisory.
+RUN apt-get update \
```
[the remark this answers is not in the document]
Scra3: Right, and it was the most serious thing wrong with the first version — thanks. Fixed in b2d34e5, though not quite the way you suggested.

**The diagnosis needs one correction.** A stale layer cannot put a vulnerable image in the registry: the publish job scans the loaded image with a blocking Trivy gate and only pushes afterwards (`docker-publish.yml`, `Scan OS packages (blocking)` → `Push by digest`). So the real failure mode is not a vulnerable release, it is a CI that goes red and **stays** red through re-runs, because every retry restores the same cached layer. Worth stating precisely, since it changes who is harmed and how they find out.

**On the fix:** I did not take the rotated-`ARG` shape as written, because "deliberately rotated" means a human has to remember. Nobody rotates it, and the person staring at a red Trivy gate has no way to discover the knob exists — so it reads as solved while behaving exactly like today. Instead CI computes the value:

```yaml
- name: Compute apt security epoch
  id: apt
  run: echo "epoch=$(date -u +%G-W%V)" >> "$GITHUB_OUTPUT"
```

passed as `build-args: APT_SECURITY_EPOCH=${{ steps.apt.outputs.epoch }}` to the validate build, the gated build and the push. Same value in the gated build and the push on purpose: a different one there would rebuild the apt layer and publish an image the scan never saw.

**Also moved the layer from `base` to `runtime`.** Identical shipped packages, since `runtime` is `FROM base`, but a buster in `base` invalidates `yarn install` and `lerna run build` beneath it. In `runtime` there is nothing below but `COPY` lines, so the weekly refresh costs ~30s instead of a cold monorepo rebuild. That was the part of your suggestion I could not take as-is.

Verified locally: the layer shows CACHED on an unchanged epoch and re-runs on a changed one, `libpcre2-8-0` lands at `10.42-1+deb12u1`, Trivy reports 0, and `docker/smoke-test.sh` passes.

packages/workflow-executor/Dockerfile
```diff
@@ -61,6 +60,22 @@ FROM base AS runtime
...
+# the ISO week, so a cached layer expires on its own. Without it the layer keys on an unchanging
+# instruction and an unchanging pinned parent, and would never contact Debian again.
+# To force a rebuild before the week turns, bump docker/apt-refresh in a PR; re-running the job on
+# its own restores the same cached layer.
+# In `runtime` rather than `base` so busting it rebuilds COPY layers, not yarn install and lerna.
+# DEBIAN_FRONTEND inline, never ENV: an ENV here would leak into the shipped image.
+ARG APT_SECURITY_EPOCH=unset
+RUN echo "apt epoch ${APT_SECURITY_EPOCH}" \
+  && DEBIAN_FRONTEND=noninteractive apt-get update \
+  && DEBIAN_FRONTEND=noninteractive apt-get upgrade -y -o Dpkg::Options::=--force-confold \
```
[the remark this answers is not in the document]
Scra3: Correct, and it is the sharpest of the three — fixed in 1e4b065. `apt-get upgrade` holds back any fix whose package gained a dependency or was split, so the upgrade reports success, the CVE survives, and the gate stays red pointing at a refresh that looks like it already ran. That failure is quiet in exactly the wrong way.

Took `--with-new-pkgs` as suggested. It is also the reason not to reach for `dist-upgrade`: both allow the additions, only `dist-upgrade` also allows removals, and an unsupervised removal in a base image is not something the smoke test would catch.

Verified: `libpcre2-8-0` at `10.42-1+deb12u1`, Trivy 0 findings, `docker/smoke-test.sh` passes.

packages/workflow-executor/Dockerfile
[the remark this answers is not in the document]
Scra3: You are right that reproducibility is not restored, and I am declining the fix rather than disputing the diagnosis — so let me be explicit about the trade instead of leaving it implicit.

Reproducible output and automatic patching are in direct opposition here. Pinning a snapshot or explicit versions gives back byte-identical builds, and in exchange the image is frozen at that snapshot until a human bumps it. That is precisely the failure this PR exists to fix: the digest pin already froze the package set, nobody noticed for months, and CI went red on two HIGH CVEs that Debian had fixed. Snapshot pinning reproduces that shape with a different knob.

There is a coherent design where both hold — pin a snapshot date and have the weekly job bump the date — but it is materially more machinery than the problem in front of us, and it still ships whatever that week's snapshot contains. It is not more reproducible in any sense that matters to an operator, only more auditable.

So the choice is deliberate: the image is reproducible from the commit **plus the build date**, not from the commit alone. What we keep is the pinned base digest, a lockfile-driven dependency tree, an SBOM on every push, and a blocking scan before publish, so a build is auditable after the fact even when it is not bit-identical.

The Dockerfile now states this in the comment rather than claiming otherwise — the first version of this PR asserted it "keeps the pin reproducible", which was simply false and is the kind of comment worth catching.

If the team later wants determinism back, the honest form is targeted `--only-upgrade` of named packages, one line per advisory, which trades the automation away openly. Worth a separate discussion, not this PR.

.github/workflows/docker-publish.yml
```diff
@@ -214,6 +234,9 @@ jobs:
           platforms: ${{ matrix.platform }}
           outputs: type=image,name=ghcr.io/forestadmin/workflow-executor,push-by-digest=true,name-canonical=true,push=true
           sbom: true
+          # Same epoch as the gated build above: a different value would rebuild the apt layer
+          # and push an image the scan never saw.
+          build-args: APT_SECURITY_EPOCH=${{ steps.apt.outputs.epoch }}
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): **Should fix**: A cache miss on the push build publishes an apt layer the blocking scan never examined.

`docker-publish.yml:228` rebuilds rather than pushing the gated image, and `Dockerfile:79` resolves versions against the live archive.

A way out: scan `ghcr.io/forestadmin/workflow-executor@${{ steps.build.outputs.digest }}` after the push and before `merge`.
</details>

1. Dispatch the workflow for a version.
2. The gated build loads an image and both scans pass.
3. The push build misses cache and re-runs apt against a newer archive.
4. GHCR receives a digest no gate examined, and `notify-server` deploys it.
</details>

packages/workflow-executor/Dockerfile
```diff
@@ -61,6 +60,25 @@ FROM base AS runtime
...
+# To force a rebuild before the week turns, bump docker/apt-refresh in a PR; re-running the job on
+# its own restores the same cached layer.
+# In `runtime` rather than `base` so busting it rebuilds COPY layers, not yarn install and lerna.
+# DEBIAN_FRONTEND inline, never ENV: an ENV here would leak into the shipped image.
+# --with-new-pkgs because plain `upgrade` holds back any fix that pulls a new dependency, which
+# would leave the CVE in place while the scan still reports it; it stops short of dist-upgrade,
+# which may also remove packages.
+ARG APT_SECURITY_EPOCH=unset
+RUN echo "apt epoch ${APT_SECURITY_EPOCH}" \
+  && DEBIAN_FRONTEND=noninteractive apt-get update \
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): **Should fix**: A failed security index fetch builds a half patched image and the step still exits zero.

`apt-get update` returns 0 when an index fails, and the base ships `/var/lib/apt/lists/` empty so nothing older covers it.

A way out: `apt-get update -o APT::Update::Error-Mode=any`, which exits non zero on any failed index.
</details>

1. A Debian security mirror is briefly unreachable during a release build.
2. `apt-get upgrade` patches what it can and the RUN chain exits 0.
3. The layer is exported to the GHA cache under this week's epoch.
4. Every build that week is served that half patched layer.
</details>

.github/workflows/docker-publish.yml
```diff
@@ -167,6 +177,15 @@ jobs:
       # exact image before it reaches the registry. Nothing is published until
       # both gates pass — otherwise a vulnerable/broken image would be pullable
       # by digest even when the gate "fails".
+      - name: Compute apt security epoch
+        id: apt
+        # Weekly, plus a serial anyone can bump in a PR. A repository variable would have been
+        # simpler but only maintainers can edit those, and the person facing the red gate is
+        # usually the PR author — a lever they cannot reach is not a lever.
+        run: |
+          echo "epoch=$(date -u +%G-W%V)-$(cat packages/workflow-executor/docker/apt-refresh)" \
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): **Should fix**: A rebuild of an already cut version tag can never bust the apt layer, and nothing reports why.

`docker-publish.yml:164` checks out the version tag, so the serial read here is the one baked into that tag.

A way out: let `workflow_dispatch` supply an epoch override, and read `apt-refresh` into its own variable.

Also moves: the validate copy at `docker-publish.yml:60` needs the same read.
</details>

1. Release 1.28.0 on Monday and its apt layer caches under this week's epoch.
2. A HIGH CVE is fixed in the Debian archive on Wednesday.
3. Dispatch a rebuild of 1.28.0 to ship the fix.
4. The tag's serial is unchanged, the cached layer returns, and the scan stays red.
</details>

packages/workflow-executor/Dockerfile
```diff
@@ -61,6 +60,25 @@ FROM base AS runtime
...
+# the ISO week, so a cached layer expires on its own. Without it the layer keys on an unchanging
+# instruction and an unchanging pinned parent, and would never contact Debian again.
+# To force a rebuild before the week turns, bump docker/apt-refresh in a PR; re-running the job on
+# its own restores the same cached layer.
+# In `runtime` rather than `base` so busting it rebuilds COPY layers, not yarn install and lerna.
+# DEBIAN_FRONTEND inline, never ENV: an ENV here would leak into the shipped image.
+# --with-new-pkgs because plain `upgrade` holds back any fix that pulls a new dependency, which
+# would leave the CVE in place while the scan still reports it; it stops short of dist-upgrade,
+# which may also remove packages.
+ARG APT_SECURITY_EPOCH=unset
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): **Should fix**: A build following this file's own usage line freezes the apt layer at whatever the first build fetched.

The default `unset` is a constant, and `Dockerfile:6` plus `example/docker-compose.executors.yml:42` both build without the arg.

A way out: drop the default and assert `APT_SECURITY_EPOCH` inside the RUN, or pass the week in both paths.
</details>

1. Follow the usage line at `Dockerfile:6` to build the image locally.
2. Build it again weeks later with a warm BuildKit cache.
3. The apt layer is served from cache and never contacts Debian again.
</details>

.github/workflows/docker-publish.yml
```diff
@@ -175,6 +194,7 @@ jobs:
           platforms: ${{ matrix.platform }}
           load: true
           tags: workflow-executor:${{ matrix.arch }}
+          build-args: APT_SECURITY_EPOCH=${{ steps.apt.outputs.epoch }}
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): **Preferential**: The two arches of one published tag can carry Debian package sets from different weeks.

`docker-publish.yml:180` runs once per matrix leg, and arm64 queues longer, so a Sunday midnight dispatch can straddle the week.

A way out: compute the epoch once in `extract-version` and expose it as a job output.
</details>

packages/workflow-executor/Dockerfile
```diff
@@ -61,6 +60,25 @@ FROM base AS runtime
...
+RUN echo "apt epoch ${APT_SECURITY_EPOCH}" \
```
hercemer42: Claude Opus 5 (claude-opus-5[1m]): **Preferential**: The echo is the whole cache busting mechanism and reads as leftover debug output.

BuildKit keys the layer on the expanded command, so dropping this `echo` would silently refreeze it.

A way out: one line in the comment block at `Dockerfile:63` saying the echo is load bearing.
</details>

hercemer42 (commented):
**Spec**: no Linear ticket is linked anywhere on this PR, so the functional check was limited to the PR description.

**Applies to:** `packages/workflow-executor/docker/README.md`
Claude Opus 5 (claude-opus-5[1m]): **Should fix**: The docker README still promises a reproducible 410 MB image that this change no longer builds.

`docker/README.md:5` claims roughly 410 MB and reproducible, while the PR measures 435 MB and `Dockerfile:63` drops reproducibility.

A way out: scope the claim to the dependency install and document the `apt-refresh` lever there.
</details>

1. Open `docker/README.md` to size the runtime image.
2. It reports 410 MB and calls the build reproducible.
3. Pull the published image and measure 435 MB with OS packages resolved at build time.
</details>


Claude Opus 5 (claude-opus-5[1m]): **Preferential**: The reproducibility trade off lives only in Dockerfile comments, so the next reader relitigates it.

`Dockerfile:63` is the only record of the trade off, and no ADR under `docs/adr/**` covers it.

A way out: run `/adr` on the decision.
</details>

hercemer42 (changes requested):

packages/workflow-executor/Dockerfile
```diff
@@ -61,6 +60,25 @@ FROM base AS runtime
...
+# APT_SECURITY_EPOCH exists only to bound how long this layer may be served from cache — CI passes
+# the ISO week, so a cached layer expires on its own. Without it the layer keys on an unchanging
+# instruction and an unchanging pinned parent, and would never contact Debian again.
+# To force a rebuild before the week turns, bump docker/apt-refresh in a PR; re-running the job on
+# its own restores the same cached layer.
+# In `runtime` rather than `base` so busting it rebuilds COPY layers, not yarn install and lerna.
+# DEBIAN_FRONTEND inline, never ENV: an ENV here would leak into the shipped image.
+# --with-new-pkgs because plain `upgrade` holds back any fix that pulls a new dependency, which
+# would leave the CVE in place while the scan still reports it; it stops short of dist-upgrade,
+# which may also remove packages.
```
hercemer42: I got to line 3 and lost the will to live 😛   Any way to make it shorter ?

.github/workflows/docker-publish.yml
Scra3: Fair in principle, but it cannot fire here. Both builds run in the same job, on the same buildx builder, with the same APT_EPOCH (d3f93d2 uses the run id). The gated build leaves every layer in that builder's local store, so the push is a hit even if the GHA cache upload failed. A miss would need the builder destroyed mid-job, and then the job fails and nothing is pushed. Skipped the post-push scan for that reason. If you see a path to a miss I am missing, say so and I will add it.

packages/workflow-executor/Dockerfile
Scra3: Taken as suggested in d3f93d2. The build accepts the flag on bookworm's apt; I did not simulate a dead mirror.

.github/workflows/docker-publish.yml
Scra3: Solved by removing the mechanism. d3f93d2 keys the layer on github.run_id and run_attempt, so a dispatched rebuild or a re-run of a failed job always refetches. The serial file is gone with it.

packages/workflow-executor/Dockerfile
Scra3: Half taken. The default is gone in d3f93d2. I did not make the arg required: local and compose builds are never published or scanned, and --no-cache already refreshes them. The comment says so in one line.

.github/workflows/docker-publish.yml
Scra3: Same value on both legs now, github.run_id, set by Actions rather than by each runner's clock. Fixed in d3f93d2.

packages/workflow-executor/Dockerfile
Scra3: The comment now says the arg is part of the cache key (d3f93d2). Kept the echo rather than relying on BuildKit invalidating any RUN after a changed ARG: that behaviour is documented, but an unused ARG reads as dead code and someone deletes it.

packages/workflow-executor/Dockerfile
Scra3: Yeah. 13 lines down to 3 in d3f93d2, and the thing it was explaining is gone too: no weekly epoch, no serial file, no compute step. CI passes the run id as a build arg and that is the whole mechanism.

Scra3:
README: fixed in d3f93d2. It now says ~435 MB with a fully pinned dependency tree, which is what the lockfile still guarantees.

ADR: skipped. The decision is one RUN line and trivially reversible, so it fails the hard-to-inverse gate.

.github/workflows/docker-publish.yml
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Accepted: the gated build leaves the layer in the job's builder store, so the push cannot miss without failing.

Both builds run in one job off the single builder created at `docker-publish.yml:158`, and both pass the same literal arg.

One ask: the deleted comment at the push step carried this reasoning and nothing replaces it.
</details>

packages/workflow-executor/Dockerfile
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Closed: an unreachable index now aborts the layer, checked against the pinned base with the network removed.

Ran the chain at `Dockerfile:68` under `--network none`: the plain form exits 0 and this one exits 100.

The property is that `APT::Update::Error-Mode=any` turns any failed index into an error, so the chain stops.

No regression test pins it, and the `docker/` build assets have never had one.
</details>

.github/workflows/docker-publish.yml
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Closed: the serial is gone and every run keys the layer on the run id, so a rebuild always refetches.

Checked that changing the arg rebuilds the `RUN` and repeating it hits cache, so the `echo` carries the key.

The property is that `github.run_id` and `run_attempt` are unique per run, so no two builds share a key.

No regression test pins it, and the `docker/` build assets have never had one.
</details>

packages/workflow-executor/Dockerfile
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Accepted: the freeze is real but confined to builds that are never published, and the file now says so.

Confirmed a build passing no arg keys on a constant string, so that layer caches until `--no-cache`.

`Dockerfile:63` records the behaviour and the escape, which is where someone hitting it will look.
</details>

.github/workflows/docker-publish.yml
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Closed: both legs now take the same run id, which Actions sets once per run rather than per runner.

The predicate was two legs computing different epochs, which one workflow expression can no longer produce.

`docker-publish.yml:179` uses that single literal across the matrix, so the week boundary case is gone.
</details>

packages/workflow-executor/Dockerfile
hercemer42: Claude Opus 5 (claude-opus-5[1m]): Closed: the comment now names the arg as part of the cache key, and the echo is kept on purpose.

The predicate was a later reader dropping the echo as dead code, which the comment now answers.

Checked that the `echo` is what carries the arg into the layer key: changing it rebuilds, repeating it caches.
</details>

hercemer42:
Claude Opus 5 (claude-opus-5[1m]): Closed: the README now reports 435 MB and scopes the pinning claim to the dependency tree.

`docker/README.md:5` no longer calls the build reproducible, which the runtime stage stopped being.

The `apt-refresh` half of the finding is moot because d3f93d2 deleted the file.

No check pins the claim, and nothing in the repo has ever guarded it.
</details>


Claude Opus 5 (claude-opus-5[1m]): Closed: the decision shrank to one RUN line in d3f93d2, which fails the hard to reverse gate.

The predicate was a decision costly to invert, and reverting is now deleting four lines.

`Dockerfile:63` keeps the reason in the file, which is what the ADR would have carried.
</details>

hercemer42 (approved):

packages/workflow-executor/Dockerfile
```diff
@@ -11,6 +11,15 @@
...
 FROM node:22-bookworm-slim@sha256:813a7480f28fdadac1f7f5c824bcdad435b5bc1322a5968bbbdef8d058f9dff4 AS base
 
+# Debian ships security fixes faster than the upstream node image rebuilds, and the digest pin means
+# we would otherwise never receive them: the Trivy gate below went red on CVE-2026-86145 and
+# CVE-2026-89161 (libpcre2) while the newest node:22-bookworm-slim still carried the vulnerable
+# 10.42-1. Applying the security updates here keeps the pin reproducible and the image patched
+# without chasing a digest per advisory.
+RUN apt-get update \
+  && apt-get upgrade -y --no-install-recommends \
+  && rm -rf /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
Scra3: Confirming what the bot already detected: addressed in b2d34e5, then refined twice.

The layer is now bounded by APT_SECURITY_EPOCH, which CI computes from the ISO week rather than from a knob someone has to remember to rotate, and it lives in `runtime` so busting it rebuilds COPY layers instead of yarn install and lerna. A developer who needs it refreshed mid-week bumps `docker/apt-refresh` in their PR — a repository variable was the first attempt and was wrong, since only maintainers can edit those and the person facing the red gate is usually the PR author.

Full reasoning, including why a stale layer could never actually publish a vulnerable image, is in the sibling thread on line 19.

PR #1901, opened 2026-09-14
fix(ai-proxy): deny gpt-live-1, a Realtime-family model OpenAI now lists

bexchauveto:
## Summary

- Root cause of the `LLM Integration Tests (ai-proxy)` CI failure on #1842 (and any other open PR right now): not a flake, a rerun would not fix it. OpenAI added `gpt-live-1` to its `/v1/models` listing; the `llm.integration` test enumerates every model that listing returns and asserts each one supports tool calls.
- `gpt-live-1` is a Realtime-family model (same as the already-denylisted `gpt-realtime`/`gpt-realtime-mini`/`gpt-audio`) — it only supports `v1/completions`, not `v1/chat/completions`, and 404s with *"This is not a chat model and thus not supported in the v1/chat/completions endpoint."* `isModelSupportingTools` didn't recognize its prefix, so it wrongly let it through.
- `isModelSupportingTools` isn't test-only — `Router` and `validate-ai-configurations` both gate on it in production, so the same gap would let a customer configure `gpt-live-1` as their `ai-proxy` model and hit the identical 404 on every request.
- Fix: add `gpt-live` to `OPENAI_UNSUPPORTED_PREFIXES`, following this file's documented convention ("If a model fails the llm.integration test, add it here").

## Test plan

- [ ] `yarn workspace @forestadmin/ai-proxy test` — 410 passing (1 new unit test added, `llm.integration` skipped locally as usual, needs real API keys)
- [ ] `yarn workspace @forestadmin/ai-proxy lint` — 0 errors
- [ ] CI's `LLM Integration Tests (ai-proxy)` job goes green (currently red on `main` and every open PR until this merges)

> [!NOTE]
> ### Deny `gpt-live` models in `OPENAI_UNSUPPORTED_PREFIXES`
> OpenAI now lists `gpt-live-1`, a Realtime-family model limited to v1 completions. Adds the `gpt-live` prefix to [supported-models.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1901/files#diff-c8703b3b65d1492f147178cbdc7ea897138fe0c4fe9c842993d4141d7a8a6cd3) so `isModelSupportingTools` classifies these models as not tool-supporting. A regression test in [supported-models.test.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1901/files#diff-63c4790a9eb972b9a4cef4054c9171052b2469ff86ad24e2badf16356682b914) asserts `gpt-live-1` returns `false`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized bacc8e0.</sup>
> 
>


bexchauveto:
Managed in PR #1900

PR #1903, opened 2026-09-15
feat(agent-bff): match a single-label subdomain wildcard in the CORS allow-list

Tonours:
`BFF_ALLOWED_ORIGINS` takes an entry whose leading host label is `*`. `https://*.apps.zdusercontent.com` matches exactly one DNS label there.

## What

| File | Change |
|---|---|
| `src/cors/origin.ts` | `parseAllowedEntry` and `matchesWildcard`, wired into `parseAllowedOrigins` and `originAllowed` |
| `src/cors/per-key-origin.ts:31` | overlap check runs both directions |
| `README.md` | the three places that said `no wildcard` |
| 3 test files | 26 cases |

One `*`, leading label only, matching exactly one label. `https://1231469.apps.zdusercontent.com` passes, `https://a.b.apps.zdusercontent.com` and the apex do not. Scheme and port match exactly. Entries with no `*` are unchanged. Refused at parse time and reported by the existing boot `Warn`: a bare `*`, a `*` elsewhere, several `*`, a host left under two non-empty labels, an IP or localhost.

## Why

An installed Zendesk app is served from `https://<app-id>.apps.zdusercontent.com`, measured 2026-09-15 on the Forest app, app-id `1231469`. An exact origin means asking the customer to copy an opaque id.

It also closes a silent failure. `https://*.apps.zdusercontent.com` parses as a valid URL, so the entry was accepted, never reached `invalidAllowedOrigins`, fired no warning, and never matched anything. `403 origin_not_allowed` on every call with nothing pointing at the cause. `git stash push -- packages/agent-bff/src` fails 12 of the new tests.

Two calls worth reviewing:

- **Layer 2 gets the wildcard too.** One matcher for the per-key `allowedOrigins`: same opaque-id problem, and the API key is the boundary there, not the origin. Cost: an illegal wildcard in a key's list is never warned about on its own, since those entries bypass `parseAllowedOrigins`.
- **This widens the allow-list.** Every Zendesk app of every publisher sits under `apps.zdusercontent.com`. The two-label minimum is not a TLD guard either: `https://*.co.uk` and `https://*.github.io` pass it and would allow every site under those suffixes. Acceptable because CORS is not the auth boundary here, a session or an API key is still required, and another origin cannot read the Forest app's tokens.

## How to test

```bash
yarn workspace @forestadmin/datasource-toolkit build   # or 13 suites fail to typecheck
```

Ran here: 97 suites / 1937 tests green, lint 0 errors (9 pre-existing warnings), build clean. Reviewed by a fresh-context Claude reviewer (Axis: Logic) plus an adversarial pass on the diff, on a different Claude model. Same-family, not cross-model.

### General

### Security

https://claude.ai/code/session_01J2dYRcxcGrtFfFBWHD8g6v


packages/agent-bff/src/cors/per-key-origin.ts
```diff
@@ -28,7 +28,11 @@ export default function createPerKeyOriginMiddleware({
   function reportOriginsThatCanNeverPass(allowedOrigins: string[], keyHash: string): void {
     if (serverAllowedOrigins.length === 0 || assessedKeys.has(keyHash)) return;
 
-    const canNeverPass = !allowedOrigins.some(entry => originAllowed(entry, serverAllowedOrigins));
+    const canNeverPass = !allowedOrigins.some(
+      entry =>
+        originAllowed(entry, serverAllowedOrigins) ||
+        serverAllowedOrigins.some(serverOrigin => originAllowed(serverOrigin, [entry])),
```
nbouliol: Claude Opus 5 (claude-opus-5): **Should fix**: An operator gets no warning for a key whose origins can never pass, only a per-request log blaming the browser.

`per-key-origin.ts:34` matches key `https://app-*.zendesk.com` against server `https://*.zendesk.com`, flipping `canNeverPass` from true before this diff to false.

A way out: have `parseAllowedEntry` report the refusal so `reportOriginsThatCanNeverPass` warns on a malformed entry in its own right.
</details>

1. Set `BFF_ALLOWED_ORIGINS=https://*.zendesk.com`.
2. Give a BFF API key the origin `https://app-*.zendesk.com` in the Forest SaaS.
3. Boot the BFF and read the startup log.
4. No warning names that key.
5. Call a data route from `https://app-123.zendesk.com` with that key.
6. Every call answers 403 `origin_not_allowed`.
</details>

packages/agent-bff/src/cors/per-key-origin.ts
```diff
@@ -28,7 +28,11 @@ export default function createPerKeyOriginMiddleware({
   function reportOriginsThatCanNeverPass(allowedOrigins: string[], keyHash: string): void {
     if (serverAllowedOrigins.length === 0 || assessedKeys.has(keyHash)) return;
 
-    const canNeverPass = !allowedOrigins.some(entry => originAllowed(entry, serverAllowedOrigins));
+    const canNeverPass = !allowedOrigins.some(
```
nbouliol: **Applies to:** `packages/agent-bff/src/cors/per-key-origin.ts:40` (not in this diff)
Claude Opus 5 (claude-opus-5): **Violates conventions**: Support cannot reach this key warning from a customer report, because the line names no rendering.

The predicate anchored here gates `per-key-origin.ts:40`, which logs `keyHash` and `keyOrigins` while `identity.renderingId` sits in scope.

A way out: thread `identity.renderingId` into `reportOriginsThatCanNeverPass`, as the sibling log at `per-key-origin.ts:52` already does.
</details>

packages/agent-bff/src/cors/origin.ts
```diff
@@ -32,6 +32,43 @@ export function normalizeOrigin(raw: string | undefined | null): string | null {
...
 
+const WILDCARD_LABEL_PREFIX = '*.';
+const MIN_WILDCARD_HOST_LABELS = 2;
+
+interface AllowedEntry {
+  origin: string;
+  wildcard: boolean;
+}
+
+function parseAllowedEntry(raw: string): AllowedEntry | null {
```
nbouliol, line 43: **Applies to:** `packages/agent/src/types.ts:167` (not in this diff)
Claude Opus 5 (claude-opus-5): **Should fix**: A customer reading the `allowedOrigins` option in their editor is told wildcards are refused, which this diff makes false.

`types.ts:167` says `No wildcard.`, and `embedded-bff.ts:61` joins that array into `BFF_ALLOWED_ORIGINS`, reaching the parser anchored here.

A way out: mirror the new README wording in the `types.ts` JSDoc on `allowedOrigins`.
</details>

1. Hover `allowedOrigins` on the embedded BFF options in an editor.
2. The JSDoc says wildcards are not supported.
3. Set the option to a list holding `https://*.apps.zdusercontent.com`.
4. The wildcard is accepted and served.
</details>

nbouliol (commented):
**Spec**: no Linear ticket is linked on this PR, so the functional check was limited to the PR description, which the diff matches.

Claude Opus 5 (claude-opus-5): **Should fix**: A reader later cannot tell why a two-label floor was chosen over a public-suffix list.

`origin.ts:36` sets `MIN_WILDCARD_HOST_LABELS = 2`, which refuses `https://*.com` and admits `https://*.co.uk`, and no accepted ADR records that choice.

A way out: run `/adr` for the wildcard admission and its `MIN_WILDCARD_HOST_LABELS` floor.
</details>

1. Read `packages/agent-bff/src/cors/origin.ts:36`.
2. Search `docs/adr/` for cors, origin, wildcard, subdomain, allow-list.
3. Nothing records the choice of a label floor over a public-suffix list.
</details>

packages/agent-bff/src/cors/per-key-origin.ts
Tonours: Confirmed and fixed in 6c4a88f16. Reproduced first: `originAllowed('https://app-*.zendesk.com', ['https://*.zendesk.com'])` returned true, because `app-*` reads as a single label. My symmetric check turned that into a silent key.

Root cause was `originAllowed` accepting a request origin carrying a `*`. It now refuses one, and the overlap check no longer goes through the request path at all: `allowedEntriesIntersect(left, right)` compares two allow-list entries, and an entry the parser refuses intersects nothing. A key holding only `https://app-*.zendesk.com` is reported at startup again. Test: `per-key-origin.test.ts`, 'reports a key whose wildcard the parser refuses'.

packages/agent-bff/src/cors/per-key-origin.ts
Tonours: Fixed in 6c4a88f16. `reportOriginsThatCanNeverPass` now takes `identity.renderingId` and logs it alongside `keyHash`, same shape as the rejection log below it. The existing assertion was tightened to the full payload rather than a partial match.

packages/agent-bff/src/cors/origin.ts
Tonours: Confirmed and fixed in 53678ed7e. `types.ts:167` said `No wildcard.` and `embedded-bff.ts:61` joins that array straight into `BFF_ALLOWED_ORIGINS`, so the JSDoc was telling customers the opposite of what ships. It now mirrors the README, including the part that `https://*.co.uk` clears the two-label floor.

Tonours:
All four findings handled, two commits.

**ADR (PR-level).** `docs/adr/2026-09-15-a-cors-wildcard-matches-one-label-over-a-host-of-at-least-two.md` records the two-label floor and why it is a label count rather than a public-suffix list: a PSL is a dependency plus a refresh cadence the BFF would have to keep current or silently drift, and the list it guards is operator-controlled. It also states the cost, `https://*.co.uk` and `https://*.github.io` clear the floor.

**Per-key silent key.** Real regression, introduced by the symmetric overlap check in this PR. Details in the thread.

**Missing renderingId** and **stale `types.ts` JSDoc**: both fixed, details in their threads.

**qlty, 6 returns in `parseAllowedEntry`:** down to 5, the two label checks merged into one condition. Not going lower, the repo's own CLAUDE.md asks for guard clauses at the top over nesting.

agent-bff 97 suites / 1945 tests, agent 88 suites / 1625 tests, lint 0 errors on both.

packages/agent-bff/src/cors/origin.ts
```diff
@@ -32,6 +32,44 @@ export function normalizeOrigin(raw: string | undefined | null): string | null {
...
+  const { hostname } = new URL(origin);
+  const stars = hostname.split('*').length - 1;
+
+  if (stars === 0) return { origin, wildcard: false };
+  if (stars > 1 || !hostname.startsWith(WILDCARD_LABEL_PREFIX)) return null;
+
+  const parentLabels = hostname.slice(WILDCARD_LABEL_PREFIX.length).split('.');
+
+  if (parentLabels.length < MIN_WILDCARD_HOST_LABELS || parentLabels.some(label => label === '')) {
+    return null;
```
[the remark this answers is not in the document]
Tonours: Not reproducible. `https://*.127.0.0.1` is already refused, and has been since the first commit, with a passing test asserting it (`origin.test.ts`, the `it.each` rejection rows).

The reason is upstream of `parseAllowedEntry`: the WHATWG host parser applies the IPv4 rule when the last label is all-numeric, `*` is not a valid IPv4 part, so `new URL('https://*.127.0.0.1')` throws and `normalizeOrigin` returns null. The entry never reaches the label checks.

Measured against the built code on this branch:

```
https://*.127.0.0.1      origins=[] invalid=["https://*.127.0.0.1"]
https://*.1.2            origins=[] invalid=["https://*.1.2"]
https://*.0x7f.1         origins=[] invalid=["https://*.0x7f.1"]
https://*.192.168        origins=[] invalid=["https://*.192.168"]
https://*.1.2.3.4        origins=[] invalid=["https://*.1.2.3.4"]
https://*.2130706433     origins=[] invalid=["https://*.2130706433"]
```

The suggested `parentLabels.every(label => /^\d+$/.test(label))` would also not have caught `https://*.0x7f.1`, which is refused anyway. Not applying it: it adds a branch that can never be true.

Resolving as no change needed, reopen if you read it differently.

packages/agent-bff/src/cors/per-key-origin.ts
nbouliol: Closed: no entry pair now reports overlap while nothing passes both layers.

`allowedEntriesIntersect` sends both sides through `parseAllowedEntry`, so a refused pattern intersects nothing and no entry is matched as a request origin.
I ran the predicate over 271441 entry pairs with labels including `app-*` and `x*y`, finding zero false silences and zero false warns.
Pinned by `per-key-origin.test.ts` 'reports a key whose wildcard the parser refuses', which fails on the pre-fix predicate.
</details>

packages/agent-bff/src/cors/per-key-origin.ts
nbouliol: Closed: the warn now carries `renderingId`, the only one of the rule's three ids this code knows.

`per-key-origin.ts:44` logs `renderingId` beside `keyHash`, matching the rejection log below it.
`ResolvedApiKeyIdentity` at `api-key-client.ts:16` holds only `user`, `renderingId` and `allowedOrigins`, so no project or environment id exists here.
No credential or payload reaches the line, and `per-key-origin.test.ts:210` now pins the full payload.
</details>

packages/agent-bff/src/cors/origin.ts
nbouliol: Closed: the JSDoc now states the shipped rules, and every clause holds against the parser.

`types.ts:167` names the single leading-label form, the exact scheme and port, the two-label floor and the `*.co.uk` cost.
Each clause checks out against `parseAllowedEntry` and `matchesWildcard`, and `embedded-bff.ts:61` still feeds `BFF_ALLOWED_ORIGINS`.
Unpinned because no test can pin a doc comment, so rule 6 has nothing to protect here.
</details>

nbouliol:
Closed: the ADR records the decision and the alternative it rejected.

`docs/adr/2026-09-15-a-cors-wildcard-matches-one-label-over-a-host-of-at-least-two.md` is accepted and indexed in `CLAUDE.md`.
It states the floor as a deliberate label count, rejects a public-suffix list for dependency and drift, and names the `*.co.uk` cost.
Unpinned because an ADR carries no regression a test could pin.
</details>

nbouliol (approved):
Approved after /verify-fixes run 1903-20260915082101-nbouliol-verify: 4 findings of 1903-20260915073020-nbouliol closed or accepted, none reopened.

PR #1906, opened 2026-09-21
feat(workflow-executor): poll automated inboxes and report segment membership

Scra3:
fixes PRD-1183, fixes PRD-1279 — part of [PRD-1177](https://linear.app/forestadmin/issue/PRD-1177/workflow-trigger-automation) (Workflow trigger automation), lane B. Server side is [forestadmin-server#8494](https://github.com/ForestAdmin/forestadmin-server/pull/8494) (lane A).

## What this delivers

The runtime half of automated inboxes. The executor asks the orchestrator which inboxes to sweep, reads their segment on the client's agent, and reports back. **It never decides what runs** — the orchestrator owns the cap of 20, the active-run check and the run start.

One cycle, per inbox, every `AUTOMATION_POLL_INTERVAL_S` (default 300s):

1. `GET /automated-inboxes?instanceId=…` — the inboxes this instance is served
2. `GET …/:inboxId/assignments` — what is already in flight
3. agent read #1 — are the finished records still in the segment? (`pk In (…)`, chunks of 50)
4. agent read #2 — first page of the segment, sized `maxConcurrentRuns + assignments`
5. `POST …/:inboxId/sync` — the two lists back

| File | Role |
|---|---|
| `automation-poller.ts` | the cycle, the per-inbox in-flight guard, drain on stop |
| `adapters/forest-server-automation-port.ts` | the three orchestrator routes |
| `adapters/agent-client-segment-reader.ts` | the three segment kinds, on the client's agent |
| `ports/automation-port.ts`, `ports/segment-reader-port.ts` | the two seams the poller is tested against |
| `adapters/step-user.ts` | `mintToken` + `toStepUser`, extracted so a run-driven step and an automation poll reach the agent as the same caller |

## Decisions worth a reviewer's eye

- **`closed` is filtered on the run being terminal** (`finished`/`aborted`), not on the assignment state. An escalation closes the automation's assignment while its run is still going; reporting that record would escalate a live run a second time. This is also the shape of the `Must fix` I left on #8494.
- **An idle inbox still syncs.** Both lists empty is still a `POST /sync` — that call is what tells the orchestrator the inbox is being polled, and it is what the "no sync in N minutes" monitor of PRD-1185 will read. An inbox with nothing to do must not look like an inbox nobody polls.
- **`dataChange` joins both trigger enums.** Without it `AvailableStepExecutionSchema` rejects every automated run as malformed — so this part is needed the moment #8494 ships, independently of the poller.
- **The record id comes from the JSON:API resource id**, which the agent already packs with `IdUtils.packId`. Rebuilding it from the attributes is wrong: the deserializer overwrites an `id` attribute with the resource id, so a composite key came back doubled (`t1|t1|5`). Caught by a test, kept as one.
- **`connectionName` is omitted, never sent empty**, for a liana that runs a bare `segmentQuery`. superagent drops `undefined` but serializes `null`, and forest-rails would read the empty value as a connection it cannot resolve.
- **A 404 on the listing reads as "this orchestrator has no such route"**, so an executor released ahead of the server stays quiet instead of logging an error every cycle. A 404 on an inbox route means the orchestrator stopped serving that inbox, and the poller drops it until the next config poll.
- **One unreadable config does not blind the poller to the others** — it is skipped and warn-logged.
- **The polling timer is `unref`'d**: a background sweep must not be the reason a process stays alive, and a 5-minute interval would otherwise hold a shutdown open.

## Verified against the real agents

`segmentQuery` and `filters` compose as an AND on all three generations, so the membership read works everywhere:

- forest-rails `resources_getter.rb:185-193` — `where(pk IN (segmentQuery))`, then `FiltersParser` on top
- forest-express-sequelize `query-options.js:151-165` — `segmentQuery` resolves to `filterByIds`, and both it and `filterByConditionTree` push into the same `_where`
- agent-nodejs `datasource-customizer/src/decorators/segment/collection.ts:44-46` — `ConditionTreeFactory.intersect`
- agent-ruby `routes/resources/list.rb:22-27` — `ConditionTreeFactory.intersect`

That covers both lianas the server allows a bare `segmentQuery` for. Not verified: agent-python.

Two things found while checking, neither blocking this PR:

- forest-express-sequelize reads a SQL segment's result as `result[primaryKey] || result.id` with `primaryKeys[0]` (`query-options.js:230-236`), so a SQL segment on a **composite** key does not work there. The server decides what it sends, so nothing to do here — worth knowing before someone configures one.
- Every agent intersects the caller's scope (`append_scope_for_user` / `getScopeForUser` / `getScope` / `get_scope`), so the poller sees the segment **through the service account's scope**. To document in PRD-1185.

## Tests

48 new, all against the ports or against real HTTP:

- `test/automation-poller.test.ts` (18) — fake ports: the `runState` filter, chunking at 50, page arithmetic, the known-record subtraction, the idle sync, a gone inbox not failing the cycle, the in-flight guard, drain on stop.
- `test/adapters/agent-client-segment-reader.test.ts` (17) — **real agent-client + nock**, so the assertions are on the query string the agent receives: `segment=`, `segmentQuery=` with and without `connectionName`, `filters=` in wire format, the composite-key OR, `fields`, `timezone`, `page[size]`.
- `test/adapters/forest-server-automation-port.test.ts` (13) — the three routes, the instance id, a skipped config, both meanings of 404.

Package suite: 1908 passing.

## Known ceilings, deliberately not addressed

Each of these is a real limit of the design rather than an oversight. Named here so nobody has to rediscover them.

- **Shutting down can still start work.** `stop()` no longer lets a cycle dispatch new inboxes, but an inbox already mid-poll runs to its `sync`, and the orchestrator starts runs off that sync. Those runs belong to the orchestrator and are picked up by the next executor, so the cost is nil — but "nothing new starts during a shutdown" would be the wrong thing to claim.
- **One slow inbox paces every other.** Inboxes are swept in parallel within a cycle, and the next cycle is only scheduled once the whole cycle ends. A large segment on a slow database therefore sets the sweep interval for the environment, not just for itself.
- **The missing-route warning is said once per process.** A wrong `forestServerUrl` looks identical to an orchestrator that predates the feature, and only the first is permanent. One line at boot is easy to miss in weeks of logs; something periodic would be more honest.
- **A half-broken agent still produces a heartbeat.** The sync is withheld only when *no* read reached the agent. One read working is enough to look alive, so a partial outage shows up in the error logs rather than in the planned staleness alert.
- **The runtime trusts the orchestrator on identity.** It holds no service-account credential, but it does hold `authSecret` and mints an agent JWT for whatever profile the orchestrator sends. That is how every other executor path already works; worth stating plainly rather than implying a stronger boundary.
- **A record the orchestrator cannot name is read past, not excluded.** The exclusion only covers what the config listing hands back, so a composite key, a v1 liana, or a set too large for a query string falls back to the padded page and its starvation ceiling. The orchestrator stops serving an inbox long before that set grows (forestadmin-server#8512), which is what makes the fallback a corner rather than the normal path.
- **Deploying leaves a gap the executor cannot close.** Each process boots with a fresh `instanceId`, so it cannot take the poller lease until the previous holder's key expires — up to `LEASE_TTL_SECONDS` (600s) with nobody sweeping. Firing a first cycle immediately would only fetch an empty list sooner. This one is server-side, raised separately.

## Follow-up pushed after review

`863189c99` answers the padding question raised on the sweep: the candidate page was sized `maxConcurrentRuns + assignments.length` so already-tracked records could sit inside it and be subtracted after. A record treated by a run but still in its segment keeps its `done` assignment for good, so on a mature inbox that reads a few hundred rows off the customer's database every five minutes to keep twenty — and once more of them sort ahead of the untreated ones than the page holds, the sweep stops seeing new records at all, silently.

The page now stays at the run cap and the records travel as a `pk not_in (...)` ANDed to the segment, the mirror of the `in` the membership read already sends.

`not_in` is not an operator every agent parses, and this is the side that speaks to agents. Support belongs to the datasource, not the liana name (PRD-1279), so the poller asks: for an agent that serves `POST /forest/_internal/capabilities` (`forest-nodejs-agent`, `agent-ruby`, `agent-python`, `agent-php`, the same list the front gates that call on), `not_in` goes out only when the primary key declares it. Every other liana pads: the v1 ones raise on `not_in` (`forest-rails` `filters_parser.rb`, `forest-express-*` `NoMatchingOperatorError`, `django-forestadmin`, `forest-laravel`), and an unknown name fails closed. Declared is not implemented, so a `not_in` read that still fails is retried padded in the same cycle, with a Warn, and nothing is remembered across cycles. Known ids are subtracted from both pages.

`6b6e99a68` answers a question nothing in the logs could: a poll that reads 320 rows off the customer's database looked exactly like one that reads 20, and nothing said which of the four conditions sent it down the padded path.

The boolean that decided now returns the clause that refused — `composite-key`, `too-many-known-records`, `unknown-liana`, `capabilities-unreadable`, `field-without-not-in`, `not-in-refused` (PRD-1279) — and the sweep log carries `candidatePageSize` alongside, with `paddedPageReason` absent when the exclusion worked. The same call builds the query and names the reason, so the log cannot drift from what was actually sent. The five candidate tests each gained the matching assertion; removing the two fields turns all five red.

Still unwritten, and outside this repo: the invariant that the orchestrator's cap of 100 untreated records fires before `MAX_EXCLUDED_RECORDS` (150) and `MAX_CANDIDATE_PAGE_SIZE` (500). Raising the server threshold silently retires both belts, and nothing records the relationship.

## Not in this PR

Datadog monitors and the developer guide are PRD-1185. The release checklist (`MIN_AUTOMATION_EXECUTOR_VERSION` on the server must point at the version this ships as) is on PRD-1183.

> [!NOTE]
> ### Add automated-inbox polling and segment membership reporting to `workflow-executor`
> - Adds `AutomationPoller` which periodically lists automated inboxes assigned to this executor instance, reads closed-assignment and candidate record IDs from agent segments, and syncs reconciliation results to Forest Server via `POST /automated-inboxes/:inboxId/sync`
> - Adds `ForestServerAutomationPort` for orchestrator communication (inbox listing, assignment reads, sync) and `AgentClientSegmentReader` for reading record IDs and field operators from remote agents, including smart, SQL, filter, and collection segment kinds with single-column and composite primary keys
> - Adds `AUTOMATION_POLL_INTERVAL_S` env var (default 300s, `0` disables sweeping) and wires the poller into the in-memory and database executor lifecycles, draining it before the workflow runner on shutdown
> - Refactors runner polling to request only available run slots (max 10 concurrent), dispatch fetched runs without awaiting, and suppress dispatch after shutdown begins
> - Behavioral Change: `WorkflowPort.getAvailableRuns` now requires a numeric `count` parameter; `TriggerType` adds `DataChange` (`dataChange`); executor shutdown stops the automation poller before the workflow runner, and a stop-during-start race no longer starts a background poller afterward
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 55a0948.</sup>
> 
>


packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,282 @@
...
+    if (this._state !== 'running') return;
+    this.pollingTimer = setTimeout(() => this.runPollCycle(), this.config.pollingIntervalS * 1000);
+    // A background sweep must not be the reason a process stays alive: the host owns that, and an
+    // interval this long would otherwise hold a shutdown open for minutes.
+    this.pollingTimer.unref?.();
+  }
+
+  private async runPollCycle(): Promise<void> {
+    try {
+      const inboxes = await this.config.automationPort.listAutomatedInboxes(this.config.instanceId);
```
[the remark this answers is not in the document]
Scra3: Fixed in a6745f357, with your guard.

Worth noting the finding landed on `5034a47a` and was half-addressed by `51684ccb` in between: `stop()` now awaits the cycle itself, not only the per-inbox promises it spawned, so the work could no longer land *outside* the drain. But you were right that it still ran — the cycle went on reading the agent and calling `/sync` while a shutdown was waiting on it. The state check after the await stops it from starting at all.

Covered by `should not dispatch inboxes fetched while stop() was already running` in `test/automation-poller.test.ts`.

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,343 @@
...
+          renderingId: config.renderingId,
+          runState,
+        });
+      }
+    }
+
+    // A closed assignment whose run is still going is an escalation in progress, and reporting it
+    // would hand a live run to a human twice. A closed assignment with no run at all is the
+    // opposite case: there is nothing to protect, so it reconciles like any other.
+    const recordIds = [
```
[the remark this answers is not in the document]
Scra3: Fixed in c757ac943, per record rather than per assignment.

I could not construct a reachable case, and I looked: `startCandidate` refuses a record that already has a `todo`/`doing`/`done` assignment, so the only shape I found was a `canceled` row plus a fresh `doing` one — and there the server cleans the canceled row correctly anyway. But nothing in the frozen contract says a record holds a single assignment, and the failure mode if it ever does is a live run escalated twice. So the rule is now: a record is only reconciled when none of its assignments has a live or unrecognised run.

One known cost: a stale `canceled` row now waits for the new run to finish before being cleaned, instead of going out on the next poll. It pads the page size slightly and resolves itself.

Covered by `should hold back a record whose other assignment still has a live run`.

packages/workflow-executor/src/adapters/server-types.ts
```diff
@@ -257,3 +260,131 @@ export interface ServerUpdateStepRequest {
...
+  | { aggregator?: string; conditions: ServerPlainConditionTree[] };
+
+const ServerPlainConditionTreeSchema: z.ZodType<ServerPlainConditionTree> = z.lazy(() =>
+  z.union([
+    z.object({ field: z.string(), operator: z.string().optional(), value: z.unknown() }),
+    z.object({
+      aggregator: z.string().optional(),
+      conditions: z.array(ServerPlainConditionTreeSchema),
+    }),
+  ]),
```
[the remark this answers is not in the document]
Scra3: Fixed in c757ac943 — branch first now.

Worth spelling out why it mattered here specifically: the leaf schema is non-strict, so it matched the ambiguous node and **stripped** `conditions` before anything downstream saw it. agent-client's `toWireFilter` already resolves this the same way and says so in a comment, but my schema ran first, so its ordering is the one that decided.

Covered by `should keep the conditions of a node that also carries a field`.

packages/workflow-executor/src/adapters/agent-client-segment-reader.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (segment.kind === 'smart') return collection.segment(segment.name);
+
+    if (segment.kind === 'sql') {
+      // A v1 liana runs the query against the single database it is bound to and never reads
+      // `connectionName`; the key is dropped rather than sent empty, which superagent would
+      // forward as a connection name the agent cannot resolve. agent-client types it required,
+      // hence the cast.
+      return collection.liveQuerySegment({
+        query: segment.query,
+        ...(segment.connectionName != null ? { connectionName: segment.connectionName } : {}),
```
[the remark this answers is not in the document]
Scra3: Fixed in c757ac943 — truthiness check.

For the record, the reachable blast radius was smaller than it looks: the server decides a SQL segment is bare on `!segment.connectionName`, so an empty string is already treated as absent there and a v2 agent never receives that descriptor. Only a v1 liana could get `connectionName=` from me, and neither forest-rails nor forest-express-sequelize reads the parameter. So nothing was actually broken.

Changed anyway, because the server and the executor were testing the same field with two different notions of "empty" — that divergence is the bug, even when today it costs nothing.

Covered by `should omit a connection name the server sent empty`.

packages/workflow-executor/src/adapters/agent-client-segment-reader.ts
```diff
@@ -0,0 +1,167 @@
...
+   * The JSON:API resource id, which the agent already builds by joining the primary key with `|`
+   * (`IdUtils.packId`) — the very form the orchestrator stores as `recordId`. Rebuilding it from the
+   * attributes would be wrong anyway: the deserializer overwrites an `id` attribute with the
+   * resource id, so a composite key would come back doubled.
+   */
+  private static readRecordId(record: Record<string, unknown>, collectionName: string): string {
+    const { id } = record;
+
+    // Guarded rather than coerced: a blank id here would start a run against the wrong record.
+    if (typeof id !== 'string' && typeof id !== 'number') {
```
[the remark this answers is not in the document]
Scra3: Fixed in c757ac943.

Agreed on the shape: `typeof '' === 'string'` sailed through a guard whose whole purpose is to stop an unusable id from travelling as a candidate. An empty id would have reached `/sync` and started a run against nothing, which is exactly what the guard exists to prevent.

Covered by `should refuse a record whose id came back empty`.

hercemer42 (commented):
**Spec (`PRD-1183`)**: conforms.

Claude Opus 5 (claude-opus-5): **Preferential**: two decisions this change carries are hard to reverse and surprising, and no accepted ADR records either.
The sweep sends record ids to the orchestrator at `automation-poller.ts:231`, and `step-user.ts:41` mints an unattended service-account token.
A way out: run `/adr` for each decision.
</details>

packages/workflow-executor/src/adapters/server-types.ts
```diff
@@ -257,3 +260,134 @@ export interface ServerUpdateStepRequest {
...
+  | { aggregator?: string; conditions: ServerPlainConditionTree[] };
+
+// Branch first, deliberately: a node carrying both `conditions` and `field` is ambiguous, and the
+// leaf schema would match it and strip the conditions, leaving a filter that reads a different
+// segment than the one configured. agent-client's own `toWireFilter` resolves it the same way.
+const ServerPlainConditionTreeSchema: z.ZodType<ServerPlainConditionTree> = z.lazy(() =>
+  z.union([
+    z.object({
+      aggregator: z.string().optional(),
+      conditions: z.array(ServerPlainConditionTreeSchema),
```
hercemer42: Claude Opus 5 (claude-opus-5): **Must fix**: an inbox whose filter segment arrives with an empty condition list reads the whole collection as its segment.
`server-types.ts:282` accepts an empty `conditions` array, and the agent reads `And` over nothing as matching every record.
A way out: `.min(1)` on the branch schema, and refuse a `filter` descriptor whose tree is an empty branch.
</details>

1. Configure an automated inbox whose filter segment has no condition left.
2. Wait for one sweep.
3. The candidate read returns records from the whole collection.
4. The orchestrator starts runs on records the segment never held.

</details>

packages/workflow-executor/src/adapters/agent-client-segment-reader.ts
```diff
@@ -0,0 +1,168 @@
...
+    return {
+      aggregator: 'or',
+      conditions: recordIds.map(recordId => {
+        const parts = recordId.split('|');
+
+        // Same limitation as the agent's own `IdUtils.packId`: a key value containing the separator
+        // cannot be unpacked. Caught here, where the id is named, rather than sent as a leaf with
+        // no field for the agent to reject.
+        if (parts.length !== primaryKeys.length) {
+          throw new CompositeRecordIdMismatchError(recordId, primaryKeys.length);
```
hercemer42, line 151: Claude Opus 5 (claude-opus-5): **Should fix**: a record id the reader proposes and cannot later split ends reconciliation for that inbox on every cycle.
`agent-client-segment-reader.ts:163` returns a packed id unchecked, and `:136` later throws on it, failing the whole chunk.
A way out: skip and warn per record, the shape `forest-server-automation-port.ts:98` uses for an unreadable config.
</details>

1. Use a collection with a two-column primary key.
2. Give one record a key value containing the `|` character.
3. The sweep proposes it and the orchestrator starts a run on it.
4. Every later sweep logs a closed-record read failure and reports nothing closed.

</details>

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,390 @@
...
+
+  private static segmentQuery(config: ServerAutomatedInboxConfig) {
+    return {
+      collectionName: config.collectionName,
+      segment: config.segment,
+      primaryKeys: config.primaryKeys,
+      user: config.serviceAccountProfile,
+      // Every executor instance must read a relative date the same way, so the machine's zone is
+      // never the fallback.
+      timezone: config.timezone ?? 'UTC',
```
hercemer42: Claude Opus 5 (claude-opus-5): **Should fix**: an empty timezone string from the orchestrator makes every segment read of that inbox fail with 400.
`automation-poller.ts:387` guards null alone, while `run-to-available-step-mapper.ts:181` also checks the zone is valid.
A way out: reuse the run path guard, `IANAZone.isValidZone` before falling back to `UTC`.
</details>

1. Have the orchestrator send an inbox config whose `timezone` is an empty string.
2. Wait for one sweep.
3. The agent answers `Missing timezone` with status 400 on both reads.
4. The inbox never sweeps and only error lines appear.

</details>

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,390 @@
...
+    config: ServerAutomatedInboxConfig,
+    assignments: ServerAutomatedInboxAssignment[],
+  ): Promise<SegmentRead<string>> {
+    // Sized so that the already-assigned records can fit inside the window alongside a full batch
+    // of new ones. No sort is imposed and each agent orders as it likes, so this is a best effort,
+    // not a guarantee: a page that comes back mostly assigned simply yields fewer candidates, and
+    // the rest are picked up by a later poll.
+    const page = await this.config.segmentReaderPort.listRecordIds({
+      ...AutomationPoller.segmentQuery(config),
+      pageSize: config.maxConcurrentRuns + assignments.length,
```
hercemer42: Claude Opus 5 (claude-opus-5): **Should fix**: a large assignment backlog asks for one page the agent cannot serve inside the client's ten second ceiling.
`automation-poller.ts:371` caps nothing, and `http-requester.ts:81` times every agent read out at ten seconds.
A way out: cap the candidate page the way `MEMBERSHIP_CHUNK_SIZE` caps the membership read.
</details>

1. Let one inbox accumulate several thousand assignments.
2. Wait for one sweep.
3. The candidate read asks for a page of that size and times out.
4. That inbox proposes no candidate again while the backlog stays.

</details>

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,390 @@
...
+        });
+      }
+    }
+
+    // A closed assignment whose run is still going is an escalation in progress, and reporting it
+    // would hand a live run to a human twice. Judged per record rather than per assignment: nothing
+    // in the contract says a record holds only one, and one terminal assignment must not speak for
+    // a sibling whose run is still alive.
+    const liveRecords = new Set(
+      assignments.filter(({ runState }) => !isTerminalRun(runState)).map(a => a.recordId),
```
hercemer42, line 353: Claude Opus 5 (claude-opus-5): **Should fix**: the warning says the record will be picked up on a later poll, and no later poll changes that.

`automation-poller.ts:293` warns on a null `runState`, while `:305` and `:374` keep that record out of both lists.

A way out: reword the warning at `automation-poller.ts:293` to say the record stays out until the executor understands the shape.

</details>

1. Have a human-created assignment on an automated inbox, which carries no `workflowRunId`.
2. Cancel it.
3. Wait for one sweep.
4. The executor warns that it leaves the record for a later poll.
5. Every later sweep repeats the warning and the record moves nowhere.

</details>

packages/workflow-executor/src/adapters/forest-server-automation-port.ts
```diff
@@ -0,0 +1,157 @@
...
+    inboxId: string,
+    body: ServerAutomatedInboxSyncRequest,
+  ): Promise<AutomatedInboxSyncResult[]> {
+    const response = await this.callPort(
+      'syncAutomatedInbox',
+      () => ServerUtils.query<unknown>(this.options, 'post', ROUTES.sync(inboxId), {}, body),
+      inboxId,
+    );
+
+    return ServerAutomatedInboxSyncResponseSchema.parse(response).results;
```
hercemer42: Claude Opus 5 (claude-opus-5): **Should fix**: a sync the orchestrator already applied is logged as a poll that never reached it.
`forest-server-automation-port.ts:132` parses outside `callPort`, so a zod error surfaces as `Automated inbox poll failed`.
A way out: `safeParse` the envelope, the shape `forest-server-workflow-port.ts:177` uses for its chained-step read.
</details>

1. Have the orchestrator answer the sync with an envelope the executor schema does not match.
2. The runs for the candidates it received have already started.
3. The executor logs `Automated inbox poll failed`.
4. An operator reads that as the sync never landing.

</details>

packages/workflow-executor/test/adapters/agent-client-segment-reader.test.ts
```diff
@@ -0,0 +1,335 @@
...
+
+      expect(captured.query).toMatchObject({ segmentQuery: 'SELECT id FROM orders' });
+      expect(JSON.parse(captured.query.filters)).toEqual({
+        field: 'id',
+        operator: 'in',
+        value: ['1'],
+      });
+    });
+
+    it('should address a composite key one record at a time', async () => {
```
hercemer42, line 214: **Applies to:** `agent-client-segment-reader.ts:136`, not in this hunk
Claude Opus 5 (claude-opus-5): **Violates conventions**: the arity guard on a composite record id has no case, so a regression in it ships unseen.
No test in the package names `CompositeRecordIdMismatchError`, while the composite happy path is covered here.
Rule `testing.md#cover-error-and-edge-paths-not-only-the-happy-path`, met by a case whose packed id splits into a different count.
</details>

packages/workflow-executor/test/adapters/agent-client-segment-reader.test.ts
```diff
@@ -0,0 +1,335 @@
...
+        .query(true)
+        .reply(200, function handler(this: { req: { headers: Record<string, string> } }) {
+          authorization = this.req.headers.authorization;
+
+          return jsonApi([]);
+        });
+
+      await reader.listRecordIds(makeQuery());
+
+      expect(authorization).toMatch(/^Bearer ey/);
```
hercemer42: Claude Opus 5 (claude-opus-5): **Violates conventions**: the service-account auth test passes for any signed token, including one minted from the wrong profile.
The assertion checks only that the header matches `^Bearer ey`, while the name claims the caller identity.
Rule `testing.md#test-name-states-the-exact-behavior-it-asserts`, met by asserting `id`, `rendering_id` and `scope`.
</details>

packages/workflow-executor/test/automation-poller.test.ts
```diff
@@ -0,0 +1,684 @@
...
+
+      const stopped = poller.stop();
+
+      expect(poller.state).toBe('draining');
+
+      release();
+      await stopped;
+
+      expect(poller.state).toBe('stopped');
+      expect(context.automationPort.sync).toHaveBeenCalled();
```
hercemer42: Claude Opus 5 (claude-opus-5): **Violates conventions**: this test passes even when the in-flight inbox syncs the wrong inbox id or an empty body.
The anchored line asserts bare called-ness on `sync`, and lines 504 and 524 pass `expect.anything`.
Rule `testing.md#assert-behavior-with-args-and-outputs`, met by `toHaveBeenCalledWith` on the id and both lists.
</details>

packages/workflow-executor/src/cli-core.ts
```diff
@@ -282,6 +283,10 @@ export function readEnvConfig(env: NodeJS.ProcessEnv, args: CliArgs): CliConfig
     httpPort: parsePositiveIntEnv('HTTP_PORT', env.HTTP_PORT) ?? DEFAULT_HTTP_PORT,
     forestServerUrl: env.FOREST_SERVER_URL,
     pollingIntervalS: parsePositiveIntEnv('POLLING_INTERVAL_S', env.POLLING_INTERVAL_S),
+    automationPollingIntervalS: parsePositiveIntEnv(
```
hercemer42: Claude Opus 5 (claude-opus-5): **Preferential**: an operator cannot turn the sweep off, so a refused environment logs an error every five minutes.
`cli-core.ts:286` rejects zero through `parsePositiveIntEnv`, so no value disables the poller.
A way out: accept a disabling value on `AUTOMATION_POLL_INTERVAL_S`.
</details>

packages/workflow-executor/src/adapters/forest-server-automation-port.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    return this.parseConfigs(response);
+  }
+
+  private parseConfigs(response: unknown): ServerAutomatedInboxConfig[] {
+    const envelope = ServerAutomatedInboxesResponseSchema.safeParse(response);
+
+    if (!envelope.success) {
+      this.logger('Error', 'Unreadable automated inbox listing', { error: envelope.error.message });
```
hercemer42: Claude Opus 5 (claude-opus-5): **Preferential**: an unreadable inbox listing logs an error nobody can tie to an environment or an executor instance.
The anchored line logs the zod message alone, while the 404 branch at `:67` carries `instanceId` and `forestServerUrl`.
A way out: carry `instanceId` and `forestServerUrl` into this log line.
</details>

packages/workflow-executor/src/automation-poller.ts
hercemer42: Claude Opus 5 (claude-opus-5): a cheaper route than a flat cap, if you want the page to stop growing at the source.

Sizing on open assignments alone drops the dominant term, since a closed record that left the segment never occupies a page slot.

A closed record still in the segment does come back, and `inbox-automation-service.ts:401` is the path that wants it, so the escalation is unchanged.

`not_in` shrinks it further on a single-column key, but a composite key has no NOT branch and the exclusion list is unbounded too.

</details>

packages/workflow-executor/src/adapters/server-types.ts
Scra3: Fixed in `46c3c479`. `aggregator` and `operator` are now required and a branch may not be empty, so a half-formed tree is refused at the boundary instead of read as a wider segment.

One correction to the reproduction, because it changes how this reads: the path you describe does not reach here through this orchestrator. `segmentFilterToConditionTree` returns `null` when every condition drops out, and `inbox-automation-service.ts:303` turns that into `empty-filter-segment`, which degrades the inbox before it is ever served. So this is the boundary doing its job on a shape the server does not emit today — worth having, since "the segment is the whole collection" is the one failure that starts runs on records the inbox never held, but not a live bug.

Covered by three cases in `forest-server-automation-port.test.ts` (empty branch, missing aggregator, missing operator), each expecting the inbox skipped.

packages/workflow-executor/src/adapters/server-types.ts
```diff
@@ -257,3 +260,134 @@ export interface ServerUpdateStepRequest {
...
+    }),
+    z.object({ field: z.string(), operator: z.string().optional(), value: z.unknown() }),
```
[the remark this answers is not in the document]
Scra3: Fixed in `46c3c479`, exactly as suggested — both discriminators are required now, plus `.min(1)` on `conditions`.

Checked before tightening rather than after: the server types make both mandatory already (`segment-filter-to-condition-tree.ts:18-27` — `WireConditionTreeLeaf.operator` and `WireConditionTreeBranch.aggregator` are non-optional), so no inbox this orchestrator serves is dropped by the stricter schema.

packages/workflow-executor/src/automation-poller.ts
Scra3: Fixed in `46c3c479`. `AutomationPoller.readTimezone` runs `IANAZone.isValidZone` and falls back to UTC, so an unusable zone is treated exactly like an absent one rather than failing both reads of every sweep.

Put in the poller rather than the schema on purpose: a bad timezone is not a reason to drop the inbox from the sweep entirely, and UTC is already the documented fallback for a project that has none.

packages/workflow-executor/src/adapters/server-types.ts
```diff
@@ -257,3 +260,137 @@ export interface ServerUpdateStepRequest {
...
+export const ServerAutomatedInboxConfigSchema = z.object({
+  inboxId: z.string().min(1),
+  renderingId: z.number().int().nonnegative(),
+  teamId: z.number().int().nonnegative().optional(),
+  workflowId: z.string().min(1).optional(),
+  collectionId: z.string().min(1).optional(),
+  collectionName: z.string().min(1),
+  primaryKeys: z.array(z.string().min(1)).min(1),
+  maxConcurrentRuns: z.number().int().positive(),
+  timezone: z.string().nullish(),
```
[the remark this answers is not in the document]
Scra3: Fixed in `46c3c479`, with one difference from what you propose.

Validated, but not by rejecting the inbox: an unusable zone now falls back to UTC in `AutomationPoller.readTimezone` (`IANAZone.isValidZone`), the same fallback an absent zone already had. Refusing the config would stop the automation over a field that only shifts how relative dates resolve, while UTC keeps it running and is already what a project without a timezone gets — and `project.timezone` is filled on well under 1% of projects.

Covered by a poller test asserting the read goes out in UTC for `Mars/Olympus`.

packages/workflow-executor/src/adapters/agent-client-segment-reader.ts
Scra3: Fixed in `46c3c479`, in the poller rather than in the reader.

Skipping inside the reader looked right and is not: `reconcileClosed` reads "not found in the segment" as "the record left", so a record silently dropped from the question would come back `stillInSegment: false`, the orchestrator would retire its assignment, and the record could then be launched a second time. Worse than the throw.

So the poller filters its own record set before asking, warns per record with the id, and leaves those records out of the report entirely — the same treatment an assignment state it does not understand gets. The reader keeps throwing, since by then the id is already in a question it cannot phrase.

Two tests: the poller drops `t1|a|5`, still reports `t2|9`, and warns; the reader still refuses a mismatched arity (`CompositeRecordIdMismatchError` had no case at all, which was your other finding).

packages/workflow-executor/src/automation-poller.ts
Scra3: Fixed in `46c3c479` with the flat cap, 500.

Your follow-up is right that sizing on open assignments alone drops the dominant term, and it is also why I did not take it: a `done` assignment retained because the record is *still in the segment* does occupy a page slot, and that is precisely the record set that grows without bound — PRD-1268 exists for it. Sizing on open rows alone would under-pad exactly the case that starves.

That ticket also lands the real fix on both sides: the executor now sends `pk not_in (...)` when the agent parses it (`863189c9`), so the page stays at the run cap whatever the inbox has accumulated, and the orchestrator stops serving an inbox past 100 untreated records. The cap here is the belt for what is left — a composite key or a v1 liana, where neither applies.

Covered by a test that builds 900 assignments and asserts the page asked for is 500, not 920.

packages/workflow-executor/src/automation-poller.ts
Scra3: Fixed in `46c3c479`. Both warnings now read "leaving the record out of every sweep until this executor knows it", which is what actually happens — the old wording promised a later poll that never comes.

packages/workflow-executor/src/adapters/forest-server-automation-port.ts
Scra3: Fixed in `46c3c479` for `sync`, and deliberately not for `listAssignments`.

`sync` now `safeParse`s and logs "Unreadable automated inbox sync response, the sync itself landed", returning no outcomes. The runs are started by then, so the old line told an operator the opposite of what happened.

`listAssignments` keeps throwing. I started by giving it the same treatment and backed it out: an empty list there is indistinguishable from an inbox with nothing assigned, so the very next step would propose every record in the segment as a fresh candidate and start duplicate runs. Failing the inbox for that cycle is the cheaper outcome. Noted in a comment on the line so the asymmetry is not read as an oversight.

packages/workflow-executor/test/adapters/agent-client-segment-reader.test.ts
Scra3: Fixed in `46c3c479` — "should refuse a packed id that does not split into as many parts as the key has" in the same describe, expecting the rejection to name the id. Deleting the arity guard fails it.

packages/workflow-executor/test/adapters/agent-client-segment-reader.test.ts
Scra3: Fixed in `46c3c479`. The test now decodes the JWT payload and asserts `id`, `email` and `rendering_id` against the profile it was built from, instead of matching `^Bearer ey` — which, as you say, any signed token satisfies.

packages/workflow-executor/test/automation-poller.test.ts
Scra3: Fixed in `46c3c479`. All three sites now assert the full call: the drain test on `('inbox-1', { closed: [], candidates: [] })`, and the two failure-isolation tests on the same shape for `inbox-2` instead of `expect.anything()`.

packages/workflow-executor/src/cli-core.ts
Scra3: Taken, in `46c3c479`. `AUTOMATION_POLL_INTERVAL_S=0` turns the sweep off: `cli-core` accepts zero for this variable alone, `build-workflow-executor` no longer folds it into the default, and `start()` logs "Automation poller disabled by configuration" and returns. The poller stays constructed and `stop()` still answers, so nothing downstream has to know.

Documented in `.env.example` and covered by two tests — the builder keeps the zero, and the poller never reaches the orchestrator with it.

packages/workflow-executor/src/adapters/forest-server-automation-port.ts
Scra3: Taken, in `46c3c479`. The line now carries `instanceId` and `forestServerUrl`, passed down to `parseConfigs` rather than kept on the instance, so it reads like its 404 sibling. The existing test asserts both.

Scra3:
Closing the ADR finding from the review body. One of the two decisions is recorded, the other is not, and the difference is the trade-off test.

**Written** — `docs/adr/2026-09-22-the-automation-sweep-sends-record-ids-to-the-orchestrator-and-decides.md`. Forest Runtime sits on the customer's infrastructure and the sweep is the one place identifiers of their records leave it, so it belongs in writing rather than inferred from a route. It names the accepted cost (primary key values transit and are stored as `inboxAssignments.recordId`, every other field stays put) and the two alternatives dropped: deciding in the runtime, which needs state no instance can hold safely, and hashing the ids, which buys nothing since the orchestrator has to name the record back to the customer's own inbox.

**Not written** — the unattended service-account token. Surprising, yes, but it fails the trade-off leg of the gate: the segment is scoped by the service account's permissions, which is the whole reason an inbox is configured with one. Minting the environment secret instead would read the collection unscoped and break the guarantee the docs promise, so there was one option, not a choice. Recording it would be a description of the code, not a decision.

packages/workflow-executor/src/adapters/server-types.ts
```diff
@@ -257,3 +260,144 @@ export interface ServerUpdateStepRequest {
...
+// Both discriminators are required and a branch may not be empty. The agent reads `And` over
+// nothing as matching every record, so a tree this schema let through half-formed would widen the
+// segment to the whole collection rather than fail — the one failure mode worth a rejected inbox.
+const ServerPlainConditionTreeSchema: z.ZodType<ServerPlainConditionTree> = z.lazy(() =>
+  z.union([
+    z.object({
+      aggregator: z.string().min(1),
+      conditions: z.array(ServerPlainConditionTreeSchema).min(1),
+    }),
+    z.object({ field: z.string().min(1), operator: z.string().min(1), value: z.unknown() }),
```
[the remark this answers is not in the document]
Scra3: Correct, and worse than a nit — this was live, not something my hardening introduced. Fixed in `b81bca42`.

`segment-filter-to-condition-tree.ts:287-289` returns `{ field, operator }` with no value at all for every operator the agent evaluates on its own: `present`, `blank`, `today`, `yesterday`, and the whole `previous_*` family. Zod requires an `unknown` key to be present, so any inbox whose segment used one of those was dropped from the listing and never swept — silently, as a config the executor "cannot read".

`value` is `z.unknown().optional()` now, which is what the contract type said all along. Covered by a case asserting a `blank` leaf survives the parse intact.

packages/workflow-executor/src/build-workflow-executor.ts
```diff
@@ -221,6 +256,8 @@ function createWorkflowExecutor(
         throw err;
       }
 
+      automationPoller.start();
```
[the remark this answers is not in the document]
Scra3: Confirmed and fixed in `b81bca42`.

`start()` awaits `runner.start()` and then `server.start()` before it arms the sweep, and `stop()` can land in either gap. The shutdown drains a poller that has not started yet, reports itself complete, and the suspended `start()` then arms one nobody will ever stop — still syncing, still launching runs, on an executor that is down.

A `stopped` flag set at the top of `stop()` and checked before `automationPoller.start()`, with an Info line so an operator sees why the sweep never came up. Covered by a test that starts, stops before the awaits resolve, and asserts the poller was never armed.

Scra3 (commented):
Review against PRD-1177 / PRD-1183 and forestadmin-server.

Outside the diff, before release:
- issue (blocking): `MIN_AUTOMATION_EXECUTOR_VERSION` is `'1.29.0'` (`inbox-automation-eligibility-service.ts:16`), the already-published version with no poller, so automation can be enabled on an environment nothing sweeps; it must point at the version this PR releases.
- issue (blocking): forestadmin-server#8508 (`DELETE /automated-inboxes/lease`) is still open, not merged, and nothing here releases the lease on `stop()` (see inline).
- issue (non-blocking): the 100 / 150 / 500 invariant does not hold: the server breaker counts only `done` dataChange rows, while `known` counts human rows (uncapped) and up to 200 cancelled ones, so past ~480 assignments the padded page yields zero candidates while syncs look healthy; the server also rejects more than 1000 `closed` items (`route-validator.ts:160`).
- suggestion: no PRD-1177 contract amendment yet for the lease DELETE, `liana`, `already-escalated` and `too-many-untreated-records`; the ADR on the unattended service-account token (`step-user.ts`) is still missing.

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,508 @@
...
+const MAX_CANDIDATE_PAGE_SIZE = 500;
+
+// The exclusion filter travels in the query string of a GET. The orchestrator stops serving an
+// inbox long before this, so it is a belt on the URL length rather than the real ceiling.
+const MAX_EXCLUDED_RECORDS = 150;
+
+// `not_in` is not an operator every agent parses: the v1 lianas raise on any operator their filter
+// parser does not list, and none of them lists it. The front draws the same line — it only offers
+// "is not in" for a field whose agent declared the operator in its capabilities.
+const LIANAS_WITHOUT_NOT_IN: ReadonlySet<string> = new Set([
```
Scra3: issue (blocking): `forest-laravel` is a v1 liana the server still knows (`apimap-controller.ts:38`) and is missing here, so a Laravel project gets `not_in`, every candidate read fails and the sync still posts `candidates: []`, an inbox that looks swept and never starts anything; an allowlist of the v2 agents (`forest-nodejs-agent`, `agent-ruby`, `agent-python`, `agent-php`) fails closed instead.

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,508 @@
...
+
+    this._state = 'running';
+    this.logger('Info', 'Automation poller started', {
+      instanceId: this.config.instanceId,
+      pollingIntervalS: this.config.pollingIntervalS,
+    });
+    this.schedulePoll();
+  }
+
+  async stop(): Promise<void> {
```
Scra3: issue (blocking): `stop()` never releases the lease, so every deploy leaves the environment unswept for up to the 600s TTL plus one interval; add `releaseLease(instanceId)` to `AutomationPort` once forestadmin-server#8508 merges, tolerate a 404, and update the PR body paragraph that calls this gap server-side.

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,508 @@
...
+      });
+
+      found.forEach(recordId => stillInSegment.add(recordId));
+    }
+
+    return {
+      outcome: 'ok',
+      items: readable.map(recordId => ({
+        recordId,
+        stillInSegment: stillInSegment.has(recordId),
```
Scra3: issue (non-blocking): any id missing from the membership answer becomes `stillInSegment: false`, so a v2 agent casting a BIGINT pk above 2^53 through `Number()` (`condition-tree-parser.ts` castToType) never finds the record, the assignment is retired and the record relaunched every cycle; at least warn when a chunk returns none of the requested ids, or file it.

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,508 @@
...
+        outcomes: results.reduce<Record<string, number>>(
+          (counts, { outcome }) => ({ ...counts, [outcome]: (counts[outcome] ?? 0) + 1 }),
+          {},
+        ),
+      });
+    } catch (error) {
+      // The orchestrator refuses this inbox for now — disabled, or degraded. The next config poll
+      // decides whether it comes back, so this is not an error.
+      if (error instanceof AutomatedInboxGoneError) {
+        this.logger('Info', 'Automated inbox no longer served, dropping it for this cycle', {
```
Scra3: issue (non-blocking): an inbox served by the listing that then 404s on its own route is a contradiction logged at Info with the cause dropped, so a route missing behind a proxy stops the inbox forever under a fresh lease; keep `cause`, name the operation that 404'd, and log Warn.

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,508 @@
...
+    logContext: Record<string, unknown>,
+    what: string,
+    read: () => Promise<SegmentRead<T>>,
+  ): Promise<SegmentRead<T>> {
+    try {
+      return await read();
+    } catch (error) {
+      this.logger('Error', `Could not read ${what} of an automated inbox`, {
+        ...logContext,
+        error: extractErrorMessage(error),
```
Scra3: issue (non-blocking): only `Agent responded with HTTP 400` reaches the log, the `AgentHttpError` body and status are lost, so a renamed segment, a 403 or an unsupported operator all read the same; log status, a truncated body and the read parameters (`requestedPageSize`, `paddedPageReason`).

packages/workflow-executor/src/adapters/forest-server-automation-port.ts
```diff
@@ -0,0 +1,177 @@
...
+    const envelope = ServerAutomatedInboxesResponseSchema.safeParse(response);
+
+    if (!envelope.success) {
+      this.logger('Error', 'Unreadable automated inbox listing', {
+        instanceId,
+        forestServerUrl: this.options.forestServerUrl,
+        error: envelope.error.message,
+      });
+
+      return [];
```
Scra3: issue (non-blocking): an unreadable listing envelope returns `[]` after the GET already renewed the lease, so every inbox of the environment vanishes while the server-side monitor reads healthy; throw it like `listAssignments` does.

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,508 @@
...
+    for (const { runState } of reconcilable) {
+      // Null belongs here too: the orchestrator binds the run before the assignment, so an
+      // assignment with no run is not a shape this executor knows how to read either.
+      if (!isTerminalRun(runState) && !isLiveRun(runState)) {
+        this.logger(
+          'Warn',
+          'Unexpected workflow run state, leaving the record out of every sweep until this executor knows it',
+          {
+            ...logContext,
+            runState,
```
Scra3: suggestion: add `recordId` and `workflowRunId` to this warning and the unknown-state one above, nobody can find the record left out of every sweep (PRD-1280) from these lines.

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,508 @@
...
+      this.logger('Warn', 'Unreadable record id, leaving the record out of the reconciliation', {
+        ...logContext,
+        recordId,
+        primaryKeyCount: config.primaryKeys.length,
+      });
+
+      return false;
+    });
+
+    if (readable.length === 0) return { outcome: 'skipped', items: [] };
```
Scra3: issue (non-blocking): turning this `'skipped'` into `'ok'` passes every test; add an inbox with no assignments whose candidate read throws and assert `sync` is never called.

packages/workflow-executor/src/adapters/agent-client-segment-reader.ts
```diff
@@ -0,0 +1,183 @@
...
+    excludedRecordIds: string[] | undefined,
+  ): AgentFilter {
+    const branches = [
+      segment.kind === 'filter' ? segment.conditionTree : null,
+      recordIds?.length
+        ? AgentClientSegmentReader.buildRecordIdFilter(primaryKeys, recordIds)
+        : null,
+      // Single-column keys only, as the port states: `not_in` takes a flat list of values, and the
+      // caller falls back to reading a wider page rather than negating a composite key.
+      excludedRecordIds?.length && primaryKeys.length === 1
```
Scra3: issue (non-blocking): dropping the `?.length` guard passes every test, yet `excludedRecordIds: []` is the first sweep of every new v2 inbox; assert it sends no `filters` param.

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,508 @@
...
+  }
+
+  private static paddedPageReason(
+    config: ServerAutomatedInboxConfig,
+    knownCount: number,
+  ): PaddedPageReason | undefined {
+    if (config.liana == null) return 'unknown-liana';
+    if (LIANAS_WITHOUT_NOT_IN.has(config.liana)) return 'liana-without-not-in';
+    if (config.primaryKeys.length !== 1) return 'composite-key';
+    if (knownCount > MAX_EXCLUDED_RECORDS) return 'too-many-known-records';
```
Scra3: suggestion: only 151 is tested, `>= 149` passes; add the exactly-150 case expecting the exclusion and `paddedPageReason: undefined`.

packages/workflow-executor/src/adapters/forest-server-automation-port.ts
```diff
@@ -0,0 +1,177 @@
...
+  async listAssignments(inboxId: string): Promise<ServerAutomatedInboxAssignment[]> {
+    const response = await this.callPort(
+      'listAutomatedInboxAssignments',
+      () => ServerUtils.query<unknown>(this.options, 'get', ROUTES.assignments(inboxId)),
+      inboxId,
+    );
+
+    // Thrown, not swallowed: an empty list here is indistinguishable from an inbox with nothing
+    // assigned, and the sweep would propose every record in the segment as a fresh candidate.
+    return ServerAutomatedInboxAssignmentsResponseSchema.parse(response).assignments;
```
Scra3: suggestion: nothing locks this throw in, a `safeParse ?? []` passes all port tests; add a malformed `{ assignments: 'x' }` case that must reject.

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,508 @@
...
+
+    return undefined;
+  }
+
+  private static readTimezone(timezone: string | null | undefined): string {
+    return timezone != null && IANAZone.isValidZone(timezone) ? timezone : 'UTC';
```
Scra3: suggestion: a `readTimezone` returning `'UTC'` unconditionally passes; assert a valid `Europe/Paris` reaches the agent unchanged.

packages/workflow-executor/src/adapters/server-types.ts
```diff
@@ -257,3 +260,152 @@ export interface ServerUpdateStepRequest {
...
+] as const;
+export type ServerAutomatedInboxSyncOutcome = (typeof SERVER_AUTOMATED_INBOX_SYNC_OUTCOMES)[number];
+
+export const ServerAutomatedInboxSyncResponseSchema = z.object({
+  results: z.array(
+    z.object({
+      recordId: z.string(),
+      // Unknown outcomes are kept as-is: the poller only counts them for logs, and refusing the
+      // whole response would stop a sync the server already applied.
+      outcome: z.string(),
```
Scra3: suggestion: tightening this to `z.enum(SERVER_AUTOMATED_INBOX_SYNC_OUTCOMES)` passes; add a case where an unknown outcome is kept, otherwise 4284c5895 changes only a type.

packages/workflow-executor/src/cli-core.ts
```diff
@@ -282,6 +283,12 @@ export function readEnvConfig(env: NodeJS.ProcessEnv, args: CliArgs): CliConfig
     httpPort: parsePositiveIntEnv('HTTP_PORT', env.HTTP_PORT) ?? DEFAULT_HTTP_PORT,
     forestServerUrl: env.FOREST_SERVER_URL,
     pollingIntervalS: parsePositiveIntEnv('POLLING_INTERVAL_S', env.POLLING_INTERVAL_S),
+    // Zero accepted here alone: an environment the orchestrator refuses logs an error on every
+    // cycle, and until now nothing let an operator stop it short of unsetting the secret.
+    automationPollingIntervalS:
+      env.AUTOMATION_POLL_INTERVAL_S?.trim() === '0'
```
Scra3: suggestion: `AUTOMATION_POLL_INTERVAL_S` has no case in `cli.test.ts` (`'0'`, `' 0 '`, `-1`, `abc`); and drop "until now" from the comment above, it describes the PR's history.

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,508 @@
...
+      // next page larger, which makes the next timeout likelier.
+      const [closed, candidates] = await Promise.all([
+        this.tryRead(logContext, 'closed records', () => this.reconcileClosed(config, assignments)),
+        this.tryRead(logContext, 'new candidates', () => this.readCandidates(config, assignments)),
+      ]);
+
+      // Reaching the agent at all is what the sync attests to. Reporting an empty poll when every
+      // read failed would tell the orchestrator this inbox is being swept while nothing is, which
+      // is the one thing a "no sync received" alert must never be lied to about. A read that
+      // legitimately had nothing to ask is a success, not a failure.
```
Scra3: nit: the last sentence is wrong, a `skipped` read counts as not-ok in the guard below, as the `SegmentRead` doc says.

packages/workflow-executor/src/automation-poller.ts
```diff
@@ -0,0 +1,508 @@
...
+      const page = await this.config.segmentReaderPort.listRecordIds({
+        ...AutomationPoller.segmentQuery(config),
+        excludedRecordIds: known,
+        pageSize: config.maxConcurrentRuns,
+      });
+
+      return { outcome: 'ok', items: page, requestedPageSize: config.maxConcurrentRuns };
+    }
+
+    // Fallback for an agent whose filters have no `not_in`, a composite key, and a set too large
```
Scra3: nit: "or", not "and", and `unknown-liana` is missing; `PaddedPageReason` already lists the cases, so the first paragraph can go.

packages/workflow-executor/src/adapters/server-types.ts
```diff
@@ -257,3 +260,152 @@ export interface ServerUpdateStepRequest {
...
+export const ServerAutomatedInboxConfigSchema = z.object({
+  inboxId: z.string().min(1),
+  renderingId: z.number().int().nonnegative(),
+  teamId: z.number().int().nonnegative().optional(),
+  workflowId: z.string().min(1).optional(),
+  collectionId: z.string().min(1).optional(),
+  collectionName: z.string().min(1),
+  primaryKeys: z.array(z.string().min(1)).min(1),
+  maxConcurrentRuns: z.number().int().positive(),
+  // Validated here rather than passed on: the agent answers 400 to an unknown zone, so an inbox
```
Scra3: nit: nothing is validated here, the schema is `z.string().nullish()` and the check lives in `readTimezone`, which already explains it.

packages/workflow-executor/src/adapters/server-types.ts
```diff
@@ -257,3 +260,152 @@ export interface ServerUpdateStepRequest {
...
+});
+export type ServerAutomatedInboxServiceAccountProfile = z.infer<
+  typeof ServerAutomatedInboxServiceAccountProfileSchema
+>;
+
+// Only what the poller reads is required. A field it merely logs must never be the reason an inbox
+// is dropped from the sweep, and `.nullable()` alone would still reject an omitted key.
+export const ServerAutomatedInboxConfigSchema = z.object({
+  inboxId: z.string().min(1),
+  renderingId: z.number().int().nonnegative(),
```
Scra3: nit: `renderingId` is only logged, so the comment above says it must not be required; make it optional or reword.

packages/workflow-executor/src/adapters/server-types.ts
```diff
@@ -257,3 +260,152 @@ export interface ServerUpdateStepRequest {
...
+  'todo',
+  'doing',
+  'done',
+  'canceled',
+  'auto-canceled',
+] as const;
+
+// States are read as plain strings, not enums: the poller only ever tests set membership, and one
+// assignment in a state a newer orchestrator introduced must not take the whole inbox down. The
+// known values live in the constants above, for the poller to compare against.
```
Scra3: nit: `SERVER_INBOX_ASSIGNMENT_STATES` is used nowhere, the poller has its own sets; delete it or use it.

packages/workflow-executor/src/automation-poller.ts
Scra3: Downgraded to non-blocking: releasing on `stop()` only covers graceful shutdowns, and the right fix (short TTL + heartbeat, release as an optional complement) is a design choice left to PRD-1274, where the options are listed. Not needed to merge this PR.

packages/workflow-executor/src/automation-poller.ts
Scra3: Fixed in a79b6b5f1 (PRD-1279): the denylist is gone. Only agents serving `_internal/capabilities` are asked, `not_in` is sent only when the primary key declares it, any other liana (forest-laravel included) pads, and a `not_in` read that still fails is retried padded in the same cycle.

packages/workflow-executor/src/automation-poller.ts
Scra3: Covered since a79b6b5f1: 'should report the candidate read failed when the padded fallback fails too' runs with a live assignment, so reconciliation is skipped while the candidate read fails, and it asserts no sync. Flipping this 'skipped' to 'ok' now turns it red.

packages/workflow-executor/src/adapters/agent-client-segment-reader.ts
Scra3: Moot since a79b6b5f1: the poller no longer sends an exclusion when nothing is known ('should read the run cap without asking for capabilities when nothing is known yet' asserts no excludedRecordIds), so an empty list never reaches this guard from the sweep.

packages/workflow-executor/src/automation-poller.ts
Scra3: Added in a79b6b5f1: 'should still exclude when exactly the maximum number of records is known' (150 ids, exclusion at the run cap, no paddedPageReason), next to the 151 case.

packages/workflow-executor/src/automation-poller.ts
Scra3: Fixed in a79b6b5f1, the sentence is gone.

packages/workflow-executor/src/automation-poller.ts
Scra3: Fixed in a79b6b5f1: the padded path moved to readPaddedCandidates and that paragraph was dropped, the PaddedPageReason union lists every case.

packages/workflow-executor/src/automation-poller.ts
Scra3: Filed as PRD-1299 (low): the lossy cast lives in the agent, not in the poller, and it only bites Snowflake-style bigint keys. Not in scope for this PR.

packages/workflow-executor/src/automation-poller.ts
Scra3: Filed as PRD-1300 (Observability, Medium), to land before the automation is switched on in production (forestadmin-server#8519). Not in this PR.

packages/workflow-executor/src/automation-poller.ts
Scra3: Filed as PRD-1301 (Observability, Medium), to land before the automation is switched on in production (forestadmin-server#8519). Not in this PR.

packages/workflow-executor/src/adapters/forest-server-automation-port.ts
Scra3: Filed as PRD-1302 (Observability, Medium), to land before the automation is switched on in production (forestadmin-server#8519). Not in this PR.

packages/workflow-executor/src/automation-poller.ts
Scra3: Filed as PRD-1303 (Observability, Medium), to land before the automation is switched on in production (forestadmin-server#8519). Not in this PR.

packages/workflow-executor/src/adapters/forest-server-automation-port.ts
Scra3: Covered in 124880eeb: 'should throw rather than read an unreadable answer as an inbox with nothing assigned'. Swapping the parse for a safeParse ?? [] now turns it red.

packages/workflow-executor/src/automation-poller.ts
Scra3: Covered since a79b6b5f1: 'should ask the agent which operators the primary key declares' asserts the fixture's Europe/Paris reaches the agent unchanged, and a readTimezone that always answers UTC turns it red.

packages/workflow-executor/src/adapters/server-types.ts
Scra3: Covered in 124880eeb: 'should keep an outcome from a newer orchestrator rather than refuse the answer'. Tightening to z.enum now turns it red.

packages/workflow-executor/src/cli-core.ts
Scra3: Covered in 124880eeb: cli.test.ts now checks '0' and ' 0 ' turn the sweep off, '45' is read as the interval, and '-1', 'abc', '1.5' are rejected. The "until now" comment is rewritten to state the reason only.

packages/workflow-executor/src/adapters/server-types.ts
Scra3: Fixed in 124880eeb, comment removed; readTimezone keeps the explanation.

packages/workflow-executor/src/adapters/server-types.ts
Scra3: Fixed in 124880eeb: renderingId is optional now, matching the comment, since the poller only logs it.

packages/workflow-executor/src/adapters/server-types.ts
Scra3: Fixed in 124880eeb: the constant and the sentence pointing at it are gone.

Scra3:
@hercemer42 on the two ADRs your review asked for ([review](https://github.com/ForestAdmin/agent-nodejs/pull/1906#pullrequestreview-5275445851)):

- The sweep sending record ids to the orchestrator is recorded: `docs/adr/2026-09-22-the-automation-sweep-sends-record-ids-to-the-orchestrator-and-decides.md`.
- The service-account token in `step-user.ts`: the ADR gate refuses this one, because it is not a decision this PR takes. `main` already mints the agent JWT with `authSecret` for whatever profile the orchestrator sends (`agent-client-agent-port.ts`, `mintToken`, used for every run-driven step). #1906 only extracts that function into `step-user.ts` and reuses it for the service account, same trust boundary, which the PR body states under "The runtime trusts the orchestrator on identity".

PR #1908, opened 2026-09-21
feat(audit-trail): gate audit detail values on the admin permission level [PRD-1260]

bexchauveto:
fixes PRD-1260 — implements the decision recorded on PRD-1259.

`previousValues` and `newValues` only reach a caller whose permission level is `admin` — exactly that level, not a privileged set. Every other caller gets each row with both objects empty: operation, author and timestamp are what makes the timeline readable, and they are not the data a non-admin isn't entitled to.

## What changes

| route | behaviour for a non-admin |
|---|---|
| `GET /_audit-trail/{collection}/{id}` | every row returned, both value objects `{}` |
| `GET /_audit-trail/{collection}/{id}/state` | **403** — nothing but reconstructed values, so there is no row left to hand back |
| `GET /_audit-trail/correlation/{key}` | same rows, same blanking |
| `GET` / `POST /_audit-trail/correlations` | same |

The gate **composes with** the existing scope withholding rather than replacing it — a row can fail both tests.

## Capability flag

`restrictsAuditTrailValuesToAdmins: true`, next to `canUseAuditTrail` in the capabilities payload. Constant, not config-driven: it states that this agent version enforces the rule itself, so the front can stop treating its own hidden expander as the guarantee. An older agent simply omits the key. **This is the name Ruby and the front mirror.**

## Notes

- `src/audit-trail/admin-gate.ts` holds the whole rule (~25 lines): `isAdmin` reads `context.state.user.permissionLevel`, straight off the Forest-signed JWT, so there is no round-trip and nothing the caller can forge.
- The cross-collection route (PRD-1257) is written on top of this rather than retrofitted.
- Existing route tests now declare `permissionLevel: 'admin'` on their caller, which is what they always meant.

## Tests

7 new cases (all five operations blanked, row identity preserved, admin unaffected, `/state` refused, both correlation routes) plus the capabilities payload. `1639 passed`, lint clean.

> [!NOTE]
> ### Gate audit-trail `previousValues` and `newValues` on admin permission level
> - Adds `isAdmin` predicate, `withholdValuesFromNonAdmin` redaction utility, and `assertCanReadAuditValues` route guard in [admin-gate.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1908/files#diff-5fd020bd42785f05291dd3c578ae8356ccfffa5c2b2194d9470b0480b136352a)
> - Applies redaction to correlation and per-record history routes so non-admin callers receive empty `previousValues` and `newValues` while retaining timeline metadata; the reconstructed-state route returns `ForbiddenError` for non-admins before querying the audit store
> - Advertises `restrictsAuditTrailValuesToAdmins: true` in every capabilities response via [capabilities.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1908/files#diff-346029c15b973e83cf5f8bee28875e579f2249197533e10d595877f285e049a0)
> - Behavioral Change: non-admin callers no longer receive audit value objects on any route; the state route now returns HTTP 403 instead of a reconstructed payload. Reviewers should check `AuditTrailRoute.handleStateAt` and `withholdValuesFromNonAdmin` for correct redaction across all operation types.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized dd415af.</sup>
> 
>


bexchauveto:
Closing — the scope of this was wrong, and the fix is to move it rather than amend it.

**What changed.** The rule is *not* "audit detail values are admin-only". It is: **a record's own history shows its diff to everybody; the project-level activity page shows it to admins only.** This PR gated the per-record route, `/state` and the two correlation routes — all record-level, all of which must stay open.

The gate now lives in #1911, on the one route it belongs to: `GET /forest/_audit-trail`, the cross-collection timeline that feeds the project-level page. The record-level routes are untouched by the whole stack again.

**Why it moved rather than shrank.** With the record routes ungated, the only consumer of the gate is the timeline route, so shipping the helper and the capability flag one PR ahead of the thing that uses them bought nothing. The stack is now three PRs, all rebased onto `main`:

- #1909 — scope answerability (now based on `main`)
- #1910 — `operation` filter
- #1911 — cross-collection timeline **+ the admin gate**

**The flag was renamed**: `restrictsAuditTrailValuesToAdmins` → **`restrictsProjectAuditValuesToAdmins`**. The old name read as global and would have invited the front to gate the record tab on it too — the exact mistake being corrected here. Nothing had adopted it yet, so the rename is free. Ruby and the front should mirror the new name.

Worth recording, since it revises the reasoning in PRD-1259: with record-level diffs open to everyone, the project-level gate is an **aggregation boundary, not a data barrier**. A non-admin who can read a collection can open any record and see the same before/after values — one record at a time, and only if they know which record. What stays admin-only is the firehose: every collection at once with no record known in advance. That is a real and defensible distinction, but it is not "they cannot see the values", which is how PRD-1259 and PRD-1260 are currently written. Both need amending.

PR #1910, opened 2026-09-21
feat(audit-trail): filter the per-record history by operation [PRD-1256]

bexchauveto:
fixes PRD-1256. Stacked on #1909 → #1908.

`GET /forest/_audit-trail/{collection}/{recordId}` accepts an `operation` query param.

## The contract (Ruby copies this verbatim)

- **param name**: `operation` (singular, like `search`; the store field is `operations`)
- **format**: comma-separated — `operation=create,delete` — consistent with `userIds` and `fields`
- **set**: closed — `create`, `update`, `delete`, `action`, `action_failed`
- **unknown value**: **400**, not dropped

That last one departs from `userIds`, which drops junk tokens. Deliberate: a silently ignored filter returns unfiltered rows into a list the caller believes is filtered — the exact failure the ticket warns about. `userIds` can afford to be lenient because a partial id list still filters; an ignored `operation` doesn't filter at all.

Empty or whitespace-only → no filter, same as `fields`.

## Pushed into the query, not applied after

`operations` lands in the shared `filters` object, so it reaches `listByRecord`, `countByRecord` **and** `listDistinctUsers`. `meta.count` and `meta.availableUsers` therefore agree with the rows, and paging stays correct — filtering the fetched page would have desynced all three.

SQL side is one clause in `buildHistoryWhereClause`; the in-memory test store gained the same filter so the double stays honest.

## Tests

6 route cases (single, comma-separated, count + author list, 400 on unknown, empty ignored) and a sql-store case covering `listByRecord` and `countByRecord` together. `1651 passed`, lint clean.

> [!NOTE]
> ### Add `operation` filter to audit-trail history endpoint
> - Adds an optional `operations` field to `AuditHistoryQuery` and the route's parsed filters, read from a comma-separated `operation` query parameter
> - Validates tokens against the closed set `AUDIT_OPERATIONS` (`create`, `update`, `delete`, `action`, `action_failed`); the first unrecognized value raises `ValidationError` and the store is not queried
> - SQL and in-memory stores now restrict history, count, and distinct-author queries to records whose operation is in the requested list; empty filters are omitted
> - Risk: callers passing an unsupported operation value now get a `ValidationError` instead of unfiltered results
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a215a4c.</sup>
> 
>


bexchauveto:
Rebased. #1908 is closed — its admin gate was scoped wrong and moved to #1911 (project-level timeline only). This PR is unchanged in substance; the inert `permissionLevel: 'admin'` that its new tests carried has been dropped, since the record route is no longer gated.

PR #1912, opened 2026-09-21
feat(workflow-executor): scope workflow start to the sub-workflow call

hercemer42:
A step running inside a Sub-workflow call now resolves "workflow start" to the record the calling step pinned, and reports no source record when that record, or the run's own record, belongs to another collection than the called workflow. The run mapper rebuilds the call frames from the raw history, unwinds a nested pin of "workflow start" out to the enclosing call, and carries the innermost frame's pin and called collection into the execution context; the record behind the pin is resolved in the shared record-step resolver, which reaches the run store. Both fields stay optional, so a run with no call scope behaves exactly as before.

## What changes

1. Before, a step inside a called workflow read "workflow start" as the record the run was launched on, whatever the calling step pinned. Now it reads the record pinned on the innermost open call, and still falls back to the run's record when the call pins none.
2. Before, a nested call pinning "workflow start" had nowhere to resolve to. Now that pin resolves in the enclosing call's own scope, frame by frame, ending on the run's record.
3. Before, a call into a workflow built on another collection handed the caller's record to the child's steps, which acted on a record of the wrong collection. Now such a step reports no source record instead — the pinned record and the run's own record alike.
4. Before, a call frame that a revision replaced or a cancellation dropped could still be read as the live one. Now revised and cancelled entries are skipped and each call is paired with its close, so a closed call pops back to the run's record and a replaced frame leaves nothing behind.

## Decisions

20 decisions. 19 of them sit on the code they concern, as review comments in the Files changed tab — answer them there. Three were rewritten after review rounds changed what they described: how a collection mismatch is reported, how it is classified, and what the call scope carries. The one with no single place in the diff:

- [x] **NOT IMPLEMENTED** · The published version the editor gates on is not recorded on the ticket

  **Spec left open:** Contract item 6 asks for the exact version of the release carrying items 1 to 4, recorded on this ticket.

  **Decided:** It is left until after the release. Semantic release assigns the version on merge, so no number exists while the code is written.

  **Impact:** Whoever builds the editor side has no floor to gate on until the published version is commented on the ticket, after this merges.

  <details><summary>For a run answering a reply</summary>

  `packages/workflow-executor/package.json:3` reads `1.29.0`, the version already released before this change, at e0a46add0.

  The last five commits on the base branch are release version bumps, checked at e0a46add0.

  PRD-1263 carried no comments when read on 2026-09-21.

  </details>

  *Tick to agree, or comment on the pull request quoting it to push back.*

  

### General

### Security

## Squash message

```
feat(workflow-executor): scope workflow start to the sub-workflow call

A step inside a called workflow read "workflow start" as the record the run was
launched on, whatever the Sub-workflow step that called it pinned. A call into a
workflow built on another collection handed the child steps a record of the
caller's collection.

"workflow start" now resolves to the record pinned on the innermost open call,
falling back to the run's record where the call pins none. A pin of "workflow
start" on a nested call resolves one frame out, ending on the run's record, and
the pin is matched only among the steps of the frame that wrote it, since a BPMN
step id is unique inside its own diagram.

A record whose collection is not the called workflow's is refused with an error
naming both collections and the Sub-workflow step where the record is set. A
source step that ran and loaded nothing is now classified empty-source rather
than as a configuration fault, so a run card reading the classification can name
the step to go back to.

Both wire fields are optional: a runtime carrying this change behaves identically
until an orchestrator starts sending them.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
```

> [!NOTE]
> ### Scope `workflow-start` record resolution to sub-workflow call scope
> - Reconstructs `CallScope` from workflow history in `toAvailableStepExecution` to pass the caller-pinned record, called collection, and originating frame to step executors.
> - Reworks `RecordStepExecutor.resolveSourceRecordRef` to use the call scope: it resolves the caller's pinned record, falls back to the run record for unpinned calls, and rejects cross-collection pins or unreached source steps.
> - Adds `empty-source` and `call-site` error kinds with `WorkflowCallSiteError` subclasses for collection mismatches and unreached source steps.
> - Behavioral Change: Empty load-related source records now classify as `empty-source` instead of `configuration`; sub-workflow collection mismatches now raise `call-site` errors instead of executing the agent read.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 2994f82.</sup>
> 
>


packages/workflow-executor/src/adapters/run-to-available-step-mapper.ts
```diff
@@ -116,6 +119,51 @@ function toPreviousSteps(
     .filter((s): s is Step => s !== null);
 }
 
+// A missing, empty or non-string wire value reads as absent: a call sending neither the pin nor the
+// called collection behaves as one that predates them.
+function toNonEmptyString(value: unknown): string | undefined {
+  return typeof value === 'string' && value.length > 0 ? value : undefined;
```
hercemer42, line 128: Claude Opus 5 (1M context): **Should fix**: a Sub-workflow call whose pin arrives empty runs the child step on the record the run was launched on.

`packages/workflow-executor/CLAUDE.md` makes an empty pin a lost target that errors, but `run-to-available-step-mapper.ts:125` reads it as absent and `record-step-executor.ts:87` uses `baseRecordRef`.

A way out: read a present-but-empty `selectedRecordStepId` as a lost pin in `toNonEmptyString`.

Also moves: `run-to-available-step-mapper.test.ts:1234` asserts an empty `selectedRecordStepId` pins no record.
</details>

1. Build a workflow calling a Sub-workflow on the run's own collection, pinning a Load Related Record step.
2. Have the orchestrator send that call with `preRecordedArgs.selectedRecordStepId` set to an empty string.
3. Run it and open the child's Update Data step.
4. It wrote the record the run was launched on, and reports success.
</details>

hercemer42 (commented):
Spec (`PRD-1263`): conforms.

packages/workflow-executor/src/executors/record-step-executor.ts
```diff
@@ -76,13 +76,36 @@ export default abstract class RecordStepExecutor<
...
+    return record;
+  }
+
+  // The record a Load Related Record step loaded, with that step's title for the messages about it.
+  // previousSteps are already restricted to the live path; in a loop the same id can appear more
+  // than once, so we take the most recent occurrence.
+  private async resolveStepRecordRef(
+    stepId: string,
+  ): Promise<{ record: RecordRef; sourceTitle?: string }> {
     const matches = this.context.previousSteps.filter(
```
hercemer42, line 139: Claude Opus 5 (1M context): **Should fix**: a step inside a Sub-workflow call can act on a record the called workflow loaded, not the one the calling step pinned.

This filter spans both workflows and `record-step-executor.ts:114` takes the most recent match, so the callee's copy of a repeated BPMN step id wins.

A way out: carry the call's own `stepIndex` in `CallScope` and bound the caller-scope lookup to steps before it.
</details>

1. Build a workflow that calls itself as a Sub-workflow, pinning its own Load Related Record step.
2. Let the called copy run that same step before a step pinned to workflow start.
3. Run it and read the target of the child's Update Data step.
4. It wrote the record the called copy loaded, not the record the call pinned.
</details>

hercemer42 (commented):
Correctness lens of run `1912-20260921183133-hercemer42`, which returned after that run's review was submitted.

packages/workflow-executor/src/adapters/run-to-available-step-mapper.ts
hercemer42: Claude Opus 5 (1M context): The empty pin still reads as absent, and `CLAUDE.md` now records why that is the call's own rule.
d925ee0
<details><summary>Why</summary>

`CallScopeSchema` rejects an empty string, so carrying one would report the whole run malformed on every poll.
Failing the single step instead needs a producer, and PRD-1262 writes that serialiser after this merges.
Deciding what an empty pin means belongs there, against the code that emits it.
</details>

packages/workflow-executor/src/executors/record-step-executor.ts
hercemer42: Claude Opus 5 (1M context): The call pin now resolves only among the steps that ran before the call that wrote it.
d925ee0
<details><summary>Why</summary>

`callScope.pinnedAtStepIndex` carries that index and the shared resolver bounds its lookup with it.
A step resolving an id it declared itself passes no bound and keeps most-recent-wins for a LinkTo loop.
</details>

packages/workflow-executor/src/adapters/run-to-available-step-mapper.ts
hercemer42: Accepted: carrying an empty pin would make the run malformed rather than failing the step, and `CLAUDE.md` now records that.

Predicate: an empty `selectedRecordStepId` reaching `CallScopeSchema` fails its `min(1)` and the mapper raises `DomainValidationError`.
`forest-server-workflow-port.ts:81` then files the run malformed, so every poll returns it and fails again.
No regression to pin, as no behaviour changed and the counter-argument is recorded in `packages/workflow-executor/CLAUDE.md`.
</details>

packages/workflow-executor/src/executors/record-step-executor.ts
hercemer42: Closed: the pin now resolves only among steps that ran before the call that wrote it.

Predicate: a Load Related Record step of the called workflow answering the caller's pin, needing a stepIndex at or above the call.
Closed by the bound, since every step inside a call carries a higher index than the call that opened it.
Pinned by a test in `read-record-step-executor.ts` that reads 99 with the bound and 77 without it, run both ways.
</details>

packages/workflow-executor/src/executors/record-step-executor.ts
```diff
@@ -76,17 +76,45 @@ export default abstract class RecordStepExecutor<
   // trigger-action "On record"). The reference is a stable BPMN step id (or the
   // WORKFLOW_START_STEP_ID sentinel), not a runtime index — so it survives the index shifts a
   // revision causes (clones keep their step id) and is knowable by the editor at build time.
-  // previousSteps are already restricted to the live path; in a loop the same id can appear more
-  // than once, so we take the most recent occurrence.
   protected async resolveSourceRecordRef(stepId: string): Promise<RecordRef> {
```
hercemer42, line 82: Claude Opus 5 (1M context)

**SHARED CODE** · The call-scoped resolution and the collection guard sit in the source resolver all four record steps share

**Spec left open:** The ticket names `resolveSourceRecordRef` and lists tests for one step type, without saying which step types the change reaches.

**Decided:** Change the one method the four record steps call, so all four read `workflow-start` through the call. Splitting it per step type would need a flag the ticket never describes.

**Impact:** All four step types use the pinned record inside a call, and refuse one of another collection, from the first run on this release.

**Alternative:** Guard only the step type a reported symptom names and leave the other three resolving to the run's record.

**Trade-off:** It narrows what a live workflow can start doing differently, at the cost of two meanings for workflow start inside one called workflow.

<details><summary>For a run answering a reply</summary>

`resolveSourceRecordRef` has four callers, in the read-record, update-record, trigger-record-action and load-related-record executors, checked at e0a46add0.

`condition-step-executor.ts:209` repeats the rule in a comment without calling the method, so it is untouched, at e0a46add0.

The executor and adapter suites pass, 1023 tests, with this change over e0a46add0.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/adapters/run-to-available-step-mapper.ts
```diff
@@ -116,6 +119,57 @@ function toPreviousSteps(
     .filter((s): s is Step => s !== null);
 }
 
+// A missing, empty or non-string wire value reads as absent: a call sending neither the pin nor the
+// called collection behaves as one that predates them.
+function toNonEmptyString(value: unknown): string | undefined {
```
hercemer42, line 127: Claude Opus 5 (1M context)

**DECISION** · An empty or null value on the call reads as absent rather than as a pin that lost its target

**Spec left open:** The ticket types both call fields as optional and treats a null collection name as absent, without saying what an empty string means.

**Decided:** Empty, null and non-string values are dropped, so such a call behaves as one predating the fields. An empty name matches no collection, so arming the guard on it would fail every step of the call.

**Impact:** A call whose pin or collection name arrives empty keeps today's behaviour instead of failing the called workflow's steps.

**Alternative:** Reject an empty value as a malformed payload and fail the run, as a record step's own cleared pin is rejected.

**Trade-off:** It surfaces a bad caller at once, at the cost of aborting runs over a field the contract makes optional.

<details><summary>For a run answering a reply</summary>

`run-to-available-step-mapper.ts:124-126` drops anything that is not a non-empty string, over e0a46add0.

`server-types.ts:234` types `timezone` as `string | null` and `toAvailableStepExecution` falls back rather than failing, at e0a46add0.

`packages/workflow-executor/CLAUDE.md:45` records the opposite rule for a record step's own pin, which errors instead, at e0a46add0.

Two parameterised tests over `run-to-available-step-mapper.ts` cover null, empty, undefined and a number, at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/executors/record-step-executor.ts
```diff
@@ -76,17 +76,45 @@ export default abstract class RecordStepExecutor<
...
-    if (stepId === WORKFLOW_START_STEP_ID) {
-      return this.context.baseRecordRef;
+    if (stepId !== WORKFLOW_START_STEP_ID) return (await this.resolveStepRecordRef(stepId)).record;
```
hercemer42: Claude Opus 5 (1M context)

**DECISION** · A step pinned to a named source step is not checked against the called workflow's collection

**Spec left open:** Contract item 5 lists what stays unchanged and does not say whether a pin naming a step is guarded inside a called workflow.

**Decided:** Only the `workflow-start` path is guarded, as the ticket scopes just that resolution. A named pin points at a record the builder chose, not at the record the call handed in.

**Impact:** A child step pinned to a Load Related Record step acts on that record even where its collection differs from the called workflow's.

**Alternative:** Check every source record a child step resolves against the called workflow's collection.

**Trade-off:** It would catch a child workflow whose own load step drifted to another collection, at the cost of failing steps that work today.

<details><summary>For a run answering a reply</summary>

`record-step-executor.ts:80` returns the named pin's record before the guard is reached, over e0a46add0.

The test named 'leaves a step pinned to a load-related step unguarded by the called collection' passes at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/executors/record-step-executor.ts
```diff
@@ -76,17 +76,45 @@ export default abstract class RecordStepExecutor<
...
+      ? await this.resolveStepRecordRef(callScope.selectedRecordStepId, callScope.pinnedAtStepIndex)
+      : { record: baseRecordRef, sourceTitle: undefined };
+
+    // A record of another collection than the called workflow is not what its steps were built
+    // against, so the step reports no source record rather than acting on the caller's.
+    if (
+      callScope.calledWorkflowCollectionName !== undefined &&
+      record.collectionName !== callScope.calledWorkflowCollectionName
+    ) {
+      throw new SourceRecordMissingError(sourceTitle);
```
hercemer42: Claude Opus 5 (1M context)

**DECISION** · A collection mismatch is reported with the existing no source record message rather than a new error

**Spec left open:** The ticket says to raise `SourceRecordMissingError` on a mismatch, pinned record and `baseRecordRef` fallback alike, without saying what the message names.

**Decided:** The pinned step's title is used where the call names a step, and the no-title wording otherwise. That error class already renders both forms, so no new wording ships.

**Impact:** An operator on a cross-collection call reads that the source step loaded no record, when that step runs.

**Alternative:** Add an error naming the called workflow's collection and the record's.

**Trade-off:** It tells the builder which two collections disagree, at the cost of a new string on an error four step types share.

<details><summary>For a run answering a reply</summary>

`errors.ts:576-589` renders the quoted title where one is given and 'its source step' otherwise, over e0a46add0.

`base-step-executor.ts:110-118` builds the error outcome from that message, at e0a46add0.

Two tests in `read-record-step-executor.test.ts` assert the two messages, for the pinned record and for the fallback, at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/executors/record-step-executor.ts
```diff
@@ -76,17 +76,45 @@ export default abstract class RecordStepExecutor<
...
+    // Inside a Sub-workflow call, "workflow start" means the record the calling step pinned; a
+    // call pinning none still starts from the record the run was launched on.
+    const { record, sourceTitle } = callScope.selectedRecordStepId
+      ? await this.resolveStepRecordRef(callScope.selectedRecordStepId, callScope.pinnedAtStepIndex)
+      : { record: baseRecordRef, sourceTitle: undefined };
+
+    // A record of another collection than the called workflow is not what its steps were built
+    // against, so the step reports no source record rather than acting on the caller's.
+    if (
+      callScope.calledWorkflowCollectionName !== undefined &&
```
hercemer42, line 115: Claude Opus 5 (1M context)

**DECISION** · A collection mismatch is classified as a configuration fault and blames no step

**Spec left open:** The ticket leaves error classification as it is, without saying what a guard mismatch writes into the step’s context.

**Decided:** It carries `configuration`, from the error family it joins, and no source step index. The workflow’s shape is wrong rather than any step it ran.

**Impact:** The run card keeps today’s framing and offers no rewind, which is right because revising the source would load the same collection again.

**Alternative:** Classify it `operator`, which is the one kind that offers a rewind.

**Trade-off:** That would drop the admin banner, at the cost of sending the operator back through a step that answers the same way.

<details><summary>For a run answering a reply</summary>

`getRewindTarget` returns null for every kind but `operator`, so no rewind is offered here, on forestadmin@594197017e.

`shouldPromptAdminContact` is true for every kind but `operator`, so `configuration` renders what an unclassified error rendered before, on forestadmin@594197017e.

No source step index is set: on the `baseRecordRef` fallback there is no source step at all, at ae8f5b7.

Rests on: A collection mismatch raises its own error naming both collections

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/executors/record-step-executor.ts
```diff
@@ -76,17 +76,45 @@ export default abstract class RecordStepExecutor<
...
+      record.collectionName !== callScope.calledWorkflowCollectionName
```
hercemer42, line 116: Claude Opus 5 (1M context)

**DECISION** · Collection names are compared exactly, with no case folding

**Spec left open:** The ticket says to compare the resolved record's collection with `calledWorkflowCollectionName` before using it, and does not say how.

**Decided:** The two names are compared as written. Both are Forest collection names from the same schema, and nothing in this package folds the case of one.

**Impact:** A call naming a collection that differs from the record's only in case is a mismatch, and the step reports no source record.

**Alternative:** Fall back to a case-insensitive match once the exact one fails.

**Trade-off:** It forgives a name whose case drifted between the editor and the schema, at the cost of joining two collections Forest treats as different.

<details><summary>For a run answering a reply</summary>

`record-step-executor.ts:95` compares with a strict inequality, over e0a46add0.

The one case-insensitive name match in this package is a polymorphic discriminator tried after an exact one, at `load-related-record-step-executor.ts:277-281`, e0a46add0.

The test named 'treats a called collection differing from the record only in case as a mismatch' passes at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/test/executors/read-record-step-executor.test.ts
```diff
@@ -1548,4 +1548,373 @@ describe('ReadRecordStepExecutor', () => {
...
+      });
+    }
+
+    function makeWorkflowStartStep(fieldNames: string[]) {
+      return makeStep({
+        preRecordedArgs: { selectedRecordStepId: WORKFLOW_START_STEP_ID, fieldNames },
+      });
+    }
+
+    it('resolves workflow-start to the record the call pinned, not the run record', async () => {
```
hercemer42, line 1611: Claude Opus 5 (1M context)

**DECISION** · The call scope tests extend the Read Record executor's test file

**Spec left open:** The ticket says the record step resolution tests land in `packages/workflow-executor/test/executors/` and names no file.

**Decided:** They extend the Read Record executor's file, which already covers the `workflow-start` sentinel and carries the mocks. The class holding the shared resolution is abstract and has no file of its own.

**Impact:** A reader looking for the guard finds it beside the other pin tests, and the four step types are covered through one of them.

**Alternative:** Add a test file for the shared record step executor with its own mocks and a local subclass.

**Trade-off:** It names the layer actually under test, at the cost of about one hundred and eighty duplicated lines of mock setup.

<details><summary>For a run answering a reply</summary>

Four executors call the shared resolution, at `read-record-step-executor.ts:27`, `update-record-step-executor.ts:208`, `trigger-record-action-step-executor.ts:145` and `load-related-record-step-executor.ts:192`, e0a46add0.

The existing sentinel test already lives in `read-record-step-executor.test.ts`, at e0a46add0.

Thirteen call scope tests now sit in that file, from line 1611, at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/test/executors/read-record-step-executor.test.ts
```diff
@@ -1548,4 +1548,373 @@ describe('ReadRecordStepExecutor', () => {
...
+        2,
+        expect.objectContaining({ collection: 'orders', id: [99] }),
+        expect.objectContaining({ id: 1 }),
+      );
+    });
+
+    // The editor keeps its own copy of this literal in
+    // app/features/workflow-editor/configuration.ts (WORKFLOW_START_STEP_ID). The two
+    // repositories share no code, so the literal is the whole agreement.
+    it('keeps the workflow-start sentinel the editor mirrors in app/features/workflow-editor/configuration.ts', () => {
```
hercemer42, line 1925: Claude Opus 5 (1M context)

**DECISION** · The editor's file is named in the title of the test pinning the `workflow-start` literal, not in an assertion message

**Spec left open:** The contract asks for a test asserting the `workflow-start` literal and naming the editor's copy of it in the assertion message.

**Decided:** The title carries the editor path and the constant name, because a Jest expectation takes no message argument. Jest prints the title on failure, so the reader still gets the path.

**Impact:** Whoever changes the literal reads the editor path in the failure output of that one test.

**Alternative:** Assert with node's assert module, whose equality check takes a message string.

**Trade-off:** It puts the path in the failure text itself, at the cost of a second assertion style in a file of Jest expectations.

<details><summary>For a run answering a reply</summary>

`WORKFLOW_START_STEP_ID` is the literal `workflow-start` at `step-definition.ts:250`, at e0a46add0.

Every assertion in the two touched test files is a Jest expectation, checked at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/adapters/run-to-available-step-mapper.ts
```diff
@@ -116,6 +119,57 @@ function toPreviousSteps(
...
+function toCallScope(
+  history: ServerStepHistory[],
+  pending: ServerStepHistory,
+): CallScope | undefined {
+  if (!pending.childrenWorkflowId) return undefined;
+
+  const openCalls: Array<{ stepIndex: number; definition: ServerStartSubWorkflow }> = [];
+
+  history.forEach(entry => {
+    if (entry.revised || entry.cancelled || entry.stepIndex >= pending.stepIndex) return;
```
hercemer42, line 150: Claude Opus 5 (1M context)

**DECISION** · A call frame counts as open even when its history entry is not marked done

**Spec left open:** The ticket says to pair the call with its close after dropping revised and cancelled entries, and says nothing about done.

**Decided:** Every entry before the pending step's own index is paired, whatever its done flag. The orchestrator marks the calling step done before the child's first step becomes pending.

**Impact:** No one observes a difference today, and a future orchestrator forking before it marks the call done would still scope the child's steps.

**Alternative:** Also require done, as the filter building `previousSteps` over the same history does.

**Trade-off:** It would match the neighbouring filter exactly, at the cost of losing the scope where a call entry stays open while its child runs.

<details><summary>For a run answering a reply</summary>

`run-to-available-step-mapper.ts:117` filters previous steps on done, revised, cancelled and index, at e0a46add0.

`run-to-available-step-mapper.ts:140` omits done from the same filter for the call frames, over e0a46add0.

Every call entry in `run-to-available-step-mapper.test.ts` sets done true, so both rules pass the tests, at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/adapters/run-to-available-step-mapper.ts
```diff
@@ -116,6 +119,57 @@ function toPreviousSteps(
...
+}
+
+// The Sub-workflow call the pending step runs inside, if any. Reconstructed from the raw history
+// because previousSteps drops the navigation steps; revised and cancelled entries are skipped and
+// each call paired with its close, as the orchestrator reads its own stack.
+function toCallScope(
+  history: ServerStepHistory[],
+  pending: ServerStepHistory,
+): CallScope | undefined {
+  if (!pending.childrenWorkflowId) return undefined;
```
hercemer42, line 138: Claude Opus 5 (1M context)

**DECISION** · The open call is found by pairing start and close entries, and used only where the pending entry names a child workflow

**Spec left open:** The ticket says to reconstruct the call scope from the raw `run.workflowHistory` and to mirror `getSubWorkflowCallStack`, without fixing the rules.

**Decided:** Each call is paired with its close over the raw history, and the innermost still open before the pending step is taken. It is attached only where the pending entry carries `childrenWorkflowId`.

- Entries flagged revised or cancelled are dropped before pairing, so a replaced call frame leaves no scope behind.
- Each close entry pops the last open call, so a closed call pops back to the record the run was launched on.
- Entries at or after the pending step's index are ignored, as the live path already is for `previousSteps`.
- A pending entry carrying no `childrenWorkflowId` gets no scope, so a step back in the calling workflow is unscoped.

**Impact:** A step of a called workflow is scoped to its call from the first run, and a step back in the calling workflow is not.

**Alternative:** Match the pending entry's `childrenWorkflowId` against the call entries' `workflowId` instead of pairing.

**Trade-off:** It would read the frame from one field, at the cost of an ambiguous match where one workflow calls the same child twice.

<details><summary>For a run answering a reply</summary>

`run-to-available-step-mapper.ts:131-165` holds the four rules, over e0a46add0.

`run-to-available-step-mapper.ts:113-119` applies the same revised, cancelled and index filter to `previousSteps`, at e0a46add0.

Seven tests in `run-to-available-step-mapper.test.ts` cover nesting, closing, a replayed frame and a revised one, at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/types/validated/execution.ts
```diff
@@ -39,6 +39,20 @@ export enum TriggerType {
 }
 export const TriggerTypeSchema = z.nativeEnum(TriggerType);
 
+// The Sub-workflow call the pending step runs inside: the record the calling step pinned and the
+// collection the called workflow is built on. Both optional — a call sending neither behaves as
+// before. Grouped rather than loose, as both describe the same call. `pinnedAtStepIndex` is the
+// index of the call that wrote the pin, which bounds where the pinned step is looked for, and so
+// travels with the pin and never without it.
+export const CallScopeSchema = z
```
hercemer42, line 48: Claude Opus 5 (1M context)

**DECISION** · The call's pin and the called collection travel as one optional field, not two

**Spec left open:** The ticket asks to carry both values through `AvailableStepExecutionSchema` as fields distinct from `baseRecordRef`, without fixing their shape.

**Decided:** Both values describe the same call, so they are grouped the way `baseRecordRef` groups its own parts. The schema rejects unknown keys, so one key is one place to extend.

**Impact:** An orchestrator author reads one optional object in the wire shape, and the executor can tell a run outside any call from one inside.

**Alternative:** Two optional top-level fields on the step execution.

**Trade-off:** It would flatten the wire shape, at the cost of losing the marker that says the step runs inside a call.

<details><summary>For a run answering a reply</summary>

`execution.ts:45-51` and `execution.ts:68` declare the object and its optional slot, both strict, over e0a46add0.

`step-executor-factory.ts:184` is the only place an execution context is built, at e0a46add0.

`toAvailableStepExecution` emits an empty object for a call sending neither value, asserted by a test at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/test/executors/step-executor-factory.test.ts
```diff
@@ -88,6 +93,32 @@ describe('StepExecutorFactory.create — MCP OAuth re-auth', () => {
     expect(result.stepOutcome).not.toHaveProperty('awaitingInputReason');
   });
 
+  // The only seam between the mapper that builds the call scope and the executor that reads it: a
+  // step of the called workflow's collection resolves, and one of another collection does not.
+  it('hands the sub-workflow call scope to the executor it builds', async () => {
```
hercemer42, line 98: Claude Opus 5 (1M context)

**DECISION** · The line handing the call scope to the executor is tested outside the two test files the ticket names

**Spec left open:** The ticket lands tests in the record step tests and the mapper tests, and covers nothing joining the two.

**Decided:** A third test file gets one case, because dropping that line disables the feature while both named suites stay green. Its test group was widened from the single topic it held.

**Impact:** A reviewer sees a change to a file the ticket does not name, whose test group is no longer only about MCP re-auth.

**Alternative:** Leave the line untested, as every other value the factory passes into the context is.

**Trade-off:** It keeps the change inside the two named files, at the cost of nothing failing if the feature is silently unwired.

<details><summary>For a run answering a reply</summary>

Removing the line turns the new test's message into 'Unexpected error during step execution', run over e0a46add0.

No other value the factory passes into the context has a test in that file, at e0a46add0.

The test group was widened from MCP OAuth re-auth to `StepExecutorFactory.create`, at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/types/validated/execution.ts
```diff
@@ -39,6 +39,20 @@ export enum TriggerType {
...
+  .object({
+    selectedRecordStepId: z.string().min(1).optional(),
+    pinnedAtStepIndex: z.number().int().nonnegative().optional(),
```
hercemer42: Claude Opus 5 (1M context)

**DECISION** · The call scope carries the index of the call that wrote the pin, a third field beyond the two the ticket names

**Spec left open:** The ticket`s scope bullet carries `selectedRecordStepId` and `calledWorkflowCollectionName` through the schema, and says nothing about which steps the pin is matched against.

**Decided:** `CallScope` also carries `pinnedAtStepIndex`, and the pin resolves only against steps that ran before it. A step id is unique inside its own diagram, while `previousSteps` flattens every frame.

**Impact:** A workflow calling a copy of itself resolves the record the caller pinned, not the callee`s same-id step.

**Alternative:** Match the pin over the whole flattened history, most-recent-wins, as the step`s own pin still does.

**Trade-off:** One more field on an internal type, against a wrong record the collection guard cannot catch.

<details><summary>For a run answering a reply</summary>

With the bound disabled the added test resolves `orders#77`, the callee`s record; with it, `orders#99`, the pinned one, on d925ee0.

The step`s own pin is unbounded and keeps most-recent-wins, so LinkTo loops are unaffected, on d925ee0.

`pinnedAtStepIndex` is built in the mapper and never leaves the runtime, so nothing on the wire changes, on d925ee0.

Without the bound a called workflow repeating the caller`s step id shadows the step the caller pinned, on d925ee0.

Rests on: The call`s pin and the called collection travel as one optional field, not two

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/adapters/run-to-available-step-mapper.ts
hercemer42, line 1: Claude Opus 5 (1M context)

**DECISION** · The nested pin is resolved when the run payload is read, not when the step runs

**Spec left open:** The ticket carries the innermost open call's `selectedRecordStepId` through to the executor, without saying which layer walks the enclosing calls.

**Decided:** `toAvailableStepExecution` walks the open calls outwards and sends one plain step id. The executor reads that id and never meets the sentinel as a pin.

**Impact:** A reader of the executor finds no frame stack, and anyone changing how nesting resolves changes `toAvailableStepExecution` alone.

**Alternative:** Carry every open call frame on the execution context and let the executor walk them.

**Trade-off:** It keeps each frame's own pin available to a later feature, at the cost of walking code in two layers.

<details><summary>For a run answering a reply</summary>

The walk drops the sentinel before the context is built, at `run-to-available-step-mapper.ts:154-157` over e0a46add0.

`record-step-executor.ts:87-89` takes one step id and reaches no enclosing frame, over e0a46add0.

Two tests of that function cover a nested pin and a chain of pins ending on nothing, at e0a46add0.

Rests on: A call pinning `workflow-start` inside another call takes the record the enclosing call pinned

</details>

*Resolve to agree, reply to push back.*

hercemer42, line 1: Claude Opus 5 (1M context)

**DECISION** · A call pinning `workflow-start` inside another call takes the record the enclosing call pinned

**Spec left open:** Contract item 3 says calls nest and the innermost wins, not what `workflow-start` means when the innermost call is the one pinning it.

**Decided:** The pin walks outwards through the open calls until one names a step, ending on the record the run was launched on. `workflow-start` already means the enclosing frame's pinned record for that frame's own steps.

**Impact:** A builder who nests two calls and pins workflow start on the inner one gets the outer call's record, from its first run.

**Alternative:** Resolve an inner `workflow-start` straight to the record the run was launched on.

**Trade-off:** It is shorter to explain, at the cost of one phrase naming two different records at two depths.

<details><summary>For a run answering a reply</summary>

`run-to-available-step-mapper.ts:152-157` holds the outward walk, over e0a46add0.

The walk ends on nothing when every open call pins the sentinel, so the resolver falls back to `baseRecordRef`, at e0a46add0.

The mapper test named 'resolves a nested workflow-start pin against the enclosing call' passes at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

hercemer42, line 1: Claude Opus 5 (1M context)

**DECISION** · A call sending neither value still marks the pending step as running inside a call

**Spec left open:** The ticket says both fields absent is today's behaviour exactly, without saying whether such a call is recorded at all.

**Decided:** `toAvailableStepExecution` emits an empty object rather than nothing, because the two resolve identically. Collapsing it would add a branch and a test for a difference nothing observes.

**Impact:** Nobody observes a difference today, and a later feature asking whether a step runs inside a call gets the answer without a new field.

**Alternative:** Return no scope at all when neither value is present.

**Trade-off:** It matches the contract wording literally, at the cost of losing the only signal that a step runs inside a call.

<details><summary>For a run answering a reply</summary>

`run-to-available-step-mapper.ts:161-164` spreads each value only when present, so both absent leaves an empty object, over e0a46add0.

The mapper test named 'carries neither value for a call that sends neither' pins the empty result, at e0a46add0.

The mapped execution stays in process and is never persisted or logged whole, checked at e0a46add0.

Rests on: The call's pin and the called collection travel as one optional field, not two

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/executors/record-step-executor.ts
hercemer42, line 1: Claude Opus 5 (1M context)

**DECISION** · A call pinning a step the run never took keeps failing as invalid parameters

**Spec left open:** The ticket adds the guard and does not say what happens where the pinned step is absent from `previousSteps`.

**Decided:** The existing branch stands, so the step reports invalid pre-configured parameters. A pin is a build-time reference, and a missing one is a broken configuration rather than a missing record.

**Impact:** A builder who deletes the pinned step sees the called workflow's steps fail with the invalid parameters message, on the next run.

**Alternative:** Report no source record instead, as the collection guard does.

**Trade-off:** It would give one message for every way a call fails to hand a record over, at the cost of hiding a configuration error.

<details><summary>For a run answering a reply</summary>

`record-step-executor.ts:139` is the pre-existing branch, now also reached through the call's pin, over e0a46add0.

The test named 'reports invalid parameters when the call pins a step the run never took' asserts the message, at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/CLAUDE.md
hercemer42, line 1: Claude Opus 5 (1M context)

**DECISION** · The change is documented in the package's own guide and nowhere public

**Spec left open:** The ticket does not say where this change is documented, or whether any public page goes stale.

**Decided:** One sentence was added to the package guide passage that already describes the sentinel. No public page describes how a step resolves its source record, so none contradicts the change.

**Impact:** A reader of the public documentation sees no change, while the package guide now says what the sentinel resolves to inside a call.

**Alternative:** Add the rule to the public workflows overview page now.

**Trade-off:** It would land the documentation with the code, at the cost of describing behaviour no orchestrator produces until the sibling ticket ships.

<details><summary>For a run answering a reply</summary>

Searches of the Forest documentation for the sub-workflow call and the step source record return only the workflows overview step list, read 2026-09-21.

`packages/workflow-executor/README.md` covers deployment and MCP only, with no step semantics, at e0a46add0.

`packages/workflow-executor/CLAUDE.md:45` was the only place describing the sentinel and the resolver, at e0a46add0.

</details>

*Resolve to agree, reply to push back.*

hercemer42:
Claude Opus 5 (1M context): **Should fix**: the collection guard reports "loaded no record" for a record that loaded fine, and in the case it was written for it names no step at all.

<details><summary>Why</summary>

`resolveSourceRecordRef` raises `SourceRecordMissingError(sourceTitle)` on a collection mismatch. That class renders *"Source step "X" loaded no record"* and *"This step uses "X" as its source, but that step didn't load any record."* Neither is true here: the source loaded a record, and it is of the wrong collection.

The worse half is the unpinned call — the scenario PRD-1263 gives as the guard's reason, a migrated workflow called from a Sub-workflow step nobody has migrated. There `callScope.selectedRecordStepId` is absent, the record is `baseRecordRef`, and `sourceTitle` is `undefined`, so the class falls back to `its source step`:

> Source step its source step loaded no record

for a run where no source step is involved at all — the record came from the run's launch. The card beside it offers *"Revise the source step"*, pointing at nothing the operator can revise, and the real cause (the calling workflow is on the AI path) is nowhere on screen.

Contract item 4 asks the step to *report no source record* rather than act on the caller's record, so reusing the error class is right; only the copy is wrong. A message naming the called workflow's collection and the record's own — and, on the unpinned path, saying the call sent none — would say what happened. Worth splitting from `SourceRecordMissingError` so the two cases cannot drift back together.

Everything else in the contract checks out: `callScope` reconstructed from the raw `workflowHistory`, the outward resolution of a `WORKFLOW_START_STEP_ID` pin, `pinnedAtStepIndex` bounding the lookup to the frame that wrote the pin, both fields optional, and `baseRecordRef` / `getAvailableRecordRefs` / the activity-log target untouched.
</details>

hercemer42:
Claude Opus 5 (1M context): **Should fix** (re-checked at `65ad33c3`, still open and now sharper): the collection guard throws with no error kind, so the very push that added `empty-source` sends the operator the admin-contact banner instead.

<details><summary>Why</summary>

`SourceRecordMissingError` extends `WorkflowExecutorError` directly, and only the two intermediate families declare a `defaultErrorKind`, so a bare construction carries `errorKind: undefined`. The guard is exactly that:

```ts
throw new SourceRecordMissingError(sourceTitle);
```

The class's own comment says *"The kind comes from the throw site: only there is it known whether a candidate was on offer"* — and this throw site passes none. What the operator gets, on the front's card:

- `shouldPromptAdminContact` is true for every kind that is not `operator`, so an unclassified error renders *"Something didn't go as planned, which may be due to permission issues or a system being unresponsive; please contact an Admin to troubleshoot this."*
- `recoveryGuidance` returns `null`, so no guidance line at all.
- The message underneath is still *"Source step "X" loaded no record"* — untrue here, the source loaded one of the wrong collection — and on the unpinned call, where `sourceTitle` is `undefined`, it reads *"Source step its source step loaded no record"* beside a "Revise the source step" button pointing at nothing.

That last case is the one the guard exists for: a migrated workflow called from a Sub-workflow step nobody has pinned. The operator is told to call an admin about a permissions problem that does not exist, for a run whose real cause is one level up in the calling workflow.

Two things worth doing while the enum is being widened anyway:

1. Classify the guard's throw. It is neither `operator` (nobody was offered anything) nor `empty-source` (a record was found) — a fifth kind, or `configuration`, which is at least honest about it being the workflow's shape rather than a data accident.
2. Give it its own message. Naming the called workflow's collection and the record's own says what happened in one line; on the unpinned path, that the call sent no record at all.

Everything else re-verified clean at this head: `callScope` from the raw history, the outward `WORKFLOW_START_STEP_ID` resolution, `pinnedAtStepIndex` bounding the lookup, both wire fields optional, `baseRecordRef` / `getAvailableRecordRefs` / the activity-log target untouched, and `empty-source` correctly returned on both empty branches of `classifyMissingSourceRecord`.
</details>

hercemer42:
Claude Opus 5 (1M context)

**Fixed in `b5dc0d88`, on two of the three points. The third does not reproduce.**

<details><summary>What changed</summary>

The guard no longer reuses `SourceRecordMissingError`. A mismatch raises
`SourceRecordCollectionMismatchError`, which extends `WorkflowConfigurationError` and therefore
carries `configuration` from the family rather than arriving unclassified — the class-level rule
that the throw site declares a kind is now honoured.

The operator-facing message names both collections and points at where the record is actually set:

> This workflow runs on invoices, but the step that called it sent a record from orders. Set the
> record on the Sub-workflow step in the calling workflow.

That disposes of both message defects: "loaded no record" was untrue when a record was found, and
the unpinned path no longer degrades to "This step uses its source step as its source".

Not a fifth kind. `configuration` already means the workflow's shape is wrong, which is exactly
this, and a kind nothing consumes is the thing being removed from this branch rather than added to
it. `empty-source` would be wrong for the same reason the comment gives: a record was found.

`yarn workspace @forestadmin/workflow-executor test` — 1838 passed, 0 failed. `build` — `tsc` clean.
`lint` — 0 errors, 9 pre-existing warnings in untouched files.
</details>

<details><summary>Where it overstates, and one thing worth being clear about</summary>

**The "Revise the source step" button does not render here.** `ReviseSourceStep.isDisplayed` is
`getRewindTarget(...) !== null`, and `getRewindTarget` returns null for every kind that is not
`operator` — so an unclassified or `configuration` error hides the component entirely. There is no
button pointing at nothing. The broken sentence was real; the button beside it was not.

**Classifying the throw changes nothing an operator sees today.** `shouldPromptAdminContact` is
`errorKind !== 'operator'`, so `configuration` and `undefined` render the identical admin-contact
banner. The value of the change is the wire being honest and PRD-1264 having something to branch
on — not the banner going away. Only `operator` removes it, and `operator` is the one kind this
must not be.

**The banner copy stays wrong for this case regardless**, and that is PRD-1264's to fix: run card
copy is out of scope on this ticket. The message underneath it is now accurate, which is the part
this layer owns.
</details>

hercemer42:
Claude Opus 5 (1M context): **Should fix**: the package's own CLAUDE.md names the error this PR replaced, in the paragraph this PR rewrote.

<details><summary>Why</summary>

The Pre-recorded args entry now ends with *"a resolved record whose collection is not the called workflow's raises `SourceRecordMissingError` instead of being used"*. It raises `SourceRecordCollectionMismatchError`, added in this PR precisely because the other class says "loaded no record" about a record that loaded fine.

It matters more here than a stale comment usually would: that file is the brief the next change to this package reads before touching it, and the two errors now differ in what the operator sees — one carries the contact-an-Admin banner, the other renders its own remedy. Someone reading the doc would reproduce the behaviour this PR fixed.

One sentence, same paragraph. PRD-1263's contract item 4 has been corrected to match.
</details>

packages/workflow-executor/src/executors/record-step-executor.ts
```diff
@@ -76,17 +78,50 @@ export default abstract class RecordStepExecutor<
...
+  private async resolveStepRecordRef(
+    stepId: string,
+    beforeStepIndex?: number,
+  ): Promise<{ record: RecordRef; sourceTitle?: string }> {
     const matches = this.context.previousSteps.filter(
       step =>
         step.stepDefinition.type === StepType.LoadRelatedRecord &&
-        step.stepOutcome.stepId === stepId,
+        step.stepOutcome.stepId === stepId &&
+        (beforeStepIndex === undefined || step.stepOutcome.stepIndex < beforeStepIndex),
```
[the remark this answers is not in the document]
hercemer42: Claude Opus 5 (1M context): The pin now resolves only among the steps of the frame that wrote it, carried as `callScope.pinnedFrameStepIndexes`.
ae8f5b7
<details><summary>Why</summary>

A frame's steps are not a span, so no single bound separates them: a nested call's body interleaves with the frame around it.
The mapper walks the call stack already, so it is the one place that knows which frame each step ran in.
Added a mapper test for a sibling call that repeats the caller's step id and closes before the pin is written.
</details>

packages/workflow-executor/src/executors/record-step-executor.ts
```diff
@@ -76,17 +78,54 @@ export default abstract class RecordStepExecutor<
...
+
+    // A record of another collection than the called workflow is not what its steps were built
+    // against, so the step refuses it rather than acting on the caller's. It names both collections:
+    // "loaded no record" would be untrue here, and on an unpinned call there is no source step to
+    // name at all.
+    if (
+      callScope.calledWorkflowCollectionName !== undefined &&
+      record.collectionName !== callScope.calledWorkflowCollectionName
+    ) {
+      throw new SourceRecordCollectionMismatchError(
```
hercemer42, line 118: Claude Opus 5 (1M context)

**DECISION** · A collection mismatch raises its own error naming both collections

**Spec left open:** The ticket says to raise `SourceRecordMissingError` on a mismatch, pinned record and `baseRecordRef` fallback alike, without saying what the message names.

**Decided:** A mismatch raises `SourceRecordCollectionMismatchError` instead. It names the record’s collection, the called workflow’s, and the Sub-workflow step where the record is set.

**Impact:** An operator on a cross-collection call reads which two collections disagree and which step to change.

**Alternative:** Keep the shared no source record error, as the ticket names it.

**Trade-off:** One more error class, against a message saying a step loaded nothing when it loaded a record.

<details><summary>For a run answering a reply</summary>

`SourceRecordMissingError` renders "Source step X loaded no record", which is untrue when the source loaded a record of another collection, at ae8f5b7.

On an unpinned call there is no source step to name, so that message degraded to a broken sentence, at ae8f5b7.

Three tests in `read-record-step-executor.test.ts` assert the new message, for the pinned record, the fallback and the case difference, at ae8f5b7.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/types/validated/execution.ts
```diff
@@ -39,6 +39,21 @@ export enum TriggerType {
...
+// The Sub-workflow call the pending step runs inside: the record the calling step pinned and the
+// collection the called workflow is built on. Both optional — a call sending neither behaves as
+// before. Grouped rather than loose, as both describe the same call. `pinnedFrameStepIndexes` holds
+// the steps of the frame that wrote the pin, which is where the pinned step is looked for, and so
+// travels with the pin and never without it. A frame's steps are not a range: a nested call's body
+// sits between them, and a closed sibling call repeating a step id is what an upper bound admits.
+export const CallScopeSchema = z
+  .object({
+    selectedRecordStepId: z.string().min(1).optional(),
+    pinnedFrameStepIndexes: z.array(z.number().int().nonnegative()).optional(),
```
hercemer42, line 51: Claude Opus 5 (1M context)

**DECISION** · The call scope carries the steps of the frame that wrote the pin, a third field beyond the two the ticket names

**Spec left open:** The ticket’s scope bullet carries `selectedRecordStepId` and `calledWorkflowCollectionName` through the schema, and says nothing about which steps the pin is matched against.

**Decided:** `CallScope` also carries `pinnedFrameStepIndexes`, and the pin resolves only among those steps. A step id is unique inside its own diagram, while `previousSteps` flattens every frame.

**Impact:** A workflow calling a copy of itself, or calling one that already closed, resolves the record its caller pinned.

**Alternative:** Bound the lookup by the index of the call that wrote the pin.

**Trade-off:** A bound is one number rather than a list, but a frame’s steps are not a span and a nested call’s body sits between them.

<details><summary>For a run answering a reply</summary>

A sibling call that repeats the caller’s step id and closes before the pin is written sits below any such bound, at ae8f5b7.

The mapper walks the call stack it already builds, so it is the only layer that knows which frame each step ran in, at ae8f5b7.

`pinnedFrameStepIndexes` is built in the mapper and never leaves the runtime, so nothing on the wire changes, at ae8f5b7.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/executors/record-step-executor.ts
```diff
@@ -27,10 +29,10 @@ function classifyMissingSourceRecord(execution?: StepExecutionData): ErrorKind |
   if (executionResult !== undefined && !('skipped' in executionResult)) return undefined;
 
   // Nothing was ever offered: only Full AI continues without pausing, so there was no choice to make.
-  if (!pendingData) return executionResult !== undefined ? 'configuration' : undefined;
+  if (!pendingData) return executionResult !== undefined ? 'empty-source' : undefined;
 
   // Whether it paused or recorded a decline, the candidate list says whether there was a choice.
-  return pendingData.availableRecordIds.length > 0 ? 'operator' : 'configuration';
+  return pendingData.availableRecordIds.length > 0 ? 'operator' : 'empty-source';
```
hercemer42, line 36: Claude Opus 5 (1M context)

**DECISION** · A source step that ran and loaded nothing gets its own error kind

**Spec left open:** The ticket’s implementation notes say error classification stays as it is, on the basis that no consumer reads it.

**Decided:** `ErrorKindSchema` gains `empty-source`, and `classifyMissingSourceRecord` returns it where a relation came back empty. The premise the note rested on does not hold.

**Impact:** Once the run card reads the kind, an empty source names the step to go back to.

**Alternative:** Leave the empty relation classified `configuration`, as it was.

**Trade-off:** A fourth value on a cross-service enum, against a permissions message for a step that simply found nothing.

<details><summary>For a run answering a reply</summary>

`getRewindTarget` and `shouldPromptAdminContact` both branch on `context.errorKind`, on forestadmin@594197017e.

The rewind on the classification is an accepted decision, `docs/adr/2026-08-18-the-source-step-rewind-is-offered-on-the-error-classification.md`, on forestadmin@594197017e.

It ships dark: a card that does not know the kind renders what it renders today, per that same record.

</details>

*Resolve to agree, reply to push back.*

packages/workflow-executor/src/executors/record-step-executor.ts
hercemer42: Is there a risk of regression ?

packages/workflow-executor/src/executors/record-step-executor.ts
hercemer42: Should be ok ?  If it's the same collection, what could mutate the case ?

packages/workflow-executor/src/types/validated/execution.ts
hercemer42: Any regression risk with this ?

packages/workflow-executor/src/types/validated/execution.ts
hercemer42: Is a reference rather than a copy not better ?  What if it mutates ?  Are race conditions possible ?

packages/workflow-executor/src/executors/record-step-executor.ts
hercemer42: Claude Opus 5 (1M context)

**No, and the reason is that the new path is unreachable for every run today.**

`resolveSourceRecordRef` returns `baseRecordRef` immediately when `context.callScope` is undefined, and the mapper only builds one when the pending step's history entry carries `childrenWorkflowId` and a `start-sub-workflow` is open above it. A run not inside a Sub-workflow call takes exactly the line it took before.

Inside a call it still matches: with no pin the resolution falls back to `baseRecordRef`, and the guard needs `calledWorkflowCollectionName`, which nothing sends until PRD-1262 merges. So all four step types keep today's behaviour on today's orchestrator, whether or not the workflow uses Sub-workflow steps.

The one change that is live is the error classification — a source step that loaded nothing now reports `empty-source` rather than `configuration`. The run card branches on `errorKind !== 'operator'`, so both render the same thing until the front reads the new kind. That one has its own decision on `classifyMissingSourceRecord`.

Package suite: 1838 passing. CI green on the head.

packages/workflow-executor/src/executors/record-step-executor.ts
hercemer42: Claude Opus 5 (1M context)

**Folding the case would be the riskier choice: Forest lets two collections differ only in case.**

The docs record it as a known limitation on activity tracking — `myCollection` and `MyCollection` can both exist. Compare case-insensitively and a record from `Orders` satisfies a guard written for `orders`, which is the exact failure this guard exists to stop.

Nothing else in this package folds a collection name either. `schemaResolver.resolve(collectionName)` and every agent call take it verbatim, so an exact comparison is the same string handling the rest of the runtime already relies on.

Your instinct points at the inverse risk, which is the real one: the same collection reaching the two sides with different casing, so the guard fires on a valid run. Both sides read the technical name rather than a display name, so it should not happen — but that is PRD-1262's to hold up, since it writes `calledWorkflowCollectionName`. Worth stating there explicitly rather than assuming it.

packages/workflow-executor/src/types/validated/execution.ts
hercemer42: Claude Opus 5 (1M context)

**No wire surface, so no compatibility risk — but there is one failure mode worth knowing.**

`CallScope` never crosses a service boundary. This package's own mapper builds it and this package's own schema parses it, both in the same process. Nothing external sends it, so no orchestrator or agent version can disagree about its shape. On a run with no call scope the field is absent and `AvailableStepExecutionSchema` parses exactly as before.

The failure mode is what `.strict()` does when the mapper is wrong. An unexpected or malformed key raises `DomainValidationError`, and `forest-server-workflow-port.ts:81` files the **whole run** malformed — so every later poll returns it and fails again, rather than one step erroring. That is why the object is closed and why an empty value on the wire reads as absent instead of being carried through: a lost pin should fail a step, not strand the run.

packages/workflow-executor/src/types/validated/execution.ts
hercemer42: Claude Opus 5 (1M context)

**There is nothing to hold a reference to, nothing that can mutate it, and no concurrency around it.**

`toAvailableStepExecution` is a pure function of one HTTP payload and runs on every dispatch. A reference would have to point into that payload, which is discarded as soon as the call returns — so it would be a reference to something shorter-lived than the copy, not longer. The copy is a handful of integers.

Mutation: the array is built in `toCallScope`, then `AvailableStepExecutionSchema.parse` returns a fresh object, and nothing writes to it afterwards. Numbers are immutable and the array is not shared with any other step, run or module.

Races: one step execution at a time in a single Node process, the context built per dispatch, and the run carries a server-side lock. The one genuine concurrency question is a run re-dispatched while a step is in flight — and that rebuilds the call scope from the newly fetched history, so a stale array cannot outlive the payload it came from.

The thing the copy does buy is that the resolution is decided in the one layer that knows the frame structure. The executor sees a set of indices and needs no view of the history at all.

packages/workflow-executor/src/executors/record-step-executor.ts
hercemer42: I'll comment on 1262

hercemer42:
Claude Opus 5.5 (1M context): The Pre-recorded args entry in `packages/workflow-executor/CLAUDE.md` now names `SourceRecordCollectionMismatchError` and says why the other error would be wrong.
a83ae90
<details><summary>Why</summary>

The later mention stays, since it describes `SourceRecordMissingError` as an error whose kind the throw site sets, which still holds.
</details>

hercemer42 (commented):
**Spec (`PRD-1263`)**: conforms to contract items 1 to 5, while the ticket's Scope, Tests and classification note still predate contract item 4.

packages/workflow-executor/CLAUDE.md
```diff
@@ -42,7 +42,7 @@ Front  ◀──▶  Orchestrator  ◀──pull/push──▶  Executor (this p
...
...and `resolveSourceRecordRef` reads it — a call pinning nothing still falls back to `baseRecordRef`, a pin of `WORKFLOW_START_STEP_ID` resolves one frame out, and a resolved record whose collection is not the called workflow's raises `SourceRecordCollectionMismatchError` instead of being used, naming both collections and the Sub-workflow step to set the record on — `SourceRecordMissingError` would say the source loaded nothing, which is untrue when a record was found. The call's own pin is the one place an empty value reads as **absent** rather than as a lost target: it arrives on the wire from an orchestrator whose vocabulary predates it, and `CallScopeSchema` rejects the empty string, so keeping it would report the whole run malformed on every poll rather than failing the one step. The pin is resolved against what ran before the call that wrote it (`callScope.pinnedAtStepIndex`) — `previousSteps` flattens every frame and a step id is unique only inside its own workflow, so a called workflow repeating one (a copy of its caller, or a call on itself) would otherwise shadow the step the caller pinned. `baseRecordRef`, `getAvailableRecordRefs` and the activity-log target are untouched.
```
hercemer42: Claude Opus 5.5 (claude-opus-5-5): **Should fix**: the package guide tells the next change to bound the call pin by one index, which reintroduces the sibling-call bug this PR fixed.

`CLAUDE.md:45` names `callScope.pinnedAtStepIndex`, the upper bound `ae8f5b7` replaced with `pinnedFrameStepIndexes` because a bound admits a closed sibling call.

A way out: name `pinnedFrameStepIndexes` as the pinning frame's own steps, and say a `workflow-start` pin keeps walking outward.
</details>

1. Open `packages/workflow-executor/CLAUDE.md` to change how a call pin resolves.
2. Search the code for `pinnedAtStepIndex`, as line 45 names it.
3. Find nothing, then restore the index bound the paragraph describes.
4. A closed sibling call repeating the pinned id now answers with its own record.
</details>

packages/workflow-executor/src/executors/record-step-executor.ts
```diff
@@ -76,17 +78,54 @@ export default abstract class RecordStepExecutor<
...
-      return this.context.baseRecordRef;
+    if (stepId !== WORKFLOW_START_STEP_ID) return (await this.resolveStepRecordRef(stepId)).record;
+
+    const { callScope, baseRecordRef } = this.context;
+    if (!callScope) return baseRecordRef;
+
+    // Inside a Sub-workflow call, "workflow start" means the record the calling step pinned; a
+    // call pinning none still starts from the record the run was launched on.
+    const { record } = callScope.selectedRecordStepId
+      ? await this.resolveStepRecordRef(
```
hercemer42: **Applies to:** `packages/workflow-executor/src/executors/record-step-executor.ts:155` (not in this diff)
Claude Opus 5.5 (claude-opus-5-5): **Should fix**: an operator sees a called workflow's step blamed for invalid parameters when the fault is the calling Sub-workflow step's pin.

This call hands the pin to line 155, whose "The pre-configured step parameters are invalid" loses its admin banner on forestadmin#9972.

A way out: throw a dedicated `WorkflowConfigurationError` on the call-pin path, worded like `SourceRecordCollectionMismatchError`.

Also moves: `read-record-step-executor.test.ts:1826`, which asserts the generic message.
</details>

1. In workflow A, pin a Sub-workflow step to a Load Related Record step on a Decision branch.
2. Run A down the other branch, so that step never runs.
3. Inside the called workflow, a step pinned to workflow start runs.
4. Its card reads "The pre-configured step parameters are invalid", and nothing points at the Sub-workflow step in A.
</details>

packages/workflow-executor/CLAUDE.md
hercemer42: Claude Opus 5.5 (1M context): `CLAUDE.md:45` now names `pinnedFrameStepIndexes` as the pinning frame's own steps and says a `workflow-start` pin keeps walking outward.
8a0f489
<details><summary>Why</summary>

It also lists all three `callScope` fields and says why a set replaces the bound, since a nested call's body sits between a frame's steps.
</details>

packages/workflow-executor/src/executors/record-step-executor.ts
hercemer42: Claude Opus 5.5 (1M context): A call pin that matches no step now raises `SourceRecordStepNotReachedError`, which points at the Sub-workflow step that set it.
8a0f489
<details><summary>Why</summary>

`resolveStepRecordRef` returns undefined on a miss so each caller names its own, and a step's own pin keeps `InvalidPreRecordedArgsError`.
The test at `read-record-step-executor.test.ts:1826` now asserts the new message and the `configuration` kind.
</details>

hercemer42:
Claude Opus 5.5 (1M context): **Should fix** (the contract text first, the code maybe not at all): contract item 4 promises the front renders the collection-mismatch error without the contact-an-Admin banner, and the front has decided it won't.

<details><summary>Why</summary>

ForestAdmin/forestadmin#9972 now keeps the banner on every `configuration` error (ADR 2026-09-22, "same kinds"), because most of that family names a fix only an Admin can make. It names `SourceRecordCollectionMismatchError` as the accepted exception, which shows the banner above its own remedy "until the executor sends a kind of its own for a remedy-bearing error".

That leaves two ways to settle it, and only one involves this PR:

- **A new kind here.** Widen `ErrorKindSchema` again and give the mismatch its own `defaultErrorKind`. The front then treats it as self-explanatory. Small, but it's a cross-service contract change, with the front and the enum moving together.
- **A copy change on the front** (suggested there). The banner is right that an Admin has to act: the operator can't edit the calling workflow. It's only wrong about the cause. Worded as a configuration problem, it stops contradicting the mismatch message and needs nothing from this repository.

Either way, PRD-1263 item 4 has to stop asserting how the card renders, because rendering is the front's to decide. What this layer owns is the class, its `configuration` kind and its message.
</details>

hercemer42:
Claude Opus 5.5 (1M context): No code change here, and PRD-1263 contract item 4 now leaves how the run card frames a kind to the front.
<details><summary>Why</summary>

This layer owns the class, its `configuration` kind and its message, and none of them makes a claim about the banner.
Item 4 already read "under the banner" when this was posted, and the `empty-source` implementation note made the same kind of claim.
Both now stop at the kind, so the front can change its ADR again without the contract moving.
</details>

hercemer42:
Claude Opus 5.5 (claude-opus-5-5): **Request: give the two call-site errors a kind of their own, so the run card can drop "contact an Admin" above them.**

**What the operator sees today.** On a step inside a called workflow, both errors render under "Something didn't go as planned, which may be due to permission issues or a system being unresponsive; please contact an Admin", directly above a message that already names the fix:

- `SourceRecordCollectionMismatchError`: "This workflow runs on rental, but the step that called it sent a record from customer. Set the record on the Sub-workflow step in the calling workflow."
- `SourceRecordStepNotReachedError`: "The Sub-workflow step that called this workflow takes its record from a step that did not run before it. Change the record on that Sub-workflow step."

That's two accounts of one failure, and the banner sends the reader to the wrong person.

**Why the front can't fix it alone.** Both arrive as `configuration`, the same kind as `NoReadableFieldsError`, `NoActionsError`, `InvalidPreRecordedArgsError` and the rest, most of which name a fix only an Admin can make. The front keeps the banner for every `configuration` error for that reason (ADR 2026-09-22-the-contact-an-admin-banner-and-the-rewind-are-offered-on-the-same-kin in forestadmin). The only other signal is the message text, which the ADR rules out: a copy change here would silently break it.

**What's needed here.**
1. A new `ErrorKindSchema` value, proposed `call-site`: the fault is in the calling workflow's Sub-workflow step, and the message names that step.
2. `SourceRecordCollectionMismatchError` and `SourceRecordStepNotReachedError` declare it as their `defaultErrorKind`. Every other `WorkflowConfigurationError` keeps `configuration`.
3. Record it in PRD-1263's contract item 7 beside `empty-source`.

No floor is needed: an executor that doesn't send it renders today's copy, banner included. The orchestrator forwards `context` opaquely, so forestadmin-server needs no change.

**The front's half** (ForestAdmin/forestadmin#9972) is one line. `call-site` joins the kinds that drop the banner, and it's not rewindable, so the card offers Skip only. That lands once the kind exists here.

hercemer42:
Claude Opus 5.5 (1M context): Both call-site errors now join a `WorkflowCallSiteError` family carrying a new `call-site` kind, recorded in PRD-1263 contract items 4 and 7.
4f3dad3
<details><summary>Why</summary>

It follows the file's one-abstract-per-kind rule, and every other `WorkflowConfigurationError` keeps `configuration`.
No code reads either family with `instanceof`, so the move changes the kind and nothing else.
`call-site` is not rewindable and needs no floor, since a card that does not know it renders today's copy.
</details>

hercemer42:
Claude Opus 5.5 (claude-opus-5-5): **Pushed `785ccac`: the called-workflow collection check now applies to pinned calls only.** Pull before continuing.

**What changes for users.** A Sub-workflow step saved before the pin existed, calling a workflow of another collection, used to start failing on release with "This workflow runs on X, but the step that called it sent a record from Y". It now keeps running exactly as it does today, on the run's record, until the Editor sets "On record". That is PRD-769's migration promise ("until 'On record' is set, the step keeps behaving exactly like today"), confirmed by Brian over the stricter version.

**What changed in code.**
- `resolveSourceRecordRef` returns `baseRecordRef` unchecked when the call scope carries no `selectedRecordStepId`. The collection check and `SourceRecordStepNotReachedError` apply only on the pinned path.
- `read-record-step-executor.test.ts`: "names both collections when the run record is of another collection" becomes "keeps the run record on an unpinned call whose called workflow is on another collection".
- `step-executor-factory.test.ts`: the call-scope seam test now pins its call, since an unpinned one no longer errors.
- `CLAUDE.md`: the fallback is described as unchecked.

**Still to do on your side:** PRD-1263's Scope bullet "Collection guard", contract item 4 and the Tests bullet all say "the pinned record and the `baseRecordRef` fallback alike". They should now say pinned records only. I've left that as a comment on the ticket rather than editing it, since you're live in it.

**Front:** ForestAdmin/forestadmin#9972 now consumes `call-site` from `4f3dad3` (`c6e1319`): no Admin banner, Skip and no rewind.

hercemer42:
**Claude Opus 5.5:** pushed `2994f8299` onto this branch. It fixes a hole my `785ccac` opened: a top-level call pinned to "workflow start" whose called workflow is on another collection ran silently on the run's record.

**What an operator saw:** a Customer workflow whose Sub-workflow step has "On record" = "Customer from workflow start" and calls a Rental workflow. The called workflow's Get Data ran on the customer (#396) instead of stopping with `SourceRecordCollectionMismatchError`. The editor blocks publishing that state, so reaching it takes an import or a drifted published workflow.

**Why:**
- `run-to-available-step-mapper` skips calls pinning `workflow-start` when it looks for the pin, so a top-level one sends no `selectedRecordStepId`.
- Since `785ccac`, `resolveSourceRecordRef` treated a missing `selectedRecordStepId` as unpinned and returned `baseRecordRef` unchecked.

**Change:**
- `CallScope` gains `isPinned: true` (optional) when the innermost call carries any pin, `workflow-start` included. The mapper sets it.
- `resolveSourceRecordRef` returns `baseRecordRef` unchecked only when the call is unpinned. A pinned call is checked against the called workflow's collection whatever its pin resolves to, including the run's record.
- Tests:
  - mapper: `isPinned` on pinned scopes, and absent on an unpinned call;
  - executor: a top-level `workflow-start` pin with a cross-collection callee is refused, and with a same-collection callee it runs on the run's record.
- The package `CLAUDE.md` sentence on call scopes now covers the new case.

**Verified:** 144/144 across the mapper, read-record and factory suites, and eslint is clean on the touched files. Live on the local stack: the same run now stops with "This workflow runs on rental, but the step that called it sent a record from customer…". Unpinned calls still run as today, as covered by the existing test.

Nothing else on the branch changed. If PRD-1263's contract lists the call-scope fields, it needs `isPinned` added.

hercemer42:
Claude Opus 5.5 (1M context): No code change here, and PRD-1263 now names `isPinned` and says a top-level `workflow-start` pin is checked against the called collection.
<details><summary>Why</summary>

The contract describes the wire, so the call-scope fields went into the Scope bullet, which was also missing `pinnedFrameStepIndexes`.
Contract item 4 and the guard bullet said "a pinned record", which did not cover a pin resolving to the run's own record.
The Tests bullet now lists the new cross-collection `workflow-start` case beside the pinned-record one.
</details>

PR #1914, opened 2026-09-22
feat(workflow-executor): persist the AI reasoning of load-related-record, update-record and MCP steps

PMerlet:
The three executors already ask the model to justify its answer and drop it. This keeps it, and records how a human ruled on it.

- Each step now carries an `aiSuggestion` of its own, written whenever an AI produced the proposal: the value reasoning for `update-record`, the compared fields plus the field and record reasoning for `load-related-record`, and `toolSelectionReasoning` for `mcp`. It sits beside `pendingData` / `executionParams` / `executionResult`, none of which it touches, because the step summary replays those into every later step's prompt.
- `aiSuggestionRuling` records how a human ruled: `kept` / `value-changed` on an update, `kept` / `record-changed` / `relation-changed` on a related record. Absent means no AI proposal was put to a human, so absent never reads as agreement.
- A relation preview rewrites `pendingData.suggestedField` to whatever the human asked to see, so the ruling reads the AI's own relation from `aiSuggestion.field` instead.
- `mcp`: a `reasoning` property is added to each candidate tool's schema before binding, under a reserved key when the tool declares its own, stripped from the args before the tool runs. A tool whose schema cannot be augmented is offered untouched rather than failing the step.
- Nothing is persisted where the AI did not decide: Manual, degraded-to-Manual, xToOne relations, pre-recorded values. Each drop site logs at `Info`.

Front counterpart: ForestAdmin/forestadmin#9974. Supersedes #1674.

No breaking change: every added field is optional, `StepOutcome` is untouched, and runs persisted by 1.29.0 are read without them.

> [!NOTE]
> ### Persist AI reasoning for `update-record`, `load-related-record`, and MCP steps
> - Adds optional `suggestion` and `ruling` fields to execution data types in [step-execution-data.ts](https://github.com/ForestAdmin/agent-nodejs/pull/1914/files#diff-7c687de53bb6a7544d145eaf135f79eb6665ca9f14337016748699238ce1f40f) for `update-record`, `load-related-record`, and MCP steps.
> - `McpStepExecutor.selectTool` dynamically injects a collision-safe `reasoning` field into Zod or JSON schemas, extracts it from the AI response, and excludes it from the actual tool invocation arguments.
> - `LoadRelatedRecordStepExecutor` and `UpdateRecordStepExecutor` capture AI reasoning during selection, persist it before executing or awaiting confirmation, and attach a `kept`, `record-changed`, `relation-changed`, or `value-changed` ruling when a user confirms or overrides the AI's choice.
> - `StepSummaryBuilder` excludes AI justification text from generated context summaries.
> - Risk: MCP schema augmentation uses duck-typing in `McpStepExecutor.withReasoningField`; tools with unsupported or missing schemas fall back to normal execution without capturing reasoning.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 939af6d.</sup>
> 
>


packages/workflow-executor/src/executors/load-related-record-step-executor.ts
```diff
@@ -539,7 +577,18 @@ export default class LoadRelatedRecordStepExecutor extends RecordStepExecutor<Lo
       stepIndex: this.context.stepIndex,
     };
 
-    return this.persistAndReturn(record, { selectedRecordRef, name, displayName }, execution);
+    const { suggestedFields, fieldsReasoning, reasoning } = pendingData;
+    const userKeptAiSuggestion =
+      suggestedFields !== undefined && selectedRecordId === pendingData.suggestedRecord?.recordId;
```
PMerlet: Claude Opus 5 (claude-opus-5): **Must fix**: An operator who confirms the AI's suggested record gets a completed step carrying no AI justification.

`load-related-record-step-executor.ts:582` compares two arrays with `===`, so it holds only when the front omits the id it always sends.

A way out: compare `serializeRecordId` on both sides, which also unpins the identity-based case at `load-related-record-step-executor.test.ts:4955`.
</details>

1. Run an AI-assisted Load Related Record step on a to-many relation.
2. Accept the record the AI suggests, then confirm.
3. Open the completed step.
4. No AI justification is shown.
</details>

packages/workflow-executor/src/executors/update-record-step-executor.ts
```diff
@@ -169,15 +180,23 @@ export default class UpdateRecordStepExecutor extends RecordStepExecutor<UpdateR
...
         // distinguish "no override" (undefined) from "override to null".
-        const rawValue =
-          userConfirmation?.value !== undefined ? userConfirmation.value : pendingData!.value;
+        const overrideValue = userConfirmation?.value;
+        const rawValue = overrideValue !== undefined ? overrideValue : pendingData!.value;
+
+        // The value comes from an `unknown` HTTP value (may be a boolean or array), so coerce
+        // it to the field's native type before updating. Idempotent on already-typed values.
+        const value = await this.coerceOverride(selectedRecordRef, pendingData, rawValue);
+        const aiValue =
```
PMerlet: Claude Opus 5 (claude-opus-5): **Should fix**: A user correcting a bad prefilled value is told their valid input is invalid, and nothing is written.

`update-record-step-executor.ts:189` re-coerces the value the user replaced, and `coerceFieldValue` throws on a pre-recorded value that no longer validates.

A way out: skip the comparison when `pendingData.reasoning` is undefined, and read a coercion failure as not equal.
</details>

1. Build an Update Record step whose pre-recorded value the field schema now rejects.
2. Run it in AI-assisted mode and type a valid value in the panel.
3. Confirm the step.
4. It errors, and the record keeps its old value.
</details>

packages/workflow-executor/src/executors/mcp-step-executor.ts
```diff
@@ -330,10 +345,51 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
...
+
+  // A tool's schema is duck-typed rather than matched with `instanceof`: ai-proxy and
+  // workflow-executor resolve different zod instances, and an MCP-server tool carries a plain
+  // JSON Schema, which is valid at runtime but does not unify with langchain's static union.
+  private static withReasoningField(tool: StructuredToolInterface): DynamicStructuredTool {
+    const { schema } = tool;
+    const isZodObject = typeof (schema as { extend?: unknown }).extend === 'function';
+    const augmented: z.ZodTypeAny = isZodObject
+      ? (schema as z.ZodObject<z.ZodRawShape>).extend({
+          [REASONING_FIELD]: z.string().describe(REASONING_FIELD_DESCRIPTION),
```
PMerlet: Claude Opus 5 (claude-opus-5): **Should fix**: An MCP tool declaring its own `reasoning` argument runs without it, and that value is stored as the AI justification.

`mcp-step-executor.ts:370` and `:387` overwrite a same-named property of a customer's tool, and `:353` strips the key before `tool.base.invoke`.

A way out: inject a reserved key such as `__forest_tool_selection_reasoning`, and strip only what `withReasoningField` added.
</details>

1. Connect an MCP server exposing a tool with its own `reasoning` parameter.
2. Run a workflow MCP step that selects that tool.
3. Inspect the call.
4. The tool ran without its own `reasoning`.
</details>

packages/workflow-executor/src/executors/update-record-step-executor.ts
```diff
@@ -305,7 +327,10 @@ export default class UpdateRecordStepExecutor extends RecordStepExecutor<UpdateR
       type: 'update-record',
       stepIndex: this.context.stepIndex,
       executionParams: { displayName, name, value },
-      executionResult: { updatedValues: updated.values },
+      executionResult: {
+        updatedValues: updated.values,
+        ...(reasoning !== undefined && { reasoning }),
```
PMerlet: Claude Opus 5 (claude-opus-5): **Should fix**: Every later AI step of the run now reads the previous update's self-justification as part of its context.

`update-record-step-executor.ts:332` puts it in `executionResult`, which `summary/step-summary-builder.ts:51` stringifies into the system message of every later step.

A way out: format `update-record` in `summary/step-execution-formatters.ts`, or move `reasoning` to the top level as the MCP step does.
</details>

1. Run a workflow whose Update Record step is followed by an AI step.
2. Read the context built for that later step.
3. The update's `reasoning` text sits inside its `Output:` line.
</details>

packages/workflow-executor/src/executors/mcp-step-executor.ts
```diff
@@ -185,10 +198,12 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
 
     // 1. Persist raw result immediately — safe state before any further network calls
     const baseExecutionResult = { success: true as const, toolResult };
+    const toolSelectionReasoning = reasoning ?? existingExecution?.toolSelectionReasoning;
     const baseData: McpStepExecutionData = {
       ...existingExecution,
       type: 'mcp',
       stepIndex: this.context.stepIndex,
+      ...(toolSelectionReasoning !== undefined && { toolSelectionReasoning }),
```
PMerlet, line 210: **Applies to:** `mcp-step-executor.ts:194`, not in this diff
Claude Opus 5 (claude-opus-5): **Should fix**: A Full AI step that fails mid-call leaves no record of why the AI picked that tool or value.

The write-ahead marker at `mcp-step-executor.ts:194` carries no justification, and `update-record-step-executor.ts:320` has the same shape.

A way out: write it in the `beforeCall` marker, beside `idempotencyPhase`.
</details>

1. Run a Full AI MCP step whose tool call fails.
2. Read the persisted step in the run store.
3. It carries `idempotencyPhase: 'executing'` and no `toolSelectionReasoning`.
</details>

packages/workflow-executor/test/executors/mcp-step-executor.test.ts
```diff
@@ -1472,4 +1475,182 @@ describe('McpStepExecutor — re-auth pause hardening', () => {
...
+        expect(invokeFn).toHaveBeenCalledWith({ message: 'Hello' });
+
+        const saved = (runStore.saveStepExecution as jest.Mock).mock.calls.at(-1)?.[1];
+        expect(saved).not.toHaveProperty('toolSelectionReasoning');
+        expect(saved.executionParams.input).toEqual({ message: 'Hello' });
+      });
+    });
+
+    describe('when the candidate tools are offered to the AI', () => {
+      it('should require a reasoning property on a JSON-Schema tool (MCP server)', async () => {
```
PMerlet: Claude Opus 5 (claude-opus-5): **Should fix**: The test pinning the MCP server schema path runs against a stub, so it cannot catch a dropped tool argument.

`mcp-step-executor.test.ts:1607` builds its tool with `MockRemoteTool`, whose schema at `:39` carries no `properties` of its own.

A way out: build the case on a real JSON Schema carrying its own `properties` and `required`.
</details>

1. Delete the `properties` spread inside `injectReasoningIntoJsonSchema`.
2. Run the MCP suite.
3. It stays green, although every MCP tool would lose its arguments.
</details>

packages/workflow-executor/src/executors/update-record-step-executor.ts
```diff
@@ -139,6 +139,17 @@ function coerceFieldValue(
...
 interface UpdateTarget extends FieldWithValue {
   selectedRecordRef: RecordRef;
+  reasoning?: string;
+}
+
+// A field value is a primitive, a string, or an array of those (Json is stored as a string), so
+// there is never a plain object to deep-compare.
+function fieldValuesEqual(a: unknown, b: unknown): boolean {
+  if (Array.isArray(a) && Array.isArray(b)) {
+    return a.length === b.length && a.every((item, i) => fieldValuesEqual(item, b[i]));
```
PMerlet, line 156: Claude Opus 5 (claude-opus-5): **Violates conventions**: The array branch of the new value comparison never runs with two arrays, so array fields stay unverified.

Rule `testing.md#cover-error-and-edge-paths-not-only-the-happy-path`: the array cases at `update-record-step-executor.test.ts:2334` pass a null `pendingValue`, so the branch is skipped.

A way out: add a confirmation case whose `pendingData` holds both an array `value` and a `reasoning`.
</details>

PMerlet (commented):
**Spec**: no Linear ticket is linked to this PR, so the functional check was limited to the PR description, which the diff matches.

packages/workflow-executor/src/executors/load-related-record-step-executor.ts
PMerlet: Claude Opus 5 (claude-opus-5): `sameRecordId` now compares the ids by value, and the trace is kept only when the suggested relation was followed.
da8274f
<details><summary>Why</summary>

The new case confirms with the suggested id echoed back, which the identity comparison could never satisfy.
</details>

packages/workflow-executor/src/executors/update-record-step-executor.ts
PMerlet: Claude Opus 5 (claude-opus-5): `userKeptAiValue` skips the re-coercion when no justification is at stake, and logs a failure instead of throwing.
da8274f
<details><summary>Why</summary>

A suggested value that no longer validates cannot be the value the user kept, so the comparison answers false.
</details>

packages/workflow-executor/src/executors/mcp-step-executor.ts
PMerlet: Claude Opus 5 (claude-opus-5): A tool declaring `reasoning` keeps it, and the justification is asked for under `__forest_tool_selection_reasoning` instead.
da8274f
<details><summary>Why</summary>

`withReasoningField` returns the key it injected, and `selectTool` strips that key alone.
</details>

packages/workflow-executor/src/executors/update-record-step-executor.ts
PMerlet: Claude Opus 5 (claude-opus-5): The justification moved to `executionParams`, and `formatUpdateRecord` keeps it out of every later step's context.
da8274f
<details><summary>Why</summary>

The front reads `pendingData.reasoning` then `executionParams.reasoning`, so the move also matches what it expects.
`step-summary-builder.test.ts` asserts the text is absent from the summary.
</details>

packages/workflow-executor/src/executors/mcp-step-executor.ts
PMerlet: Claude Opus 5 (claude-opus-5): Both write-ahead markers now carry the justification and the attempt, so a failed call still records why.
da8274f
<details><summary>Why</summary>

`mcp-step-executor.ts` and `update-record-step-executor.ts` write it in `beforeCall`, beside `idempotencyPhase`.
</details>

packages/workflow-executor/test/executors/mcp-step-executor.test.ts
PMerlet: Claude Opus 5 (claude-opus-5): The case builds a real JSON Schema and asserts the tool's own `message` property survives the injection.
da8274f

packages/workflow-executor/src/executors/update-record-step-executor.ts
PMerlet: Claude Opus 5 (claude-opus-5): Two confirmation cases on a `[String]` field now run the array branch, one keeping the justification and one dropping it.
da8274f

packages/workflow-executor/src/executors/mcp-step-executor.ts
```diff
@@ -330,10 +345,51 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
...
+      schema: augmented,
+      func: undefined,
+    });
+  }
+
+  private static injectReasoningIntoJsonSchema(schema: JsonSchemaObject): JsonSchemaObject {
+    return {
+      ...schema,
+      properties: {
+        ...(schema.properties ?? {}),
```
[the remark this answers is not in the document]
PMerlet: Claude Opus 5 (claude-opus-5): A tool declaring no schema is now offered untouched, so a parameterless tool runs instead of failing the step.
ee6ca16
<details><summary>Why</summary>

The throw came earlier than the suggested line, at the `.extend` probe in `withReasoningField`, so optional chaining on `properties` would not have prevented it.
Such a tool keeps no selection justification, which is the trade-off for leaving a working class of tools exactly as it was.
</details>

packages/workflow-executor/src/executors/mcp-step-executor.ts
```diff
@@ -330,10 +350,83 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
       ),
     ];
 
-    return this.invokeWithTools(
+    const reasoningKeys = new Map<string, string>();
+    const augmentedTools = tools.map(t => {
+      const { tool, reasoningKey } = McpStepExecutor.withReasoningField(t.base);
+      reasoningKeys.set(t.base.name, reasoningKey);
```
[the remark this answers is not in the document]
PMerlet: Claude Opus 5 (claude-opus-5): The key of the first tool with that name is kept now, which is the tool the selection itself resolves to.
0690ff4
<details><summary>Why</summary>

Source-qualifying the key is not reachable at strip time, since the model answers with a name alone and carries no `sourceId`.
Matching `tools.find` on first-wins keeps both sides on one rule, and a new case fails without it.
</details>

packages/workflow-executor/src/executors/mcp-step-executor.ts
```diff
@@ -330,10 +350,83 @@ export default class McpStepExecutor extends BaseStepExecutor<McpStepDefinition>
...
+      }),
+      reasoningKey,
+    };
+  }
+
+  private static injectReasoningIntoJsonSchema(
+    schema: JsonSchemaObject,
+    reasoningKey: string,
+  ): JsonSchemaObject {
+    return {
```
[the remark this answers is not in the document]
PMerlet: Claude Opus 5 (claude-opus-5): No change made, since a `$ref`-rooted schema is inlined before this code ever sees it.
<details><summary>Why</summary>

`@langchain/mcp-adapters@1.1.3` builds each tool as `simplifyJsonSchemaForLLM(dereferenceJsonSchema(tool.inputSchema))` in `dist/tools.js:452`.
That resolver inlines a root `$ref` into `#/$defs` or `#/definitions`, and Forest connector tools carry zod schemas rather than JSON Schema.
A root `$ref` that survives it points outside the document, so the model cannot read the tool at all, with or without the injection.
</details>

PR #1920, opened 2026-09-23
fix(agent-bff): keep the text of a textarea or option in action html

Tonours:
### General

### Security

## What

The action html sanitizer now keeps the **text** of a `<textarea>` or `<option>` it strips, instead of dropping the text with the tag.

`sanitize-html` discards the content of its default `nonTextTags` (`script`, `style`, `textarea`, `option`, `xmp`) along with the tag. We now pass `nonTextTags: ['script', 'style', 'xmp']`, so a disallowed `textarea`/`option` loses its tag but its text survives, entity-escaped. This applies to the execute result `html` and to form `htmlBlock` content, which share the same options.

## Why

Forest's own "Generate reset password link" Smart Action returns the link inside `<textarea class="c-beta-input" rows="10" disabled>…</textarea>`. Through the BFF it came back as:

```json
{"type":"success","message":"","invalidated":[],"html":"\n            \n          "}
```

The Zendesk app showed nothing, while every run still regenerated the token and invalidated the previous link. Seen on staging against agent-bff 1.33.0.

## Decisions

- **`textarea` is not added to `allowedTags`.** When `textarea` is allowed, sanitize-html passes its inner text through raw and unescaped (`index.js:575`). `<textarea></textarea>` would come out with live-looking markup. That output is inert in a browser, but it becomes live for any client that strips the tag and keeps the content. Keeping only the text is enough: the Zendesk app strips action html down to text anyway.
- **`xmp` stays in `nonTextTags`.** sanitize-html 2.17.5 documents that a disallowed `xmp` re-emits its raw content unescaped.

## Scope and safety

- `script`/`style` content is still dropped. Allowed tags, attributes, classes and styles are unchanged.
- Kept text is escaped: `<textarea></textarea>` → `&lt;img src=x onerror="alert(1)"&gt;`.
- Known cosmetic effect: an entity already encoded inside a textarea comes out double-encoded, because textarea is a raw-text element and is not decoded on input.
- `sanitize-html` is pinned to `2.17.5` exactly, so the defaults this overrides cannot drift.

## How to test

```bash
cd packages/agent-bff && yarn jest test/action
```

Three new tests fail without the one-line change. The full agent-bff suite passes (108 suites, 2120 tests), and tsc and eslint are clean. It has not been run against a deployed BFF yet.


nbouliol (approved):
Approved after /validator run 1920-20260923153356-nbouliol: nothing at or above `should-fix` posted, 0 suppressed.

PR #1922, opened 2026-09-23
fix(workflow-executor): sweep at most five automated inboxes at once

Scra3:
fixes PRD-1314

Stacked on #1906.

## Problem

`runPollCycle` swept every automated inbox of the listing at once. With 30 inboxes in one environment (a real case in production), one cycle fires 90 to 120 segment reads at the customer's agent at the same instant. Under that burst the reads hit agent-client's 10 s timeout and whole inboxes skip their sync.

## Fix

At most 5 inboxes are swept at once (`MAX_CONCURRENT_INBOX_POLLS`, hardcoded). Five workers pull from a shared queue, so a slot goes to the next inbox as soon as one finishes. Every inbox is still swept in the same cycle.

Unchanged: per-inbox error isolation (`pollInbox` catches everything, so a failing inbox just frees its slot) and the drain. Once `stop()` starts, queued inboxes are not started, and the ones in flight are drained as before.

## Tests

- 12 inboxes: never more than 5 read at once, all 12 synced in the cycle.
- A failing inbox hands its slot to the next one right away.
- `stop()` mid-cycle: the 5 in flight finish and sync, the 7 queued are never read.

The first and last fail without the bound, the second fails against fixed batches of 5.

> [!NOTE]
> ### Limit `AutomationPoller` sweep to five concurrent inbox polls
> - Replaces the eager per-inbox poll launch in `AutomationPoller.polling-cycle` with a shared queue and at most five worker loops, so queued inboxes start only as active polls settle.
> - Each worker stops taking new work once the poller leaves the running state, but the cycle still waits for all in-flight workers before scheduling the next cycle.
> - Adds tests for the five-inbox concurrency cap, slot reuse after a failed inbox read, and shutdown draining active polls without starting queued work.
> - Behavioral Change: sweeps that previously polled all automated inboxes at once now poll no more than five concurrently, which reduces peak load but may lengthen total sweep time for large inbox counts.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ad212bf.</sup>
> 
>


matthv:
Non-blocking: bounding the sweep to 5 concurrent inboxes stretches cycle duration from `max(per inbox)` to roughly `ceil(N/5) × per inbox` — a 6× multiplier at the 30-inbox environment this PR is written for. On the forestadmin-server side, the automation poller-lease is only renewed at the start of each cycle (via `listAutomatedInboxes`), and the next cycle only gets scheduled once the current one fully resolves — so the real renewal interval is now `poll interval + cycle duration`. With `LEASE_TTL_SECONDS = 600` and a 300s default poll interval, a degraded cycle (slow agent, large backlog — exactly the conditions this PR exists to mitigate) can push past the lease TTL, letting a second instance win the election and both sweep concurrently. That's tolerated (duplicate polls are absorbed by the assignment/run unique indexes), not a correctness bug, but the lease comment's stated invariant ("a live holder always renews before expiry") is no longer guaranteed by construction, and nothing on the server side documents this cross-repo coupling. Might be worth a one-line note in `automation-poller.ts`'s comment, or flagging to whoever owns the lease.

matthv (approved):

Scra3:
Good catch, and it holds up on both sides. `automation-master-lease.ts` renews only on `tryHold`, which the config call drives, so the renewal interval really is `poll interval + cycle duration` — and the lease's own comment ("Twice the executor's default poll interval, so a live holder always renews before expiry") was written when a cycle cost `max(per inbox)`.

Noted in `022668b08`, on the `MAX_CONCURRENT_INBOX_POLLS` comment: capping the sweep buys the customer's database room by making cycles longer, which can outlast the TTL and let a second instance sweep alongside this one — reads, not correctness, since the assignment and run unique indexes absorb it.

Left the server side alone here rather than widen this PR. Its comment now overstates an invariant it no longer guarantees by construction, so it wants either a renewal that does not ride on the config call or an honest comment. Happy to open that one if you'd rather it not sit as a known gap.

matthv (approved):

